PR #1, opened 2014-08-06
Fix GPIO init on MicroBlaze

DenisKuzovin:
Add missing initialization of GPIO


commodo:
@DenisKuzovin 
can you re-spin this with a Signed-off-by tag ?
This looks good from my side and I'd merge

Sorry for the late reply.

Thanks
Alex

commodo:
after talking a bit with @dbogdan he made a point that `gpio_instance` is only used/needed for Zynq and not for Microblaze;

Microblaze uses the `gpio_config` instance directly to access GPIOs;
so adding a `gpio_instance` is kind of redundant

this change has been around for a while;
sorry for the lateness of this reply

will close it

thanks
Alex

PR #4, opened 2015-02-07
Fix RFPLL frequency calculation

tmineno:
Fix ad9361_rfpll_recalc_rate and ad9361_rfpll_set_rate.
After this fix, these methods return or set correct value even when user use SDM Bypass or SDM PD (0x232[7:6] or 0x272[7:6]).


rgetz:
The no-os driver is a derived work of the Linux driver. I think this section is a copy/paste, so the appropriate path for is for Michael to review, if it's necessary - commit to the Linux driver, and then Dragos will pull from that.

tmineno:
Thanks.
I'll send same pull req to linux repos.

rgetz:
Thanks -- closing this pull request then

PR #14, opened 2016-07-27
Fix lvds_bias_ctrl

JohnSchig:
Must subtract 75 from init_params->lvds_bias_mV before dividing by 75 in order to agree with the register definition which states: LVDS driver amplitude control. |VOD| = 75 mV to 450 mV; 75 mV/LSB.


dbogdan:
Hi John,
Thanks for letting us know about the issue and for fixing it.
Dragos

PR #18, opened 2016-10-24
2015 r2

davidabisset:
1st pull of no_os pn 2015_r2


commodo:
Closing PR.
There is no need to do a PR from an older branch to master.
Master should contain all that's needed and if not, we should cherry-pick them into master.

PR #28, opened 2017-08-14
Use the Euclidean algorithm in gcd() function across all drivers

dNechita:
This improves the efficiency of the gcd() function which returns the
greatest common divisor of two numbers.

Signed-off-by: Dan Nechita <Dan.Nechita@analog.com>

### DIFF
--- a/drivers/adf4350/adf4350.c
+++ b/drivers/adf4350/adf4350.c
@@ -151,17 +151,16 @@ int32_t adf4350_tune_r_cnt(struct adf4350_state *st, uint16_t r_cnt)
 *******************************************************************************/
 uint32_t gcd(uint32_t x, uint32_t y)
 {
-	int32_t tmp;
+	uint32_t tmp;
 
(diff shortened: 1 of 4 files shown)


drivers/adf4350/adf4350.c
```diff
@@ -153,15 +153,14 @@ uint32_t gcd(uint32_t x, uint32_t y)
 {
 	int32_t tmp;
```
gregoryng: this should probably be a uint32_t.

fmcomms1/ADF4351/ADF4351.c
```diff
@@ -145,15 +145,14 @@ uint32_t gcd(uint32_t x, uint32_t y)
 {
 	int32_t tmp;
```
gregoryng: this should probably be a uint32_t.

dNechita: Absolutely. Thanks!

PR #32, opened 2017-10-20
Update ad9361.c

JohnSchig:
Fixed debug output


mhennerich:
looks good - Thanks for reporting.

commodo:
Sorry for the late reply.
We've fixed this similarly in PR #139

PR #43, opened 2018-04-11
Corrected a wrong shift

hakan-akan:
According to the data sheet POWER_MODE register (0x4) bits 4-5, instead the macro is creating mask for bits 5-6.


dbogdan:
This looks good. Can you please update the pull request in order to merge this to master branch instead of dev?

Thanks,
Dragos

hakan-akan:
Done

commodo:
@hakan-akan 
simply changing the target branch will also bring other (undesired) commits into master

i suspect you used Github's web-UI to edit the code;
in this case, it would be simpler to close this PR and open another one basing from master
otherwise you need to do an interactive rebase from the command line tool to remove all commits except for your commit ; 

could you please re-spin this ? :)
[ either as a new PR or rebasing and removing all other commits is fine ]

thanks
Alex

PR #54, opened 2018-05-18
ad4110: update to new platform driver and coding style

adrimbarean:
Updated to work with the new platform driver and to keep linux coding
style.

Signed-off-by: Andrei Drimbarean <andrei.drimbarean@analog.com>


commodo:
Will merge this later today if no objections

PR #55, opened 2018-05-25
ad5770r: Initial commit of AD5770R driver

mirceacaprioru:
The AD5770R is a 6-channel, 14-bit resolution, low noise, programmable
current output digital-to-analog converter (DAC) for photonics control
applications. The chip contains five 14-bit resolution sourcing DAC
channels and one 14-bit resolution current sourcing/sinking DAC channel.

The driver enables the user to set the output of any of the five DAC
channels with hardware LDAC control or software LDAC.

### DIFF
--- a/drivers/ad5770r/ad5770r.c
+++ b/drivers/ad5770r/ad5770r.c
@@ -1,0 +1,650 @@
+/***************************************************************************//**
+ *   @file   ad5770r.c
+ *   @brief  Implementation of ad5770r Driver.
+ *   @author Mircea Caprioru (mircea.caprioru@analog.com)
+********************************************************************************
+ * Copyright 2018(c) Analog Devices, Inc.
+ *
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *  - Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ *  - Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in
+ *    the documentation and/or other materials provided with the
+ *    distribution.
+ *  - Neither the name of Analog Devices, Inc. nor the names of its
+ *    contributors may be used to endorse or promote products derived
+ *    from this software without specific prior written permission.
+ *  - The use of this software may or may not infringe the patent rights
+ *    of one or more patent holders.  This license does not release you
+ *    from the requirement that you obtain separate licenses from these
+ *    patent holders to use this software.
+ *  - Use of the software either in source or binary form, must be run
+ *    on or directly connected to an Analog Devices Inc. component.
+ *
+ * THIS SOFTWARE IS PROVIDED BY ANALOG DEVICES "AS IS" AND ANY EXPRESS OR
+ * IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, NON-INFRINGEMENT,
+ * MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED.
+ * IN NO EVENT SHALL ANALOG DEVICES BE LIABLE FOR ANY DIRECT, INDIRECT,
+ * INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, INTELLECTUAL PROPERTY RIGHTS, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+*******************************************************************************/
+
+/******************************************************************************/
+/***************************** Include Files **********************************/
+/******************************************************************************/
+#include <stdint.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <stdbool.h>
+#include <string.h>
+#include "platform_drivers.h"
+#include "ad5770r.h"
+
+/******************************************************************************/
+/************************** Functions Implementation **************************/
+/******************************************************************************/
+/**
+ * Read from device.
+ * @param dev - The device structure.
+ * @param reg_addr - The register address.
+ * @param reg_data - The register data.
+ * @return 0 in case of success, negative error code otherwise.
+ */
+int32_t ad5770r_spi_reg_read(struct ad5770r_dev *dev,
+			     uint8_t reg_addr,
+			     uint8_t *reg_data)
+{
+	uint8_t buf[3];
+	int32_t ret;
+
+	buf[0] = AD5770R_REG_READ(reg_addr);
+	buf[1] = 0x00;
+
+	ret = spi_write_and_read(dev->spi_desc, buf, 2);
+	*reg_data = buf[1];
+
+	return ret;
+}
+
+/**
+ * Multibyte read from device. A register read begins with the address
+ * and autoincrements for each aditional byte in the transfer.
+ * @param dev - The device structure.
+ * @param reg_addr - The register address.
+ * @param reg_data - The register data.
+ * @param count - Number of bytes to read.
+ * @return 0 in case of success, negative error code otherwise.
+ */
+int32_t ad5770r_spi_reg_read_multiple(struct ad5770r_dev *dev,
+				      uint8_t reg_addr,
+				      uint8_t *reg_data,
+				      uint16_t count)
+{
+	uint8_t buf[1024];
+	uint16_t i;
+	int32_t ret;
+
+	/* Check if multibyte enabled */
+
+	buf[0] = AD5770R_REG_READ(reg_addr);
+	memset(&buf[1], 0x00, count);
+
+	ret = spi_write_and_read(dev->spi_desc, buf, count + 1);
+
+	for (i = 0; i < count; i++)
+		reg_data[i] = buf[i+1];
+
+	return ret;
+}
+/**
+ * Write to device.
+ * @param dev - The device structure.
+ * @param reg_addr - The register address.
+ * @param reg_data - The register data.
+ * @return 0 in case of success, negative error code otherwise.
+ */
+int32_t ad5770r_spi_reg_write(struct ad5770r_dev *dev,
+			      uint8_t reg_addr,
+			      uint8_t reg_data)
+{
+	int32_t ret;
+	uint8_t buf[2];
+
+	buf[0] = AD5770R_REG_WRITE(reg_addr);
+	buf[1] = reg_data & 0xFF;
+
+	ret = spi_write_and_read(dev->spi_desc, buf, 2);
+
+	return ret;
+}
+
+/**
+ * Multibyte write from device. A register write begins with the address
+ * and autoincrements for each additional byte in the transfer.
+ * @param dev - The device structure.
+ * @param reg_addr - The register address.
+ * @param reg_data - The register data.
+ * @param count - Number of bytes to read.
+ * @return 0 in case of success, negative error code otherwise.
+ */
+int32_t ad5770r_spi_reg_write_multiple(struct ad5770r_dev *dev,
+				       uint8_t reg_addr,
+				       uint8_t *reg_data,
+				       uint16_t count)
+{
+	int32_t ret;
+	uint8_t data[1024];
+
+	data[0] = AD5770R_REG_WRITE(reg_addr);
+	memcpy(&data[1], reg_data, count);
+
+	ret = spi_write_and_read(dev->spi_desc, data, count + 1);
+
+	return ret;
+}
+
+/**
+ * SPI write to device using a mask.
+ * @param dev - The device structure.
+ * @param reg_addr - The register address.
+ * @param mask - The mask.
+ * @param data - The register data.
+ * @return 0 in case of success, negative error code otherwise.
+ */
+int32_t ad5770r_spi_write_mask(struct ad5770r_dev *dev,
+			       uint8_t reg_addr,
+			       uint32_t mask,
+			       uint8_t data)
+{
+	uint8_t reg_data;
+	int32_t ret;
+
+	ret = ad5770r_spi_reg_read(dev, reg_addr, &reg_data);
+	reg_data &= ~mask;
+	reg_data |= data;
+	ret |= ad5770r_spi_reg_write(dev, reg_addr, reg_data);
+
+	return ret;
+}
+
+/**
+ * Set device spi settings.
+ * @param dev - The device structure.
+ * @param spi_settings - The structure that contains the device spi
+ *		       parameters.
+ * @return SUCCESS in case of success, negative error code otherwise.
+ */
+int32_t ad5770r_set_device_spi(struct ad5770r_dev *dev,
+			       struct device_spi_settings spi_settings)
+{
+	int32_t ret;
+
+	ret = ad5770r_spi_write_mask(dev,
+				     AD5770R_INTERFACE_CONFIG_A,
+				     AD5770R_INTERFACE_CONFIG_A_ADDR_ASCENSION_MSB_MSK,
+				     AD5770R_INTERFACE_CONFIG_A_ADDR_ASCENSION_MSB(spi_settings.addr_ascension));
+
+	ret |= ad5770r_spi_write_mask(dev,
+				      AD5770R_INTERFACE_CONFIG_B,
+				      AD5770R_INTERFACE_CONFIG_B_SINGLE_INST_MSK,
+				      AD5770R_INTERFACE_CONFIG_B_SINGLE_INST(spi_settings.single_instruction));
+
+	ret |= ad5770r_spi_reg_write(dev,
+				     AD5770R_STREAM_MODE,
+				     spi_settings.stream_mode_length);
+
+	if (ret)
+		return -1;
+
+	dev->dev_spi_settings.addr_ascension =
+		spi_settings.addr_ascension;
+	dev->dev_spi_settings.single_instruction =
+		spi_settings.single_instruction;
+	dev->dev_spi_settings.stream_mode_length =
+		spi_settings.stream_mode_length;
+
+	return ret;
+}
+
+/**
+ * Set the channel configuration.
+ * @param dev - The device structure.
+ * @param channel_config - The structure that contains the channel
+ *			configuration.
+ * @return SUCCESS in case of success, negative error code otherwise.
+ */
+int32_t ad5770r_channel_config(struct ad5770r_dev *dev,
+			       bool *channel_config)
+{
+	int32_t ret;
+
+	ret = ad5770r_spi_reg_write(dev,
+				    AD5770R_CHANNEL_CONFIG,
+				    AD5770R_CHANNEL_CONFIG_CH0_SHUTDOWN_B(channel_config[0]) |
+				    AD5770R_CHANNEL_CONFIG_CH1_SHUTDOWN_B(channel_config[1]) |
+				    AD5770R_CHANNEL_CONFIG_CH2_SHUTDOWN_B(channel_config[2]) |
+				    AD5770R_CHANNEL_CONFIG_CH3_SHUTDOWN_B(channel_config[3]) |
+				    AD5770R_CHANNEL_CONFIG_CH4_SHUTDOWN_B(channel_config[4]) |
+				    AD5770R_CHANNEL_CONFIG_CH5_SHUTDOWN_B(channel_config[5]) |
+				    AD5770R_CHANNEL_CONFIG_CH0_SINK_EN(channel_config[6]));
+
+	if (ret)
+		return -1;
+
+	dev->channel_config = channel_config;
+
+	return ret;
+}
+
+/**
+ * Set the channel configuration.
+ * @param dev - The device structure.
+ * @param output_mode - The structure that contains the channel output
+ *			parameters.
+ * @return SUCCESS in case of success, negative error code otherwise.
+ */
+int32_t ad5770r_set_output_mode(struct ad5770r_dev *dev,
+				struct output_range output_mode,
+				enum ad5770r_channels channel)
+{
+	return ad5770r_spi_reg_write(dev,
+				     AD5770R_OUTPUT_RANGE_CH0 + channel,
+				     AD5770R_OUTPUT_RANGE_OUTPUT_SCALING(output_mode.output_scale) |
+				     AD5770R_OUTPUT_RANGE_MODE(output_mode.output_range_mode));
+};
+
+/**
+ * Set reference configuration.
+ * @param dev - The device structure.
+ * @param output_mode - The structure that contains the channel output
+ *			parameters.
+ * @return SUCCESS in case of success, negative error code otherwise.
+ */
+int32_t ad5770r_set_reference(struct ad5770r_dev *dev,
+			      bool external_reference,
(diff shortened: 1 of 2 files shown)


drivers/ad5770r/ad5770r.c
```diff
@@ -0,0 +1,650 @@
...
+ * @param dev - The device structure.
+ * @param reg_addr - The register address.
+ * @param reg_data - The register data.
+ * @return 0 in case of success, negative error code otherwise.
+ */
+int32_t ad5770r_spi_reg_read(struct ad5770r_dev *dev,
+			     uint8_t reg_addr,
+			     uint8_t *reg_data)
+{
+	uint8_t buf[3];
```
commodo: nitpick: you can make this `buf[2]` ; since you are reading/writing only 2 bytes

drivers/ad5770r/ad5770r.c
```diff
@@ -0,0 +1,650 @@
...
+	int32_t ret;
+
+	buf[0] = AD5770R_REG_READ(reg_addr);
+	buf[1] = 0x00;
+
+	ret = spi_write_and_read(dev->spi_desc, buf, 2);
+	*reg_data = buf[1];
```
commodo: since this function can be called outside of this file [as an API or other] maybe add a checks for `dev` and `reg_data`
similarly, this can be done for other similar functions;
usually functions return `-EINVAL` when nulls are passed as args

drivers/ad5770r/ad5770r.c
```diff
@@ -0,0 +1,650 @@
...
+				      uint16_t count)
+{
+	uint8_t buf[1024];
+	uint16_t i;
+	int32_t ret;
+
+	/* Check if multibyte enabled */
+
+	buf[0] = AD5770R_REG_READ(reg_addr);
+	memset(&buf[1], 0x00, count);
```
commodo: in this case, maybe make sure `count < sizeof(buf) - 1`

drivers/ad5770r/ad5770r.c
```diff
@@ -0,0 +1,650 @@
...
+
+	ret = spi_write_and_read(dev->spi_desc, buf, count + 1);
+
+	for (i = 0; i < count; i++)
+		reg_data[i] = buf[i+1];
```
commodo: you can use `memcpy()` here

drivers/ad5770r/ad5770r.c
```diff
@@ -0,0 +1,650 @@
...
+			      uint8_t reg_addr,
+			      uint8_t reg_data)
+{
+	int32_t ret;
+	uint8_t buf[2];
+
+	buf[0] = AD5770R_REG_WRITE(reg_addr);
+	buf[1] = reg_data & 0xFF;
+
+	ret = spi_write_and_read(dev->spi_desc, buf, 2);
```
commodo: nitpick: more of a style-comment; you can do `return spi_write_and_read(dev->spi_desc, buf, 2);` directly here

commodo: also, a nitpick [feel free to disregard]: you could pass `sizeof(buf)` instead of `2`; since many other drivers already pass constant values, I'm not feeling strongly about this ; so, i'll leave this up to you

drivers/ad5770r/ad5770r.c
```diff
@@ -0,0 +1,650 @@
...
+ * @param count - Number of bytes to read.
+ * @return 0 in case of success, negative error code otherwise.
+ */
+int32_t ad5770r_spi_reg_write_multiple(struct ad5770r_dev *dev,
+				       uint8_t reg_addr,
+				       uint8_t *reg_data,
+				       uint16_t count)
+{
+	int32_t ret;
+	uint8_t data[1024];
```
commodo: this also applies to other similar cases; but 1024 bytes on the call-stack can be a lot [depending on the user/application];
if we consider applications [like a few in the automotive], their use of the stack is limited via coding standard [or at least it used to be]; or if we consider some deployments that include micro-controllers ;

maybe using malloc() here would be slightly better;
but I'm open for discussion here;

drivers/ad5770r/ad5770r.c
```diff
@@ -0,0 +1,650 @@
...
+
+	data[0] = AD5770R_REG_WRITE(reg_addr);
+	memcpy(&data[1], reg_data, count);
+
+	ret = spi_write_and_read(dev->spi_desc, data, count + 1);
```
commodo: you could do `return spi_write_and_read(dev->spi_desc, data, count + 1);`   here directly;

drivers/ad5770r/ad5770r.c
```diff
@@ -0,0 +1,650 @@
...
+{
+	uint8_t reg_data;
+	int32_t ret;
+
+	ret = ad5770r_spi_reg_read(dev, reg_addr, &reg_data);
+	reg_data &= ~mask;
+	reg_data |= data;
+	ret |= ad5770r_spi_reg_write(dev, reg_addr, reg_data);
+
+	return ret;
```
commodo: `return ad5770r_spi_reg_write(dev, reg_addr, reg_data);`

drivers/ad5770r/ad5770r.c
```diff
@@ -0,0 +1,650 @@
...
+
+/**
+ * Set device spi settings.
+ * @param dev - The device structure.
+ * @param spi_settings - The structure that contains the device spi
+ *		       parameters.
+ * @return SUCCESS in case of success, negative error code otherwise.
+ */
+int32_t ad5770r_set_device_spi(struct ad5770r_dev *dev,
+			       struct device_spi_settings spi_settings)
```
commodo: pass this via const pointer;
```
int32_t ad5770r_set_device_spi(struct ad5770r_dev *dev,
			       const struct device_spi_settings *spi_settings)
```

the way this is done now, copies the entire `struct device_spi_settings` on the call-stack

drivers/ad5770r/ad5770r.c
```diff
@@ -0,0 +1,650 @@
...
+		return -1;
+
+	dev->dev_spi_settings.addr_ascension =
+		spi_settings.addr_ascension;
+	dev->dev_spi_settings.single_instruction =
+		spi_settings.single_instruction;
+	dev->dev_spi_settings.stream_mode_length =
+		spi_settings.stream_mode_length;
+
+	return ret;
```
commodo: if you don't plan to add anything later here, this could be `return 0` ;

drivers/ad5770r/ad5770r.c
```diff
@@ -0,0 +1,650 @@
...
+				    AD5770R_HW_LDAC,
+				    AD5770R_HW_LDAC_MASK_CH(mask_hw_ldac[0], AD5770R_CH0) |
+				    AD5770R_HW_LDAC_MASK_CH(mask_hw_ldac[1], AD5770R_CH1) |
+				    AD5770R_HW_LDAC_MASK_CH(mask_hw_ldac[2], AD5770R_CH2) |
+				    AD5770R_HW_LDAC_MASK_CH(mask_hw_ldac[3], AD5770R_CH3) |
+				    AD5770R_HW_LDAC_MASK_CH(mask_hw_ldac[4], AD5770R_CH4) |
+				    AD5770R_HW_LDAC_MASK_CH(mask_hw_ldac[5], AD5770R_CH5));
+
+	if (ret)
+		return -1;
```
commodo: maybe do,  
```
if (ret)
    return ret;
```
in all places;
that way we pass the error code up to the caller; so it doesn't get lost

drivers/ad5770r/ad5770r.c
```diff
@@ -0,0 +1,650 @@
...
+/**
+ * Set the channel configuration.
+ * @param dev - The device structure.
+ * @param output_mode - The structure that contains the channel output
+ *			parameters.
+ * @return SUCCESS in case of success, negative error code otherwise.
+ */
+int32_t ad5770r_set_output_mode(struct ad5770r_dev *dev,
+				struct output_range output_mode,
+				enum ad5770r_channels channel)
```
commodo: same for this struct; pass it via pointer;

drivers/ad5770r/ad5770r.c
```diff
@@ -0,0 +1,650 @@
...
+{
+	int32_t ret;
+	uint8_t data[2];
+
+	data[1] = (uint8_t)AD5770R_CH_DAC_DATA_LSB(page_mask.dac_data_page_mask);
+	data[0] = (uint8_t)((page_mask.dac_data_page_mask & 0x3FC0) >> 6);
+
+	ret = ad5770r_spi_reg_write_multiple(dev,
+					     AD5770R_DAC_PAGE_MASK_MSB,
+					     data, 2);
```
commodo: maybe it would be a good idea to do a `if (ret) return ret;`  here

drivers/ad5770r/ad5770r.c
```diff
@@ -0,0 +1,650 @@
...
+
+/**
+ * Initialize the device.
+ * @param device - The device structure.
+ * @param init_param - The structure that contains the device initial
+ *		       parameters.
+ * @return SUCCESS in case of success, negative error code otherwise.
+ */
+int32_t ad5770r_init(struct ad5770r_dev **device,
+		     struct ad5770r_init_param init_param)
```
commodo: pass this via pointer

drivers/ad5770r/ad5770r.h
```diff
@@ -0,0 +1,377 @@
...
+struct output_range {
+	uint8_t		output_scale;
+	uint8_t		output_range_mode;
+};
+
+struct device_spi_settings {
+	bool			addr_ascension;
+	bool			single_instruction; // for multibyte read/write
+	uint8_t		stream_mode_length;
+};
```
lclausen-adi: What about those structs above, should they be prefixed to avoid namespace issues?

drivers/ad5770r/ad5770r.h
```diff
@@ -0,0 +1,377 @@
...
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+*******************************************************************************/
+
+#ifndef AD5770R_H_
+#define AD5770R_H_
+
+/******************************************************************************/
+/********************** Macros and Constants Definitions **********************/
+/******************************************************************************/
+#define BIT(x)					(1UL << (x))
```
commodo: it's a good idea to have this guarded via #ifndef
```
#ifndef BIT
#define BIT(x)
#endif
```
to prevent this colliding with other similar definitions

drivers/ad5770r/ad5770r.c
```diff
@@ -0,0 +1,650 @@
...
+					 init_param.mon_setup);
+
+	*device = dev;
+
+	if (!ret)
+		printf("ad5770r successfully initialized\n");
+	else
+		printf("ad5770r initialization error (%d)\n", ret);
+
+	mdelay(1000);
```
commodo: maybe remove this delay; some users may not want it; or would like to add their own delay

drivers/ad5770r/ad5770r.c
```diff
@@ -0,0 +1,650 @@
...
+}
+
+/**
+ * Set channel select.
+ * @param dev - The device structure.
+ * @param mask_channel_sel - The array contains channel selection options.
+ * @return SUCCESS in case of success, negative error code otherwise.
+ */
+int32_t ad5770r_set_mask_channel(struct ad5770r_dev *dev,
+				 bool *mask_channel_sel)
```
commodo: i'm seeing a few cases where a predefined array of `bool` ( with 6 elements or more ) is passed ;
it's not a good idea to leave it like this for others to use, as it can be ambiguous for users about the number of elements in the array they should use;

there are 2 ways to do this:
1) document this [in the header or other docs ]
2) enforce a fixed typedef/struct;

[An example] You could have, 
```
struct ad5770r_channel_switches {
     bool en0, en1, en2, en3, en4, en5;
}
```
then
```
int32_t ad5770r_set_mask_channel(struct ad5770r_dev *dev,
				 struct ad5770r_channel_switches *ch)

ret = ad5770r_spi_reg_write(dev,
				    AD5770R_CH_SELECT,
				    AD5770R_CH_SELECT_SEL_CH(ch->en0, AD5770R_CH0) |
				    AD5770R_CH_SELECT_SEL_CH(ch->en1, AD5770R_CH1) |
				    AD5770R_CH_SELECT_SEL_CH(ch->en2, AD5770R_CH2) |
				    AD5770R_CH_SELECT_SEL_CH(ch->en3, AD5770R_CH3) |
				    AD5770R_CH_SELECT_SEL_CH(ch->en4, AD5770R_CH4) |
				    AD5770R_CH_SELECT_SEL_CH(ch->en5, AD5770R_CH5));
```

drivers/ad5770r/ad5770r.c
```diff
@@ -0,0 +1,650 @@
...
+				    AD5770R_CH_SELECT_SEL_CH(mask_channel_sel[1], AD5770R_CH1) |
+				    AD5770R_CH_SELECT_SEL_CH(mask_channel_sel[2], AD5770R_CH2) |
+				    AD5770R_CH_SELECT_SEL_CH(mask_channel_sel[3], AD5770R_CH3) |
+				    AD5770R_CH_SELECT_SEL_CH(mask_channel_sel[4], AD5770R_CH4) |
+				    AD5770R_CH_SELECT_SEL_CH(mask_channel_sel[5], AD5770R_CH5));
+
+	if (ret)
+		return -1;
+
+	dev->mask_channel_sel = mask_channel_sel;
```
commodo: also, passing references like this here, should be documented, or you should keep a copy of the data in the driver;
if the user passes a value that is on the stack, the `mask_channel_sel` pointer will point to something invalid at some point in time;

drivers/ad5770r/ad5770r.c
```diff
@@ -0,0 +1,650 @@
...
+
+/**
+ * Set the channel monitor configuration.
+ * @param dev - The device structure.
+ * @param mon_setup - The structure that contains the monitor setup
+ *		values.
+ * @return SUCCESS in case of success, negative error code otherwise.
+ */
+int32_t ad5770r_set_monitor_setup(struct ad5770r_dev *dev,
+				  struct monitor_setup mon_setup)
```
commodo: pass this via pointer; this also allocs `sizeof(struct monitor_setup)` on the call-stack

drivers/ad5770r/ad5770r.h
```diff
@@ -0,0 +1,377 @@
...
+int32_t ad5770r_set_channel_en(struct ad5770r_dev *dev,
+			       bool *channel_enable);
+int32_t ad5770r_get_status(struct ad5770r_dev *dev,
+			   uint8_t *status);
+int32_t ad5770r_set_monitor_setup(struct ad5770r_dev *dev,
+				  struct monitor_setup mon_setup);
+int32_t ad5770r_init(struct ad5770r_dev **device,
+		     struct ad5770r_init_param init_param);
+
+#endif // AD5770R_H_0
```
commodo: nitpick: this looks like a typo;  `AD5770R_H_0` -> `AD5770R_H_`; 
since this line may change, maybe use comment style `/* AD5770R_H_ */`

drivers/ad5770r/ad5770r.c
lclausen-adi: We just shouldn't do it. Taking references to other objects and keeping the reference without any object lifetime management is asking for trouble.

Besides the driver never uses this, so maybe just drop it.

mirceacaprioru:
Changelog V1 -> V2

- replaced return -1 with return ret
- added prefix to structures
- replaced bool arrays with struct ad5770r_channel_switches
- device structure no longer takes references to objects 
- replaced malloc with calloc
- 1024 arrays are now represented by a pointer that has memory allocated depending on count value. 
- removed delay at the end of init function
- added checks for dev and reg_data NULL pointer case

drivers/ad5770r/ad5770r.c
```diff
@@ -0,0 +1,658 @@
...
+int32_t ad5770r_spi_reg_read_multiple(struct ad5770r_dev *dev,
+				      uint8_t reg_addr,
+				      uint8_t *reg_data,
+				      uint16_t count)
+{
+	uint8_t *buf;
+	int32_t ret;
+
+	if (!dev | !reg_data)
+		return FAILURE;
```
commodo: why not `-EINVAL` ?

drivers/ad5770r/ad5770r.c
```diff
@@ -0,0 +1,658 @@
...
+	uint8_t buf[2];
+	int32_t ret;
+
+	if (!dev | !reg_data)
+		return FAILURE;
+
+	buf[0] = AD5770R_REG_READ(reg_addr);
+	buf[1] = 0x00;
+
+	ret = spi_write_and_read(dev->spi_desc, buf, sizeof(buf));
```
commodo: do error checking here ?
```
if (ret)
    return ret;
```

drivers/ad5770r/ad5770r.c
```diff
@@ -0,0 +1,658 @@
...
+
+	if (!dev | !reg_data)
+		return FAILURE;
+
+	buf = (uint8_t*)calloc(count + 1, sizeof(uint8_t));
+
+	buf[0] = AD5770R_REG_READ(reg_addr);
+	memset(&buf[1], 0x00, count);
+
+	ret = spi_write_and_read(dev->spi_desc, buf, count + 1);
```
commodo: error checking here ?

drivers/ad5770r/ad5770r.c
```diff
@@ -0,0 +1,658 @@
...
+	uint8_t *buf;
+	int32_t ret;
+
+	if (!dev | !reg_data)
+		return FAILURE;
+
+	buf = (uint8_t*)calloc(count + 1, sizeof(uint8_t));
+
+	buf[0] = AD5770R_REG_READ(reg_addr);
+	memset(&buf[1], 0x00, count);
```
commodo: `memset(0)` is redundant if you're using `calloc()`

drivers/ad5770r/ad5770r.c
```diff
@@ -0,0 +1,658 @@
...
+
+	ret = spi_write_and_read(dev->spi_desc, buf, count + 1);
+
+	memcpy(reg_data, buf + 1, count);
```
commodo: you need to `free(buf)` here, and on all exit paths if changing stuff

drivers/ad5770r/ad5770r.c
```diff
@@ -0,0 +1,658 @@
...
+{
+	uint8_t *data;
+
+	if (!dev | !reg_data)
+		return FAILURE;
+
+	data = (uint8_t*)calloc(count + 1, sizeof(uint8_t));
+
+	data[0] = AD5770R_REG_WRITE(reg_addr);
+	memcpy(&data[1], reg_data, count);
```
commodo: `free(data)` before exiting

drivers/ad5770r/ad5770r.c
```diff
@@ -0,0 +1,658 @@
...
+				    AD5770R_CHANNEL_CONFIG_CH2_SHUTDOWN_B(channel_config->en2) |
+				    AD5770R_CHANNEL_CONFIG_CH3_SHUTDOWN_B(channel_config->en3) |
+				    AD5770R_CHANNEL_CONFIG_CH4_SHUTDOWN_B(channel_config->en4) |
+				    AD5770R_CHANNEL_CONFIG_CH5_SHUTDOWN_B(channel_config->en5) |
+				    AD5770R_CHANNEL_CONFIG_CH0_SINK_EN(channel_config->sink0));
+
+	if (ret)
+		return ret;
+
+	dev->channel_config = *channel_config;
```
commodo: this still passes a reference to an external object, which may have limited life-time

commodo: wait; my bad
it de-references the object;

drivers/ad5770r/ad5770r.c
lclausen-adi: Long term it might be good to add a spi\_write\_**then**\_read() to our set of library functions. The function would take two buffers and two counts, one for write and one for read.

This seems to be a common pattern and would make the whole alloc/memcpy/free dance unnecessary.

drivers/ad5770r/ad5770r.c
commodo: it would be a good idea to do `memcpy(&dev->channel_config, channel_config, sizeof(dev->channel_config))`
this is more certain to work well on most C compilers;

lclausen-adi: No, keep it the way it is. This should work with all compilers either way.

drivers/ad5770r/ad5770r.c
```diff
@@ -0,0 +1,658 @@
...
+ * Set the hardware ldac configuration.
+ * @param dev - The device structure.
+ * @param mask_hw_ldac - The array contains HW LDAC channel masks.
+ * @return SUCCESS in case of success, negative error code otherwise.
+ */
+int32_t ad5770r_set_hw_ladc(struct ad5770r_dev *dev,
+			    const struct ad5770r_channel_switches *mask_hw_ldac)
+{
+	int32_t ret;
+
```
commodo: you could to null-checks here:
```
if (!dev || !mask_hw_ldac)
    return -EINVAL;
```
and in similar places

drivers/ad5770r/ad5770r.c
```diff
@@ -0,0 +1,658 @@
...
+		     const struct ad5770r_init_param *init_param)
+{
+	struct ad5770r_dev	*dev;
+	uint8_t product_id_l, product_id_h;
+	int32_t ret;
+	enum ad5770r_channels i;
+
+	dev = (struct ad5770r_dev *)calloc(1, sizeof(*dev));
+	if (!dev)
+		return FAILURE;
```
commodo: you could use `-ENOMEM` here

drivers/ad5770r/ad5770r.c
```diff
@@ -0,0 +1,658 @@
...
+ * @return SUCCESS in case of success, negative error code otherwise.
+ */
+int32_t ad5770r_init(struct ad5770r_dev **device,
+		     const struct ad5770r_init_param *init_param)
+{
+	struct ad5770r_dev	*dev;
+	uint8_t product_id_l, product_id_h;
+	int32_t ret;
+	enum ad5770r_channels i;
+
```
commodo: you could check if `device` is NULL here;

drivers/ad5770r/ad5770r.c
```diff
@@ -0,0 +1,658 @@
...
+	if (product_id_l != 0x04 || product_id_h != 0x40) {
+		printf("failed to read id (0x%X : 0x%X)\n", product_id_l,
+		       product_id_h);
+		return FAILURE;
+	}
+
+	/* Device settings */
+	ret = ad5770r_set_device_spi(dev,
+				     &init_param->dev_spi_settings);
```
commodo: there is no case to handle  `init_param` being null;
there are 2 ways to do it:
1) if `init_param` is NULL, return -EINVAL
2) if `init_param` is NULL pass a default static `init_param` with default values for this device

your preference which one to choose

drivers/ad5770r/ad5770r.c
```diff
@@ -0,0 +1,658 @@
...
+
+	ret |= ad5770r_set_reference(dev,
+				     init_param->external_reference,
+				     init_param->reference_selector);
+	ret |= ad5770r_set_alarm(dev, &init_param->alarm_config);
+
+	ret |= ad5770r_set_hw_ladc(dev,
+				   &init_param->mask_hw_ldac);
+
+	for( i = AD5770R_CH0; i <=  AD5770R_CH5; i++) {
```
commodo: hmm; maybe also run a style-check on this;

mirceacaprioru:
Changelog V2 -> V3

- added checks for dev and other necessary structures
- added free(data) before exiting function
- remove memset(0) after calloc

drivers/ad5770r/ad5770r.c
```diff
@@ -0,0 +1,718 @@
...
+ * @param dev - The device structure.
+ * @param mon_setup - The structure that contains the monitor setup
+ *		values.
+ * @return SUCCESS in case of success, negative error code otherwise.
+ */
+int32_t ad5770r_get_status(struct ad5770r_dev *dev,
+			   uint8_t *status)
+{
+
+	if (!dev)
```
commodo: nitpick: extra line before this check;
also, `ad5770r_spi_reg_read()` handles NULL, so you could remove this check

PR #56, opened 2018-05-30
Update ad7124_regs.c

ArjunMurthyS:
We had a issue with writing to the register after initialization it was because of we had 0x0044 at AD7124_Error_En... then changed it to 0x0040... then it is working.


adrimbarean:
Thank you! The issue has been solved.

PR #59, opened 2018-06-21
adxl372: Add missing self test commands

wko27:
Several of the self test commands were missing from the ADXL 372 driver :)


drivers/adxl372/adxl372.h
```diff
@@ -204,6 +204,11 @@
 #define ADXL372_INT2_MAP_LOW_MSK		BIT(7)
 #define ADXL372_INT2_MAP_LOW_MODE(x)		(((x) & 0x1) << 7)
 
+#define ADXL372_SELF_TEST_ST_DONE_MSK		BIT(1)
+#define ADXL372_SELF_TEST_ST_DONE_MODE(x)	(((x) & 0x1) << 1)
+#define ADXL372_SELF_TEST_USER_ST_MSK		BIT(2)
+#define ADXL372_SELF_TEST_USER_ST_MODE(x)	(((x) & 0x1) << 2)
+
```
stefpopa: I am not sure if these commands are of any use, since ST_DONE and USER_ST bits are read only and these masks are used together with adxl372_spi_write_mask() function.
However, we can add a command for initializing the self test, something like this:
```
#define ADXL372_SELF_TEST_ST_MSK	BIT(0)
#define ADXL372_SELF_TEST_ST_MODE(x)	(((x) & 0x1) << 0)
```

stefpopa: If we want to check if ST_DONE and USER_ST bits are set, we can do it something like this: 
```
#define ADXL372_SELF_TEST_ST_DONE(x)		(((x) >> 1) & 0x1)
#define ADXL372_SELF_TEST_USER_ST(x)		(((x) >> 2) & 0x1)
```

drivers/adxl372/adxl372.h
```diff
@@ -133,6 +133,9 @@
 #define ADXL372_POWER_CTL_MODE_MSK		GENMASK(1, 0)
 #define ADXL372_POWER_CTL_MODE(x)		(((x) & 0x3) << 0)
 
+#define ADXL372_OFFSET_Z_MODE_MSK               GENMASK(3, 0)
+#define ADXL372_OFFSET_Z_MODE(x)                (((x) & 0xF) << 0)
+
```
stefpopa: Please also add a use case for this macros

mirceacaprioru:
Any updates on this PR? Also the CLA license must be signed before merging this.

wko27:
Hello!

Unfortunately, it looks like my company's legal counsel is unwilling to sign off on the CLA license :(

Apologies!
Andrew

PR #69, opened 2018-07-11
ad6676-ebz: add SPI type initialization for multiple carriers

adrimbarean:
SPI type must correspond to the current carrier. Add initialization for
MICROBLAZE, ZINQ_PSU and NIOS_II.


dbogdan:
It's a good point of view. We should do this.

PR #70, opened 2018-08-06
ad9371: platform_xilinx: fix GPIO_OFFSET for Microblaze

commodo:
Reported via EZ:
  https://ez.analog.com/thread/109040-ad9375-platformdriversh-modification-using-microblaze

The GPIOs for Microblaze are part of 2 x 32 bit registers, so there cannot
be a GPIO number higher than 63.

The GPIO_OFFSET is a left-over from the implementation of the driver for
the Zynq platform.

Signed-off-by: Alexandru Ardelean <alexandru.ardelean@analog.com>


commodo:
Will merge this later today if no objections

PR #74, opened 2018-08-09
Master ad5676 support

stefpopa:
This series of patches adds support for AD5671R/72R/75R/76/76R/86/94/96 DACs as part of the ad5686 driver.  An older version of the ad5676 driver existed, but was not updated to the new platform drivers, therefore it was removed to avoid confusion.

Adding support for all these similar devices inside ad5686 is also consistent with the implementation of the Linux driver.


drivers/ad5686/ad5686.h
```diff
@@ -113,9 +108,21 @@ enum comm_type {
...
+	AD5686_CH_4,
+	AD5686_CH_5,
+	AD5686_CH_6,
+	AD5686_CH_7,
+};
+
 struct ad5686_chip_info {
 	uint8_t		resolution;
 	enum comm_type	communication;
+	const unsigned int *channel_addr;
```
adrimbarean: Use fixed width integers to avoid ambiguity. 'const uint32_t' instead of 'const unsigned int'.

stefpopa:
changes in v2:
* changed `const unsigned int *channel_addr; `to const `uint32_t *channel_addr;`

Issue #75, opened 2018-08-09
Why does ad9361_dig_tune return 0 when max_freq != 0 in the case of failure?


codypiersall:
I'm using `ad9361_dig_tune` to re-tune the interface whenever the sampling frequency is changed.  I pass in 
0 for the parameter `max_freq` so that only the current sampling rate is tuned.  I was expecting to receive an error code whenever the interface could not be tuned, but to my surprise the function returned 0!  I traced this down to [line 547 of ad9361_conv.c](https://github.com/analogdevicesinc/no-OS/blob/d31cb84db241a35b1a8b03adeb627e386952f412/ad9361/sw/ad9361_conv.c#L547).  What is the reason for that check, instead of just returning the error whenever tuning doesn't work?

buha:
Closing due to inactivity. Feel free to reopen if you still think it's applicable.

PR #76, opened 2018-08-10
Make ad9361_dig_tune report failure to tune.

codypiersall:
ad9361_dig_tune didn't return an error code when max_freq == 0, making
it impossible to know whether tuning the digital interface to a specific
frequency succeeded. Now a caller of ad9361_dig_tune can determine
whether tuning to the frequency was successful or not.

### DIFF
--- a/ad9361/sw/ad9361_conv.c
+++ b/ad9361/sw/ad9361_conv.c
@@ -544,8 +544,6 @@ int32_t ad9361_dig_tune(struct ad9361_rf_phy *phy, uint32_t max_freq,
 
 		if (ret == -EIO)
 			restore = true;
-		if (!max_freq)
-			ret = 0;
 	}
 
 	if (restore) {


codypiersall:
This fixes #75.

mhennerich:
This fix will break FMCOMMS5 setup where the salve core (ADI_REG_ID > 0) is expected to fail -

	ret = ad9361_dig_tune(phy, (axiadc_read(st, ADI_REG_ID)) ?
		0 : 61440000, flags);
	if (ret < 0)
		goto error;

So far max frequency has been used to support this case.

codypiersall:
Ah, interesting.  I'm not using any of the FMCOMMS boards, so I may be breaking some assumptions or conditions that weren't meant to be broken; one other thing to note is that the parameter `max_freq` is not the maximum frequency that will be tuned to, but a flag that causes all the rates in the `rates` array to be tuned to.  (I'm sure you're aware of that, but it took me a while to figure it out!)    I'm using the `max_freq` flag to tune to a user-specified sampling frequency.  I think replacing that code with something like this would work (untested!)

```c
bool is_slave = axiadc_read(st, ADI_REG);
ret = ad9361_dig_tune(phy,  is_slave ? 0 : 61440000, flags & BE_VERBOSE);
if (ret < 0 && !is_slave)
    return ret;
}
```

I've already changed what I needed to change for my project, but the less deviation from upstream the better for me.

mirceacaprioru:
Since this will not be applied, i thinks it is safe to close this PR.

PR #77, opened 2018-08-10
AD9144 PLL support

lclausen-adi:
Add support for the internal PLL in the AD9144. It can be used to synthesize a DAC clock from a slower external reference clock.

Also add an example to the DAQ2 project that shows how to use the PLL mode.

### DIFF
--- a/drivers/ad9144/ad9144.c
+++ b/drivers/ad9144/ad9144.c
@@ -272,6 +272,95 @@ int32_t ad9144_setup_jesd204_link(struct ad9144_dev *dev,
 	return 0;
 }
 
+/*
+ * PLL fixed register writes according to table 17 of the
+ * AD9144 datasheet Rev. B.
+ */
+static const struct ad9144_reg_seq ad9144_pll_fixed_writes[] = {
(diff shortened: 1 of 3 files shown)


drivers/ad9144/ad9144.c
```diff
@@ -272,6 +272,95 @@ int32_t ad9144_setup_jesd204_link(struct ad9144_dev *dev,
...
+	{ 0x1bf, 0x8e },
+	{ 0x1c0, 0x2a },
+	{ 0x1c1, 0x2a },
+	{ 0x1c4, 0x7e },
+};
+
+static int ad9144_pll_setup(struct ad9144_dev *dev,
+			    const struct ad9144_init_param *init_param)
+{
+	unsigned int fref, fdac;
```
adrimbarean: Using fixed width integer types will help reduce code ambiguity when using different compilers and platforms. So we should use 'uint32_t' instead of 'unsigned int' and 'int32_t' instead of 'int'.

lclausen-adi:
Changes in v2:
* Use fixed width integer types

PR #79, opened 2018-08-15
fmcdaq2 SPI improvements

david-mle:
No functional change.
Commits can be applied separately if desired.


mirceacaprioru:
Sorry for the delayed response to this PR. The change is good but since it has been so long between this one and our work there are some conflicts on applying this on master. I would do a rebase of the master and then i would apply it. 

Thanks and sorry again,
Mircea

PR #80, opened 2018-08-17
ad9144: Keep link disabled during configuration

lclausen-adi:
Currently when calling the ad9144_setup() function there is a certain
chance that the AD9144 will not be correctly initialized. The symptoms can
range from the AD9144 not reporting any errors, but also not generating any
signal at the DAC output, to recording a lot of JESD204 character errors
and constantly restarting the link.

This appears to be caused by the link being enabled while the link is being
configured. Make sure that the link is kept disabled during link
configuration.

After this patch has been applied erratic behavior after converter
initialization can no longer be observed and the AD9144 properly
initializes every time.

Signed-off-by: Lars-Peter Clausen <lars@metafoo.de>


commodo:
LGTM
Will merge this in the next 1-2 days if no objections

PR #86, opened 2018-08-27
ad4111: add AD4111 driver

adrimbarean:
The AD4111 is a low power, low noise, 24-bit, sigma delta (Σ-Δ) analog-
to-digital converter (ADC) that integrates an analog front end (AFE) for
fully differential or single-ended rail-to-rail, buffered bipolar, ±10 V
voltage inputs, and 0 mA to 20 mA current inputs.


adrimbarean:
Device is too similar to the AD717x family to require different drivers. Support for this device will be added in the ad717x driver.

PR #87, opened 2018-08-29
Ad4111 driver dev

adrimbarean:
Make small fixes to the AD717X driver and add AD4111 support to the AD717X driver.

The AD4111 is a low power, low noise, 24-bit, sigma delta (Σ-Δ) analog-to-digital converter (ADC) that integrates an analog front end (AFE) for fully differential or single-ended rail-to-rail, buffered bipolar, ±10 V voltage inputs, and 0 mA to 20 mA current inputs.

The AD4111 is very similar to the AD717X family so support for it is added to that family driver instead of creating new driver with duplicate code.

### DIFF
--- a/drivers/ad717x/ad717x.c
+++ b/drivers/ad717x/ad717x.c
@@ -439,6 +439,11 @@ int32_t AD717X_Init(ad717x_dev **device,
                 preg ++;
         }
 
+        /* Read ID register to identify the part */
+        ret = AD717X_ReadRegister(dev, AD717X_ID_REG);
+        if(ret < 0)
+                return ret;
+
         *device = dev;
 
         return ret;
--- a/drivers/ad717x/ad717x.c
+++ b/drivers/ad717x/ad717x.c
@@ -285,13 +285,54 @@ int32_t AD717X_ReadData(ad717x_dev *device,
         if (!dataReg)
                 return INVALID_VAL;
 
+        /* Update the data register length with respect to device and options */
+        ret = AD717X_ComputeDataregSize(device);
+
         /* Read the value of the Status Register */
-        ret = AD717X_ReadRegister(device, AD717X_DATA_REG);
+        ret |= AD717X_ReadRegister(device, AD717X_DATA_REG);
 
         /* Get the read result */
         *pData = dataReg->value;
 
         return ret;
+}
+
+/***************************************************************************//**
+* @brief Computes data register read size to account for bit number and status
+* 		 read.
+*
+* @param device - The handler of the instance of the driver.
+*
+* @return 0in case of success or negative code in case of failure.
+*******************************************************************************/
+int32_t AD717X_ComputeDataregSize(ad717x_dev *device)
+{
+	ad717x_st_reg *reg_ptr;
+	ad717x_st_reg *datareg_ptr;
+	uint16_t case_var;
+
+	/* Get interface mode register pointer */
+	reg_ptr = AD717X_GetReg(device, AD717X_IFMODE_REG);
+	/* Get data register pointer */
+	datareg_ptr = AD717X_GetReg(device, AD717X_DATA_REG);
+	case_var = reg_ptr->value & (AD717X_IFMODE_REG_DATA_STAT |
+			AD717X_IFMODE_REG_DATA_WL16);
+
+	/* Compute data register size */
+	datareg_ptr->size = 3;
+	if ((case_var & AD717X_IFMODE_REG_DATA_WL16) == AD717X_IFMODE_REG_DATA_WL16)
+		datareg_ptr->size--;
+	if ((case_var & AD717X_IFMODE_REG_DATA_STAT) == AD717X_IFMODE_REG_DATA_STAT)
+		datareg_ptr->size++;
+
+	/* Get ID register pointer */
+	reg_ptr = AD717X_GetReg(device, AD717X_ID_REG);
+
+	/* If the part is 32/24 bit wide add a byte to the read */
+	if((reg_ptr->value & AD717X_ID_REG_MASK) == AD7177_2_ID_REG_VALUE)
+		datareg_ptr->size++;
+
+	return 0;
 }
 
 /***************************************************************************//**
--- a/drivers/ad717x/ad717x.h
+++ b/drivers/ad717x/ad717x.h
@@ -301,6 +301,10 @@ int32_t AD717X_WaitForReady(ad717x_dev *device,
 int32_t AD717X_ReadData(ad717x_dev *device,
 			int32_t* pData);
 
+/*! Computes data register read size to account for bit number and status
+ *  read. */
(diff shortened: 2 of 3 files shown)


drivers/ad717x/ad717x.h
```diff
@@ -158,7 +158,7 @@ typedef struct {
 #define AD717X_STATUS_REG_ADC_ERR  (1 << 6)
 #define AD717X_STATUS_REG_CRC_ERR  (1 << 5)
 #define AD717X_STATUS_REG_REG_ERR  (1 << 4)
-#define AD717X_STATUS_REG_CH(x)    ((x) & 0x03)
+#define AD717X_STATUS_REG_CH(x)    ((x) & 0x0f)
```
stefpopa: nitpick: use 0x0F instead of 0x0f for consistency sake

drivers/ad717x/ad717x.c
```diff
@@ -285,6 +285,9 @@ int32_t AD717X_ReadData(ad717x_dev *device,
         if (!dataReg)
                 return INVALID_VAL;
 
+        /* Update the data register length with respect to device and options */
+        ret |= AD717X_ComputeDataregSize(device);
```
stefpopa: ret will be overwritten bellow

drivers/ad717x/ad717x.c
```diff
@@ -439,6 +439,9 @@ int32_t AD717X_Init(ad717x_dev **device,
                 preg ++;
         }
 
+        /* Read ID register to identify the part */
+        ret = AD717X_ReadRegister(dev, AD717X_ID_REG);
```
stefpopa: where is the ID register further used?

drivers/ad717x/ad717x.c
```diff
@@ -294,6 +297,58 @@ int32_t AD717X_ReadData(ad717x_dev *device,
...
+* @brief Computes data register read size to account for bit number and status
+* 		 read.
+*
+* @param device - The handler of the instance of the driver.
+*
+* @return 0in case of success or negative code in case of failure.
+*******************************************************************************/
+int32_t AD717X_ComputeDataregSize(ad717x_dev *device)
+{
+	int32_t ret = 0;
```
stefpopa: you don't need this variable since it is never updated

drivers/ad717x/ad717x.c
```diff
@@ -294,6 +297,58 @@ int32_t AD717X_ReadData(ad717x_dev *device,
...
+	/* Get interface mode register pointer */
+	reg_ptr = AD717X_GetReg(device, AD717X_IFMODE_REG);
+	/* Get data register pointer */
+	datareg_ptr = AD717X_GetReg(device, AD717X_DATA_REG);
+
+	case_var = reg_ptr->value & (AD717X_IFMODE_REG_DATA_STAT |
+			AD717X_IFMODE_REG_DATA_WL16);
+
+	/* Compute data register size */
+    switch(case_var) {
```
stefpopa: indentation of the switch is strange

drivers/ad717x/ad717x.c
```diff
@@ -294,6 +297,58 @@ int32_t AD717X_ReadData(ad717x_dev *device,
...
+    default:
+    	datareg_ptr->size = 3;
+    	break;
+    }
+
+	/* Get ID register pointer */
+    reg_ptr = AD717X_GetReg(device, AD717X_ID_REG);
+
+    /* If the part is 32/24 bit wide add a byte to the read */
+    if((reg_ptr->value & 0xfff0) == 0x4fd0)
```
stefpopa: use a define instead of 0x4fd0 'magic number'; use a mask instead of 0xfff0

drivers/ad717x/ad717x.c
```diff
@@ -294,6 +297,58 @@ int32_t AD717X_ReadData(ad717x_dev *device,
...
+	int32_t ret = 0;
+	ad717x_st_reg *reg_ptr;
+	ad717x_st_reg *datareg_ptr;
+	uint16_t case_var;
+
+	/* Get interface mode register pointer */
+	reg_ptr = AD717X_GetReg(device, AD717X_IFMODE_REG);
+	/* Get data register pointer */
+	datareg_ptr = AD717X_GetReg(device, AD717X_DATA_REG);
+
```
stefpopa: one extra blank space

drivers/ad717x/ad717x.c
```diff
@@ -294,6 +297,58 @@ int32_t AD717X_ReadData(ad717x_dev *device,
...
+    case AD717X_IFMODE_REG_DATA_STAT:
+    	datareg_ptr->size = 4;
+    	break;
+    case (AD717X_IFMODE_REG_DATA_STAT | AD717X_IFMODE_REG_DATA_WL16):
+    	datareg_ptr->size = 3;
+    	break;
+    default:
+    	datareg_ptr->size = 3;
+    	break;
+    }
```
stefpopa: this can be simplified:
```
datareg_ptr->size = 3; // default
if (case_var == AD717X_IFMODE_REG_DATA_WL16)
        datareg_ptr->size--;
else if (case_var == AD717X_IFMODE_REG_DATA_STAT)
         datareg_ptr->size--;
```
or something like this

drivers/ad717x/ad717x.c
```diff
@@ -398,7 +398,7 @@ int32_t AD717X_Init(ad717x_dev **device,
         dev->num_regs = init_param.num_regs;
 
         /* Initialize the SPI communication. */
-        ret = spi_init(&dev->spi_desc, &init_param.spi_init);
+        ret = spi_init(&dev->spi_desc, init_param.spi_init);
```
commodo: are you sure ?
which `spi_init` function are you looking at ?

drivers/ad717x/ad717x.c
commodo: this can also be problematic, since this changes the value of `ret` to non-zero, which can be interpreted as error

adrimbarean:
Change v0->v1:
- use upper-case letters for hex numbers
- change AD717X_ComputeDataregSize switch to the recommended if logic
- used defines instead of magic numbers
- erase redundant ret variable
- remove extra blank spaces

adrimbarean:
Changes v1->v2:
- remove spi_init fix commit because it was wrong

drivers/ad717x/ad717x.c
commodo: ideally, you should also use the value read here;
either print it, or store it somewhere

adrimbarean:
Changes v2->v3:
- add conditional return after reading ID register in AD717X_Init function

drivers/ad717x/ad717x.c
commodo: my bad here; i didn't realize that registers get stored in a regmap/reg list

commodo:
Will merge next week if no objections

PR #90, opened 2018-09-04
Master adxl372 i2c support

stefpopa:
This series of patches adds support for I2C communication, but also adds a few miscellaneous, no functional changes which should improve readability.

### DIFF
--- a/drivers/adxl372/adxl372.c
+++ b/drivers/adxl372/adxl372.c
@@ -675,20 +675,32 @@ int32_t adxl372_init(struct adxl372_dev **device,
 		     struct adxl372_init_param init_param)
 {
 	struct adxl372_dev	*dev;
-	uint8_t dev_id, part_id;
+	uint8_t dev_id, part_id, rev_id;
 	int32_t ret;
 
 	dev = (struct adxl372_dev *)malloc(sizeof(*dev));
 	if (!dev)
 		return -1;
 
-	dev->reg_read = adxl372_spi_reg_read;
-	dev->reg_write = adxl372_spi_reg_write;
-	dev->reg_read_multiple = adxl372_spi_reg_read_multiple;
+	dev->comm_type = init_param.comm_type;
+	if (dev->comm_type == SPI) {
+		/* SPI */
+		ret = spi_init(&dev->spi_desc, &init_param.spi_init);
+		dev->reg_read = adxl372_spi_reg_read;
+		dev->reg_write = adxl372_spi_reg_write;
+		dev->reg_read_multiple = adxl372_spi_reg_read_multiple;
+	} else { /* I2C */
+		ret = i2c_init(&dev->i2c_desc, &init_param.i2c_init);
+		dev->reg_read = adxl372_i2c_reg_read;
+		dev->reg_write = adxl372_i2c_reg_write;
+		dev->reg_read_multiple = adxl372_i2c_reg_read_multiple;
 
-	/* SPI */
-	ret = spi_init(&dev->spi_desc, &init_param.spi_init);
-
+		ret |= adxl372_read_reg(dev, ADXL372_REVID, &rev_id);
+		/* Starting with the 3rd revision an I2C chip bug was fixed */
+		if (rev_id < 3) {
+			printf("I2C might not work properly with other "
+			       "devices present on the bus\n");
+	}
 	/* GPIO */
 	ret |= gpio_get(&dev->gpio_int1,
 			init_param.gpio_int1);
--- a/drivers/adxl372/adxl372.c
+++ b/drivers/adxl372/adxl372.c
@@ -1,6 +1,6 @@
 /***************************************************************************//**
  *   @file   adxl372.c
- *   @brief  Implementation of adxl372 Driver.
+ *   @brief  Implementation of adxl372 Core Driver.
  *   @author SPopa (stefan.popa@analog.com)
 ********************************************************************************
  * Copyright 2018(c) Analog Devices, Inc.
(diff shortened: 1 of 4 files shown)


drivers/adxl372/adxl372.c
```diff
@@ -58,7 +58,7 @@
  * @param reg_data - The register data.
  * @return 0 in case of success, negative error code otherwise.
  */
-int32_t adxl372_spi_reg_read(adxl372_dev *dev,
+int32_t adxl372_spi_reg_read(struct adxl372_dev *dev,
```
commodo: This looks more like a preference here; and this change adds a bit of noise.
Personally, I don't mind whether we use struct/enum prefixes [Linux style] or go with typedefs;

I guess we'd need to come [internally] to an agreement about which coding style to use for types.

That being said, it's fine from me to add this change.

stefpopa: I think we should stick with the Linux coding style. In my opinion, the argument of not using typedefs makes sense: https://www.kernel.org/doc/html/v4.10/process/coding-style.html#typedefs

stefpopa:
Changes in v2:
* changed the file description for adxl372_i2c.c and adxl372_spi.c

drivers/adxl372/adxl372_spi.c
```diff
@@ -0,0 +1,118 @@
...
+ * @param reg_data - The register data.
+ * @param count - Number of bytes to read.
+ * @return 0 in case of success, negative error code otherwise.
+ */
+int32_t adxl372_spi_reg_read_multiple(struct adxl372_dev *dev,
+				      uint8_t reg_addr,
+				      uint8_t *reg_data,
+				      uint16_t count)
+{
+	uint8_t buf[1024];
```
commodo: just a thought: 1024 is a bit big to put on the stack of certain systems;

drivers/adxl372/adxl372_spi.c
```diff
@@ -0,0 +1,118 @@
...
+	uint16_t i;
+	int32_t ret;
+
+	buf[0] = ADXL372_REG_READ(reg_addr);
+	memset(&buf[1], 0x00, count);
+
+	ret = spi_write_and_read(dev->spi_desc, buf, count + 1);
+
+	for (i = 0; i <= count; i++)
```
commodo: here you could use `memcpy()`

drivers/adxl372/adxl372_i2c.c
```diff
@@ -0,0 +1,128 @@
...
+
+	ret = i2c_write(dev->i2c_desc, buf, 1, 0);
+	if (ret < 0)
+		return ret;
+
+	ret = i2c_read(dev->i2c_desc, buf, count, 0);
+	if (ret < 0)
+		return ret;
+
+	for (i = 0; i < count; i++)
```
commodo: memcpy() here ?

stefpopa:
changes in v3:
* changed the buffer size for the fifo from 1024 to 512
* used memcpy()

commodo:
Will merge tomorrow if no objections.

drivers/adxl372/adxl372_spi.c
```diff
@@ -0,0 +1,117 @@
...
+	uint8_t buf[512];
+	uint16_t i;
+	int32_t ret;
+
+	buf[0] = ADXL372_REG_READ(reg_addr);
+	memset(&buf[1], 0x00, count);
+
+	ret = spi_write_and_read(dev->spi_desc, buf, count + 1);
+
+	memcpy(reg_data, &buf[1], count + 1);
```
commodo: curios: would it be `memcpy(reg_data, &buf[1], count);` here ?

drivers/adxl372/adxl372_spi.c
```diff
@@ -0,0 +1,117 @@
...
+				      uint8_t reg_addr,
+				      uint8_t *reg_data,
+				      uint16_t count)
+{
+	uint8_t buf[512];
+	uint16_t i;
+	int32_t ret;
+
+	buf[0] = ADXL372_REG_READ(reg_addr);
+	memset(&buf[1], 0x00, count);
```
commodo: it would probably be a good idea to check that `count < sizeof(buf)` and return an err here

drivers/adxl372/adxl372_i2c.c
```diff
@@ -0,0 +1,127 @@
...
+				      uint8_t reg_addr,
+				      uint8_t *reg_data,
+				      uint16_t count)
+{
+	uint8_t buf[512];
+	uint16_t i;
+	int32_t ret;
+
+	buf[0] = reg_addr;
+	memset(&buf[1], 0x00, count);
```
commodo: it would probably be a good idea to check that `count < sizeof(buf)` and return an err here too

commodo: sorry;  `count < sizeof(buf) - 1`

drivers/adxl372/adxl372.c
```diff
@@ -772,12 +808,14 @@ int32_t adxl372_init(adxl372_dev **device,
 
 	*device = dev;
```
commodo: maybe move this under `if (!ret)`  ; otherwise you coudl return with a pointer to something that was free'd

drivers/adxl372/adxl372_spi.c
```diff
@@ -0,0 +1,123 @@
...
+				      uint8_t reg_addr,
+				      uint8_t *reg_data,
+				      uint16_t count)
+{
+	uint8_t buf[513];
+	int32_t ret;
+
+	if (count > 512) {
+		count = 512;
+		printf("warning: multiple read has been limited to 512 bytes");
```
commodo: why not return err ?

stefpopa: The maximum number of bytes you would ever want to read is 512, as this is the FIFO size. If the user will try to read more than this, the sw assumes that he wants to read the maximum number. Therefore a warning is printed instead of returning an error. However, this is just a preference.

PR #94, opened 2018-09-07
Fix issues after running static analysis: part 4

amiclaus:
Fourth set of fixed issues after running cppcheck tool.

### DIFF
--- a/adum7701_fmc/clkgen_core.c
+++ b/adum7701_fmc/clkgen_core.c
@@ -404,22 +404,18 @@ uint32_t clkgen_get_rate(clkgen_device *dev,
 int32_t clkgen_setup(clkgen_device **clkgen_dev,
 		clkgen_init_params init_params)
 {
-	clkgen_device	*adc_clkgen;
+	clkgen_device	adc_clkgen;
 	uint32_t	reg_val;
 	int32_t		status = 0;
 
-	adc_clkgen = (clkgen_device *)malloc(sizeof(*adc_clkgen));
-        if (!adc_clkgen)
-                return -1;
-
-	adc_clkgen->base_addr = init_params.base_addr;
-	clkgen_set_rate(adc_clkgen,
+	adc_clkgen.base_addr = init_params.base_addr;
+	clkgen_set_rate(&adc_clkgen,
 			init_params.rate,
 			init_params.parent_rate);
 
 	mdelay(1);
 
-	clkgen_read(adc_clkgen, CLKGEN_REG_STATUS, &reg_val);
+	clkgen_read(&adc_clkgen, CLKGEN_REG_STATUS, &reg_val);
 	if ((reg_val & CLKGEN_STATUS) == 0x0) {
 		printf("clkgen_setup failed");
 		status--;
--- a/common_drivers/platform_drivers/platform_drivers.c
+++ b/common_drivers/platform_drivers/platform_drivers.c
@@ -196,6 +196,7 @@ int32_t spi_init(spi_desc **desc,
 		break;
 #endif
 	default:
+		free(dev);
 		return FAILURE;
 	}
 
@@ -207,10 +208,12 @@ int32_t spi_init(spi_desc **desc,
 	m_spi_config = XSpiPs_LookupConfig(dev->device_id);
 
 	if (m_spi_config == NULL) {
+		free(dev);
 		return FAILURE;
 	}
 
 	if (XSpiPs_CfgInitialize(&m_spi, m_spi_config, m_spi_config->BaseAddress) != 0) {
+		free(dev)
 		return FAILURE;
 	}
 	#endif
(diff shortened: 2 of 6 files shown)


drivers/adf4350/adf4350.c
```diff
@@ -248,7 +248,7 @@ int64_t adf4350_set_freq(adf4350_dev *dev,
 		(dev->pdata->r2_user_settings & (ADF4350_REG2_PD_POLARITY_POS |
 		ADF4350_REG2_LDP_6ns | ADF4350_REG2_LDF_INT_N |
 		ADF4350_REG2_CHARGE_PUMP_CURR_uA(5000) |
-		ADF4350_REG2_MUXOUT(0x7) | ADF4350_REG2_NOISE_MODE(0x9)));
+		ADF4350_REG2_MUXOUT(0x7) | ADF4350_REG2_NOISE_MODE(0b101)));
```
dbogdan: ADF4350_REG2_NOISE_MODE has a length of 2 bits, so 0b101 value is incorrect. I would also specify this in its definition (e.g., (((x) & 0x3) << 29)).
In this place, ADF4350_REG2_NOISE_MODE should be set 0x3 to enable the low spur mode.

amiclaus:
Changelog: v1->v2:
* added "Signed-off by:" to each commit.
* added description to each commit.
* modified macro definition for ADF4350_REG2_NOISE_MODE and set value to enable low spur mode.

adum7701_fmc/clkgen_core.c
```diff
@@ -425,5 +425,7 @@ int32_t clkgen_setup(clkgen_device **clkgen_dev,
 		status--;
 	}
 
+	free (adc_clkgen);
```
commodo: I think in this case you could probably change the type of the `adc_clkgen` variable;
to `clkgen_device adc_clkgen;`

that way it's automatically allocated on & free'd from the stack

commodo: this can be done here, also because you are allocating a fixed size memory block, so malloc() doesn't add much value

common_drivers/platform_drivers/platform_drivers.c
```diff
@@ -196,6 +196,7 @@ int32_t spi_init(spi_desc **desc,
 		break;
 #endif
 	default:
+		free(dev);
```
commodo: this is good;
but i think you may also need to add the` free()` at lines 210-211 & 214-215 below

common_drivers/platform_drivers/platform_drivers.c
```diff
@@ -315,6 +316,7 @@ int32_t gpio_get(gpio_desc **desc,
 		return FAILURE;
 
 	if (gpio_number < 32) {
+		free(dev);
 		return FAILURE;
 	}
```
commodo: in this case, if you move this code block before the malloc() you don't need to do the free()

drivers/linux_platform/platform_drivers.c
```diff
@@ -192,20 +193,23 @@ int32_t spi_init(spi_desc **desc, const spi_init_param *param)
...
 		printf("%s: Can't open device\n\r", __func__);
+		free(descriptor);
 		return FAILURE;
 	}
 
 	ret = ioctl(descriptor->fd, SPI_IOC_WR_MODE,
 			&param->mode);
 	if (ret == -1) {
 		printf("%s: Can't set mode\n\r", __func__);
+		free(descriptor);
```
commodo: i think here you may also need to close the `fd` before free-ing the descriptor
so,
```
close(descriptor->fd);
free(descriptor);
```
similarly, this should be done at lines 212 & 300

amiclaus:
Changelog v2->v3:
 * remove unnecessary memory deallocation by moving 'if' statement.
 * close 'fd 'before the memory is freed

adum7701_fmc/clkgen_core.c
```diff
@@ -404,14 +404,10 @@ uint32_t clkgen_get_rate(clkgen_device *dev,
...
-	clkgen_device	*adc_clkgen;
+	clkgen_device	adc_clkgen;
 	uint32_t	reg_val;
 	int32_t		status = 0;
 
-	adc_clkgen = (clkgen_device *)malloc(sizeof(*adc_clkgen));
-        if (!adc_clkgen)
-                return -1;
-
 	adc_clkgen->base_addr = init_params.base_addr;
```
commodo: if you change the type here, then this also changes to `adc_clkgen.base_addr = init_params.base_addr;`
and you need to call it with `clkgen_set_rate(&adc_clkgen,` and `clkgen_read(&adc_clkgen,`

commodo:
rest looks good

amiclaus:
Changelog v3->v4:
 * fix the way `adc_clkgen` structure is accessed after changing the variable type.

PR #95, opened 2018-09-07
Fix issues after running static analysis: part 5

amiclaus:
Fifth set of fixed issues after running cppcheck tool.

### DIFF
--- a/common_drivers/fru_tools/fru_tools.c
+++ b/common_drivers/fru_tools/fru_tools.c
@@ -610,6 +610,7 @@ static unsigned int insert_str(unsigned char *buf, const unsigned char * str, bo
 			}
 			buf[0] = tmp | (FRU_STRING_ASCII << 6);
 			memcpy(&buf[1], &str[1], tmp);
+			free(six);
 			return tmp + 1;
 		}
 		if (tmp > 0x3F) {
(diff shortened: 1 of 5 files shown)


amiclaus:
Changelog: v1->v2:
* added "Signed-off by:" to each commit.
* added description to each commit.

fmcadc5/fmcadc5.c
```diff
@@ -198,6 +198,7 @@ int main(void)
 	i5g_init_param.ad9625_cs_0 = 1;
 	i5g_init_param.ad9625_cs_1 = 2;
 	i5g_init_param.regs = XPAR_AXI_FMCADC5_SYNC_BASEADDR;
+	i5g_init_param.sysref_delay = 0;
```
commodo: this is good;
another way to do it, is to do a `memset(&i5g_init_param, 0, sizeof(i5g_init_param))` right before these assignments ; 
this typically makes sure that all memory is 0, except for the parameters which are specifically initialized;

common_drivers/fru_tools/fru_tools.c
```diff
@@ -610,6 +610,7 @@ static unsigned int insert_str(unsigned char *buf, const unsigned char * str, bo
 			}
 			buf[0] = tmp | (FRU_STRING_ASCII << 6);
 			memcpy(&buf[1], &str[1], tmp);
+			free(six);
 			return tmp + 1;
 		}
```
commodo: i think we may also need to free this after line 615, to make sure we don't leak in case `tmp >= 0`

commodo: my bad;
i didn't see it at line 623

fmcadc5/fmcadc5.c
commodo: ack: disregard my comment :)

commodo:
LGTM

commodo:
Will merge later today if no objections

PR #96, opened 2018-09-10
AD5755 extreme case fix.

adrimbarean:
Values lower than negative full scale caused the output to swing to positive full scale instead of clamping to negative full scale.

The return value was not calculated using the value written to the data register.


commodo:
[LGTM](http://78.media.tumblr.com/e9123561b66d76ceae3ee6c9ff16a532/tumblr_oay3lbssM91tyncywo1_400.gif)

PR #99, opened 2018-09-18
Fix AD9361 build warnings 

amiclaus:
* Add 'Werror' flag in AD9361 makefiles
 * Fix raised warnings for AD9361 builds


ad9361/sw/ad9361_api.c
```diff
@@ -601,7 +601,7 @@ int32_t ad9361_get_en_state_machine_mode (struct ad9361_rf_phy *phy,
 int32_t ad9361_set_rx_rf_gain (struct ad9361_rf_phy *phy,
 							   uint8_t ch, int32_t gain_db)
 {
-	struct rf_rx_gain rx_gain = {0};
+	struct rf_rx_gain rx_gain = {0, 0, 0, 0, 0, 0, 0, 0, 0};
```
commodo: hmm ; which compiler gives this error ?
typically, `rx_gain = {0};` & `rx_gain = {0, 0, 0, 0, 0, 0, 0, 0, 0};` should be equivalent ; I do know that for GCC this is true

amiclaus: Clang compiler for linux build : https://travis-ci.org/amiclaus/no-OS/jobs/429935069#L664

commodo: thanks for the info ;
i suspected this behavior may be only GCC specific ;

ad9361/sw/ad9361_conv.c
```diff
@@ -512,7 +512,7 @@ int32_t ad9361_dig_tune(struct ad9361_rf_phy *phy, uint32_t max_freq,
 		return -ENODEV;
 
 	dev_dbg(&phy->spi->dev, "%s: freq %lu flags 0x%X\n", __func__,
```
commodo: you could change the printf specifier to `%u` ; you don't really need the case

ad9361/sw/ad9361.c
```diff
@@ -4346,12 +4350,12 @@ static int32_t ad9361_validate_trx_clock_chain(struct ad9361_rf_phy *phy,
 	}
 
 	for (i = 1; i <= 3; i++) {
-		if (abs(rx_path_clks[ADC_FREQ] / i - data_clk) < 4)
+		if (labs((int64_t)(rx_path_clks[ADC_FREQ] / (i - (int64_t)data_clk))) < 4)
```
commodo: i am starting to think here that maybe it's simpler/easier to just convert both `i` & `data_clk` to int64_t;
it would simplify the nature of these operations and remove some of the explicit casts;

amiclaus:
Changelog v1->v2: 
 * remove explicit cast and change printf specifier
 * add macro to perform absolute value on subtraction between unsigned types

PR #104, opened 2018-10-02
Implementation of AD5683 Driver.

amiclaus:
No-OS Driver compatible with AD5683R/AD5682R/AD5681R/AD5693R/AD5692R/AD5691R.

### DIFF
--- a/drivers/ad5683/ad5683.c
+++ b/drivers/ad5683/ad5683.c
@@ -1,0 +1,334 @@
+/***************************************************************************//**
+ *   @file   ad5683.c
+ *   @brief  Implementation of AD5683 Driver.
+ *   @author Antoniu Miclaus (antoniu.miclaus@analog.com)
+********************************************************************************
+ * Copyright 2018(c) Analog Devices, Inc.
+ *
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *  - Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ *  - Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in
+ *    the documentation and/or other materials provided with the
+ *    distribution.
+ *  - Neither the name of Analog Devices, Inc. nor the names of its
+ *    contributors may be used to endorse or promote products derived
+ *    from this software without specific prior written permission.
+ *  - The use of this software may or may not infringe the patent rights
+ *    of one or more patent holders.  This license does not release you
+ *    from the requirement that you obtain separate licenses from these
+ *    patent holders to use this software.
+ *  - Use of the software either in source or binary form, must be run
+ *    on or directly connected to an Analog Devices Inc. component.
+ *
+ * THIS SOFTWARE IS PROVIDED BY ANALOG DEVICES "AS IS" AND ANY EXPRESS OR
+ * IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, NON-INFRINGEMENT,
+ * MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED.
+ * IN NO EVENT SHALL ANALOG DEVICES BE LIABLE FOR ANY DIRECT, INDIRECT,
+ * INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, INTELLECTUAL PROPERTY RIGHTS, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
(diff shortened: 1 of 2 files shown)


drivers/ad5683/ad5683.c
```diff
@@ -0,0 +1,334 @@
...
+ *                         0 - SPI peripheral was initialized and the
+ *                             device is present.
+*/
+int32_t ad5683_init(struct ad5683_dev **device,
+		    struct ad5683_init_param init_param)
+{
+	struct ad5683_dev *dev;
+	int32_t ret;
+
+	dev = (struct ad5683_dev *)malloc(sizeof(*dev));
```
commodo: would `calloc()`  be a better idea here ?
calloc() guarantees that the memory s initialized to 0; malloc() doesn't
i agree that there are many places where malloc() is used (though calloc()) would be better suited ;

drivers/ad5683/ad5683.h
```diff
@@ -0,0 +1,196 @@
...
+
+#define MAX_RESOLUTION             16 // Maximum resolution of the supported devices
+
+/* Power-down operation modes masks */
+#define AD5683_PWRM_NORMAL         0	// Power Mode Normal
+#define AD5683_PWRM_1K             1	// Power-down mode 1kOhm to GND
+#define AD5683_PWRM_100K           2	// Power-down mode 100kOhm to GND
+#define AD5683_PWRM_THREESTATE     3	// Three-State
+
+#define AD5683_DC_EN               1
```
commodo: as a practice : for `<x>_EN` and `<x>_DIS` value pairs, you could just define the behavior of `0 = disable ; non-zero = enable`

stefpopa:
To me, it looks that the support for these devices can be added to the already existing ad5686 driver, as they are very similar, the only obvious difference being the number of bits. This is the way it was done for the Linux driver.

drivers/ad5683/ad5683.c
```diff
@@ -0,0 +1,334 @@
...
+	return ad5683_set_shift_reg(dev, AD5683_CMD_WR_CTRL_REG,
+				    AD5683_CTRL_INT_REF(value));
+}
+
+/**
+ * @brief Set DCEN (daisy-chain enable)
+ *
+ * @param dev   - The device structure.
+ * @param value - Enable or disable daisy-chain mode
+ *                Example : 'AD5683_DC_EN' - daisy-chain enable
```
commodo: related to commit https://github.com/analogdevicesinc/no-OS/pull/104/commits/d1914dad852f657681586dd757949bae0c63a60e#r221907052  you could just do  `uint8_t enable`, where 0 disables and non-zero enables

drivers/ad5683/ad5683.c
```diff
@@ -0,0 +1,334 @@
...
+ *
+ * @return read_back_data - value read from register.
+*/
+uint32_t ad5683_read_back_register(struct ad5683_dev *dev)
+{
+
+	uint32_t read_back_data = 0;
+	uint32_t offset = MAX_RESOLUTION - chip_info[dev->act_device].resolution;
+
+	if(chip_info[dev->act_device].communication == SPI) {
```
commodo: maybe check if `dev` is NULL before doing anything here with it

drivers/ad5683/ad5683.c
amiclaus: The structure is initialized afterwards in the same function with init_params. Does it make sense to initialize it twice? Moreover, I think `malloc` is a bit faster that `calloc`.

drivers/ad5683/ad5683.c
```diff
@@ -0,0 +1,334 @@
...
+ * 		       parameters.
+ *
+ * @return ret - The result of the initialization procedure.
+ *               Example: -1 - SPI peripheral was not initialized or the
+ *                             device is not present.
+ *                         0 - SPI peripheral was initialized and the
+ *                             device is present.
+*/
+int32_t ad5683_init(struct ad5683_dev **device,
+		    struct ad5683_init_param init_param)
```
commodo: pass this as pointer ?

drivers/ad5683/ad5683.c
commodo: agreed; malloc() is faster ; but calloc() is less error prone ;
i admit i haven't looked at the init of the device, but, if it's fully initialized, and no field member is omitted, then malloc() is fine

commodo:
i would 2nd Stefan's comment:  https://github.com/analogdevicesinc/no-OS/pull/104#issuecomment-426232929
i didn't check to see if there is an existing driver already ;

but it would make sense to extend that if possible

amiclaus:
> To me, it looks that the support for these devices can be added to the already existing ad5686 driver, as they are very similar, the only obvious difference being the number of bits. This is the way it was done for the Linux driver.

The input shift register content is pretty much different from the one in the AD5686. The AD5683 has a Write control register that is accessed via a single command and contains multiple functionalities, while in the AD5686 driver each functionality is accessed via command bits.

commodo:
register addresses can be made configurable per device type ;

we could have a `ad5686_init()` and `ad5683_init()` functions in the same driver, if the drivers differ only by a few elements, but most of the functionality is similar

amiclaus:
Creating new PR with support for these devices in the AD5686 driver.

PR #106, opened 2018-10-18
Add initial ADRV9009 support

dbogdan:


commodo:
There are some pertinent comments from Codacy.
Most of them are trivial/simple [I'm referring to `Variable 'halError' is reassigned a value before the old one has been used.` ]

The more serious one seems to be `Either the condition 'dout==0' is redundant or there is division by zero at line 404.`

For the astyle stuff, we'd have to find a different approach altogether, so we can ignore them for now.
Will talk to @amiclaus about it and implement it.
The thing is, that different versions of astyle produce different output, which makes it hard to have a consistent/reproduce-able build [from local to Travis-CI].

Let's fix the more serious report for now, and adapt our plan/CI moving forward.

projects/adrv9009/src/devices/adi_hal/LICENSE.txt
```diff
@@ -0,0 +1,102 @@
...
+			 UNDER ANY LEGAL THEORY (INCLUDING WITHOUT LIMITATION CONTRACT, WARRANTY, TORT,
+					 NEGLIGENCE, STRICT OR PRODUCT LIABILITY),
+			 EVEN IF IT HAS BEEN INFORMED OF THE POSSIBILITY OF SUCH DAMAGES. Some
+			 jurisdictions do not permit the exclusion
+				 or limitation of liability for consequential, incidental or other damages,
+				 and, as such, some portion of the above limitation may not apply. In such
+				 jurisdictions, liability is limited to the greatest extent permitted by law.
+
+		 Third Party Software:
+				 The software may be accompanied by or include software made available by one
```
commodo: I think something here is misformatted.

projects/adrv9009/src/devices/talise/LICENSE.txt
```diff
@@ -0,0 +1,102 @@
...
+					 and conditions (“Third Party Licenses”). The Third Party Licenses for Third
+					 Party Software delivered with the software are set forth or identified (by url
+							 or otherwise) in (i) Appendix A to this license (if any),
+									 (ii) the applicable software header
+									 or footer text, (iii) a text file located in the directory of the applicable
+									 Third Party Software component and/
+									 or (iv) such other location customarily used for licensing terms. The use of
+									 each portion of Third Party Software is subject to the Third Party Licenses,
+									      and you agree that your use of any Third Party Software is bound by the
+									      applicable Third Party License. You agree to review
```
commodo: this also looks mis-formatted

dbogdan:
I've fixed the issues reported by Codacy and updated the license files.

PR #107, opened 2018-10-22
Ultrascale xcvr projects update

adrimbarean:
After the xcvr_core updates some structure members changed names making projects using xcvr_core incorrect. Adjust new names, apply AStyle on the sources and tow minor bugs: one introduced with the xcvr_core update and one missing semicolon.

### DIFF
--- a/ad-fmcjesdadc1-ebz/ad_fmcjesdadc1_ebz.c
+++ b/ad-fmcjesdadc1-ebz/ad_fmcjesdadc1_ebz.c
@@ -134,7 +134,7 @@ int main(void)
 
 	ad9250_xcvr.reconfig_bypass = 0;
 	ad9250_xcvr.lane_rate_kbps = ad9250_0_param.lane_rate_kbps;
-	ad9250_xcvr.ref_clock_khz = 245760;
+	ad9250_xcvr.ref_rate_khz = 245760;
 
 	ad9250_jesd204.scramble_enable = 1;
 	ad9250_jesd204.octets_per_frame = 1;
--- a/fmcadc4/fmcadc4.c
+++ b/fmcadc4/fmcadc4.c
@@ -213,7 +213,7 @@ int main(void)
 	ad9680_1_core.no_of_channels = 2;
 	ad9680_1_core.resolution = 14;
 
-    // receiver DMA configuration
+	// receiver DMA configuration
 
 #ifdef ZYNQ
 	rx_xfer.start_address = XPAR_DDR_MEM_BASEADDR + 0x800000;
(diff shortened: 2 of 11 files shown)


adrimbarean:
V0 -> V1:
- run AStyle on platform_drivers.c and platform_drivers.h

fmcdaq2/fmcdaq2.c
```diff
@@ -572,7 +572,7 @@ int main(void)
 	xcvr_setup(&ad9680_xcvr);
 #endif
 #ifdef XILINX
-	if (ad9144_xcvr.dev.qpll_enable) {	// DAC_XCVR controls the QPLL reset
+	if (!ad9144_xcvr.dev.cpll_enable) {	// DAC_XCVR controls the QPLL reset
```
commodo: this looks like a bug-fix, or ?

fmcadc4/fmcadc4.c
```diff
@@ -238,9 +238,9 @@ int main(void)
 
 	xcvr_getconfig(&ad9680_xcvr);
 	ad9680_xcvr.reconfig_bypass = 0;
-	ad9680_xcvr.ref_clock_khz = 500000;
-	ad9680_xcvr.lane_rate_kbps = ad9680_0_param.lane_rate_kbps;
-	ad9680_xcvr.dev.qpll_enable = 1;
+	ad9680_xcvr.ref_rate_khz = 500000;
+	ad9680_xcvr.lane_rate_khz = ad9680_0_param.lane_rate_kbps;
```
commodo: this seems a bit weird now  ; i.e: assiging rate_kbps to rate_khz ; 
maybe also update that field as well ?

ad-fmcjesdadc1-ebz/ad_fmcjesdadc1_ebz.c
```diff
@@ -133,8 +133,8 @@ int main(void)
 	xcvr_getconfig(&ad9250_xcvr);
 
 	ad9250_xcvr.reconfig_bypass = 0;
-	ad9250_xcvr.lane_rate_kbps = ad9250_0_param.lane_rate_kbps;
-	ad9250_xcvr.ref_clock_khz = 245760;
+	ad9250_xcvr.lane_rate_khz = ad9250_0_param.lane_rate_kbps;
```
commodo: this also would need some semantic update; rate_khz  <= rate_kbps

adrimbarean:
V1 -> V2:
- change lane_rate_khz back to lane_rate_kbps
- separate name-change commits from logic changing commmits

adrimbarean:
V2 -> V3:
- update fmcdaq3 project to use the new xcvr_core driver
- change #endif-#ifdef pair with #elif

fmcdaq3/fmcdaq3.c
```diff
@@ -129,18 +129,15 @@ int main(void)
 	ad9528_spi_param.type = ZYNQ_PS7_SPI;
 	ad9152_spi_param.type = ZYNQ_PS7_SPI;
 	ad9680_spi_param.type = ZYNQ_PS7_SPI;
-#endif
```
commodo: actually this change is not quite without any functional changes.
there is a semantic difference between   `#if` and `#ifdef` (which is short for `#if defined()`)
in many cases you don't notice the difference, because typically people do `#define ZYNQ_PS7   1` ; this evaluates true for `#if` and `#ifdef`, while `#define ZYNQ_PS7`  would evaluate true only for `#ifdef ZYNQ_PS7`

this PR is pretty big as is;
i would recommend we defer this change for another PR/changeset; and we can review it there

adrimbarean:
V3 -> V4:
- remove #ifdef commit

commodo:

[LGTM](https://i.pinimg.com/564x/d3/ee/a1/d3eea1f40b83eab128cb64c82582992c.jpg)

PR #108, opened 2018-11-01
drivers: ad5592r: Fix read value reference

mirceacaprioru:
Variable in which to read the register value should be passed by reference.


commodo:
that depends;
different versions of astyle produce different output ; which is annoying ;

we'll have to update our build to use a single fixed version of astyle so that output is consistent
@amiclaus will take care of this

commodo:
[LGTM](https://i.pinimg.com/564x/d5/14/9e/d5149e7d0ced07a3b633a21b4a7e066e.jpg)

PR #109, opened 2018-11-02
general: use fixed astyle version on single build

amiclaus:
Coding style is checked now on a single travis build, using a fixed
version of AStyle. This commit solves the issue of getting different
coding style outputs on different builds.


ci/travis/before_install_linux
```diff
@@ -2,4 +2,14 @@
...
-sudo apt-get install libmatio-dev cppcheck astyle
+sudo apt-get install libmatio-dev cppcheck 
+
+if [ -n "$ASTYLE_FORMAT"]
+then
+    wget "https://sourceforge.net/projects/astyle/files/astyle/astyle 3.1/astyle_3.1_linux.tar.gz"
+    tar -xzf astyle_3.1_linux.tar.gz
+    cd ./astyle/build/gcc
+    make
+    sudo make install
```
commodo: i wouldn't install it on the system;
maybe just append to the PATH env var, the path to the `astyle` executable 
that way, this can be run locally

.travis.yml
```diff
@@ -17,4 +21,4 @@ before_install:
 script:
   - ./ci/scripts/build_projects.sh
   - ./ci/scripts/cppcheck.sh
-  - ./ci/scripts/astyle.sh
+  - if [[ -n "$ASTYLE_FORMAT" ]]; then ./ci/scripts/astyle.sh; fi
```
commodo: you could move this check inside the `./ci/scripts/astyle.sh` script

ci/scripts/astyle.sh
```diff
@@ -24,6 +28,20 @@ is_source_file() {
 	return 1
 }
 
+if [ ! -d build/astyle/build/gcc ]
+then	
+	mkdir -p build
+	pushd build
+	wget "https://sourceforge.net/projects/astyle/files/astyle/astyle 3.1/astyle_3.1_linux.tar.gz"
+    tar -xzf astyle_3.1_linux.tar.gz
```
commodo: indentation is inconsistent here

commodo:
you're also committing a binary file: astyle_3.1_linux.tar.gz
please remove it

.travis.yml
```diff
@@ -2,19 +2,24 @@ language: c
...
+    - os: linux
+      compiler: gcc
+    - os: linux
+      compiler: clang
+    - os: osx
+      compiler: gcc
+    - os: osx
+      compiler: clang
+    - os: linux
+      env: ASTYLE_FORMAT=true
```
commodo: maybe move this build to be the first in the list
it seems Travis-CI takes them in top-to-bottom order, and it would be neat if this runs first

amiclaus:
Changelog v1 -> v2:

- remove astyle system installation
- create separate build job for code style check

.travis.yml
commodo: I guess in this case, a wrapper script would look nicer.
One that calls all 3 scripts based on the variable.

One example is:
* https://github.com/analogdevicesinc/linux/blob/master/.travis.yml#L13
* https://github.com/analogdevicesinc/linux/blob/master/ci/travis/run-build.sh#L33

amiclaus:
changelog v2->v3:
- add wrapper script

ci/scripts/run_build.sh
```diff
@@ -0,0 +1,14 @@
...
+
+build_default() {
+    ./ci/scripts/build_projects.sh
+    ./ci/scripts/cppcheck.sh
+}
+
+build_astyle() {
+    ./ci/scripts/astyle.sh
+}
+
```
commodo: you could do a `BUILD_TYPE=${BUILD_TYPE:-default}` which means that the default build will be used if `BUILD_TYPE` is unspecified ; that way, you can remove `env: BUILD_TYPE=default` from `.travis.yml`

ci/scripts/run_build.sh
```diff
@@ -0,0 +1,14 @@
...
+set -e
+
+build_default() {
+    ./ci/scripts/build_projects.sh
+    ./ci/scripts/cppcheck.sh
+}
+
+build_astyle() {
+    ./ci/scripts/astyle.sh
```
commodo: this is fine as-is; 
but only as a note, you could also just include the scripts with `. ./ci/scripts/astyle.sh` or `source ./ci/scripts/astyle.sh`  [ the latter only works on `bash` ]

the difference is that
* `./ci/scripts/astyle.sh` will run the script as a separate executable
* `. ./ci/scripts/astyle.sh` will include the script code into this script and run it as part of the same script

commodo:
[LGTM](https://i.pinimg.com/564x/52/96/75/52967586c13318cd8274f0e4ef6b9ca6.jpg)

PR #120, opened 2018-11-29
drivers: ad5686: Add support for AD5674R/AD5679R

mirceacaprioru:
AD5674R/AD5679R are low power, 16-channel, 12-/16-bit buffered voltage
output digital-to-analog converters (DACs). They include a 2.5 V internal
reference.

They are very similar to the AD5686 devices aside from having 16 channels
for output voltage.


commodo:
[LGTM](https://i.pinimg.com/originals/49/97/fb/4997fbfd8edcf54953d72f1a08acfa3a.gif)

drivers/ad5686/ad5686.c
```diff
@@ -474,6 +545,8 @@ void ad5686_power_mode(struct ad5686_dev *dev,
 	uint8_t address = chip_info[dev->act_device].channel_addr[channel];
 
 	if(chip_info[dev->act_device].register_map == AD5686_REG_MAP) {
+		if (channel > AD5686_CH_7)
```
stefpopa: A comment here would be useful. Rest looks good.

commodo:

mirceacaprioru:
Changelog V1 -> V2
- added descriptive comment for channel subtraction

drivers/ad5686/ad5686.c
```diff
@@ -474,6 +545,9 @@ void ad5686_power_mode(struct ad5686_dev *dev,
 	uint8_t address = chip_info[dev->act_device].channel_addr[channel];
 
 	if(chip_info[dev->act_device].register_map == AD5686_REG_MAP) {
+		/* AD5674R/AD5679R have 16 channels and 2 powerdown registers */
```
amiclaus: Might be worth mentioning in the header comment that "100k to GND" Power-Down mode(`AD5686_PWRM_100K`) is not available for AD5674R/AD5679R.

mirceacaprioru: Ack

PR #124, opened 2019-01-07
2016 r2

chengye123:
thanks


commodo:
We don't typically merge old releases into master.
master should have all code from release branches.

PR #125, opened 2019-01-09
travis: rework

amiclaus:
The PR contains the following modifications:

- Move all travis scripts under 'ci/travis'
- Add script that downloads the common scripts that are shared between multiple repositories
- Make 'astyle_config' file path customizable
- Option to extend cppcheck command properties


ci/travis/astyle.sh
```diff
@@ -43,7 +43,7 @@ fi
 for file in $FILES; do
 	if is_source_file $file
 	then 
-		./build/astyle/build/gcc/bin/astyle --options=./ci/scripts/astyle_config $file
+		./build/astyle/build/gcc/bin/astyle --options=$ASTYLE_CONFIG_PATH $file
```
commodo: why not use `"$(get_script_path astyle_config)` here ?

amiclaus: Because currently 'lib.sh' is used only by other repos that require same scripts. 

Currently seems pretty redundant to use it in this repo, but I can add it for the sake of similarity.

amiclaus:
Changelog v1->v2:
- use `get_script_path` to get config file path

amiclaus:
Changelog v2-v3:

- improve astyle script to work if file paths contain spaces

commodo:

PR #129, opened 2019-01-23
add astyle ignore

amiclaus:
Similar to .gitignore file, the user can now specifiy files that astyle
ignores through .astyleignore file found in the root folder

### DIFF
--- a/ci/travis/astyle.sh
+++ b/ci/travis/astyle.sh
@@ -19,6 +19,16 @@ then
 	fi
 fi
 
+is_valid_file(){
+	[[ -f ".astyleignore" ]] &&
+	while read -r fpath
+	do
+		[[ -z "$fpath" ]] && continue
+		[[ "$file" == *"$fpath"* ]] && return 1
+	done < ".astyleignore"
+	return 0
+}
+
 is_source_file() {
 	local file="$1"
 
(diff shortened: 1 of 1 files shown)


ci/travis/astyle.sh
```diff
@@ -19,6 +21,13 @@ then
 	fi
 fi
 
+is_valid_file(){
+        for aign in $ASTYLE_IGNORE; do
```
commodo: does this work for paths that include spaces ?

ci/travis/astyle.sh
```diff
@@ -19,6 +21,13 @@ then
 	fi
 fi
 
+is_valid_file(){
```
commodo: maybe let's name this function `is_ignored`

ci/travis/astyle.sh
amiclaus: Not always. I'll add this feature.

ci/travis/astyle.sh
amiclaus: This won't make sense when interpreting line:
`if is_source_file "$file" && is_ignored "$file"`

Natural language: if the current file is a source file and is ignored then run astyle

commodo: it would be in the sense of `if is_source_file "$file" && ! is_ignored "$file"`
but no strong opinion; we can leave it like this for now

it could also be implemented something like where everything is moved under `is_valid_file` and `is_valid_file` could check if it is ignored, a source file, etc; it would also be extendable;

amiclaus: Yep, that's how I initially did it: check both if ignored and if it s source file within same function i.e. `is_source_file`. I can implement it like that.

commodo: let's leave it like this for now

commodo:
[LGTM](https://i.pinimg.com/originals/a0/55/a1/a055a1366346a941e0f5640d94ec711a.jpg)

PR #130, opened 2019-01-23
ad7606: Add basic support for AD7606 family of devices

stefpopa:
The AD7606/AD7606-6/AD7606-4 is a 16-bit 8/6/4 channel Analog to
Digital Converter (ADC). It can accommodate ± 10V and ±5V true bipolar
inputs. The AD7605-4 is part of the same family, the main difference is
that it does not support oversampling.

This patch adds basic serial support for these devices.

Datasheet:
Link: https://www.analog.com/media/en/technical-documentation/data-sheets/ad7606_7606-6_7606-4.pdf
Link: https://www.analog.com/media/en/technical-documentation/data-sheets/ad7605-4.pdf

### DIFF
--- a/drivers/ad7606/ad7606.c
+++ b/drivers/ad7606/ad7606.c
@@ -1,0 +1,220 @@
+/***************************************************************************//**
+ *   @file   ad7606.c
+ *   @brief  Implementation of ad7606 Driver.
+ *   @author Stefan Popa (stefan.popa@analog.com)
+********************************************************************************
+ * Copyright 2019(c) Analog Devices, Inc.
+ *
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *  - Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ *  - Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in
+ *    the documentation and/or other materials provided with the
+ *    distribution.
+ *  - Neither the name of Analog Devices, Inc. nor the names of its
+ *    contributors may be used to endorse or promote products derived
+ *    from this software without specific prior written permission.
+ *  - The use of this software may or may not infringe the patent rights
+ *    of one or more patent holders.  This license does not release you
+ *    from the requirement that you obtain separate licenses from these
+ *    patent holders to use this software.
+ *  - Use of the software either in source or binary form, must be run
+ *    on or directly connected to an Analog Devices Inc. component.
+ *
+ * THIS SOFTWARE IS PROVIDED BY ANALOG DEVICES "AS IS" AND ANY EXPRESS OR
+ * IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, NON-INFRINGEMENT,
+ * MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED.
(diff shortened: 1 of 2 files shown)


drivers/ad7606/ad7606.h
```diff
@@ -0,0 +1,111 @@
...
+ * MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED.
+ * IN NO EVENT SHALL ANALOG DEVICES BE LIABLE FOR ANY DIRECT, INDIRECT,
+ * INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, INTELLECTUAL PROPERTY RIGHTS, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+*******************************************************************************/
+enum ad7606_supported_device_ids {
```
commodo: curios: do we usually provide header guards for no-OS drivers ?

drivers/ad7606/ad7606.c
```diff
@@ -0,0 +1,220 @@
...
+	return gpio_set_value(dev->gpio_os2, ((osr & 0x04) >> 2));
+}
+
+int32_t ad7606_init(struct ad7606_dev **device,
+		    struct ad7606_init_param *init_param)
+{
+	struct ad7606_dev *dev;
+	int32_t ret;
+
+	dev = (struct ad7606_dev *)malloc(sizeof(*dev));
```
commodo: most drivers use `malloc()` in no-OS;
i'd argue that we could use `calloc()` to make sure that memory is 0

drivers/ad7606/ad7606.c
```diff
@@ -0,0 +1,220 @@
...
+	printf("ad7606 successfully initialized\n");
+
+	return 0;
+error:
+	printf("ad7606 initialization failed\n");
+	ad7606_remove(dev);
+	return ret;
+}
+
+int32_t ad7606_remove(struct ad7606_dev *dev)
```
commodo: the naming here would need a bit of work
something like `ad7606_destroy()` or `ad7606_uninit()` ; the `remove` term is more Linux-y

drivers/ad7606/ad7606.h
stefpopa: Yes, we do. I forgot to add them. Thanks!

drivers/ad7606/ad7606.c
stefpopa: fine with me

drivers/ad7606/ad7606.c
stefpopa: This just follows the convention that we agreed upon in `platform_drivers`. i.e. `gpio_remove()`, `spi_remove()`

commodo: ack
let's leave it like this

commodo:
[LGTM](https://i.pinimg.com/originals/31/57/83/315783114b157774ab559d65b0320a10.png)

PR #135, opened 2019-01-29
drivers: ad411x: rename ad4111_regs.h file to ad411x_regs.h

adrimbarean:
The AD4111 and AD4112 devices have the register structure identical,
with the exception of one bit in the GPIOCON register. Rename the
register file to include the whole family not just AD4111.


commodo:
I'm vague here.
Why rename the file to a non-standard filename ?

adrimbarean:
I was thinking the file is intended to be used with both AD4111 and AD 4112 and to not create confiusion because it is named only ad4111_regs.h

amiclaus:
Retrigger PR

commodo:
[LGTM](https://i.pinimg.com/564x/6d/7e/cb/6d7ecb7b20cd01db5e78f16281703021.jpg)

commodo:
astyle has some complaints

PR #136, opened 2019-01-29
Ad7768 1 rework

stefpopa:
- The main purpose of this PR is to rework the ad7768-1 driver in order to make it generic. The current driver implementation is dependent to the Xilinx platform as it uses specific spi_engine functions. However, the driver should also be able to interface with a microcontroller using a classic SPI.

- Secondly, the driver was updated to work with the latest version of spi_engine. 

- The remaining patches are just miscellaneous cleanups that make the driver more readable.


stefpopa:
changes in v2:
* when CRC is enabled, the CRC result is placed on the forth byte of the read buffer (`buf[3]`, **NOT** `buf[4`])

ad7768-1fmcz/spi_engine.c
```diff
@@ -439,8 +496,8 @@ int32_t spi_eng_transfer_multiple_msgs(spi_dev *dev, uint8_t no_of_messages)
 /***************************************************************************//**
 * @brief spi_eng_setup
 *******************************************************************************/
-int32_t spi_eng_setup(spi_dev **device,
```
mirceacaprioru: Should we have this modification as a separate commit since it relates to the spi_engine more?

stefpopa: The latest version of spi_engine also uses `spi_init `instead of `spi_eng_setup`.

ad7768-1fmcz/spi_engine.c
```diff
@@ -102,6 +102,63 @@ int32_t spi_eng_dma_read(spi_dev *dev,
 	return 0;
 }
 
+/***************************************************************************//**
```
commodo: is there a common `spi_engine.c` file that could be used instead ?

stefpopa: Not yet, but the latest and best version that we have so far is part of the [ad400x ](https://github.com/analogdevicesinc/no-OS/blob/master/ad400x-fmcz/spi_engine.c)project. We can probably use it in our future projects.

PR #137, opened 2019-01-31
ad7606: Add support for AD7606B ADC

stefpopa:
The AD7606B is a 16-bit ADC that supports simultaneous sampling of 8
channels. It is pin compatible to AD7606, but adds extra modes by
writing to the register map.

The AD7606B can be configured to work in software mode by setting all
oversampling pins to high. The oversampling ratio is configured from the
OS_MODE register (address 0x08) with the addition of OS=128 and OS=256
that were not available in hardware mode.

Moreover, in software mode, the range gpio has no longer its function.
Instead, the scale can be configured individually for each channel from
the RANGE_CH registers (address 0x03 to 0x06). Besides the already
supported ±10 V and ±5 V ranges, software mode can also accommodate the
±2.5 V range.

### DIFF
--- a/drivers/ad7606/ad7606.c
+++ b/drivers/ad7606/ad7606.c
@@ -64,7 +64,66 @@ static const struct ad7606_chip_info ad7606_chip_info_tbl[] = {
 		.num_channels = 8,
 		.has_oversampling = true,
 	},
+	[ID_AD7606B] = {
+		.num_channels = 8,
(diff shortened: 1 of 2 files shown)


commodo:
[LGTM](https://i.pinimg.com/564x/c8/8e/89/c88e89441974c21ab2730047f96c30d0.jpg)

PR #142, opened 2019-02-12
drivers:ltc2312: Add support

adrimbarean:
The LTC2312-14 is a 14-bit or 12-bit, 500ksps, serial sampling A/D
converter that draws only 3.2mA from a single 3V or 5V supply.

This driver is a modified LTC legacy driver and adds support for both
the 14-bit and 12-bit versions. The change from the original is only the
addition of the no-OS style API (setup and remove functions).

Part datasheet:
https://www.analog.com/media/en/technical-documentation/data-sheets/231214fa.pdf

Original driver:
https://github.com/analogdevicesinc/Linduino/tree/master/LTSketchbook/libraries/LTC2315

### DIFF
--- a/drivers/ltc2312/ltc2312.c
+++ b/drivers/ltc2312/ltc2312.c
@@ -1,0 +1,219 @@
+/*!
+REVISION HISTORY
+$Revision: 1785 $
+$Date: 2013-07-25 08:47:24 -0700 (Thu, 25 Jul 2013) $
+
+Copyright (c) 2013, Linear Technology Corp.(LTC)
+All rights reserved.
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions are met:
+
+1. Redistributions of source code must retain the above copyright notice, this
+   list of conditions and the following disclaimer.
+2. Redistributions in binary form must reproduce the above copyright notice,
+   this list of conditions and the following disclaimer in the documentation
+   and/or other materials provided with the distribution.
+
+THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR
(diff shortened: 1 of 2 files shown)


stefpopa:
Should we add the source of the driver in the commit message? I am not sure what the procedure should be. 
Also, have you added any modifications to the original driver?

adrimbarean:
There are modifications from the original driver. I don't know why I cannot find the original driver I took, but I sourced to one very similar. The difference between this one and the one I took is that the one I took had the shifting of the adc code built in.

V0 -> V1:
 - mention the original driver and changes made to it.

drivers/ltc2312/ltc2312.c
```diff
@@ -0,0 +1,216 @@
...
+ *
+ * @param [in] dev - The device structure.
+ *
+ * @return 0 in case of success, negative error code otherwise.
+ */
+int32_t ltc2312_remove(struct ltc2312_dev *dev)
+{
+	int32_t ret;
+
+	ret = spi_remove(dev->spi_desc);
```
commodo: maybe check `dev` for null

drivers/ltc2312/ltc2312.c
```diff
@@ -0,0 +1,216 @@
...
+ * bytes are actual data, depending on the part. The rest of the bits are 0, so
+ * the data must be shifted to the right by 2 for LTC2312-14 and by 4 for
+ * LTC2312-12.
+ *
+ * @param [in]  dev 		 - The device structure.
+ * @param [out] ptr_adc_code - Read data.
+ *
+ * @return void
+ */
+static void ltc2312_get_shift(struct ltc2312_dev *dev, uint8_t *shift)
```
commodo: this could be  made to return the shifted value
```
static uint8_t ltc2312_get_shift(struct ltc2312_dev *dev)
```

drivers/ltc2312/ltc2312.h
```diff
@@ -0,0 +1,105 @@
...
+/* Reads the LTC2315 and returns 32-bit data in offset binary format. */
+int32_t ltc2312_read(struct ltc2312_dev *dev, uint16_t *ptr_adc_code);
+
+/* Calculates the LTC2315 input voltage given the binary data and LSB weight. */
+void ltc2312_code_to_voltage(struct ltc2312_dev *dev, uint16_t adc_code,
+			     float vref, float *voltage);
+
+#endif  //  LTC2315_H
+
+
```
commodo: these 2 empty lines could be removed

commodo:
rest looks fine

adrimbarean:
V1 -> V2:
 - check for NULL descriptor pointer in the remove function
 - make ltc2312_get_shift return the shift
 - remove empty lines

commodo:
[LGTM](https://i.redd.it/6bb7tb4ofnh21.jpg)

PR #143, opened 2019-02-12
Ad9361

cristianpop-analog:
Add tinyiiod with serial and network interface


projects/ad9361/src/tinyiiod.c
```diff
@@ -0,0 +1,225 @@
+/*
```
commodo: it would be nice at some point to just reference https://github.com/analogdevicesinc/libtinyiiod  vs copying files over

projects/drivers/util/util.c
```diff
@@ -144,3 +144,17 @@ uint32_t hweight8(uint32_t word)
 
 	return count;
 }
+
+/**
+ * Calculate the quotient and the remainder of an integer division.
+ */
+uint64_t do_div(uint64_t* n,
```
commodo: should `do_div` be called `do_div64` ?
the idea is to make it clearer that it's u64, in case there would be a u32 `do_div` equivalent

projects/ad9361/src/tinyiiod.c
cristianpop-analog: I think you are right, probably the initial version of tinyiiod that I use and modified, was taken from here.

projects/drivers/util/util.c
cristianpop-analog: we can rename it to "do_div64" when a  "do_div32" is implemented?

commodo: fine from my side

projects/drivers/network/network.c
```diff
@@ -0,0 +1,502 @@
...
+	tcp_setprio(newpcb, TCP_PRIO_MAX);
+
+	es = (struct network_instance *)mem_malloc(sizeof(struct network_instance));
+	if (es != NULL) {
+		es->state = ES_ACCEPTED;
+		es->pcb = newpcb;
+		es->retries = 0;
+		es->p = NULL;
+		es->instance_id = inst_id;
+		HASH_ADD_INT( instances, instance_id, es );
```
commodo: according to cppcheck there is an un-initialized variable here and in other places;
should we check if the generated code is ok ?
when following the macros, i am bit vague about it, whether the reported error is valid or not

projects/drivers/network/lwip_init.c
```diff
@@ -0,0 +1,246 @@
...
+	init_platform();
+
+#if LWIP_IPV6==0
+#if LWIP_DHCP==1
+	ipaddr.addr = 0;
+	gw.addr = 0;
+	netmask.addr = 0;
+#else
+	/* initliaze IP addresses to be used */
+//	IP4_ADDR(&ipaddr,  10, 168,   1, 10);
```
mirceacaprioru: This commented code i think could be removed altogether.

projects/drivers/network/network.c
```diff
@@ -0,0 +1,502 @@
...
+#ifdef DEBUG_NETWORK
+		printf("new clientconnected: %d\n", inst_id);
+#endif // DEBUG_NETWORK
```
mirceacaprioru: This comments could be changed to /* */ since that is the norm for C style.

projects/drivers/network/network.c
```diff
@@ -0,0 +1,502 @@
...
+		network_close(tpcb, es);
+		ret_err = ERR_OK;
+	} else {
+		/* unkown es->state, trash data  */
+		tcp_recved(tpcb, p->tot_len);
+		es->p = NULL;
+		pbuf_free(p);
+		ret_err = ERR_OK;
+	}
+	return ret_err;
```
mirceacaprioru: Usually there is a blank line before return to separate from the rest of the code.

projects/drivers/network/network.c
commodo: this looks fine; maybe here, cppcheck has a limitation;
i suggest using inline suppression comments;
so, if you add `/* cppcheck-suppress uninitvar ;  cppcheck reports this as a false positive */` , cppcheck should consider the comment and ignore this

commodo:
Codacy reports quite a number of issues that are reasonable enough to handle.
I would make another PR to handle them

.cppcheckignore
```diff
@@ -10,4 +10,5 @@
 // as libraries/modules and are outside of the scope of this repository. 
 //
 // Another alternative is using inline suppressions. Please see Cppcheck manual
-// for further information: https://cppcheck.sourceforge.net/manual.pdf 
+// for further information: https://cppcheck.sourceforge.net/manual.pdf
+Uninitialized variable:projects/drivers/network/network.c
```
commodo: please don't add entire files to `.cppcheckignore` unless needed
in this case you can use inline suppressions

projects/drivers/network/network.c
commodo: related to comment https://github.com/analogdevicesinc/no-OS/pull/143/files#r257196836   please take a look at this comment;
you can try 
```
/* cppcheck-suppress uninitvar ; cppcheck reports this as a false positive */
```
which avoids cppcheck going over this line only

projects/ad9361/src/tinyiiod.c
cristianpop-analog: Reference to the project is added:
https://github.com/analogdevicesinc/no-OS/pull/143/commits/45bd09c8e49deb15a0e0819fa8928ded7ba8c6d9

PR #158, opened 2019-03-21
Update xstc.tcl script to increase heap size for MICROBLAZE

adrimbarean:
For MICROBLAZE platforms most projects require an increase in HEAP size
for the application to function properly. Increase the size of the heap and
the stack on the MICROBLAZE platforms to match the default sizes of the
ZYNQ platforms.

### DIFF
--- a/scripts/xsct.tcl
+++ b/scripts/xsct.tcl
@@ -33,7 +33,7 @@ if {$m_mode == "library"} {
 if {$m_mode == "defines"} {
 
   sdk configapp -app sw build-config debug
-  for {set i 1} {$i < $argc} {incr i} {
+  for {set i 2} {$i < $argc} {incr i} {
     set compiler_define [lindex $argv $i]
     if { [regexp [ 0-9] $compiler_define ] } {
 	regsub -all { } $compiler_define {=} compiler_define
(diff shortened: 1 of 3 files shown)


scripts/xsct.tcl
```diff
@@ -12,6 +12,19 @@ if {$m_mode == "init"} {
   sdk createhw -name hw -hwspec $hdf_file
   sdk createapp -name xilsw -hwproject hw -proc $cpu_name -bsp bsp -app {Hello World}
   sdk createapp -name sw -hwproject hw -proc $cpu_name -bsp bsp -app {Empty Application}
+  if { $cpu_name == "sys_mb" } {
```
commodo: isn't there some project settings somewhere, where this can be modified without doing linker tricks ?

adrimbarean:
V0 -> V1:
 - increase heap size of MICROBLAZE projects by defining _HEAP_SIZE symbol for the linker script to use as the heap memory size instead of the default value and avoid modifying the linker script

adrimbarean:
V1  -> V2:
 - change heap and stack sizes for MICROBLAZE to 0x2000 to match the ZYNQ default values

adrimbarean:
V2 -> V3:
 - update commit and PR descriptions

PR #162, opened 2019-04-05
projects:drivers:xilinx_platform: add AXI I2C implementation

adrimbarean:
Add AXI I2C implementation to the xilinx platform drivers. The
implementation is for the ZYNQ PS7 family.

### DIFF
--- a/projects/drivers/xilinx_platform/platform_drivers.c
+++ b/projects/drivers/xilinx_platform/platform_drivers.c
@@ -57,15 +57,48 @@
 int32_t i2c_init(struct i2c_desc **desc,
 		 const struct i2c_init_param *param)
 {
-	if (desc) {
-		// Unused variable - fix compiler warning
-	}
+	i2c_desc *dev;
+	int32_t ret;
 
-	if (param->type) {
-		// Unused variable - fix compiler warning
-	}
+	dev = calloc(1, sizeof *dev);
+	if(!dev)
+		return FAILURE;
+
+	dev->type = param->type;
+	dev->id = param->id;
+	dev->max_speed_hz = param->max_speed_hz;
+	dev->slave_address = param->slave_address;
+
+	dev->config = XIic_LookupConfig(dev->id);
+	if (dev->config == NULL)
+		goto error;
+
+	ret = XIic_CfgInitialize(&dev->instance, dev->config,
+				 dev->config->BaseAddress);
+	if(ret != 0)
+		goto error;
+
+	Xil_ExceptionInit();
+
+	Xil_ExceptionRegisterHandler(XIL_EXCEPTION_ID_IRQ_INT,
+				     XIic_InterruptHandler, &dev->instance);
+
+	ret = XIic_Start(&dev->instance);
+	if(ret != 0)
+		goto error;
+
+	ret = XIic_SetAddress(&dev->instance, XII_ADDR_TO_SEND_TYPE,
+			      dev->slave_address);
+	if(ret != 0)
+		goto error;
+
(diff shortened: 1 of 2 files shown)


projects/drivers/xilinx_platform/platform_drivers.c
```diff
@@ -97,21 +132,18 @@ int32_t i2c_write(struct i2c_desc *desc,
...
+	if(stop_bit == 1)
+		teminate_option = XIIC_STOP;
+	else
+		teminate_option = XIIC_REPEATED_START;
 
-	if (stop_bit) {
-		// Unused variable - fix compiler warning
-	}
+	ret = XIic_Send(desc->instance.BaseAddress, desc->slave_address, data,
+			bytes_number, teminate_option);
```
dbogdan: Since teminate_option is used only once, I would remove it and use something like:
stop_bit ? XIIC_STOP : XIIC_REPEATED_START
Similar for i2c_read().

projects/drivers/xilinx_platform/platform_drivers.c
```diff
@@ -57,15 +57,46 @@
...
+		goto error;
+
+	ret = XIic_CfgInitialize(&dev->instance, dev->config,
+				 dev->config->BaseAddress);
+	if(ret != XST_SUCCESS)
+		goto error;
+
+	Xil_ExceptionRegisterHandler(XIL_EXCEPTION_ID_IRQ_INT,
+				     XIic_InterruptHandler, &dev->instance);
+
```
dbogdan: We have to make sure that this is the right approach (e.g.,  shouldn't we call Xil_ExceptionInit() in advance and Xil_ExceptionEnable() after? ).

projects/drivers/xilinx_platform/platform_drivers.c
```diff
@@ -57,15 +57,46 @@
...
+	dev->max_speed_hz = param->max_speed_hz;
+	dev->slave_address = param->slave_address;
+
+	dev->config = XIic_LookupConfig(dev->id);
+	if (dev->config == NULL)
+		goto error;
+
+	ret = XIic_CfgInitialize(&dev->instance, dev->config,
+				 dev->config->BaseAddress);
+	if(ret != XST_SUCCESS)
```
dbogdan: For consistency with the other functions, I would use 0 instead of XST_SUCCESS. Otherwise, I would update similarly the other functions. :)

projects/drivers/xilinx_platform/platform_drivers.c
adrimbarean: This document on page 70 notes that for ARM Cortex-A53, Cortex-R5, and Cortex-A9 the exception handlers are being initialized statically and this function is only used for backwards compatibility. Should we add it to have this feature as well?

https://github.com/Xilinx/embeddedsw/blob/master/lib/bsp/standalone/doc/standalone.pdf

adrimbarean: I will add Xil_ExceptionInit() just to eliminate unnecessary risks. I would add a function call and a return, but the application has no constrains on speed that will be affected by this.

adrimbarean:
V0 -> V1:
 - make implementation non-generic by passing the i2c device id through the initialization routine;
 - check for 0 instead of XST_SUCCESS to maintain consistency;
 - use conditional operator in the i2c read and write functions;
 - add exception driver initialization function (Xil_ExceptionInit()) to reduce risk.

mirceacaprioru:
Any new updates on this thread ? @dbogdan

PR #166, opened 2019-04-24
Master ad9371 dac dma example

stefpopa:
Commit f9992d9 offers an example of how to send custom data using the
DMA and adrv9009.
This patch updates the headless file in the same manner in order to work
with the ad9371 project. The main difference is that the configuration
for ad9371 uses the FIFO, while adrv9009 bypasses it. At the moment, the
DMA cyclic mode is not supported in the hdl.


dbogdan:
Fixing the code style issues should be the only required change.

stefpopa:
changes in v2:
* fixed astyle warnings

PR #171, opened 2019-05-24
ad713x: spi_engine: Pass init structure by reference instead of by value

stefpopa:
Signed-off-by: Istvan Csomortani <istvan.csomortani@analog.com>


mirceacaprioru:
There are some astyle complaints that should be addressed before merging.

PR #174, opened 2019-05-28
Platform drivers spi

cristiantpop:
When writing to SPI, some mask is applied to chip_select value. This seems
to be wrong.

### DIFF
--- a/projects/drivers/xilinx_platform/platform_drivers.c
+++ b/projects/drivers/xilinx_platform/platform_drivers.c
@@ -259,7 +259,7 @@ int32_t spi_write_and_read(struct spi_desc *desc,
 			   XSPIPS_CLK_PHASE_1_OPTION : 0));
 
 	XSpiPs_SetSlaveSelect(&desc->instance,
-			      0xf & ~desc->chip_select);
+			      desc->chip_select);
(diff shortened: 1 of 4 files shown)


projects/drivers/xilinx_platform/platform_drivers.c
```diff
@@ -259,7 +259,7 @@ int32_t spi_write_and_read(struct spi_desc *desc,
 			   XSPIPS_CLK_PHASE_1_OPTION : 0));
 
 	XSpiPs_SetSlaveSelect(&desc->instance,
-			      0xf & ~desc->chip_select);
+			      desc->chip_select);
```
dbogdan: Since this change affects other projects too, they should be updated as well by the same commit.

cristiantpop:
V2:

- Change CLK_CS values accordingly

cristiantpop:
V3:

- Fix slave select value for Microblaze platform

cristiantpop:
V4: 
-  Substract "-1"  in "spi_ad_desc->chip_select = spiSettings->chipSelectIndex - 1;" to obtain all possible values for CS.

cristiantpop:
V5:

- revert CS values in ad9371 parameters.h, since they are fixed by V4 changes in this pull request

PR #176, opened 2019-06-06
AD5592/3r byte swap fix

adrimbarean:
Changes in this patch:
 - fix SPI byte swap error when reading registers;
 - fix compiler error about a static function treated as non-static.


mirceacaprioru:
LGTM

Issue #31, opened 2017-10-16
 no-OS/drivers/ad717x/ad717x.c AD717X_WaitForReady


Silhouette2022:
Driver file  no-OS/drivers/ad717x/ad717x.c  has a bug inside AD717X_WaitForReady function.
According to manual ready bit in status register set to 0 if conversion has been completed and set to 1 if data is awaiting. Metioned function test that bit for 1 and return success status. So, instead of testing for 1 it should test for 0 value.

dbogdan:
Thanks for letting us know about this and sorry for the inconvenience.
The issue should be fixed now: https://github.com/analogdevicesinc/no-OS/commit/8251dfc96b85e1b1f3875dcd8a35633cea075e42

PR #178, opened 2019-06-10
drivers: ad7124: Fix ad7124_wait_for_conv_ready()

spiccinini:
Hi!
The RDY bit of the status register is cleared when data is ready to be read and the code does the opposite.

It seems that this is the same issue but in another part https://github.com/analogdevicesinc/no-OS/issues/31 and this was the fix https://github.com/analogdevicesinc/no-OS/commit/8251dfc96b85e1b1f3875dcd8a35633cea075e42

Tested with hardware.


adrimbarean:
Hello @spiccinini,
Good catch! Thank you for the input!
Can you please sign the commit?

PR #179, opened 2019-06-12
drivers:ad5592/3r: initialize internal reference

adrimbarean:
The internal reference is a core feature for the AD5592R and AD5593R
devices. The driver must have the posibility to initialize the internal
reference so that applications can activate it or leaving it deactivated in
a transparent manner. This helps applications to deliver a more complete
out-of-the-box experience.

### DIFF
--- a/drivers/adc-dac/ad5592r/ad5593r.c
+++ b/drivers/adc-dac/ad5592r/ad5593r.c
@@ -211,9 +211,11 @@ int32_t ad5593r_gpio_read(struct ad5592r_dev *dev, uint8_t *value)
  * @param dev - The device structure.
  * @return 0 in case of success, negative error code otherwise
  */
-int32_t ad5593r_init(struct ad5592r_dev *dev)
+int32_t ad5593r_init(struct ad5592r_dev *dev,
+		     struct ad5592r_init_param *init_param)
 {
 	int32_t ret;
+	uint16_t temp_reg_val;
 
 	if (!dev)
 		return FAILURE;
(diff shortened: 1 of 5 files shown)


drivers/ad5592r/ad5592r-base.h
```diff
@@ -91,6 +91,15 @@ struct ad5592r_rw_ops {
 	int32_t (*gpio_read)(struct ad5592r_dev *dev, uint8_t *value);
 };
 
+enum ad5592r_reference {
+	AD5592R_INT_REF_ACTIVE,
+	AD5592R_INT_REF_DEACTIVE
+};
+
+struct ad5592r_init_param {
+	enum ad5592r_reference int_ref;
```
stefpopa: I would rather use a boolean variable instead of an enum. Something like `bool use_int_ref;`

drivers/ad5592r/ad5592r.c
```diff
@@ -227,5 +229,19 @@ int32_t ad5592r_init(struct ad5592r_dev *dev)
 	if (ret < 0)
 		return ret;
 
-	return ad5592r_set_channel_modes(dev);
+	ret = ad5592r_set_channel_modes(dev);
+
+	if(init_param->int_ref == AD5592R_INT_REF_ACTIVE) {
+		if (ret < 0)
+			return ret;
```
stefpopa: hmm.. shouldn't his ret check be done after `ret = ad5592r_set_channel_modes(dev);` ?
```
ret = ad5592r_set_channel_modes(dev);
if (ret < 0)
        return ret;
```

adrimbarean: I was thinking there is no need to check it if the internal reference is not activated because the next line would be `return ret;` either way.

mirceacaprioru: Yes but  this evaluates an extra if statement that would not be necessary if ad5592r_set_channel_modes fails.

mirceacaprioru: The concept is to exit as soon as something fails do not do extra operations.

adrimbarean:
V0 -> V1;
 - make `use_int_ref` boolean;
 - move ret checking on the logical place.

drivers/adc-dac/ad5592r/ad5593r.c
```diff
@@ -222,5 +224,19 @@ int32_t ad5593r_init(struct ad5592r_dev *dev)
 	if (ret < 0)
 		return ret;
 
-	return ad5592r_set_channel_modes(dev);
-}
\ No newline at end of file
+	ret = ad5592r_set_channel_modes(dev);
+
+	if(init_param->int_ref) {
+		if (ret < 0)
```
mirceacaprioru: you still have this here. should be changed same as above with ret after ad5592r_set_channel_modes

adrimbarean:
V1 -> V2:
 - move ret checking in the logical place for both initialization routines.

mirceacaprioru:
will merge tomorrow

PR #180, opened 2019-06-14
d9361: Clear ad9361 state, when channels mode is selected

cristiantpop:
In this way, ad9361 gain table will be reconfigured, as before calling
"ad9361_set_no_ch_mode()" function.


dbogdan:
The commit message is not exactly clear.
Since the device is reset by ad9361_set_no_ch_mode(), its state should be also cleared.

ad9361/sw/ad9361.h
```diff
@@ -3541,4 +3541,5 @@ int32_t ad9361_get_temp(struct ad9361_rf_phy *phy);
 int ad9361_synth_lo_powerdown(struct ad9361_rf_phy *phy,
```
mirceacaprioru: there is a typo in the commit message ad9361 instead of d9361. The message is vague, it should contain more information about what it does.

cristiantpop:
v2:

- fix typo in commit message;

- add more details in commit message, about what the commit does.

PR #182, opened 2019-06-21
projects/drivers/util: Add additional macros

cristiantpop:
Necessary for ad9172 driver implementation


projects/drivers/util/util.h
```diff
@@ -83,6 +83,15 @@
...
 
+#define BITS_PER_LONG 64
+
+#define GENMASK(h, l) \
+	(((~0UL) - (1UL << (l)) + 1) & (~0UL >> (BITS_PER_LONG - 1 - (h))))
+
+#define U16_MAX		((uint16_t)~0U)
+#define S16_MAX		((int16_t)(U16_MAX>>1))
+
+#define DIV_U64(x, y) (x / y)
```
stefpopa: You might also add FIELD_PREP macro here as well. Might prove to be useful in the future. However, you should know better.

PR #183, opened 2019-06-21
hmc7044: Extend driver API

cristiantpop:
hmc7044_read - read device register
hmc7044_remove - release device resources

### DIFF
--- a/drivers/hmc7044/hmc7044.c
+++ b/drivers/hmc7044/hmc7044.c
@@ -49,6 +49,7 @@
 /********************** Macros and Constants Definitions **********************/
 /******************************************************************************/
 #define HMC7044_WRITE		(0 << 15)
+#define HMC7044_READ		(1 << 15)
 #define HMC7044_CNT(x)		(((x) - 1) << 13)
 #define HMC7044_ADDR(x)		((x) & 0xFFF)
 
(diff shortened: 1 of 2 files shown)


drivers/hmc7044/hmc7044.c
```diff
@@ -235,6 +236,33 @@ static int hmc7044_write(struct hmc7044_dev *dev,
...
 }
 
+/**
+ * SPI register read from device.
+ * @param dev - The device structure.
+ * @param reg - The register address.
+ * @param val - The register data.
+ * @return SUCCESS in case of success, negative error code otherwise.
+ */
+static int hmc7044_read(struct hmc7044_dev *dev,
```
mirceacaprioru: From this single commit PR i don't see the need for this extension. Usually in a PR you should also add the use of these functions that you created or at least say in the commit message that they are intended for future use.

cristiantpop:
v2:

- add following to commit message: "This functions will be used in AD9172 demo project"

drivers/hmc7044/hmc7044.c
```diff
@@ -548,3 +576,13 @@ int32_t hmc7044_init(struct hmc7044_dev **device,
 
 	return hmc7044_setup(dev);
 }
+
+int32_t hmc7044_remove(struct hmc7044_dev *device) {
```
dbogdan: Please add the function's description.

cristiantpop:
v3:

- Added the function's description

drivers/hmc7044/hmc7044.c
```diff
@@ -235,6 +236,33 @@ static int hmc7044_write(struct hmc7044_dev *dev,
...
+	cmd = HMC7044_READ | HMC7044_CNT(1) | HMC7044_ADDR(reg);
+	buf[0] = cmd >> 8;
+	buf[1] = cmd & 0xFF;
+	buf[2] = 0;
+
+	ret = spi_write_and_read(dev->spi_desc, buf, ARRAY_SIZE(buf));
+
+	*val = buf[2];
+
+	return ret;
```
stefpopa: Curious: does it make sense to update `val ` if the `spi_write_and_read()` has failed?
```
ret = spi_write_and_read(dev->spi_desc, buf, ARRAY_SIZE(buf));
if (ret < 0)
        return ret;

*val = buf[2];

return 0;
```

cristiantpop:
v4:

- return immediately with error code, from "hmc7044_read" if "spi_write_and_read" fails

PR #191, opened 2019-06-27
drivers: ad5686 & ad7768

dbogdan:
drivers: ad5686: Fix compilation error
drivers: ad7768: Update to current platform driver


dbogdan:
v2:
   -  Fix "Variable 'ret' is reassigned a value before the old one has been used.
          ret = spi_write_and_read(dev->spi_desc, buf, 2);" issue.

dbogdan:
v3:
   - Split GPIO and SPI changes.

PR #195, opened 2019-07-15
2nd reorg

dbogdan:


dbogdan:
v2:
   - Fix code style issues.

dbogdan:
v3:
   - Fix all code style issues. :)

drivers/platform/generic/delay.c
```diff
@@ -0,0 +1,72 @@
+/***************************************************************************//**
```
commodo: this all looks like entirely new code;
well, from a patch perspective point of view;
why not move these file from an existing place?

the diff would be much smaller and easier to track;

drivers/accel/adxl345/adxl345.c
```diff
@@ -40,7 +40,6 @@
 /******************************************************************************/
 /***************************** Include Files **********************************/
 /******************************************************************************/
-#include <stdint.h>
```
commodo: i'm not entirely sure about this change as-is;
that is to say: it's not bad, but I don't know about the added value it adds;
nor the implications of doing it this way or another;

reasons for my un-sureness:
1. patch title is `drivers/accel: Make the header files self-sufficient` without providing details about the implications here; all it seems to do, is move stdint.h, which makes me assume it's for the int32_t/int8_t/etc types;
there isn't sufficient argumentation about why only stdint.h is moved and not other headers as well
2. since this code is not Linux, it gets copied/integrated all over, people will get various types/implementations of `stdint.h`; some header collisions could occur or missing definitions;
we may have to provide our own `stdint.h` for this code-base at some point to be a bit more comfortable with this point; but that's another point for another time

that being said: i am not against this as-is; nor am i convinced by this patch as-is

dbogdan: 1. I agree - the description was entirely missing.
2. I guess that this should be the subject of another discussion. The current patch just moves the include from .c to .h .

drivers/platform/generic/delay.c
dbogdan: Well, not all the drivers were updated to use this new approach - moving the code will mean that the old platform_drivers gets removed, right?

mirceacaprioru: Isn't this entirely new code ? Did we have this format before, or everything was in platform_data ?

dbogdan: I was mainly a copy-paste process.
Have a look here: https://github.com/analogdevicesinc/no-OS/blob/master/drivers/platform/generic/platform_drivers.c

dbogdan:
v4:
   - Rebase to master branch.
   - Add a description for the "drivers/accel: Make the header files self-sufficient" patch.

drivers/platform/generic/delay.c
commodo: why not do this a sequential split?
or, remove the file when these files get introduced in the same commit

i don't insist on doing it;
but it does show up neater if the commit has the original/old file on which to anchor the "new" code/files

dbogdan: Sorry, it was a misunderstanding. I've pushed some changes.

dbogdan:
v5:
   - Remove platform_drivers.c.
   - Add a separate patch for each independent change.

PR #198, opened 2019-07-18
ad9172: Add DAC_DMA_EXAMPLE

cristiantpop:
Select AXI_DAC_DATA_SEL_DMA as source for DAC. In this way custom data can
be sent at the output of AD9172.

### DIFF
--- a/projects/ad9172/src/main.c
+++ b/projects/ad9172/src/main.c
@@ -50,6 +50,10 @@
 #include "parameters.h"
 #include "inttypes.h"
 
+#ifdef DAC_DMA_EXAMPLE
+#include "axi_dmac.h"
+#endif /* DAC_DMA_EXAMPLE */
+
 int main(void)
 {
 	struct spi_init_param hmc7044_spi_param = {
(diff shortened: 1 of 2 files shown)


projects/ad9172/src/main.c
```diff
@@ -192,6 +206,15 @@ int main(void)
...
 	}
 
+#ifdef DAC_DMA_EXAMPLE
+	extern const uint32_t sine_lut_iq[1024];
+	axi_dac_load_custom_data(tx_dac, sine_lut_iq,
+				 ARRAY_SIZE(sine_lut_iq),
+				 DDR_MEM_BASEADDR);
+	axi_dmac_init(&tx_dmac, &tx_dmac_init);
+	axi_dmac_transfer(tx_dmac, DDR_MEM_BASEADDR,
+			  sizeof(sine_lut_iq) * 2);
```
dbogdan: I guess that this multiplication by 2 is related to the number of channels. I would use that  information instead since it should be available at this point.

cristiantpop:
V2:

- add new commit to update README file

projects/ad9172/src/main.c
cristiantpop: Well..."num_channels" in "tx_dac_init" is 4, so an other division will be needed.

cristiantpop:
V3:

- Add commit for README file fix

- Use "tx_dac->num_channels", in calculating the size of data that needs to be transferred by DMA

- Update README, with DMA corresponding files, in the second commit

PR #200, opened 2019-07-22
Ad9208 prepare

stefpopa:
This PR comes before adding support for the AD9208


projects/drivers/jesd204/axi_adxcvr.c
```diff
@@ -219,39 +214,47 @@ int32_t adxcvr_clk_set_rate(struct adxcvr *xcvr,
...
-
-		ret = xilinx_xcvr_write_rx_clk25_div(&xcvr->xlx_xcvr,
-						     ADXCVR_DRP_PORT_CHANNEL_BCAST, clk25_div);
-	} else {
-		ret = xilinx_xcvr_write_tx_clk25_div(&xcvr->xlx_xcvr,
-						     ADXCVR_DRP_PORT_CHANNEL_BCAST, clk25_div);
 	}
-	if (ret < 0)
-		return ret;
+
```
mirceacaprioru: nitpick: extra line

projects/drivers/jesd204/axi_adxcvr.c
```diff
@@ -80,9 +80,8 @@
 #define ADXCVR_DRP_PORT_ADDR_COMMON		0x00
```
dbogdan: We should state the fact that the BROADCAST feature will write the same value to all channels. Since some times not all channels are configured the same (e.g., clock source is different), an update of some register's bits should be made independently for each channel to not overwrite other settings. The BCAST read might also fail since it returns the ORed between all channels.

projects/drivers/util/util.c
```diff
@@ -158,3 +158,21 @@ uint64_t do_div(uint64_t* n,
 
 	return mod;
 }
+
+uint64_t div64_u64_rem(uint64_t dividend, uint64_t divisor, uint64_t *remainder)
+{
+	*remainder = dividend % divisor;
+	return dividend / divisor;
```
dbogdan: For consistency with the rest of the file I would add a new line.

dbogdan: ... and for the same reason, please add a short comment before the implementation.

stefpopa:
Changes in v2:
* Added a more explicit commit message for 5df46a3064623084d202101f96ec4a933eb59476
* Added a new line before returning
* Added comments for the implementation of the util helper functions

stefpopa:
Will merge later if there are no objections.

PR #203, opened 2019-07-23
drivers: adc: include shared header files - Part 1

amiclaus:
Include shared header files accross all platforms

Include stdint.h in driver's header file

Using one commit since the modifications are small for each driver separately.


dbogdan:
Moving stdint.h should be done in a different commit - it has a different reason - to make the header files self-sufficient.

amiclaus:
Changelog v1 -> v2:

- use different commit for moving stdint.h header files

PR #207, opened 2019-07-24
Reorg drivers - Part 3

amiclaus:
Include shared header files across all platforms.

Move stdint.h into header files.


drivers/temperature/adt7420/adt7420.h
```diff
@@ -39,6 +39,11 @@
 #ifndef __ADT7420_H__
 #define __ADT7420_H__
 
+/******************************************************************************/
+/***************************** Include Files **********************************/
+/******************************************************************************/
+#include "i2c.h"
```
commodo: maybe this is late to mention; it just occurred to me,  but it may be a good idea to open a discussion of `#include <i2c.h>` vs `#include "i2c.h"`

`#include "i2c.h"` works fine in general, but this usually means [or works like]:
* if there is a `"i2c.h"` file near the C file, include it
* if not, use the `"i2c.h"` file in the build include paths [which is usually specified via `-Iinclude -I/usr/include`, etc]

`#include <i2c.h>` will ignore the  `"i2c.h"` file near the C file, and directly try to include from the build include paths [ `-Iinclude -I/usr/include`, etc]

the implications of this can vary; and none of these 2 options is considerably better than the other;
for the Linux project [as an example], the convention of using `#include <i2c.h>` vs `#include "i2c.h"` seems to be enforced:
* if using headers from the include path, use  `#include <i2c.h>`
* if using private headers [specific to the driver/module/etc] use `#include "i2c.h'`

amiclaus: The platform generic source files (i.e. `i2c.c`) contain mainly only the header functions required. Without implementation. I suppose the implementation will be done particularly for each driver when using it. Therefore will be included in the same folders with the driver source/header files.

mirceacaprioru: This kind of depends on what is being used, some projects use a base src and include folder. This is how we use them on ADICUP3029, 360 so the i2c.h will be in a different folder and it will follow the build include paths. 

But some clients may use a different structure. Since #include "i2c.h" is more general and we don't enforce anything on no-OS i would go with this one. 

But this is just my opinion.

dbogdan: I vote for #include "file" too.
https://gcc.gnu.org/onlinedocs/cpp/Include-Syntax.html - our files are not exactly system header files.

commodo: ack;
no strong preference;
maybe we should just keep this convention in mind for later;

PR #212, opened 2019-07-31
ad7091: move driver to Xilinx SDP ADC project

amiclaus:
The implementation of AD7091 is using ADC core interface which is
currently specific to SDP ADC project.

The AD7091R no-OS driver will be used instead both for AD7091/AD7091R
since the software requirements correspond.


amiclaus:
Changelog v1->v2: 

- fix code style issues

PR #214, opened 2019-08-08
Adds STATUS on DATA readback, and fixes issue writing to part after reset

mike-bradley:
Refactoring CRC calculation code to meet astyle formatting requirements

### DIFF
--- a/drivers/adc/ad7124/ad7124.c
+++ b/drivers/adc/ad7124/ad7124.c
@@ -386,7 +386,9 @@ uint8_t ad7124_compute_crc8(uint8_t * p_buf, uint8_t buf_size)
 
 	while(buf_size) {
 		for(i = 0x80; i != 0; i >>= 1) {
-			if(((crc & 0x80) != 0) != ((*p_buf & i) != 0)) { /* MSB of CRC register XOR input Bit from Data */
+			bool cmp1 = (crc & 0x80) != 0;
+			bool cmp2 = (*p_buf & i) != 0;
+			if(cmp1 != cmp2) { /* MSB of CRC register XOR input Bit from Data */
 				crc <<= 1;
 				crc ^= AD7124_CRC8_POLYNOMIAL_REPRESENTATION;
 			} else {


drivers/adc/ad7124/ad7124.c
```diff
@@ -68,6 +79,7 @@ int32_t ad7124_no_check_read_register(struct ad7124_dev *dev,
 	uint8_t i = 0;
 	uint8_t check8 = 0;
 	uint8_t msg_buf[8] = {0, 0, 0, 0, 0, 0, 0, 0};
+	int32_t add_status_length = 0;
```
mirceacaprioru: this is not just cleanup, you actually add functionality here. I would do this in a separate commit. Also uint8_t will be enough and you can pair it with check8.

uint8_t check8 = 0, add_status_length = 0;

drivers/adc/ad7124/ad7124.c
```diff
@@ -231,10 +260,15 @@ int32_t ad7124_reset(struct ad7124_dev *dev)
 				 wr_buf,
 				 8);
 
-	/* Wait for the reset to complete */
+	/* This is to ensure the CRC is disabled after reset */
+	dev->use_crc = AD7124_DISABLE_CRC;
```
mirceacaprioru: this disable CRC is again unrelated to the commit.

drivers/adc/ad7124/ad7124.c
```diff
@@ -8,7 +7,8 @@
 *
 * All rights reserved.
 *
-* Redistribution and use in source and binary forms, with or without modification,
+* Redistribution and use in source and binary forms, with
```
mirceacaprioru: squash this with previous commit since they are both astyle modifications.

drivers/adc/ad7124/ad7124.c
```diff
@@ -1,3 +1,4 @@
+/***************************************************************************//**
```
mirceacaprioru: squash this commit too.

drivers/adc/ad7124/ad7124.h
```diff
@@ -247,6 +250,9 @@
 #define AD7124_CFG_REG_REF_SEL(x)  ((x) & 0x3) << 3
 #define AD7124_CFG_REG_PGA(x)      (((x) & 0x7) << 0)
 
+/* PGA Gain Value */
+#define AD7124_PGA_GAIN(x)     (1 << (x))
```
mirceacaprioru: again not related to this commit.

drivers/adc/ad7124/ad7124.c
```diff
@@ -420,7 +420,8 @@ uint8_t ad7124_compute_crc8(uint8_t * p_buf, uint8_t buf_size)
 
 	while(buf_size) {
 		for(i = 0x80; i != 0; i >>= 1) {
-			if(((crc & 0x80) != 0) != ((*p_buf & i) != 0)) { /* MSB of CRC register XOR input Bit from Data */
+			if(((crc & 0x80) != 0) != ((*p_buf & i) !=
+					                   0)) { /* MSB of CRC register XOR input Bit from Data */
```
commodo: nitpick; this is isn't the nicest wai to break this line;
admittedly, it's also not the nicest code either [from our side]

one way to do it is to use temp vars;
let's say:
```
cmp1 = (crc & 0x80) != 0);
cmp2 = (*p_buf & i);

if (cmp1 != cmp2) {
```

drivers/adc/ad7124/ad7124.h
mike-bradley: will split this and related #define changes out into separate pull request

drivers/adc/ad7124/ad7124.c
mike-bradley: will split out into separate pull request to ensure CRC setting is disabled after reset

drivers/adc/ad7124/ad7124.c
mike-bradley: will do

drivers/adc/ad7124/ad7124.c
mike-bradley: will review this pull request so it covers just the adding support appending 4th byte of STATUS to DATA

drivers/adc/ad7124/ad7124.c
mike-bradley: not a problem, can edit code as suggested.

PR #215, opened 2019-08-08
projects: drivers: xilinx: update platform drivers

amiclaus:
Include shared header files.

Reorganize platform drivers using the sturctures available in shared
header files.

Use `extra` pointers to access platforms specific attributes.

### DIFF
--- a/projects/drivers/xilinx_platform/platform_drivers.c
+++ b/projects/drivers/xilinx_platform/platform_drivers.c
@@ -43,6 +43,7 @@
 #include <stdlib.h>
 #include <sleep.h>
 #include "platform_drivers.h"
+#include "xil_drivers.h"
 
 /******************************************************************************/
 /************************ Functions Definitions *******************************/
@@ -59,36 +60,47 @@ int32_t i2c_init(struct i2c_desc **desc,
 {
 #ifdef XIIC_H
 	i2c_desc *dev;
+	xil_i2c_desc *xil_dev;
+	xil_i2c_init_param *xil_param;
 	int32_t ret;
 
 	dev = calloc(1, sizeof *dev);
 	if(!dev)
 		return FAILURE;
 
-	dev->type = param->type;
-	dev->id = param->id;
+	dev->extra = calloc(1, sizeof *xil_dev);
+	if(!(dev->extra)){
+		free(dev);
+		return FAILURE;
+	}
+
+	xil_dev = dev->extra;
+	xil_param = param->extra;
+
+	xil_dev->type = xil_param->type;
+	xil_dev->id = xil_param->id;
 	dev->max_speed_hz = param->max_speed_hz;
 	dev->slave_address = param->slave_address;
 
-	dev->config = XIic_LookupConfig(dev->id);
-	if (dev->config == NULL)
+	xil_dev->config = XIic_LookupConfig(xil_dev->id);
+	if (xil_dev->config == NULL)
 		goto error;
 
-	ret = XIic_CfgInitialize(&dev->instance, dev->config,
-				 dev->config->BaseAddress);
+	ret = XIic_CfgInitialize(&xil_dev->instance, xil_dev->config,
+				 xil_dev->config->BaseAddress);
 	if(ret != 0)
 		goto error;
 
 	Xil_ExceptionInit();
 
 	Xil_ExceptionRegisterHandler(XIL_EXCEPTION_ID_IRQ_INT,
-				     XIic_InterruptHandler, &dev->instance);
+				     XIic_InterruptHandler, &xil_dev->instance);
 
-	ret = XIic_Start(&dev->instance);
+	ret = XIic_Start(&xil_dev->instance);
 	if(ret != 0)
 		goto error;
 
-	ret = XIic_SetAddress(&dev->instance, XII_ADDR_TO_SEND_TYPE,
+	ret = XIic_SetAddress(&xil_dev->instance, XII_ADDR_TO_SEND_TYPE,
 			      dev->slave_address);
 	if(ret != 0)
 		goto error;
@@ -114,8 +126,10 @@ int32_t i2c_remove(struct i2c_desc *desc)
 {
 #ifdef XIIC_H
 	int32_t ret;
+	xil_i2c_desc *xil_desc;
+	xil_desc = desc->extra;
 
-	ret = XIic_Stop(&desc->instance);
+	ret = XIic_Stop(&xil_desc->instance);
 	if(ret != 0)
 		return FAILURE;
 
@@ -140,7 +154,9 @@ int32_t i2c_write(struct i2c_desc *desc,
 		  uint8_t stop_bit)
 {
 #ifdef XIIC_H
-	return XIic_Send(desc->instance.BaseAddress, desc->slave_address, data,
+	xil_i2c_desc *xil_desc;
+	xil_desc = desc->extra;
+	return XIic_Send(xil_desc->instance.BaseAddress, desc->slave_address, data,
 			 bytes_number, stop_bit ? XIIC_STOP : XIIC_REPEATED_START);
 #else
 	return SUCCESS;
@@ -163,7 +179,10 @@ int32_t i2c_read(struct i2c_desc *desc,
 		 uint8_t stop_bit)
 {
 #ifdef XIIC_H
-	return XIic_Recv(desc->instance.BaseAddress, desc->slave_address, data,
+	xil_i2c_desc *xil_desc;
+	xil_desc = desc->extra;
+
+	return XIic_Recv(xil_desc->instance.BaseAddress, desc->slave_address, data,
 			 bytes_number, stop_bit ? XIIC_STOP : XIIC_REPEATED_START);
 #else
 	return SUCCESS;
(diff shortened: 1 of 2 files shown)


projects/drivers/xilinx_platform/platform_drivers.h
```diff
@@ -81,141 +74,56 @@ typedef enum i2c_type {
 	XILINX_I2C
 } i2c_type;
 
-typedef struct i2c_init_param {
+typedef struct extra_i2c_init_param {
```
commodo: why the `extra_` prefix ?
shouldn't `i2c_init_param` from i2c.h be sufficient?

commodo: if this is xilinx specific, then maybe move this entirely to `platform_drivers.c` and rename to `xil_i2c_init_param`

projects/drivers/xilinx_platform/platform_drivers.c
```diff
@@ -59,36 +59,42 @@ int32_t i2c_init(struct i2c_desc **desc,
 {
```
mirceacaprioru: you have a typo in the commit message "sturctures"

projects/drivers/xilinx_platform/platform_drivers.c
```diff
@@ -180,55 +193,62 @@ int32_t spi_init(struct spi_desc **desc,
 		 const struct spi_init_param *param)
 {
 	spi_desc *descriptor;
+	extra_spi_desc *extra_descriptor;
+	extra_spi_init_param *extra_param;
 	int32_t ret;
 
 	descriptor = (struct spi_desc *) calloc(1, sizeof(*descriptor));
-	if (!descriptor)
+	extra_descriptor = (struct extra_spi_desc *) calloc(1,
```
mirceacaprioru: i don't think you need casts when allocating memory. I has no actual value. Check this in other places also.

projects/drivers/xilinx_platform/platform_drivers.c
```diff
@@ -180,55 +193,62 @@ int32_t spi_init(struct spi_desc **desc,
...
+			   sizeof(*extra_descriptor));
+	if (!descriptor || !extra_descriptor)
 		return FAILURE;
 
+	descriptor->extra = extra_descriptor;
+	extra_param = param->extra;
```
mirceacaprioru: why use an extra variable for param->extra ?

projects/drivers/xilinx_platform/platform_drivers.c
```diff
@@ -180,55 +193,62 @@ int32_t spi_init(struct spi_desc **desc,
...
 
-	XSpiPs_SetClkPrescaler(&descriptor->instance,
+	XSpiPs_SetClkPrescaler(&extra_descriptor->instance,
 			       XSPIPS_CLK_PRESCALE_64);
 
-	XSpiPs_SetSlaveSelect(&descriptor->instance, 0xf);
+	XSpiPs_SetSlaveSelect(&extra_descriptor->instance, 0xf);
 #else
-	ret = XSpi_Initialize(&descriptor->instance, param->id);
+	ret = XSpi_Initialize(&extra_descriptor->instance, extra_param->id);
```
mirceacaprioru: in some places you use extra_param->id and in others extra_descriptor->id. 

This is happening because of the extra variable you don't actually need, and you end up with inconsistencies.

projects/drivers/xilinx_platform/platform_drivers.c
```diff
@@ -267,30 +287,33 @@ int32_t spi_write_and_read(struct spi_desc *desc,
...
 {
+	extra_spi_desc *extra_desc;
+	extra_desc = desc->extra;
+
 #ifdef _XPARAMETERS_PS_H_
-	XSpiPs_SetOptions(&desc->instance,
+	XSpiPs_SetOptions(&extra_desc->instance,
 			  XSPIPS_MASTER_OPTION |
-			  ((desc->flags & SPI_CS_DECODE) ?
+			  ((((extra_spi_desc *)desc->extra)->flags & SPI_CS_DECODE) ?
```
mirceacaprioru: extra_desc ?

projects/drivers/xilinx_platform/platform_drivers.c
```diff
@@ -309,23 +332,28 @@ int32_t gpio_get(struct gpio_desc **desc,
 		 uint8_t gpio_number)
 {
 	gpio_desc *descriptor;
+	extra_gpio_desc *extra_descriptor;
 	int32_t ret;
 
 	descriptor = (struct gpio_desc *) malloc(sizeof(*descriptor));
-	if (!descriptor)
+	extra_descriptor = (struct extra_gpio_desc *) calloc(1,
+			   sizeof(*extra_descriptor));
```
mirceacaprioru: i would use the structure used to define extra_descriptor, not referencing the actual pointer.

projects/drivers/xilinx_platform/platform_drivers.c
amiclaus: Maybe we should keep the casts for the sake of consistency.

projects/drivers/xilinx_platform/platform_drivers.h
dbogdan: @commodo is right, the header file should be platform agnostic.

amiclaus: Question: should I remove the header file completely within this commit? Since, I suppose, we will use the generic header file: https://github.com/analogdevicesinc/no-OS/blob/master/drivers/platform/generic/platform_drivers.h ?

dbogdan: I guess we can do this step now.

amiclaus:
Changelog v1->v2:

- fix typo in commit message
- use platform specific prefix for structures (`xil_`)
- make header file platform agnostic (delete current header file, since https://github.com/analogdevicesinc/no-OS/blob/master/drivers/platform/generic/platform_drivers.h will be used from now on)
- remove intermediary variable `extra_param`
- allocate memory for the extra structure and define extra descriptor based on it.

projects/drivers/xilinx_platform/platform_drivers.c
dbogdan: There are arguments for keeping the casting (https://www.gnu.org/software/libc/manual/html_node/Basic-Allocation.html), but since we are following the Linux kernel coding guidelines, we shouldn't do it.

projects/drivers/xilinx_platform/platform_drivers.c
```diff
@@ -40,10 +40,87 @@
 /******************************************************************************/
 /***************************** Include Files **********************************/
 /******************************************************************************/
+#include <stdint.h>
```
dbogdan: Do we really need this here?

projects/drivers/xilinx_platform/platform_drivers.c
```diff
@@ -40,10 +40,87 @@
...
+
+#define SPI_CS_DECODE	0x01
+
+/******************************************************************************/
+/*************************** Types Declarations *******************************/
+/******************************************************************************/
+
+typedef enum i2c_type {
+	XILINX_I2C
+} i2c_type;
```
dbogdan: This structure seems to be useless. Same as spi_type and gpio_type.

projects/drivers/xilinx_platform/platform_drivers.c
amiclaus: using param->extra requires casting each time it is accessed.
In my opinion, adding an intermediary variable will improve the code readability. 

I.e. 
`xil_dev->id = ((xil_i2c_desc *)param->extra)->id;`
vs
`xil_dev->id = xil_param->id;`

amiclaus: In this manner we avoid casting as much as possible.

projects/drivers/xilinx_platform/platform_drivers.c
amiclaus: These enumerations are used as attributes inside the descriptor/init_param structures. I suppose their purpose is to support extensions for multiple i2c types.

projects/drivers/xilinx_platform/platform_drivers.c
amiclaus: Nope.

amiclaus:
Changelog v2->v3:

- remove `stdint.h`
- create separate header file `xil_drivers.h` containing xilinx specific header files and structures that are used across multiple source files of the project.

projects/drivers/xilinx_platform/platform_drivers.c
```diff
@@ -180,55 +195,64 @@ int32_t spi_init(struct spi_desc **desc,
 		 const struct spi_init_param *param)
 {
 	spi_desc *descriptor;
+	xil_spi_desc *xil_descriptor;
+	xil_spi_init_param *xil_param;
 	int32_t ret;
 
-	descriptor = (struct spi_desc *) calloc(1, sizeof(*descriptor));
+	descriptor = calloc(1, sizeof(*descriptor));
```
adrimbarean: nitpick: no need for parenthesis in `sizeof(*descriptor)`; it can be just `sizeof *descriptor`.
This can be checked in other places as well. `sizeof` only requires parenthesis when it is done with a type (e.g. `sizeof (int)`), or with an expression (e.g. `sizeof (a+b)`).

amiclaus:
Changelog v3->v4:

- adding back `platform_drivers.h` generic header after new discussions on this topic.

amiclaus:
Changelog v4-v5:

- minor fix: remove unnecessary brackets.

PR #218, opened 2019-08-19
ad7124: Adding #defines for configuration registers bits

mike-bradley:

### DIFF
--- a/drivers/adc/ad7124/ad7124.h
+++ b/drivers/adc/ad7124/ad7124.h
@@ -208,6 +208,7 @@
 #define AD7124_ERR_REG_SPI_WRITE_ERR      (1 << 3)
 #define AD7124_ERR_REG_SPI_CRC_ERR        (1 << 2)
 #define AD7124_ERR_REG_MM_CRC_ERR         (1 << 1)
+#define AD7124_ERR_REG_ROM_CRC_ERR        (1 << 0)
 
 /* Error_En Register bits */
 #define AD7124_ERREN_REG_MCLK_CNT_EN           (1 << 22)
@@ -231,6 +232,7 @@
 #define AD7124_ERREN_REG_SPI_WRITE_ERR_EN      (1 << 3)
 #define AD7124_ERREN_REG_SPI_CRC_ERR_EN        (1 << 2)
 #define AD7124_ERREN_REG_MM_CRC_ERR_EN         (1 << 1)
+#define AD7124_ERREN_REG_ROM_CRC_ERR_EN        (1 << 0)
 
 /* Channel Registers 0-15 bits */
 #define AD7124_CH_MAP_REG_CH_ENABLE    (1 << 15)


drivers/adc/ad7124/ad7124.h
```diff
@@ -247,6 +249,9 @@
 #define AD7124_CFG_REG_REF_SEL(x)  ((x) & 0x3) << 3
 #define AD7124_CFG_REG_PGA(x)      (((x) & 0x7) << 0)
 
+/* PGA Gain Value */
+#define AD7124_PGA_GAIN(x)     (1 << (x))
```
mirceacaprioru: why not just use AD7124_CFG_REG_PGA for PGA configuration. 

This definition will only allow gain: 1, 2, 4, 16.
Check page 90 of https://www.analog.com/media/en/technical-documentation/data-sheets/AD7124-8.pdf

mike-bradley: The intent of AD7124_PGA_GAIN is to provide a method to calculate the actual gain values (1, 2, 4, 8, 16, 32, 64, 128) that correspond to the 3-bit config values stored in the PGA bit field of the CFG_REG. The AD7124_PGA_GAIN 'function' is useful when converting raw DATA register values to a voltage.

adrimbarean:
Commit needs a `Signed-of-by` tag and a mention to the driver affected by the change.

mirceacaprioru:
I still don't understand the need for this. It could just be me not figuring out where this will be used. This is the problem when there is an isolated commit in a PR with nothing actually using this added functionality. 

I think the best thing here would be to do more commits in this pull request with the usage of this.

commodo:
> I still don't understand the need for this. It could just be me not figuring out where this will be used. This is the problem when there is an isolated commit in a PR with nothing actually using this added functionality.
> 
> I think the best thing here would be to do more commits in this pull request with the usage of this.

Personally, I'm a bit divided/undecided as well when adding register definitions that are not used.

For @mike-bradley to get a feeling/understanding about what/why we're talking about here: a lot of the comments we provide here, are derived from feedback we got from the Linux community.
Now, one might suspect that the Linux community is as colorful as the number of maintainers there are (and that's actually true), but some common rules do emerge, as well as exceptions to the rules.

One rule is: when adding register definitions, or functions, or code, it's a good idea to have a user of this added code/stuff. Mostly so that people understand why it's being added.
Naturally, some other maintainers don't mind having definitions of all registers & bits; this makes the code look like a C-version of the datasheet, and that is fine.

In this case, we could go with the latter (#defines as C-versions of the datasheet), also because the no-OS code is used in many ways/places that we don't know about, so having the definitions already in here, could help with other projects/users of this code.
For Linux it's generally easy (with respect to users of a piece of code), as the kernel is pretty much self-contained.

mike-bradley:
Thanks @commodo for the background and philosophy behind the feedback - that really helps set the bar and criteria for for acceptance of changes.  Based on this, I'd break this pull request into 2 parts.
1) the two register bits #defines AD7124_ERR_REG_ROM_CRC_ERR and AD7124_ERREN_REG_ROM_CRC_ERR_EN are adding bits that are missing, but could be used in the future.  Adding these helps to improve the completeness of the 'C register data sheet'. 
2) the AD7124_PGA_GAIN(x)  was added to aid with converting raw counts to voltages.  I'm using a function similar to this to do the conversion:
`convertedValue = (DATA_REG_SAMPLE * AD7124_REF_VOLTAGE)/(AD7124_PGA_GAIN(channelPGA) * (1 << AD7124_ADC_N_BITS));`
Based on the input above, it would make more sense to include this #define in a future pull request that includes the conversion support functions that need it.

mirceacaprioru:
Should we close this PR then ? I am basing this on your comment above but would like your input. 

Thanks.

mike-bradley:
I would prefer to revise this pull request so it only adds the missing #defines for AD7124_ERR_REG_ROM_CRC_ERR and AD7124_ERREN_REG_ROM_CRC_ERR_EN bits. All the other bits in those registers are #define'd in the header, so that would complete the register definitions in ad714.h.

The AD7124_PGA_GAIN(x) #define would not be included as part of this of this pull request, so I can remove it.

PR #221, opened 2019-08-22
ad9361: update project to latest platform drivers

amiclaus:
Use the reorganized Xilinx platform drivers for the AD9361 project.

### DIFF
--- a/projects/ad9361/src/ad9361_api.c
+++ b/projects/ad9361/src/ad9361_api.c
@@ -43,6 +43,7 @@
 #include "ad9361.h"
 #include "ad9361_api.h"
 #include "platform_drivers.h"
+#include "xil_drivers.h"
 #include "axi_adc_core.h"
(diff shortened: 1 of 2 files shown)


projects/ad9361/src/ad9361_api.c
```diff
@@ -125,7 +126,7 @@ int32_t ad9361_init (struct ad9361_rf_phy **ad9361_phy,
 	phy->dev_sel = init_param->dev_sel;
 
 	/* Identification number */
-	phy->spi->id = init_param->id_no;
+	((xil_spi_desc*)phy->spi->extra)->id = init_param->id_no;
```
dbogdan: The API must remain platform agnostic.

mirceacaprioru: I think for the Xilinx projects a general direction should be established, otherwise we just keep modifying things all around with no consistency.

PR #224, opened 2019-08-26
Read the status register with each conversion of the ADC.

malavikasajikumar:
- Enabled the data status bit in the ADC control register.
- Increased the SPI received data size by one.

### DIFF
--- a/drivers/adc/ad7124/ad7124_regs.c
+++ b/drivers/adc/ad7124/ad7124_regs.c
@@ -2,8 +2,8 @@
 
 struct ad7124_st_reg ad7124_regs[AD7124_REG_NO] = {
 	{0x00, 0x00,   1, 2}, /* AD7124_Status */
-	{0x01, 0x0000, 2, 1}, /* AD7124_ADC_Control */
-	{0x02, 0x0000, 3, 2}, /* AD7124_Data */
+	{0x01, 0x0400, 2, 1}, /* AD7124_ADC_Control */
+	{0x02, 0x0000, 4, 2}, /* AD7124_Data */
 	{0x03, 0x0000, 3, 1}, /* AD7124_IOCon1 */
 	{0x04, 0x0000, 2, 1}, /* AD7124_IOCon2 */
 	{0x05, 0x02,   1, 2}, /* AD7124_ID */


adrimbarean:
We need the Contributor License Agreement signed to merge the PR.

drivers/adc/ad7124/ad7124_regs.c
```diff
@@ -2,8 +2,8 @@
```
adrimbarean: The commit needs to be signed, so do `git commit -s` when commiting.
This will add text in the form of:
`Signed-off-by: Name Surname <name.surname@domain.com>`

drivers/adc/ad7124/ad7124_regs.c
```diff
@@ -2,8 +2,8 @@
 
 struct ad7124_st_reg ad7124_regs[AD7124_REG_NO] = {
 	{0x00, 0x00,   1, 2}, /* AD7124_Status */
-	{0x01, 0x0000, 2, 1}, /* AD7124_ADC_Control */
-	{0x02, 0x0000, 3, 2}, /* AD7124_Data */
+	{0x01, 0x0400, 2, 1}, /* AD7124_ADC_Control */
+	{0x02, 0x0000, 4, 2}, /* AD7124_Data */
```
adrimbarean: This change in the initialization structure of the driver can work for an individual application on a case by case basis, but I don't think it's fit in the general driver. If another user were to want to read data without the status register he would have to undo this change and recompile the program, because the `ad7124_read_data()` API will always try to read 4 bytes from now on.

In my opinion, if we want to add support for reading status register with the data read, we should add an API for that, so that the driver always works regardless of how the user wants to use it. This can be done a number of ways:

1. My personal go-to would be to add a specific function to enable the data with status read with the prototype something like `ad7124_data_status_set(struct ad7124_dev *dev, bool enable)`. Then add a member to the driver handler that always keeps track of the status of this bit and update the `ad7124_read_data()` function to check this member and read one more byte if it(the member) is asserted.
Pros: Easiest to use for the user.
Cons: If the register in the device changes through other ways than than specialized API, this implementation may not work.

2. Just update the `ad7124_read_data()` function to check the data_status bit in the register table and turn this feature on and off using the already existing `ad7124_write_register()` API. This, in theory, does not have the same con as the last point, but you can never be sure unless you read the ADC control register every time you want to read the data. And that can get troubling too.
Pros: Easier to implement than 1;
Cons: Requires more effort from the user to use;

3. Other implementation that I did not think of.

mike-bradley: I submitted a  pull request #214 with implementation #2, but pulled that change out as it was too large, but in the process of re-submitting the DATA with STATUS as part of a series of smaller changes to this driver at the moment.  Could review option 2 then, but agree opt 1 may be better.

adrimbarean:
Change covered in another patch:
https://github.com/analogdevicesinc/no-OS/pull/228

Closing.

PR #225, opened 2019-08-27
ad7124: Adding ADI license to register setting files

mike-bradley:

### DIFF
--- a/drivers/adc/ad7124/ad7124_regs.c
+++ b/drivers/adc/ad7124/ad7124_regs.c
@@ -1,3 +1,44 @@
+/***************************************************************************//**
+*   @file    ad7124-regs.c
+*   @brief   AD7124 register file.
+*   @devices AD7124-4, AD7124-8
+*
+********************************************************************************
+* Copyright 2015-2019(c) Analog Devices, Inc.
+*
+* All rights reserved.
+*
+* Redistribution and use in source and binary forms, with or without modification,
+* are permitted provided that the following conditions are met:
+*  - Redistributions of source code must retain the above copyright
+*    notice, this list of conditions and the following disclaimer.
+*  - Redistributions in binary form must reproduce the above copyright
+*    notice, this list of conditions and the following disclaimer in
+*    the documentation and/or other materials provided with the
+*    distribution.
+*  - Neither the name of Analog Devices, Inc. nor the names of its
+*    contributors may be used to endorse or promote products derived
+*    from this software without specific prior written permission.
+*  - The use of this software may or may not infringe the patent rights
+*    of one or more patent holders.  This license does not release you
+*    from the requirement that you obtain separate licenses from these
+*    patent holders to use this software.
+*  - Use of the software either in source or binary form, must be run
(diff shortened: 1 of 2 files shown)


adrimbarean:
Looks good to me. Just one comment: I think it's better if we can tell by the commit message which files are changed. I suggest writing the title like `ad7124: Adding ADI license files` or mentioning this in the body of the commit.

mirceacaprioru:
One question should we have  Copyright 2015-2019(c) Analog Devices, Inc. ?

mike-bradley:
Date of 2015 was based on when the files would have been created originally, can edit to be 2015-2019 if required

drivers/adc/ad7124/ad7124_regs.c
```diff
@@ -1,3 +1,44 @@
+/***************************************************************************//**
+*   @file    AD7124-regs.c
```
commodo: it's a good idea to keep filename consistency here;
the filename is `ad7124_regs.c`, so it would be a good idea to also name this the same;

drivers/adc/ad7124/ad7124_regs.c
```diff
@@ -1,3 +1,44 @@
...
+*   @brief   AD7124 register file.
+*   @devices AD7124-4, AD7124-8
+*
+********************************************************************************
+* Copyright 2015(c) Analog Devices, Inc.
```
commodo: curios: why 2015?
why not something newer;
assuming it's a year range [because the driver was probably written in 2015], you could also do `2015-2019`

drivers/adc/ad7124/ad7124_regs.h
```diff
@@ -1,3 +1,44 @@
+/***************************************************************************//**
+*   @file    AD7124-regs.h
```
commodo: same comment about filename

drivers/adc/ad7124/ad7124_regs.h
```diff
@@ -1,3 +1,44 @@
...
+*   @brief   AD7124 register file.
+*   @devices AD7124-4, AD7124-8
+*
+********************************************************************************
+* Copyright 2015(c) Analog Devices, Inc.
```
commodo: same comment about year;

commodo:
> Date of 2015 was based on when the files would have been created originally, can edit to be 2015-2019 if required

oh: i just saw this comment;
disregard mine

drivers/adc/ad7124/ad7124_regs.c
mike-bradley: done

drivers/adc/ad7124/ad7124_regs.c
mike-bradley: done.

drivers/adc/ad7124/ad7124_regs.h
mike-bradley: done.

drivers/adc/ad7124/ad7124_regs.h
mike-bradley: done

PR #226, opened 2019-08-27
ad7124: Ensuring CRC is disabled after reset

mike-bradley:
need to make sure CRC is disabled after a reset, otherwise there may be communication issues.

### DIFF
--- a/drivers/adc/ad7124/ad7124.c
+++ b/drivers/adc/ad7124/ad7124.c
@@ -231,6 +231,9 @@ int32_t ad7124_reset(struct ad7124_dev *dev)
 				 wr_buf,
 				 8);
 
+	/* CRC is disabled after reset */
+	dev->use_crc = AD7124_DISABLE_CRC;
+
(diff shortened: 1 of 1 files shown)


drivers/adc/ad7124/ad7124.c
```diff
@@ -231,6 +231,9 @@ int32_t ad7124_reset(struct ad7124_dev *dev)
 				 wr_buf,
 				 8);
 
+	/* CRC is disabled after reset */
+	dev->use_crc = AD7124_DISABLE_CRC;
```
adrimbarean: Good catch! Should we use `ad7124_update_crcsetting` to do this or is it better like this?

mike-bradley: I think it needs to be explicit like this, the ad7124_update_crcsettings function conditionally enables/disables the CRC based on a the dev->regs stored value of a bit in the Error_En register.  After a reset, the CRC is always disabled, so the conditional check can't be guaranteed valid.

PR #228, opened 2019-08-27
ad7124: add support to read DATA reg with STATUS byte when enabled

mike-bradley:
AD7124: Implementation to support STATUS byte appended to ADC DATA register read, taking different approach to #224.

### DIFF
--- a/drivers/adc/ad7124/ad7124.c
+++ b/drivers/adc/ad7124/ad7124.c
@@ -66,7 +66,7 @@ int32_t ad7124_no_check_read_register(struct ad7124_dev *dev,
 	int32_t ret = 0;
 	uint8_t buffer[8] = {0, 0, 0, 0, 0, 0, 0, 0};
 	uint8_t i = 0;
-	uint8_t check8 = 0;
+	uint8_t check8 = 0, add_status_length = 0;
 	uint8_t msg_buf[8] = {0, 0, 0, 0, 0, 0, 0, 0};
 
 	if(!dev || !p_reg)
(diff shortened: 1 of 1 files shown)


adrimbarean:
Looks good to me. Just a comment: I think we should be able to tell from the commit message what files or what drivers are affected by the change.

mirceacaprioru:
As a general remark you could put more commits into a single pull request. Especially if the commits have common functionality and refer to the same device. Right now they are hard to follow.

We usually create PR containing up to 5 commits. This will make it easier for the review process.

drivers/adc/ad7124/ad7124.c
```diff
@@ -76,29 +76,46 @@ int32_t ad7124_no_check_read_register(struct ad7124_dev *dev,
 	buffer[0] = AD7124_COMM_REG_WEN | AD7124_COMM_REG_RD |
 		    AD7124_COMM_REG_RA(p_reg->addr);
 
+	/*
+	 * If this is an AD7124_DATA register read, and the DATA_STATUS bit is set
+	 * in ADC_CONTROL, need to read 4, not 3 bytes for DATA with STATUS
+	 */
+	if ((p_reg->addr == AD7124_DATA_REG) &&
+	    (dev->regs[AD7124_ADC_Control].value & AD7124_ADC_CTRL_REG_DATA_STATUS)) {
+		add_status_length = 1;
```
commodo: i understood the gist of this change;
but [the change] points out that this code should have been written in a neater fashion [when it was originally written];

nothing needs to change here; this change is fine as-is;
just a thought :)

commodo:
will merge this tomorrow if no objections

PR #229, opened 2019-08-28
Revert "changing how POR is handled"

mirceacaprioru:
Reverts analogdevicesinc/no-OS#227


mirceacaprioru:
This need some explanation, why the delay is necessary. Right now it still waits for POR_FLAG the delay is not needed.

PR #235, opened 2019-09-04
ADV7511: move project to projects/ folder

adrimbarean:
Make this change first before reworking to avoid confusions.


adrimbarean:
Will merge later today if everything is good.

PR #240, opened 2019-09-05
projects: adrv9009: Fix README

amiclaus:
Fix `ifdef` statement.


dbogdan:
Please do the same for: https://github.com/analogdevicesinc/no-OS/blob/master/projects/ad9371/src/README#L4

PR #242, opened 2019-09-20
Ad fmcjesdadc1 ebz fix

cristiantpop:
Fix SPI init value for spi_init.type
This parameter is platform dependent and some compiler switch has been
added.

### DIFF
--- a/ad-fmcjesdadc1-ebz/ad9250.c
+++ b/ad-fmcjesdadc1-ebz/ad9250.c
@@ -105,6 +105,10 @@ int32_t ad9250_setup(struct ad9250_dev **device,
 
 	/* Setup SPI descriptor */
 	ret = spi_init(&dev->spi_dev, &init_param.spi_init);
+	if(ret < 0) {
+		ad_printf("Error spi_init()\n");
+		return ret;
+	}
 
 	dev->id_no = init_param.id_no;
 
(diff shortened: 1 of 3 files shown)


ad-fmcjesdadc1-ebz/ad9250.c
```diff
@@ -105,6 +105,10 @@ int32_t ad9250_setup(struct ad9250_dev **device,
 
     /* Setup SPI descriptor */
     ret = spi_init(&dev->spi_dev, &init_param.spi_init);
+	if(ret < 0) {
+		ad_printf("Error spi_init()\n");
```
mirceacaprioru: if you just want to print the function name you can use __func__ if the ad_printf accepts parameters like %s

ad_printf("Error %s\n", __func__);

ad-fmcjesdadc1-ebz/ad_fmcjesdadc1_ebz.c
```diff
@@ -88,9 +88,29 @@ int main(void)
 	rx_xfer_1.start_address = XPAR_DDR_MEM_BASEADDR + 0x900000;
 #endif
 
+#ifdef ZYNQ_PS7
```
mirceacaprioru: we should try to keep these compiler switches to a minimum. If someting can be done without them then use that option. I'm not sure in this case what could be done.

cristiantpop:
v2:
Fix travis errors

ad-fmcjesdadc1-ebz/ad9250.c
cristiantpop: In this case the current function is not spi_init but ad9250_setup, but the error is from SPI

mirceacaprioru:
will merge later today

PR #254, opened 2019-09-25
Drivers & ADRV9009 project updates.

amiclaus:
1. Update drivers using `axi_io` wrapper functions. Make drivers platform agnostic.

2. Update ADRV9009 main function using `axi_io` functions and platform specific initialization parameters.

### DIFF
--- a/projects/drivers/axi_adc_core/axi_adc_core.h
+++ b/projects/drivers/axi_adc_core/axi_adc_core.h
@@ -52,12 +52,16 @@ struct axi_adc {
 	uint32_t base;
 	uint8_t	num_channels;
 	uint64_t clock_hz;
+	int32_t (*axi_io_read)(uint32_t, uint32_t, uint32_t*);
+	int32_t (*axi_io_write)(uint32_t, uint32_t, uint32_t);
 };
 
 struct axi_adc_init {
 	const char *name;
 	uint32_t base;
 	uint8_t	num_channels;
+	int32_t (*axi_io_read)(uint32_t, uint32_t, uint32_t*);
+	int32_t (*axi_io_write)(uint32_t, uint32_t, uint32_t);
 };
 
 enum axi_adc_pn_sel {
--- a/projects/drivers/axi_dac_core/axi_dac_core.c
+++ b/projects/drivers/axi_dac_core/axi_dac_core.c
@@ -44,12 +44,6 @@
 #include <stdio.h>
 #include <inttypes.h>
 #include "platform_drivers.h"
-#ifdef ALTERA_PLATFORM
-#include "io.h"
-#else
-#include "xil_io.h"
-#include "xil_cache.h"
-#endif
 #include "util.h"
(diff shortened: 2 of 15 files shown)


amiclaus:
Changelog v1->v2: 

- fix code style issues

projects/drivers/axi_adc_core/axi_adc_core.h
```diff
@@ -52,12 +52,16 @@ struct axi_adc {
 	uint32_t base;
 	uint8_t	num_channels;
 	uint64_t clock_hz;
+	int32_t (*axi_io_read)(uint32_t, uint32_t, uint32_t*);
```
mirceacaprioru: why does ad9528 has the extra parameter and here we have the read/write function prototype ?

amiclaus: The `extra` void pointer is used to point at platform specific parameters (i.e. Xilinx/Altera spi config params) while `axi_io_read` and `axi_io_write` are pointers to functions that wrap read/write platform specific functions for AXI IO interface.

projects/drivers/axi_dac_core/axi_dac_core.c
```diff
@@ -796,16 +772,9 @@ int32_t axi_dac_set_buff(struct axi_dac *dac,
...
 		data_q = (buff[index + 1] << 16);
-#ifdef ALTERA_PLATFORM
-		IOWR_32DIRECT(address, index * 2, data_i | data_q);
-#else
-		Xil_Out32(address + index * 2, data_i | data_q);
-#endif
-	}
 
-#ifndef ALTERA_PLATFORM
-	Xil_DCacheFlush();
```
adrimbarean: This functionality is completely removed. Don't we still need this?

projects/drivers/jesd204/axi_jesd204_rx.h
```diff
@@ -73,6 +75,8 @@ struct jesd204_rx_init {
 	uint8_t subclass;
 	uint32_t device_clk_khz;
 	uint32_t lane_clk_khz;
+	int32_t (*axi_io_read)(uint32_t, uint32_t, uint32_t*);
+	int32_t(*axi_io_write)(uint32_t, uint32_t, uint32_t);
```
adrimbarean: Missing space here. Does it still compile like this?

projects/drivers/axi_dac_core/axi_dac_core.c
amiclaus: The project is working without the Cache flush. Keeping these functionalities will imply adding a special wrapper for this function to make the driver platform agnostic (besides axi_io read/write). So, I believe this subject is debatable.

dbogdan: We need it - it should be moved to the main() function.

amiclaus:
Changelog v2->v3:

- add missing space
- perform cache flush inside main function after loading custom data.

PR #266, opened 2019-10-08
irq: add irq interface header

cristiantpop:
This will abstract platform specific implementation

### DIFF
--- a/include/irq.h
+++ b/include/irq.h
@@ -1,7 +1,7 @@
 /***************************************************************************//**
- *   @file   xilinx_platform_drivers.h
- *   @brief  Header file of Xilinx Platform Drivers.
- *   @author Antoniu Miclaus (antoniu.miclaus@analog.com)
+ *   @file   irq.h
+ *   @brief  Header file of IRQ interface.
+ *   @author Cristian Pop (cristian.pop@analog.com)
 ********************************************************************************
  * Copyright 2019(c) Analog Devices, Inc.
  *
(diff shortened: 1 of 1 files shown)


include/irq.h
```diff
@@ -0,0 +1,85 @@
...
+
+#ifndef IRQ_H_
+#define IRQ_H_
+#include <stdint.h>
+
+/******************************************************************************/
+/*************************** Types Declarations *******************************/
+/******************************************************************************/
+
+typedef struct irq_init_param {
```
adrimbarean: Do we still want to keep typedefs for new implementations? The [Linux kernel coding style](https://www.kernel.org/doc/html/v4.10/process/coding-style.html#typedefs) we want to abide by, suggests against this type of usage.

cristiantpop:
V2:
- remove typedefs

include/irq.h
```diff
@@ -0,0 +1,85 @@
...
+
+/* Initialize the interrupt request peripheral. */
+int32_t irq_init(struct irq_desc **desc,
+		 const struct irq_init_param *param);
+
+/* Free the resources allocated by irq_init(). */
+int32_t irq_remove(struct irq_desc *desc);
+
+/* Registers a generic IRQ handling function */
+int32_t irq_register(struct irq_desc *desc, uint32_t irq_id,
```
dbogdan: irq_desc already contains the irq_id information. Why do we need it here?

include/irq.h
```diff
@@ -0,0 +1,85 @@
...
+	void		*extra;
+} irq_desc;
+
+/******************************************************************************/
+/************************ Functions Declarations ******************************/
+/******************************************************************************/
+
+/* Initialize the interrupt request peripheral. */
+int32_t irq_init(struct irq_desc **desc,
+		 const struct irq_init_param *param);
```
dbogdan: What's the difference between irq_init() and irq_register()? Can't we have only one function? I fell that both are mandatory to be called.

projects/drivers/xilinx_platform/xilinx_platform_drivers.h
```diff
@@ -65,6 +66,10 @@
 /*************************** Types Declarations *******************************/
 /******************************************************************************/
 
+struct xil_irq_desc {
+	XScuGic 		*gic;
+} xil_irq_desc;
```
dbogdan: This won't compile on MicroBlaze platforms.

include/irq.h
cristiantpop: irq_init is like i2c_init, where the global interrupt system is initialized. 
irq_register is used to register a secondary interrupt like: UART, SPI, Timer...

include/irq.h
cristiantpop: "irq_desc already contains the irq_id": this is the global interrupt ID.
"irq_id" parameter in "irq_register" function refers to the id of the interrupt that needs to be register. For UART, an irq_id needs to be provided.

projects/drivers/xilinx_platform/xilinx_platform_drivers.h
dbogdan: We should at least add "#ifdef _XPARAMETERS_PS_H_".

include/irq.h
dbogdan: The naming is a little bit confusing for me. :)

cristiantpop:
V3:
- Add "#ifdef XPARAMETERS_PS_H".

mirceacaprioru:
if there are no other remarks will merge later today.

PR #267, opened 2019-10-09
fifo: add fifo implementation

cristiantpop:
This is used in UART communication for tinyiiod project.
Save data received from UART, into a fifo as soon as they arrive, so that the UART buffer
doesn't get overwritten.

### DIFF
--- a/projects/drivers/util/fifo.c
+++ b/projects/drivers/util/fifo.c
@@ -1,0 +1,136 @@
+/***************************************************************************//**
+ *   @file   fifo.c
+ *   @brief  Implementation of fifo.
+ *   @author Cristian Pop (cristian.pop@analog.com)
+********************************************************************************
+ * Copyright 2019(c) Analog Devices, Inc.
+ *
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *  - Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ *  - Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in
+ *    the documentation and/or other materials provided with the
+ *    distribution.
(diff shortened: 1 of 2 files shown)


projects/drivers/util/fifo.h
```diff
@@ -0,0 +1,69 @@
...
+
+/******************************************************************************/
+/*************************** Types Declarations *******************************/
+/******************************************************************************/
+
+struct fifo {
+	struct fifo *next;
+	char *data;
+	uint16_t len;
+	uint16_t index;
```
adrimbarean: `index` might be a reserved name in some cases. It might be safer to use something else.

projects/drivers/util/fifo.c
```diff
@@ -0,0 +1,140 @@
...
+struct fifo * fifo_remove_head(struct fifo *p_fifo)
+{
+	struct fifo *p = p_fifo;
+
+	if (p_fifo != NULL) {
+		p_fifo = p_fifo->next;
+		free(p->data);
+		p->len = 0;
+		p->next = NULL;
+		p->data = NULL;
```
adrimbarean: Are these still necessary if we `free(p)` later anyway?

mirceacaprioru: these don't count anymore.

projects/drivers/util/fifo.h
```diff
@@ -0,0 +1,69 @@
...
+/******************************************************************************/
+
+#include <stdint.h>
+
+/******************************************************************************/
+/*************************** Types Declarations *******************************/
+/******************************************************************************/
+
+struct fifo {
+	struct fifo *next;
```
dbogdan: This should be a fifo element.

cristiantpop:
v2:
- remove set values with NULL when removing fifo head element.
- remove index from fifo_element
- rename functions to: "fifo_remove_head", "fifo_insert_tail" to "fifo_remove", "fifo_insert"

mirceacaprioru:
will merge later today if there are no remarks.

PR #271, opened 2019-10-09
projects:  update README

amiclaus:
This patch adds the new source file paths required by some no-OS projects
after spliting platform drivers into separate modules and making drivers
platform agnostic.

### DIFF
--- a/projects/ad9172/src/README
+++ b/projects/ad9172/src/README
@@ -1,8 +1,19 @@
 
 # Additional required source files:
 
-	cp ../../drivers/xilinx_platform/platform_drivers.h	./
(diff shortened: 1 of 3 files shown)


mirceacaprioru:
If we have three README modifications and these are similiar you could include all 3 commits in one PR without the extra noise of doing each one in their own PR. Also please add a bit more context to the commit messages something like:

> The new version of the projects has been moved to a new location and the drivers used have been modified to become platform agnostic. This patch adds the new files required in the README with their correct path.

amiclaus:
Changelog v1->v2:

- add all commits into one single PR.
- Improve description of the patches in the commit messages.

PR #272, opened 2019-10-10
xml: Add xml generator

cristiantpop:
This is used in tinyiiod project to generate xml configuration string that
describes a device: channels, attributes.
This string is sent to iio-oscilloscope(libiio) when a "PRINT"
command is received.

### DIFF
--- a/projects/drivers/util/xml.c
+++ b/projects/drivers/util/xml.c
@@ -1,0 +1,336 @@
+/***************************************************************************//**
+ *   @file   xml.c
+ *   @brief  Implementation of xml
+ *   @author Cristian Pop (cristian.pop@analog.com)
+********************************************************************************
+ * Copyright 2019(c) Analog Devices, Inc.
+ *
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *  - Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ *  - Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in
+ *    the documentation and/or other materials provided with the
+ *    distribution.
+ *  - Neither the name of Analog Devices, Inc. nor the names of its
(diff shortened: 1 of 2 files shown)


projects/drivers/util/xml.h
```diff
@@ -0,0 +1,102 @@
...
+#include "stdio.h"
+
+/******************************************************************************/
+/*************************** Types Declarations *******************************/
+/******************************************************************************/
+
+struct xml_attribute {
+	char *name;
+	char *value;
+} xml_attribute;
```
adrimbarean: I think this is an instantiation of a variable with the name `xml_attribute` of the type `struct xml_attribute`. I don't think it belongs here, it maybe is just a remainder from the `typedef` days.

projects/drivers/util/xml.c
```diff
@@ -0,0 +1,281 @@
...
+			return FAILURE;
+	}
+
+	len = strlen(node->name) + 3;
+	char *buff = realloc((*document)->buff, (*document)->index + len);
+	if (!buff)
+		goto error;
+	(*document)->buff = buff;
+	(*document)->index += sprintf(&(*document)->buff[(*document)->index], "<%s ",
+				      node->name);
```
adrimbarean: Here, if `node->name` is `NULL` this could be a weird error and would cause the index to go backwards. Can that happen?

cristiantpop:
V2:
- remove remaining from typedef.

projects/drivers/util/xml.c
cristiantpop: Checking for NULL values is not done anywhere in the code. Should we check  for NULL values here?

mirceacaprioru: since this implementation is pretty complex with arrays of pointers it would be wise to check for null.

cristiantpop:
V4:

- check pointers for NULL values;
- check "sprintf" return values;
- implement a "xml_print_to_doc" function.

cristiantpop:
Anything new here?

mirceacaprioru:
will merge later today if there are no remarks.

PR #274, opened 2019-10-14
ad9371: Use /projects/ad9371 sources instead

dbogdan:
/projects/ad9371 will be the supported version from now on.


mirceacaprioru:
will merge later today if there are no remaks.

PR #275, opened 2019-10-15
legacy: Move ad-dyno1-ebz

dbogdan:
The projects for which the support has been discontinued will be moved
to legacy folder. They will be kept here for the moment only for legacy
purposes.


mirceacaprioru:
If we are just keeping them for legacy purposes i don't think we should bother with astyle.

PR #283, opened 2019-10-17
uart: Add UART implementation for Xilinx platform

cristiantpop:
This is used in tinyiiod project to communicate with iio-oscilloscope.
This version uses a fifo when receiving, data since there is no handshake
for smaller chunks of data.

### DIFF
--- a/drivers/platform/xilinx/uart.c
+++ b/drivers/platform/xilinx/uart.c
@@ -1,0 +1,400 @@
+/******************************************************************************
+*
+* Copyright (C) 2010 - 2014 Xilinx, Inc.  All rights reserved.
+*
+* Permission is hereby granted, free of charge, to any person obtaining a copy
+* of this software and associated documentation files (the "Software"), to deal
+* in the Software without restriction, including without limitation the rights
+* to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+* copies of the Software, and to permit persons to whom the Software is
+* furnished to do so, subject to the following conditions:
+*
+* The above copyright notice and this permission notice shall be included in
+* all copies or substantial portions of the Software.
+*
+* THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+* IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+* FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL
+* XILINX  BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY,
+* WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF
+* OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+* SOFTWARE.
+*
+* Except as contained in this notice, the name of the Xilinx shall not be used
+* in advertising or otherwise to promote the sale, use or other dealings in
+* this Software without prior written authorization from Xilinx.
+*
+******************************************************************************/
+
+/******************************************************************************/
+/***************************** Include Files **********************************/
+/******************************************************************************/
+
+#include <xparameters.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include "error.h"
+#include "fifo.h"
+#include "irq.h"
+#include "uart.h"
+#include "uart_extra.h"
+#ifdef XPAR_XUARTPS_NUM_INSTANCES
+#include <xil_exception.h>
+#include <xuartps.h>
+#endif
+
+/******************************************************************************/
+/************************ Functions Definitions *******************************/
+/******************************************************************************/
+
+/**
+ * @brief Save received data into fifo.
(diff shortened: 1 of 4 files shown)


drivers/platform/xilinx/uart.c
```diff
@@ -0,0 +1,338 @@
...
+		return FAILURE;
+	}
+
+	/*
+	 * Initialize the UART driver so that it's ready to use
+	 * Look up the configuration in the config table, then initialize it.
+	 */
+	config = XUartPs_LookupConfig(descriptor->device_id);
+	if (!config) {
+		return FAILURE;
```
mirceacaprioru: free ? also in the next ones.

drivers/platform/xilinx/uart.c
```diff
@@ -0,0 +1,338 @@
...
+	}
+
+	status = XUartPs_CfgInitialize(xil_uart_desc->instance, config,
+				       config->BaseAddress);
+	if (status != XST_SUCCESS) {
+		return FAILURE;
```
mirceacaprioru: free?

drivers/platform/xilinx/uart.c
```diff
@@ -0,0 +1,338 @@
...
+
+/******************************************************************************/
+/***************************** Include Files **********************************/
+/******************************************************************************/
+
+#include <xparameters.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include "error.h"
+#include "fifo.h"
```
adrimbarean: Do we want the FIFO module dependency? Will we always use the UART module along with the FIFO, or we might want to use just the UART sometimes? I am perfectly OK if we will always use the UART with FIFO, just wanted it to be a conscious decision on our part, I guess.

drivers/platform/xilinx/uart.c
```diff
@@ -0,0 +1,338 @@
...
+ * @param call_back_ref - Instance of UART.
+ * @param event - Event that caused the interrupt.
+ * @param data_len - number of bytes received.
+ */
+static void uart_irq_handler(void *call_back_ref, uint32_t event,
+			     uint32_t data_len)
+{
+	struct xil_uart_desc *xil_uart_desc = call_back_ref;
+	/* All of the data has been received */
+	if (event == XUARTPS_EVENT_RECV_DATA) {
```
adrimbarean: Would this string of `if`'s be better as a `switch` clause?

cristiantpop:
V3:

- fix memory leaks

drivers/platform/xilinx/uart.c
cristiantpop: Probably, I will change that.

drivers/platform/xilinx/uart.c
cristiantpop: We want as less dependencies as possible, but when receiving large amounts of data(from iio-oscilloscope), we are not able to process those fast enough and data gets lost. Introducing fifo fixes this.

drivers/platform/xilinx/uart.c
```diff
@@ -0,0 +1,339 @@
...
+static int32_t uart_irq_init(struct uart_desc *descriptor)
+{
+	uint32_t uart_irq_mask;
+	struct xil_uart_desc *xil_uart_desc = descriptor->extra;
+
+	int32_t status = irq_register(xil_uart_desc->irq_desc, xil_uart_desc->irq_id,
+				      (Xil_ExceptionHandler) XUartPs_InterruptHandler, xil_uart_desc->instance);
+	if (status < 0)
+		return status;
+	XUartPs_SetHandler(xil_uart_desc->instance, uart_irq_handler, xil_uart_desc);
```
mirceacaprioru: will this work with any xilinx platform ?

include/uart.h
```diff
@@ -0,0 +1,78 @@
...
+*******************************************************************************/
+#ifndef UART_H_
+#define UART_H_
+
+/******************************************************************************/
+/*************************** Types Declarations *******************************/
+/******************************************************************************/
+
+struct uart_init_par {
+	uint32_t	device_id;
```
mirceacaprioru: is uint32 needed for device id?

cristiantpop:
V4:

- Declare "uint8_t device_id".

- Add platform checks to see what platform it is used, and call platform specific functions. 

- Use "switch" statement instead of "if"s in "uart_irq_handler".

cristiantpop:
V5:

- update commit message.

- update comment to: "/* Read data to UART. */" in uart.h

- change type from "uint32_t" to "uint8_t" for "device_id" in uart.h

drivers/platform/xilinx/uart.c
```diff
@@ -0,0 +1,400 @@
+/******************************************************************************
+*
+* Copyright (C) 2010 - 2014 Xilinx, Inc.  All rights reserved.
```
mirceacaprioru: i would include in the commit message that this patch offers support only for specific xilinx products, not the entire platform.

cristiantpop:
V5:

- update commit message: specify supported platform.

dbogdan:
Since include/uart.h is not Xilinx platform dependent, I would add it in a separate commit, that will maybe include a uart.c implementation for the generic platform.

cristiantpop:
V6:

- Add a separate commit for "uart.h" generic header.

cristiantpop:
V7:

- add UART generic implementation.

PR #285, opened 2019-10-18
drivers: platform: xilinx: Updated i2c driver

scuciurean:
Removed platform_drivers.h dependency and updated compiler
switches to support both PS and PL implementation of a i2c
controller.
The driver is now updated to use the platform agnostic extra
features passed in the i2c descriptor.
Changed 'stop_bit' param to 'options'.

Signed-off-by: Sergiu Cuciurean sergiu.cuciurean@analog.com

### DIFF
--- a/drivers/platform/xilinx/i2c.c
+++ b/drivers/platform/xilinx/i2c.c
@@ -42,9 +42,18 @@
 /******************************************************************************/
 
 #include <stdlib.h>
+
+#include <xparameters.h>
+#ifdef XPAR_XIIC_NUM_INSTANCES
+#include <xiic.h>
+#endif
+#ifdef XPAR_XIICPS_NUM_INSTANCES
+#include <xiicps.h>
+#endif
+
 #include "error.h"
 #include "i2c.h"
-#include "xilinx_platform_drivers.h"
+#include "i2c_extra.h"
 
 /******************************************************************************/
 /************************ Functions Definitions *******************************/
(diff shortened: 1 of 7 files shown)


drivers/platform/xilinx/i2c.c
```diff
@@ -57,65 +67,98 @@
...
-		goto error;
 
-	*desc = dev;
+		XIicPs_SetSClk(xdesc->instance, init_param->max_speed_hz);
 
+		break;
 #endif
+	/* Intended fallthrough */
+	default:
+		goto not_defined;
```
dbogdan: Shouldn't you goto "error" instead?

drivers/platform/xilinx/i2c.c
```diff
@@ -57,65 +67,98 @@
...
+					 ((XIic_Config*)xdesc->config)
+					 ->BaseAddress);
+		if(ret != SUCCESS)
+			goto error;
 
-	Xil_ExceptionInit();
+		Xil_ExceptionInit();
+		Xil_ExceptionRegisterHandler(XIL_EXCEPTION_ID_IRQ_INT,
+					     XIic_InterruptHandler,
+					     &xdesc->instance);
```
adrimbarean: This part is a bit tricky. If the core should be the only one to generate interrupts in the system it is OK to connect it directly to the ALU interrupt handler. But if there are more cores need to generate interrupts there would need to be an interrupt controller that will be connected to the ALU handler and all the other ones are connected to the interrupt controller. For example when using the I2C core and a timer, or an UART core.

Maybe the driver could check if there is an interrupt controller available, and if there is connect to it instead, and if there is not then do what it does now. Maybe this check can be done with and XPAR_NUM_OF_INSTANCES like the availability of the I2C cores?

drivers/platform/xilinx/i2c_extra.h
```diff
@@ -0,0 +1,69 @@
...
+/***************************** Include Files **********************************/
+/******************************************************************************/
+
+#include <stdint.h>
+
+/******************************************************************************/
+/*************************** Types Declarations *******************************/
+/******************************************************************************/
+
+typedef enum xil_i2c_type {
```
adrimbarean: Should we still be using `typedef`s?

scuciurean: @dbogdan Should we?

drivers/platform/xilinx/i2c.c
scuciurean: Got updated.

scuciurean:
Changelog:

1. Fixed _goto error_ statement in i2c_init to free the memory properly.

mirceacaprioru:
there is a codacy complaint.

drivers/platform/xilinx/i2c_extra.h
dbogdan: We should comply with the Linux kernel coding style: https://github.com/analogdevicesinc/linux/blob/master/Documentation/process/coding-style.rst#5-typedefs

drivers/platform/xilinx/i2c.c
```diff
@@ -152,16 +214,55 @@ int32_t i2c_remove(struct i2c_desc *desc)
...
+	switch (xdesc->type) {
+	case IIC_PL:
 #ifdef XIIC_H
-	xil_i2c_desc *xil_desc;
-	xil_desc = desc->extra;
-	return XIic_Send(xil_desc->instance.BaseAddress, desc->slave_address, data,
-			 bytes_number, stop_bit ? XIIC_STOP : XIIC_REPEATED_START);
-#else
-	return SUCCESS;
+		ret = XIicPs_SetOptions(xdesc->instance,
```
adrimbarean: `XIicPs_SetOptions` is from the PS driver not PL. The PL equivalent would be `XIic_SetOptions` and it returns void, so the return value cannot be tested here. Check for the `i2c_write` function too.

drivers/platform/xilinx/i2c.c
```diff
@@ -57,65 +67,98 @@
...
+					 ->BaseAddress);
+		if(ret != SUCCESS)
+			goto pl_error;
 
-	xil_dev = dev->extra;
-	xil_param = param->extra;
+		Xil_ExceptionInit();
+		Xil_ExceptionRegisterHandler(XIL_EXCEPTION_ID_IRQ_INT,
+					     XIic_InterruptHandler,
+					     &xdesc->instance);
```
adrimbarean: Also the `instance` member of `xdesc` is already a pointer (void pointer) to the struct we need. Giving it by reference here would make it a double pointer. Reference needs to be removed and maybe cast is needed.

drivers/platform/xilinx/i2c.c
```diff
@@ -57,65 +67,98 @@
...
+		ret = XIic_Start(&xdesc->instance);
```
adrimbarean: `instance` member of `xdesc` is already a pointer (void pointer) to the struct we need. Giving it by reference here would make it a double pointer. Reference needs to be removed and maybe cast is needed.

drivers/platform/xilinx/i2c.c
```diff
@@ -57,65 +67,98 @@
...
+		if(ret != SUCCESS)
+			goto pl_error;
 
-	xil_dev->type = xil_param->type;
-	xil_dev->id = xil_param->id;
-	dev->max_speed_hz = param->max_speed_hz;
-	dev->slave_address = param->slave_address;
+		ret = XIic_SetAddress(&xdesc->instance,
```
adrimbarean: Double pointer here too, as above.

scuciurean:
Changelog -> V3:

1. Removed the interrupt assignment to the main IRQ.
2. Updated how the i2c descriptor is passed in several functions.
3. Removed status check for functions that don't return a status.

dbogdan:
The README changes have a different reason: I2C is not used in those projects, so we should have a different commit for it.
Good guidelines regarding this can be found here: https://github.com/analogdevicesinc/linux/blob/master/Documentation/process/submitting-patches.rst

dbogdan:
Don't forget to also update the body of the explanation line wrapping - if I'm not wrong, it is now set to 60 columns for your commit.

scuciurean:
Changelog -> V4:

1. Moved README update to a different commit

PR #286, opened 2019-10-22
projects:drivers:axi_damc: AXI DMAC update

adrimbarean:
Changes this patch:
 - add "remove" function to the driver;
 - make register write and read available for the user;
 - remove "platform_drivers.h" outdated dependency and add "error.h" include;
 - correct small struct error.


adrimbarean:
V0 -> V1:
Add extra change to the patch:
 - remove "platform_drivers.h" outdated dependency and add "error.h" include;

projects/drivers/axi_dmac/axi_dmac.c
```diff
@@ -151,3 +151,16 @@ int32_t axi_dmac_init(struct axi_dmac **dmac_core,
 
 	return SUCCESS;
 }
```
mirceacaprioru: remove the extra : at the start of the commit title

adrimbarean:
V1 -> V2:
 - removed extra : in the commit titles.

PR #288, opened 2019-10-24
axi_adc_core: Add "axi_adc_get_sampling_freq()" function.

cristiantpop:
This is needed in tinyiiod project.

### DIFF
--- a/projects/drivers/axi_adc_core/axi_adc_core.c
+++ b/projects/drivers/axi_adc_core/axi_adc_core.c
@@ -89,6 +89,24 @@ int32_t axi_adc_set_pnsel(struct axi_adc *adc,
 }
 
 /***************************************************************************//**
+ * @brief axi_adc_get_sampling_freq
+*******************************************************************************/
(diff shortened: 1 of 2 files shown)


projects/drivers/axi_adc_core/axi_adc_core.c
```diff
@@ -88,6 +88,24 @@ int32_t axi_adc_set_pnsel(struct axi_adc *adc,
...
+ * @brief axi_adc_get_sampling_freq
+*******************************************************************************/
+int32_t axi_adc_get_sampling_freq(struct axi_adc *adc,
+				  uint32_t chan,
+				  uint64_t *sampling_freq)
+{
+	uint32_t freq;
+	uint32_t ratio;
+
+	axi_adc_read(adc, AXI_ADC_REG_CLK_FREQ, &freq);
```
mirceacaprioru: can't we check the return of this functions ?

dbogdan: Since no other function is doing it, this can be the subject of a future patch.

PR #303, opened 2019-10-31
Iio types update

cristiantpop:
Add "iio_interface_init_par" structure.

This structure will be used together with "iio_register_interface" function
and contains the mapping between a device instance and a "iio_device", that
describes the channels and attributes of a device. Callbacks for read/write
operations, to device instance are also contained here.


cristiantpop:
V2:

- changes in "iio_interface_init_par": add read and write callbacks.

mirceacaprioru:
will merge today.

PR #304, opened 2019-10-31
Update ad5686 driver

ssmith73:
While using the ad5686 driver for mbed firmware I added a couple of features and spotted a small bug - please review.


ssmith73:
I have signed the CLA  - and when i click on the details, it confirms that i have signed it - not sure why this is still pending?

dbogdan:
@ssmith73, it is fine now. The process seems to need some time.

PR #312, opened 2019-11-04
iio: add iio core

cristiantpop:
This module implements read/write ops, required by libtinyiiod and further
calls show/store functions, corresponding to device/channel/attribute.

### DIFF
--- a/iio/iio.c
+++ b/iio/iio.c
@@ -1,0 +1,838 @@
+/***************************************************************************//**
+ *   @file   iio.c
+ *   @brief  Implementation of iio.
+ *   This module implements read/write ops, required by libtinyiiod and further
+ *   calls show/store functions, corresponding to device/channel/attribute.
+ *   @author Cristian Pop (cristian.pop@analog.com)
+********************************************************************************
+ * Copyright 2019(c) Analog Devices, Inc.
+ *
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *  - Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ *  - Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in
+ *    the documentation and/or other materials provided with the
+ *    distribution.
+ *  - Neither the name of Analog Devices, Inc. nor the names of its
+ *    contributors may be used to endorse or promote products derived
+ *    from this software without specific prior written permission.
+ *  - The use of this software may or may not infringe the patent rights
+ *    of one or more patent holders.  This license does not release you
+ *    from the requirement that you obtain separate licenses from these
+ *    patent holders to use this software.
+ *  - Use of the software either in source or binary form, must be run
+ *    on or directly connected to an Analog Devices Inc. component.
+ *
+ * THIS SOFTWARE IS PROVIDED BY ANALOG DEVICES "AS IS" AND ANY EXPRESS OR
+ * IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, NON-INFRINGEMENT,
+ * MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED.
+ * IN NO EVENT SHALL ANALOG DEVICES BE LIABLE FOR ANY DIRECT, INDIRECT,
+ * INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, INTELLECTUAL PROPERTY RIGHTS, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+*******************************************************************************/
+
+/******************************************************************************/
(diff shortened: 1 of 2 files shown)


iio/iio.c
```diff
@@ -0,0 +1,789 @@
...
+			"<!ATTLIST context name CDATA #REQUIRED description CDATA #IMPLIED>"
+			"<!ATTLIST context-attribute name CDATA #REQUIRED value CDATA #REQUIRED>"
+			"<!ATTLIST device id CDATA #REQUIRED name CDATA #IMPLIED>"
+			"<!ATTLIST channel id CDATA #REQUIRED type (input|output) #REQUIRED name CDATA #IMPLIED>"
+			"<!ATTLIST scan-element index CDATA #REQUIRED format CDATA #REQUIRED scale CDATA #IMPLIED>"
+			"<!ATTLIST attribute name CDATA #REQUIRED filename CDATA #IMPLIED>"
+			"<!ATTLIST debug-attribute name CDATA #REQUIRED>"
+			"<!ATTLIST buffer-attribute name CDATA #REQUIRED>"
+			"]>"
+			"<context name=\"xml\" description=\"Linux analog 4.9.0-g2398d50 #189 SMP PREEMPT Tue Jun 26 09:52:32 IST 2018 armv7l\" >"
```
dbogdan: Let's update the description for no-OS.

cristiantpop:
V2:

- Update the description for no-OS, in xml header.

iio/iio.c
```diff
@@ -0,0 +1,789 @@
...
+	const char *device_name;
+	const char *channel_name;
+	const char *attribute_name;
+	bool ch_out;
+};
+
+/**
+ * iio_read_attr(), iio_write_attr() functions, they need to know about iio_interfaces
+ */
+static struct iio_interfaces *iio_interfaces = NULL;
```
adrimbarean: Should the interface table exist here? It would seem to me that the interface table should exist in the layer above this one and instead the functions here take the pointer parameter to the interfaces table.

cristiantpop: That is a good question. This means that the following functions:
ops->read_attr = iio_read_attr;
	ops->write_attr = iio_write_attr;
	ops->ch_read_attr = iio_ch_read_attr;
	ops->ch_write_attr = iio_ch_write_attr;
	ops->transfer_dev_to_mem = iio_transfer_dev_to_mem;
	ops->read_data = iio_read_dev;
	ops->transfer_mem_to_dev = iio_transfer_mem_to_dev;
	ops->write_data = iio_write_dev;
should receive a "struct iio_interface *" parameter. This will affect the libtinyiiod interface. 
Since this doesn't have much benefits, only changes, we decided in the weekly meeting that we shouldn't change anything, regarding this.
Multiple interfaces (UART, USB, Network), will be supported, we only need to call "tinyiiod_create" multiple times to create multiple "struct tinyiiod" elements corresponding to each interface (UART, USB, Network). Each "struct tinyiiod" instance will use the same list of
"static struct iio_interfaces *iio_interfaces".

iio/iio.c
cristiantpop: Updated.

cristiantpop:
V3:

- Update some of the header comments.

cristiantpop:
V4:

- add "iio_unregister" function. This removes an interface from the list.

cristiantpop:
V5:

- Fix some comments.

cristiantpop:
V6:

Fix some comments.

iio/iio.c
```diff
@@ -0,0 +1,837 @@
...
+			"<!ATTLIST context name CDATA #REQUIRED description CDATA #IMPLIED>"
+			"<!ATTLIST context-attribute name CDATA #REQUIRED value CDATA #REQUIRED>"
+			"<!ATTLIST device id CDATA #REQUIRED name CDATA #IMPLIED>"
+			"<!ATTLIST channel id CDATA #REQUIRED type (input|output) #REQUIRED name CDATA #IMPLIED>"
+			"<!ATTLIST scan-element index CDATA #REQUIRED format CDATA #REQUIRED scale CDATA #IMPLIED>"
+			"<!ATTLIST attribute name CDATA #REQUIRED filename CDATA #IMPLIED>"
+			"<!ATTLIST debug-attribute name CDATA #REQUIRED>"
+			"<!ATTLIST buffer-attribute name CDATA #REQUIRED>"
+			"]>"
+			"<context name=\"xml\" description=\"Linux analog 4.9.0-g2398d50 #189 SMP PREEMPT Tue Jun 26 09:52:32 IST 2018 armv7l\" >"
```
buha: XML context shows Linux again after last changes ;)

cristiantpop: Thank You, I will fix this.

cristiantpop:
V7:

- Fixed context element in xml.

iio/iio.c
```diff
@@ -0,0 +1,838 @@
...
+				iio_interfaces->interfaces[i]->iio);
+		if (ret < 0)
+			goto error;
+
+		length = strlen(xml);
+		tmp_xml2 = realloc(xml, strlen(xml) + strlen(tmp_xml) + 1);
+		if (!tmp_xml2)
+			goto error;
+
+		xml = tmp_xml2;
```
adrimbarean: I think this can be written easier like:
`xml = realloc(xml, strlen(xml) + strlen(tmp_xml) + 1);
if (!tmp_xml2)
    goto error;`
No need for the extra `tmp_xml2`.
And I'm not sure, but this might implementation might be somewhat of an issue if the new allocated memory is somewhere else than the previous. That would mean that we change the value of `xml` without freeing the memory in the first place.

iio/iio.c
```diff
@@ -0,0 +1,838 @@
...
+		return FAILURE;
+	}
+
+	for(i = 0; i < iio_interfaces->num_interfaces; i++) {
+		if (!strcmp(device_name, iio_interfaces->interfaces[i]->name)) {
+			free(iio_interfaces->interfaces[i]);
+			deleted = 1;
+			continue;
+		}
+		interfaces->interfaces[i - deleted] = iio_interfaces->interfaces[i];
```
mirceacaprioru: this seems a bit off. 
if i = 0  and device_name == iio_interfaces->interfaces[i]->name then you will get 
interfaces->interfaces[0 - 1] = iio_interfaces->interfaces[0];

iio/iio.c
adrimbarean: Never mind. What I suggested is bad practice, actually. No change needed.

PR #316, opened 2019-11-04
util: Move util files into the new location.

cristiantpop:
Move only new files that won't affect other projects.


cristiantpop:
V2:

- remove "util.h" and "util.c" from this commit.

PR #318, opened 2019-11-05
ad9361_parameters: Add UART parameters.

cristiantpop:
They will be used with for tinyiiod server application, with UART
interface.


cristiantpop:
V2:

- rename "UART_1_DEVICE_ID" to "UART_DEVICE_ID"

PR #321, opened 2019-11-06
Xilinx timer platform driver

adrimbarean:
Add support for PS and PL timers for Xilinx.

### DIFF
--- a/drivers/platform/xilinx/timer.c
+++ b/drivers/platform/xilinx/timer.c
@@ -1,0 +1,284 @@
+/***************************************************************************//**
+*   @file   timer.c
+*   @brief  Timer and GIC control module source.
+*   @author Andrei Drimbarean (andrei.drimbarean@analog.com)
+********************************************************************************
+* Copyright 2019(c) Analog Devices, Inc.
+*
+* All rights reserved.
+*
+* Redistribution and use in source and binary forms, with or without
+* modification, are permitted provided that the following conditions are met:
+*  - Redistributions of source code must retain the above copyright
+*    notice, this list of conditions and the following disclaimer.
+*  - Redistributions in binary form must reproduce the above copyright
+*    notice, this list of conditions and the following disclaimer in
+*    the documentation and/or other materials provided with the
+*    distribution.
+*  - Neither the name of Analog Devices, Inc. nor the names of its
+*    contributors may be used to endorse or promote products derived
+*    from this software without specific prior written permission.
+*  - The use of this software may or may not infringe the patent rights
+*    of one or more patent holders.  This license does not release you
+*    from the requirement that you obtain separate licenses from these
+*    patent holders to use this software.
+*  - Use of the software either in source or binary form, must be run
+*    on or directly connected to an Analog Devices Inc. component.
+*
+* THIS SOFTWARE IS PROVIDED BY ANALOG DEVICES "AS IS" AND ANY EXPRESS OR IMPLIED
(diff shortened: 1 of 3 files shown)


include/timer.h
```diff
@@ -0,0 +1,90 @@
...
+#include <stdbool.h>
+#include <stdint.h>
+
+/******************************************************************************/
+/*************************** Types Declarations *******************************/
+/******************************************************************************/
+
+struct timer_desc {
+	uint32_t load;
+	uint32_t freq;
```
cristiantpop: We should have here only have "load" or "freq" since they are dependent. 
Keep "freq" simpler to use, keep "load" for precision. Probably "freq" is enough.
Please specify a measurement unit for frequency.

include/timer.h
```diff
@@ -0,0 +1,90 @@
...
+	bool auto_reload;
+	void *extra;
+};
+
+struct timer_init_param {
+	uint32_t start_freq;
```
cristiantpop: Rename "start_freq" to "freq".

include/timer.h
```diff
@@ -0,0 +1,90 @@
...
+void timer_start(struct timer_desc *desc);
+
+/* Stop a timer from counting. */
+void timer_stop(struct timer_desc *desc);
+
+/* Enable timer interrupt. */
+void timer_enable_interrupt(struct timer_desc *desc);
+
+/* Disable timer interrupt. */
+void timer_disable_interrupt(struct timer_desc *desc);
```
cristiantpop: Probably we should use "irq_source_enable" and "irq_source_disable" from "irq.h", for a minimal interface.

adrimbarean:
V0 -> V1:
 - remove unnecessary load member of the timer handler structure;
 - rename the various "freq" members of the structs with "freq_hz" to keep consistency and specify the measurement unit of the variable;
 - remove unnecessary interrupt API functions.

include/timer.h
dbogdan: I agree, if possible, we should use the irq.h here.

adrimbarean: You are right. Change is done.

drivers/platform/xilinx/timer.c
```diff
@@ -0,0 +1,284 @@
...
+	switch(xil_timer_dev->type) {
+	case TIMER_PL:
+		;
+#ifdef XTMRCTR_H
+		uint8_t tmr_options = 0;
+
+		load = MB_AXI_TIMER_CLOCK / dev->freq_hz;
+		xil_timer_dev->active_tmr = xil_timer_extra_init->active_tmr;
+
+		xil_timer_dev->instance = (XTmrCtr *)calloc(1, sizeof(XTmrCtr));
```
mchindri: Shouldn't you check if the allocation succeeded?

drivers/platform/xilinx/timer.c
```diff
@@ -0,0 +1,284 @@
...
+		break;
+pl_error:
+		free(xil_timer_dev->instance);
+#endif
+		goto error;
+	case TIMER_PS:
+#ifdef XSCUTIMER_H
+		load = CORE_PRIVATE_TIMER_CLOCK / dev->freq_hz;
+
+		xil_timer_dev->instance = (XScuTimer *)calloc(1, sizeof(XScuTimer));
```
mchindri: same as before

adrimbarean:
Continue the pull request at https://github.com/analogdevicesinc/no-OS/pull/396

PR #322, opened 2019-11-07
altera: update drivers and projects

amiclaus:
This patch contains the following modifications for Altera platform
drivers:

1. Split `altera_platform_drivers.h` into `spi_extra.h`, `i2c_extra.h`
and `gpio_extra.h`

2. Add extra parameters `type` and `base address` to spi/i2c/gpio
descriptors. which will be used for read/write operations. (replaces
hardcoded values).

3. Add `gpio_init` function that helps initialization for gpio
descriptors with type/address attributes. (NOTE: this implementation
will change the `gpio.h` header file from `/include` but won't affect
the functionality of other projects/drivers.)

4. Update drivers using the latest added parameters (type, base address)

5. Update projects which are Altera platform dependent from `projects`
folder.

### DIFF
--- a/drivers/platform/altera/gpio.c
+++ b/drivers/platform/altera/gpio.c
@@ -44,6 +44,7 @@
 #include <stdlib.h>
 #include <altera_avalon_spi_regs.h>
 #include "gpio.h"
+#include "gpio_extra.h"
 #include "error.h"
 #include "parameters.h"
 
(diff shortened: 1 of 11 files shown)


amiclaus:
Changelog v1->v2:

- fix memory leak

amiclaus:
Changelog v1->v2:

- fix Codacy issues

dbogdan:
Can we skip for the moment the GPIO changes? I'm not sure if adding the gpio_init() is the way we should go.
Thanks,
Dragos

PR #326, opened 2019-11-12
iio_types: Add channel type information to "struct iio_channel".

cristiantpop:
This is necessary to differentiate, between input and output channels, with
the same name, but have different lists of attributes. This is useful when
reading or writing all the attributes of a channel.


cristiantpop:
I will merge later today, if there are no other comments.

PR #331, opened 2019-11-15
drivers:dac:ad5686: fix driver

amiclaus:
1. Replace undefined macros for i2c stop bit.

2. Fix gpio init param types.


dbogdan:
2 unrelated changes in 1 commit.

amiclaus:
There are 2 fixes in the same driver. One without another still makes the driver broken.

mirceacaprioru:
I also think this should go in 2 commits. Even if the first one does not fix the driver alone it fixes the part it is referring to.

amiclaus:
v2:
- split into 2 commits

PR #336, opened 2019-11-16
travis: build drivers

amiclaus:
This patch integrates the drivers builds into TravisCI.

Builds are available on Linux OS.

Platform drivers and AXI Core drivers are skipped.

Build errors are fixed with #331 and #332.

### DIFF
--- a/ci/travis/before_install_linux
+++ b/ci/travis/before_install_linux
@@ -2,5 +2,5 @@
 set -e
 
 sudo apt-get update
-sudo apt-get install libmatio-dev cppcheck 
+sudo apt-get install libmatio-dev cppcheck gcc-arm-none-eabi libnewlib-arm-none-eabi
 
--- a/ci/travis/build_projects.sh
+++ b/ci/travis/build_projects.sh
@@ -4,7 +4,12 @@ set -e
 
 if [ "$TRAVIS_OS_NAME" == "linux" ] 
 then 
+	echo "Building AD9361-Linux"
 	make -C ./ad9361/sw -f Makefile.linux
+	echo "Building Drivers"
+	make -C ./drivers -f Makefile
 fi
 
+echo "Building AD9361-Generic"
 make -C ./ad9361/sw -f Makefile.generic
+
--- a/drivers/Makefile
+++ b/drivers/Makefile
@@ -1,12 +1,14 @@
 CC = arm-none-eabi-gcc
-CFLAGS = -c -Wall -I../include/ \
+CFLAGS = -c -Wall -std=gnu11 -Wformat=0 \
+	 -I../include/ \
 	 -I../projects/drivers/util/ \
 	 -I./dac/ad917x/ad917x_api/ \
 	 -I../projects/ad9361/src/ \
-	 -I./adc/ad9208/ad9208_api \
-	 -I../projects/drivers/axi_adc_core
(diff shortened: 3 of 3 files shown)


ci/travis/before_install_linux
```diff
@@ -2,5 +2,5 @@
 set -e
 
 sudo apt-get update
-sudo apt-get install libmatio-dev cppcheck 
+sudo apt-get install libmatio-dev cppcheck gcc-arm-none-eabi libnewlib-arm-none-eabi
```
commodo: should we try the hard-float compilers?
i.e. ` gcc-arm-none-eabihf` ?
also, maybe later, it would be interesting to also add a `aarch64` [i.e. ARM64] build run

ci/travis/build_projects.sh
```diff
@@ -4,7 +4,12 @@ set -e
 
 if [ "$TRAVIS_OS_NAME" == "linux" ] 
 then 
+	echo "Building AD9361-Linux"
```
commodo: why this `echo "Building AD9361-Linux"` ? ; it sounds like the entire no-OS project is the `AD9361-Linux` project

commodo: well, to maybe later, let's move this single project in a more organized way;
like maybe each project build should be it's own function;
something like `build_ad9361_linux()`; then we would have another `build_ad9371_generic()` function, etc

drivers/Makefile
```diff
@@ -1,12 +1,14 @@
...
 	 -I./dac/ad917x/ad917x_api/ \
 	 -I../projects/ad9361/src/ \
-	 -I./adc/ad9208/ad9208_api \
-	 -I../projects/drivers/axi_adc_core
+	 -I./adc/ad9208/ad9208_api/ \
+	 -I./axi_core/axi_adc_core/
 
-SUBDIRS = $(shell find . -path ./platform -prune -o -name '*.c')
+SUBDIRS = $(shell find . -path ./platform -prune -o \
+	  -path ./axi_core -prune -o -name '*.c')
```
commodo: doing hard-coded skip stuff in Makefile looks messy; 
maybe add an env-var that make can use, which contains a list of folder/projects/stuff to skip

drivers/Makefile
```diff
@@ -1,12 +1,14 @@
 CC = arm-none-eabi-gcc
```
commodo: `CC = arm-none-eabi-gcc` <- this looks hard-coded;
maybe change it to `CC ?= arm-none-eabi-gcc`, so that you can configure this externally;

ci/travis/build_projects.sh
commodo: now i am a bit confused;
why is this patch changing `projects` build code, when the commit description tells that it's handling `drivers` build?

drivers/Makefile
```diff
@@ -1,12 +1,14 @@
...
-CFLAGS = -c -Wall -I../include/ \
+CFLAGS = -c -Wall -std=gnu11 -Wformat=0 \
```
commodo: these CFLAGS don't seem ok;
`-std=gnu11` means you are trying to build stuff for a really-new compiler, and ignoring older ones;

no-OS is kind of a generic set of drivers/code; it should probably also run with C89 [for some customers]

amiclaus:
Closing this one. New PR: #341

PR #340, opened 2019-11-19
drivers: Makefile rework

amiclaus:
Add variables for includes and for folders to be skipped.

Skip `axi_core` drivers. (require BSP generated header files).

Disable Wformat flag (for now). There are multiple warnings regarding
printf format which are in conflict with Codacy.


amiclaus:
Creating new PR.

PR #341, opened 2019-11-19
travis: update

amiclaus:
This patch contains the following modfications for TravisCI.

1. Create independent jobs for each project/build type.

2. Get rid of small shell scripts and integrate them as functions in
`run_build.sh`.

3. Integrate `drivers` build.


.travis.yml
```diff
@@ -2,22 +2,13 @@ language: c
...
-      compiler: gcc
-    - os: osx
-      compiler: clang
-
-before_install:
-  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then ./ci/travis/before_install_darwin ; fi
-  - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then ./ci/travis/before_install_linux ; fi
+env:
+  matrix:
+  - BUILD_TYPE=astyle
```
scuciurean: nitpick: Maybe we can replace BUILD_TYPE with JOB since not all the functions presume to build something.

amiclaus: I think this i more like a naming preference. 
Each BUILD_TYPE env from matrix determines what should be run in the Build Job.
I used this name to keep consistency with other repos where TravisCI is integrated.

amiclaus:
New PR: #346

Issue #350, opened 2019-11-22
Makefile rule to build sources only


buha:
Currently, we perform `make` from `projects/xxxx` which creates `build/` directory where the bsp and all the necessary source files are copied. However, invoking `make` again at this point fails because it retries to use the SDK to create a project in the same `build/` dir.

We need a Makefile rule to build only the sources already existing in `build/`.
Or even better, a smart `all:` rule that only creates the `build/` dir and the SDK project in it when the directory is missing.
When `build/` is on the filesystem, it would only compile whatever is in it.

PR #354, opened 2019-11-25
tools: scripts: Makefile: Added altera build

scuciurean:
The makefile now has the possibility to build and upload the no-Os software, supporting Intel and Xilinx carriers.
Added environment scripts that are used when building with an SDK and updated the project's source makefile.
Added bsp handling so it will be generated only once (issue #350).


tools/scripts/Makefile
```diff
@@ -1,7 +1,14 @@
-export PROCESSOR_CORE 
+#------------------------------------------------------------------------------
```
mirceacaprioru: this modification is not related to altera build, i would do it in a separate commit

tools/scripts/Makefile
```diff
@@ -29,25 +42,38 @@ MUTE = @
...
+#                            COMMON COMPILER FLAGS                             
+#------------------------------------------------------------------------------
 CFLAGS = -Wall 								\
 	 -Wmissing-field-initializers					\
 	 -Wclobbered 							\
 	 -Wempty-body 							\
 	 -Wignored-qualifiers 						\
 	 -Wmissing-parameter-type					\
+	 -Wno-format  							\
```
mirceacaprioru: not related to this commit

buha:
I tested this on Xilinx only and these work:
- Upload functionality works for me
- 2nd `make` does indeed only build the source files

Some things which could perhaps be improved before the merge:
- Running `make` consecutively should only build modified source files in `build/` dir. It currently builds everything even if there were no changes.
- Running `make` after a `make clean` on a bare project without any `.hdf` file yields this output. The `cat ... arch.txt` line shouldn't be there.
```
$ make
[11:32:38] Evaluating hardware : 
[11:32:38] Platform not found
cat: /home/dberghe/Work/no-OS/projects/adrv9009/../../projects/adrv9009/build/tmp/arch.txt: No such file or directory
[11:32:38] Platform : none
[11:32:38] Can't generate the bsp
make: *** [../../tools/scripts/Makefile:274: bsp] Error 1
```

tools/scripts/Makefile
```diff
@@ -372,10 +377,24 @@ post-build:
 		mv -f sw.map $(TEMP_DIR);				\
 	fi;
 
-xilinx:	xil-pre-cleanup eval-hardware pre-build xilinx-bsp
-	$(call copy_srcs,xilinx)
-	@$(MAKE) -s compile
-	rm -rf $(BUILD_DIR)/SDK.log
+# Upload the bitstream and elf file to the board
+.SILENT:run
+run: eval-hardware
```
buha: `run` should depend on the source compilation rule, otherwise it might upload an old binary

scuciurean:
Changelog -> v2:

1. Splitted altera build commit.

tools/scripts/Makefile
scuciurean: Doing this means that the user can run the binary only once per build.

tools/scripts/platform/intel/environment.sh
```diff
@@ -0,0 +1,28 @@
...
+:$1/$2/embedded/host_tools/altera/mkimage\
+:$1/$2/embedded/host_tools/altera/mkpimag\
+:$1/$2/embedded/host_tools/altera/device_tee\
+:$1/$2/embedded/host_tools/altera/diskutils\
+:$1/$2/embedded/host_tools/altera/imagecat\
+:$1/$2/embedded/host_tools/altera/securebot\
+:$1/$2/embedded/host_tools/gnu/dtc\
+:$1/$2/embedded/ds-5/sw/gcc/bin\
+:$1/$2/embedded/ds-5/sw/ARMCompler5.06u1/bin\
+:$1/$2/embedded/ds-5/bin"
```
dbogdan: Let's include at this point only the mandatory tools. Initially, we will support only Nios2 platforms.

tools/scripts/platform/xilinx/environment.sh
```diff
@@ -0,0 +1,19 @@
...
+:$1/Vivado/$2/bin\
+:$1/SDK/$2/bin\
+:$1/SDK/$2/gnu/microblaze/lin/bin\
+:$1/SDK/$2/gnu/arm/lin/bin\
+:$1/SDK/$2/gnu/microblaze/linux_toolchain/lin64_le/bin\
+:$1/SDK/$2/gnu/aarch32/lin/gcc-arm-linux-gnueabi/bin\
+:$1/SDK/$2/gnu/aarch32/lin/gcc-arm-none-eabi/bin\
+:$1/SDK/$2/gnu/aarch64/lin/aarch64-linux/bin\
+:$1/SDK/$2/gnu/aarch64/lin/aarch64-none/bin\
+:$1/SDK/$2/gnu/armr5/lin/gcc-arm-none-eabi/bin\
```
dbogdan: Same comment for the Altera tools is valid for the Xilinx ones.

tools/scripts/Makefile
dbogdan: I still see unrelated changes...

scuciurean: This flag is used by default in the intel sdk

scuciurean:
Changelog -> V2:

1. Stripped down the environment scripts.

buha:
I did some research on how `rsync` might be used instead of `cp` to copy only modified files (newer filesystem timestamp) from repo to `build/` dir. It turns out a simple rule like this will do this:
```
rsync-srcs:                                                                      
    printf '%s\n' $(subst LOCAL_PLATFORM,$(PLATFORM),$(SRCS)) >  $(BUILD_DIR)/srclist.txt
    printf '%s\n' $(subst LOCAL_PLATFORM,$(PLATFORM),$(INCS)) >> $(BUILD_DIR)/srclist.txt
    rsync -t --no-relative --files-from=$(BUILD_DIR)/srclist.txt / $(BUILD_DIR)/app/src
```
rsync exists on all linuxes. As a first step, we could do a simple check to use `copy-srcs` on Windows and `rsync-srcs` on Linux. Then, if there is interest, I'm sure it can be implemented on Windows too.

I still think this feature is a must. The whole point of `make` is to build only modified sources and we need this in our project.

scuciurean:
Changelog -> V4:

1. Split in more commits
2. Fixed conflicts

scuciurean:
Changelog -> V5:

1. Renamed platform folder from intel to altera.

dbogdan:
If you decided to do the rename, please update tools/scripts/platform/intel/environment.sh too.

dbogdan:
Otherwise, it looks good.

scuciurean:
> If you decided to do the rename, please update tools/scripts/platform/intel/environment.sh too.

I forgot to push the changes

PR #356, opened 2019-11-25
projects: ad9208: Moved source files in src folder

scuciurean:
Moved the source files in the src directory in order to mantain the common
project structure, used by the makefiles.


amiclaus:
If we want to maintain the common structure, then I think also the README file should be moved and updated.

scuciurean:
Changelog -> v2

1. Updated project readme

scuciurean:
Changelog -> v3

1. Fixed relative paths

PR #358, opened 2019-11-26
Drivers: platform: modify type of parameter bytes_number from uint8_t to uint16

mchindri:
Modify parameter type of _bytes_number_ from _uint8_t_ to _uint16_t_ in
function _spi_write_and_read_. Modify file in includes and in all platform
files.
_Uint8_t_ data type is not enough to represent the size of a data block
(512 bytes) for reading and writing data in a SD card over SPI.
This modification doesn't affect the existing spi drivers because in the
implementation there are no constrains for this parameter to be _uint8_t_

Signed-off-by: Chindris <Mihail.Chindris@analog.com>

### DIFF
--- a/drivers/platform/altera/spi.c
+++ b/drivers/platform/altera/spi.c
@@ -114,7 +114,7 @@ int32_t spi_remove(struct spi_desc *desc)
 
 int32_t spi_write_and_read(struct spi_desc *desc,
 			   uint8_t *data,
-			   uint8_t bytes_number)
+			   uint16_t bytes_number)
 {
 	uint32_t i;
 	struct altera_spi_desc *altera_desc;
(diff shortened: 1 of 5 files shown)


drivers/platform/altera/spi.c
```diff
@@ -114,7 +114,7 @@ int32_t spi_remove(struct spi_desc *desc)
 
 int32_t spi_write_and_read(struct spi_desc *desc,
 			   uint8_t *data,
-			   uint8_t bytes_number)
```
mirceacaprioru: you have a few typos in the commit message: 
This modifiction don't affect

mchindri:
V2
- Fix typo in commit message

mchindri:
V3
 - Fix the other typo.

PR #359, opened 2019-11-27
uart: Rename structure to "uart_ini_param" and parameter to "param".

cristiantpop:
This is necessary, for consistency with other modules: spi, irq, i2c.


amiclaus:
LGTM.

PR #360, opened 2019-11-28
drivers:platform:xilinx: SPI transmit fix

adrimbarean:
When the option XSPIPS_FORCE_SSELECT_OPTION is used for the transfer, as is
the case here the CS line will be asserted as soon as it is picked via the
XSpiPs_SetSlaveSelect method. If the slave is not deselected after the
transfer is complete the CS line will remain asserted potentially creating
problems, for example with new transfers, or with devices that only
update registers on CS deassert.

Add this patch to deselect the CS and deassert the CS line after every
transfer.

Signed-off-by: Drimbarean <Andrei.Drimbarean@analog.com>


adrimbarean:
Will merge today if no objections are raised.

adrimbarean:
V0 -> V1:
 - add deassert value define and replace function parameter wit said defne.

PR #363, opened 2019-12-02
drivers: axi_dac_core: add channel customization

amiclaus:
Previous implementation did not permit customization of the dac channels
by the user.

This patch allows the user setup the dac channels using the
`axi_dac_channel` structure.

The default configuration is used if the channel is not defined.

### DIFF
--- a/drivers/axi_core/axi_dac_core/axi_dac_core.c
+++ b/drivers/axi_core/axi_dac_core/axi_dac_core.c
@@ -87,6 +87,8 @@
 #define AXI_DAC_DDS_INCR(x)				(((x) & 0xFFFF) << 0)
 #define AXI_DAC_TO_DDS_INCR(x)			(((x) >> 0) & 0xFFFF)
 
+#define DAC_REG_DATA_PATTERN(c)			(0x0410 + (c) * 0x40)
+
 #define AXI_DAC_REG_DATA_SELECT(c)		(0x0418 + (c) * 0x40)
 #define AXI_DAC_DATA_SELECT(x)			(((x) & 0xF) << 0)
 #define AXI_DAC_TO_DATA_SELECT(x)		(((x) >> 0) & 0xF)
@@ -824,7 +826,6 @@ int32_t axi_dac_init(struct axi_dac **dac_core,
 	uint32_t reg_data;
 	uint32_t freq;
 	uint32_t ratio;
-	uint8_t ch;
 
 	dac = (struct axi_dac *)malloc(sizeof(*dac));
 	if (!dac)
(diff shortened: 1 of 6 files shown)


amiclaus:
V2:

- Rename: `DAC_REG_DATA_PATTERN` -> `AXI_DAC_REG_DATA_PATTERN`

drivers/axi_core/axi_dac_core/axi_dac_core.c
```diff
@@ -877,6 +871,45 @@ int32_t axi_dac_init(struct axi_dac **dac_core,
...
 }
 
+int32_t axi_dac_data_setup(struct axi_dac *dac)
+{
+	struct axi_dac_channel *chan;
+	uint32_t i;
+
+	for (i = 0; i < dac->num_channels; i++) {
+		chan = &dac->channels[i];
+		if (chan) {
```
adrimbarean: Possible undefined behavior. Can be solved with double pointer or adding an initialized channel mask.

amiclaus: You are right. Will check for best solution.

amiclaus:
V3:
- add `dac_data_default` parameter to `axi_dac` structures. The parameter determines if the channels are configured manually or by default.
- update `axi_dac` dependent projects.

amiclaus:
V4:
- fix code style

amiclaus:
Updating the driver separately may affect the functionality of the projects, since they are dependent on the driver.

amiclaus:
V5:

- Update: remove `dac_data_default` parameter. The first channel is checked for initialization. If it is empty, it is assumed that no other channel is initialized and the default configuration will be applied.

drivers/axi_core/axi_dac_core/axi_dac_core.h
```diff
@@ -52,12 +53,14 @@ struct axi_dac {
 	uint32_t base;
 	uint8_t	num_channels;
 	uint64_t clock_hz;
+	struct axi_dac_channel *channels;
```
mchindri: You can comment the structures to understand how to use the `channels` parameter.

amiclaus: Commenting the entire structure means creating separate commit/PR, since this is responsible only of adding a new member to the struct. I will comment here only the added member to the struct.

amiclaus:
V6:

- Remove `stdbool.h`
- Add comments for newly added members.

amiclaus:
V7:
- perform rebase

PR #364, opened 2019-12-02
adrv9009 cleanup

buha:
This commit sequence simplifies the headless.c and prepares the talise setup sequence to be multi-instance capable.


projects/adrv9009/src/app/headless.c
```diff
@@ -30,6 +30,9 @@
 #include "app_jesd.h"
 #include "app_transceiver.h"
 #include "app_talise.h"
+#ifndef ALTERA_PLATFORM
+#include "xil_cache.h"
+#endif
```
dbogdan: Isn't this an error (the subject of this commit states that some warnings are fixed)? Anyway, I would split this commit into multiple patches and describe exactly what is wrong.

projects/adrv9009/src.mk
```diff
@@ -49,31 +49,31 @@ INCS :=	$(PROJECT)/src/app/app_config.h					\
...
+	$(DRIVERS)/rf-transceiver/talise/talise_reg_addr_macros.h		\
+	$(DRIVERS)/rf-transceiver/talise/talise_rx.h			\
+	$(DRIVERS)/rf-transceiver/talise/talise_rx_types.h			\
+	$(DRIVERS)/rf-transceiver/talise/talise_tx.h			\
+	$(DRIVERS)/rf-transceiver/talise/talise_tx_types.h			\
+	$(DRIVERS)/rf-transceiver/talise/talise_types.h			\
+	$(DRIVERS)/rf-transceiver/talise/talise_user.h			\
+	$(DRIVERS)/rf-transceiver/talise/talise_version.h			\
 	$(PROJECT)/src/firmware/talise_arm_binary.h			\
 	$(PROJECT)/src/firmware/talise_stream_binary.h			\
```
dbogdan: The API is firmware dependent, so both should be kept together.

projects/adrv9009/src/app/headless.c
buha: You're right, I will commit separately

projects/adrv9009/src.mk
buha: Correct, it needs to be moved together

drivers/Makefile
```diff
@@ -8,7 +8,8 @@ INCLUDES = -I../include/ \
 	 -I./dac/ad917x/ad917x_api/ \
 	 -I../projects/ad9361/src/ \
 	 -I./adc/ad9208/ad9208_api \
-	 -I./axi_core/axi_adc_core
+	 -I./axi_core/axi_adc_core \
+	 -I../projects/adrv9009/src/devices/adi_hal
```
dbogdan: These changes should go to previous commit, since they are only fixing an error introduced there.

buha: True, I squashed these commits.

buha:
There were several forced pushes that changed the following:
- moved `projects/adrv9009/src/firmware` to `drivers/rf-transceiver/talise/firmware`
- squashed the commit that updates `drivers/Makefile` into a more relevant commit
- moved the xil_cache.h include from the "warning fix" commit to a more relevant commit

dbogdan:
One more thing: For consistency with the rest of the files, I would keep the init/remove naming instead of init/deinit.

buha:
> One more thing: For consistency with the rest of the files, I would keep the init/remove naming instead of init/deinit.

@dbogdan I intentionally used different scheme because these are not drivers, but application level wrappers for initialization setup sequences. app_talise is not even consistent with the rest - the init function is talise_setup, to keep in line with the linux wrapper naming. If you still consider init/remove would be better, I'll do the refactoring.

PR #365, opened 2019-12-03
adrv9009 warnings fix

buha:
Fixed adrv9009 xilinx compilation warnings coming from our own code.
Some warnings (related to talise and the autogenerated talise_config) could not be fixed at this point, they need to be fixed upstream.

### DIFF
--- a/drivers/axi_core/axi_dmac/axi_dmac.c
+++ b/drivers/axi_core/axi_dmac/axi_dmac.c
@@ -125,7 +125,7 @@ int32_t axi_dmac_transfer(struct axi_dmac *dmac,
 	/* Wait until the transfer with the ID transfer_id is completed. */
 	do {
 		axi_dmac_read(dmac, AXI_DMAC_REG_TRANSFER_DONE, &reg_val);
-	} while((reg_val & (1 << transfer_id)) != (1 << transfer_id));
+	} while((reg_val & (1u << transfer_id)) != (1u << transfer_id));
 
 	return SUCCESS;
 }
--- a/drivers/axi_core/clk_axi_clkgen/clk_axi_clkgen.c
+++ b/drivers/axi_core/clk_axi_clkgen/clk_axi_clkgen.c
@@ -368,12 +368,12 @@ void axi_clkgen_calc_params(struct axi_clkgen *axi_clkgen,
 			dout = DIV_ROUND_CLOSEST(fvco, fout);
 			dout = clamp(dout, 1, 128);
 			f = fvco / dout;
-			if (abs(f - fout) < abs(best_f - fout)) {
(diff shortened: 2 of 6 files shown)


drivers/platform/xilinx/spi.c
```diff
@@ -189,7 +189,6 @@ int32_t spi_init(struct spi_desc **desc,
  */
 int32_t spi_remove(struct spi_desc *desc)
 {
-	int32_t			ret;
```
dbogdan: Are you sure?
#ifdef XSPI_H
   ret = XSpi_Stop((XSpi *)(xdesc->instance));
   if(ret != SUCCESS)
      goto error;
#endif

drivers/axi_core/axi_dmac/axi_dmac.c
```diff
@@ -125,7 +125,7 @@ int32_t axi_dmac_transfer(struct axi_dmac *dmac,
 	/* Wait until the transfer with the ID transfer_id is completed. */
 	do {
 		axi_dmac_read(dmac, AXI_DMAC_REG_TRANSFER_DONE, &reg_val);
-	} while((reg_val & (1 << transfer_id)) != (1 << transfer_id));
+	} while((reg_val & (1 << transfer_id)) != (uint32_t)(1 << transfer_id));
```
amiclaus: Maybe use `1u << transfer_id` instead of type casting the entire operation.

buha: I agree, it's cleaner.

drivers/platform/xilinx/spi.c
buha: Well seen. I edited the commit.

drivers/axi_core/clk_axi_clkgen/clk_axi_clkgen.c
```diff
@@ -368,12 +368,12 @@ void axi_clkgen_calc_params(struct axi_clkgen *axi_clkgen,
 			dout = DIV_ROUND_CLOSEST(fvco, fout);
 			dout = clamp(dout, 1, 128);
 			f = fvco / dout;
-			if (abs(f - fout) < abs(best_f - fout)) {
+			if (abs((int32_t)(f - fout)) < abs((int32_t)(best_f - fout))) {
```
dbogdan: This looks a little bit strange to me... Since fout was already divided by 1000, maybe the safest thing would be to cast fout instead of the entire result.

buha: True. I edited the commit and casted `fout` only instead.
It seems safe because it is guaranteed to be less than 2^31-1 (max int32_t value) due to the division you mentioned.

PR #366, opened 2019-12-04
libraries: add FatFs module

mchindri:
More information: http://elm-chan.org/fsw/ff/00index_e.html

This module will be used by future projects.


buha:
I would suggest the following:
- the folder should be `libraries/fatfs` or `libraries/FatFs` and not `libraries/ff`
- tinyiiod is also a library and resides in its own repository. no-OS uses it as a git submodule. I think it makes sense for FatFs to be used also as a git submodule because it can may used in other ADI projects too.
```
git submodule
 0ed85fe498517c9bbdf2e120a84821f58b00cf36 libraries/libtinyiiod (heads/master)
```

dbogdan:
FatFs doesn't have an official git repository. ff is the name of their zip archive, but I don't mind naming the folder fatfs instead of ff if this is the preference.

mchindri:
V2:
 -Renaming folder _libraries/ff_ -> _libraries/FatFs_

mchindri:
V3:
 -Rename folder `libraries/FatFs` -> `libraries/fatfs`

PR #377, opened 2019-12-06
Add /libraries/fatfs to .astyleignore

mchindri:


mchindri:
Added signed-off to commit

PR #379, opened 2019-12-09
drivers:platform:generic: add timer driver

adrimbarean:
Add generic timer platform driver and the generic header file for it.
This generic driver includes no implementation, but serves as a form factor
for implementing the platform-specific timer drivers that many no-OS
projects use, for example: AD9172, ADV7511, many of the ADICUP3029 projects, etc.

Signed-off-by: Drimbarean <Andrei.Drimbarean@analog.com>

### DIFF
--- a/drivers/platform/generic/timer.c
+++ b/drivers/platform/generic/timer.c
@@ -1,0 +1,184 @@
+/***************************************************************************//**
+*   @file   timer.c
+*   @brief  Timer control module source.
+*   @author Andrei Drimbarean (andrei.drimbarean@analog.com)
+********************************************************************************
+* Copyright 2019(c) Analog Devices, Inc.
+*
+* All rights reserved.
+*
+* Redistribution and use in source and binary forms, with or without
+* modification, are permitted provided that the following conditions are met:
+*  - Redistributions of source code must retain the above copyright
+*    notice, this list of conditions and the following disclaimer.
+*  - Redistributions in binary form must reproduce the above copyright
+*    notice, this list of conditions and the following disclaimer in
+*    the documentation and/or other materials provided with the
+*    distribution.
+*  - Neither the name of Analog Devices, Inc. nor the names of its
+*    contributors may be used to endorse or promote products derived
+*    from this software without specific prior written permission.
+*  - The use of this software may or may not infringe the patent rights
+*    of one or more patent holders.  This license does not release you
+*    from the requirement that you obtain separate licenses from these
+*    patent holders to use this software.
+*  - Use of the software either in source or binary form, must be run
+*    on or directly connected to an Analog Devices Inc. component.
+*
+* THIS SOFTWARE IS PROVIDED BY ANALOG DEVICES "AS IS" AND ANY EXPRESS OR IMPLIED
+* WARRANTIES, INCLUDING, BUT NOT LIMITED TO, NON-INFRINGEMENT, MERCHANTABILITY
+* AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED.
+* IN NO EVENT SHALL ANALOG DEVICES BE LIABLE FOR ANY DIRECT, INDIRECT,
+* INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+* LIMITED TO, INTELLECTUAL PROPERTY RIGHTS, PROCUREMENT OF SUBSTITUTE GOODS OR
+* SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+* CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+* OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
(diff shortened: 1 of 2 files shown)


drivers/platform/generic/timer.c
```diff
@@ -0,0 +1,184 @@
...
+		// Unused variable - fix compiler warning
+	}
+
+	return SUCCESS;
+}
+
+/**
+ * @brief Get the timer clock frequency.
+ * @param [in]  desc      - Pointer to the device handler.
+ * @param [out] count_clk - The new value of the counter register.
```
buha: This is a bit confusing to me, for the following reasons:

- what is the measurement unit, Hz / ticks ? It should be specified.
- If by `count_clk` you mean frequency, i think it would be more clear to name it as such ('freq', 'frequency' etc.). This comment applies also to `timer_count_clk_get` and `timer_count_clk_set`.

drivers/platform/generic/timer.c
```diff
@@ -0,0 +1,184 @@
...
+ * @param [in] new_val - The new value of the counter register.
+ * @return 0 in case of success, error code otherwise.
+ */
+int32_t timer_counter_set(struct timer_desc *desc, uint32_t new_val)
+{
+	if(desc) {
+		// Unused variable - fix compiler warning
+	}
+
+	if(new_val) {
```
buha: I know this is just a template implementation, but this comparison is always true ;)

drivers/platform/generic/timer.c
```diff
@@ -0,0 +1,184 @@
...
+
+	if(count_clk) {
+		// Unused variable - fix compiler warning
+	}
+
+	return SUCCESS;
+}
+
+/**
+ * @brief Set the timer clock frequency or source.
```
buha: > frequency or source
I think you meant just frequency, it cannot mean both.

drivers/platform/generic/timer.c
```diff
@@ -0,0 +1,184 @@
+/***************************************************************************//**
+*   @file   timer.c
+*   @brief  Timer and GIC control module source.
```
buha: So far, this is just a Timer control driver template.
I think it would make sense to add GIC text in the file description only when some sort of GIC functionality will be implemented in timer.c/.h.

drivers/platform/generic/timer.c
adrimbarean: You are right, I will update this.

drivers/platform/generic/timer.c
adrimbarean: True, this text is a mistake from a previous implementation. Will remove GIC mentions.

drivers/platform/generic/timer.c
adrimbarean: This will not be true if `new_val` is 0.

drivers/platform/generic/timer.c
adrimbarean: I agree that there should be no uncertainty in the comments describing the function, but I wanted to emphasize that the variable can be either a clock source ID or a target frequency as in some cases is used one and in other cases the other. Are we sure we want to change this?

drivers/platform/generic/timer.c
buha: For some reason I wanted to comment about this check being pointless (looks like a copy-paste from the other functions where the second parameter is a pointer and is checked against NULL) and I commented something completely wrong hah :D 

What I mean is, with this function, any value should possible, the user can set the counter to any value from uint32_t including 0 so the test is pointless.

dbogdan: Well, this is not actually a test - it is only a workaround for fixing the compiler warning.

drivers/platform/generic/timer.c
```diff
@@ -0,0 +1,184 @@
...
+ * @breif Initialize hardware timer and the handler structure associated with
+ *        it.
+ * @param [out] desc - Pointer to the reference of the device handler.
+ * @param [in] init_param - Initialization structure.
+ * @return 0 in case of success, negative error code otherwise
+ */
+int32_t timer_init(struct timer_desc **desc,
+		   struct timer_init_param *init_param)
+{
+	if(desc) {
```
dbogdan: The coding style asks for an additional space character: _if (desc) {_

drivers/platform/generic/timer.c
buha: Oh ok, I totally misunderstood the point although there is a comment "fix compiler warning".

drivers/platform/generic/timer.c
adrimbarean: Will update in the whole file. Thanks!

adrimbarean:
V0 -> V1:
 - fix comments;
 - fix parameter names to be more explicit;
 - insert space in `if ()`.

drivers/platform/generic/timer.c
```diff
@@ -0,0 +1,184 @@
...
+
+/**
+ * @breif Initialize hardware timer and the handler structure associated with
+ *        it.
+ * @param [out] desc - Pointer to the reference of the device handler.
+ * @param [in] init_param - Initialization structure.
+ * @return 0 in case of success, negative error code otherwise
+ */
+int32_t timer_init(struct timer_desc **desc,
+		   struct timer_init_param *init_param)
```
cristiantpop: Rename "init_param" to "param", for consistency with other drivers.

adrimbarean: Will rename.

adrimbarean:
V1 -> V2:
 - Rename `init_param` to `param`, for consistency with other drivers.

adrimbarean:
Will merge today if there are no more comments.

PR #380, opened 2019-12-11
Timer generic update and fix

adrimbarean:
Changes in this patch:
 - fix variable name and comments in the header file;
 - add generic field to the initialization and handler structures.


adrimbarean:
Will merge tomorrow if all is good.

PR #386, opened 2019-12-13
include: add type comments

amiclaus:
Add comments for types to easily integrate with Doxygen.

The `///<` sequence is used to document struct members and enums.

Outlook:
Struct members:
Screenshot from 2019-12-13 14-36-08

Enums:
Screenshot from 2019-12-13 14-43-18

### DIFF
--- a/include/fifo.h
+++ b/include/fifo.h
@@ -49,10 +49,14 @@
 /*************************** Types Declarations *******************************/
 /******************************************************************************/
 
+/**
+ * @struct fifo_element
+ * @brief Structure holding the fifo element parameters.
+ */
 struct fifo_element {
-	struct fifo_element *next;
-	char *data;
-	uint32_t len;
+	struct fifo_element *next; ///< next FIFO element
+	char *data; ///< FIFO data pointer
+	uint32_t len; ///< FIFO length
 };
 
 /******************************************************************************/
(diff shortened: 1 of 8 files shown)


include/fifo.h
```diff
@@ -49,10 +49,14 @@
...
 
+/**
+ * @struct fifo_element
+ * @brief Structure holding the fifo element parameters.
+ */
 struct fifo_element {
-	struct fifo_element *next;
-	char *data;
-	uint32_t len;
+	struct fifo_element *next; ///< next FIFO element
```
dbogdan: Why did you decide to place the documentation block after the member (shouldn't we do it before, the same as for the function's parameters for example)?

dbogdan: Why did you decide to place the documentation block after the member (shouldn't we do it before, the same as for the function's parameters for example)?

amiclaus: I forgot to mention in the PR description that this is a proposal, not a decision. This approach populates the already available Doxygen fields (Member Data Documentation/Enumeration Type Documentation) with the inserted comments VS original approach which keeps everything in the brief description of the structure/enum. So it depends on how "sophisticated" we want the documentation to be.

amiclaus: Update: I added an alternative that comments the members above the implementation. Please see #390. After a decision is made, this PR can be closed.

amiclaus:
Moved to #390

PR #390, opened 2019-12-14
include: add type comments in header

amiclaus:
Add type comments for sturct/enum members in the header comment.

This patch is an alernative to #386 

### DIFF
--- a/include/fifo.h
+++ b/include/fifo.h
@@ -49,6 +49,16 @@
 /*************************** Types Declarations *******************************/
 /******************************************************************************/
 
+/**
+ * @struct fifo_element
+ * @brief Structure holding the fifo element parameters.
+ * @var fifo_element::next
+ * next FIFO element
+ * @var fifo_element::data
+ * FIFO data pointer
+ * @var fifo_element::len
+ * FIFO length
+ */
 struct fifo_element {
 	struct fifo_element *next;
 	char *data;
--- a/include/gpio.h
+++ b/include/gpio.h
@@ -59,11 +59,27 @@
 /*************************** Types Declarations *******************************/
 /******************************************************************************/
 
+/**
+ * @struct gpio_init_param
+ * @brief Structure holding the parameters for GPIO initialization.
+ * @var gpio_init_param::number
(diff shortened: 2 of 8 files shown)


amiclaus:
**UPDATE:** The approach that is currently used on our repository to comment types is the format proposed by Linux Kernel: https://www.kernel.org/doc/html/v4.11/doc-guide/kernel-doc.html
In that case, the doc generation was done using http://www.sphinx-doc.org/en/master/

In our case, we use Doxygen for documentation generation, therefore one of the following approaches for doc format should be taken into consideration: 

https://github.com/analogdevicesinc/no-OS/blob/c3265161d8ab292514b9795490187c7fc8ddb919/include/fifo.h#L52-L66

OR

https://github.com/analogdevicesinc/libiio/blob/1036e51decb0e2f1d284af7a9e641ee3890be8bc/iio.h#L1464-L1498

OR

https://github.com/analogdevicesinc/libm2k/blob/1e081afcc910626032de20192347fb5f36e2c378/include/libm2k/enums.hpp#L144-L154

mchindri:
Is it posible the libio format but just comment? 
Like:
"/*Right-shift to apply when converting sample */
unsigned int shift;"

Not: "/** @brief Right-shift to apply when converting sample */
	unsigned int shift;"

buha:
I would go with libiio style for 2 reasons:
- when you jump through code and need the structure member declaration and an explanation, you land on the parameter and you have the explanation right next to it.
- this approach is more flexible than the other two - the comment may contain more than the @brief section, but a full-blown description for members that require more documentation

amiclaus:
> 
> 
> Is it posible the libio format but just comment?
> Like:
> 
> ```
> /*Right-shift to apply when converting sample */
> unsigned int shift;
> ```
> 
> Not:
> 
> ```
> /** @brief Right-shift to apply when converting sample */
> 	unsigned int shift;
> ```

You can do it like this: `/** Right-shift to apply when converting sample */` with double `**`

amiclaus:
Closing ths PR.
Please see #412 for the new format.

PR #391, opened 2019-12-14
drivers: rename duplicate file names

amiclaus:
Support Doxygen integration by adding partial path for files with
same name.


amiclaus:
I will merge it later if no objections.

PR #395, opened 2019-12-16
ad9361: Fix ad9361_get_split_table_gain.

### LINKED ISSUE #394: AD9361: Cannot read gain value for split gain table mode (memory access errors)
**What is happening?** I'm attempting to use the split gain table mode of the AD9361, and running into some issues.  I believe I'm using the API correctly and that the implementation of [ad9361_get_split_table_gain][get_gt] is buggy.  This would also apply to the [implementation in the kernel module][kernel-impl], although I have no way to test that.  After compiling this program and the no-OS library with [`-fsanitize=address`][fsan] and reading the gain with [`ad9361_get_rx_rf_gain`][ad9361_get_rx_rf_gain], there are some memory access violations.

The full main.c reproducer is at the bottom of this issue (expand the arrow to see it), but the main gist is that there are memory access violations in the following situation:

1. Compile everything with `-fsanitize=address` (this just *exposes* the buggy memory access, doesn't cause it, of course).
2. Use manual gain control (this is a prerequisite to use the split gain table at all).
3. Use split table mode (specified in the `InitParam` struct).
4. Try to read the gain using [`ad9361_get_rx_rf_gain`](https://github.com/analogdevicesinc/no-OS/blob/9578dc028e70ece512368dee651e6a61f4a4d7f7/ad9361/sw/ad9361_api.h#L373).

**Why is this happening**?  I have a partial answer.  The access violations are happening [when calculating the lmt gain][calc]: the index into `lna_table` is invalid in split gain table mode.  For example, if the RF LO frequency is set to 3.07GHz and I'm in split gain mode, `ad9361_gt(phy)` is going to return `4`.  But the `lna_table` only has a length of 3, so we're reading beyond it.

So I get *how* this is happening, but I don't understand what the whole ad9361_get_split_table_gain function is doing.  I think it should just be doing what `ad9361_get_full_table_gain` function is doing, i.e. just using the `abs_gain_tbl`.  I will send in a PR that does this so we can discuss a concrete change.

[calc]: https://github.com/analogdevicesinc/linux/blob/eadedb6d5a4c235caa1edddca4a209c96698fc3f/drivers/iio/adc/ad9361.c#L1534
[kernel-impl]: https://github.com/analogdevicesinc/linux/blob/eadedb6d5a4c235caa1edddca4a209c96698fc3f/drivers/iio/adc/ad9361.c#L1515-L1550
[fsan]: https://gcc.gnu.org/onlinedocs/gcc/Instrumentation-Options.html
[ad9361_get_rx_rf_gain]: https://github.com/analogdevicesinc/no-OS/blob/9578dc028e70ece512368dee651e6a61f4a4d7f7/ad9361/sw/ad9361_api.h#L373
[get_gt]: https://github.com/analogdevicesinc/no-OS/blob/9578dc028e70ece512368dee651e6a61f4a4d7f7/ad9361/sw/ad9361.c#L1790-L1834

<summary><b>minimal reproducer.</b></summary>
<details>

```c
#include "config.h"
#include "ad9361_api.h"
#include "parameters.h"
#include "platform.h"
#include "adc_core.h"
#include "dac_core.h"

// the ad9361_globals.h includes the phy initialization struct, etc.
// must set manual gain control and split gain table mode!
#include "ad9361_globals.h"

struct ad9361_rf_phy *ad9361_phy;

 * @brief main
int main(void)
{
    // boilerplate init stuff
    default_init_param.gpio_resetb = GPIO_RESET_PIN;
    default_init_param.gpio_sync = -1;
    default_init_param.gpio_cal_sw1 = -1;
    default_init_param.gpio_cal_sw2 = -1;

    gpio_init(default_init_param.gpio_resetb);
    gpio_direction(default_init_param.gpio_resetb, 1);

    spi_init(SPIDEV_DEV, SPI_DEVICE_ID, 1, 0);
    ad9361_init(&ad9361_phy, &default_init_param);
    if (!ad9361_phy) {
        fprintf(stderr, "Could not initialize AD9361!\n");
        exit(1);
    }

    ad9361_set_tx_fir_config(ad9361_phy, tx_fir_config);
    ad9361_set_rx_fir_config(ad9361_phy, rx_fir_config);

    // use gain table or not?
    ad9361_phy->pdata->split_gt = 1;
    // do not skip tuning
    ad9361_phy->pdata->dig_interface_tune_skipmode = 0;
    for (int i = 0; i < 48; i++) {
        int cur_gain;
        int ret = ad9361_set_rx_rf_gain(ad9361_phy, 0, i);
        if (ret) {
            continue;
        }
        ret = ad9361_get_rx_rf_gain(ad9361_phy, 0, &cur_gain);
        if (ret) {
            fprintf(stderr, "bad deal for reading gain %d\n", i);
            continue;
        }
        cur_gain = (int8_t) cur_gain;
        fprintf(stderr, "gain %d, %d\n", i, cur_gain);
    }

    return 0;
}

```
</details>

Thanks for your time.

codypiersall:
Fixes #394.

The implementation was based on ad9361_get_full_table_gain.  This PR also removes the now unused static tables.

I believe ad9361_get_split_table_gain could be simplified further if we didn't need to fill out the lna_index, mixer_index, or tia_index of the `lpf_gain` index of the `rx_gain` struct.  

Alternatively, `ad9361_get_rx_gain` could be implemented to do what `ad9361_get_split_table_gain` was trying to do: see which table index is active, read the gains for that table index, and calculate the gain based on that.  Then there wouldn't need to be two different methods for split table/full table calculations, just a conditional to add the LPF gain in split gain table mode.  I could submit a PR implementing that if that would be preferable.

### DIFF
--- a/ad9361/sw/ad9361.c
+++ b/ad9361/sw/ad9361.c
@@ -659,16 +659,6 @@ static const uint8_t gm_st_ctrl[16] = { 0x0, 0xD, 0x15, 0x1B, 0x21, 0x25, 0x29,
 					0x2C, 0x2F, 0x31, 0x33, 0x34, 0x35, 0x3A, 0x3D, 0x3E
 				      };
 
-static const int8_t lna_table[RXGAIN_TBLS_END][4] = {
-	{5, 17, 19, 24}, {3, 14, 17, 21}, {-4, 10, 13, 14}
-};
-static const int8_t tia_table[] = { -6, 0 };
-static const int8_t mixer_table[RXGAIN_TBLS_END][16] = {
-	{0, 3, 9, 13, 14, 15, 16, 17, 18, 19, 20, 21, 22, 23, 24, 25},
-	{0, 3, 9, 14, 15, 16, 17, 18, 19, 20, 21, 22, 23, 24, 25, 26},
-	{0, 3, 8, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21, 22, 23, 24}
-};
-
 static const uint32_t gain_step_calib_reg_val[4][5] = {
 	{0xC0, 0x2E, 0x10, 0x06, 0x00},	// LO Frequency Range: 600 to 1300 MHz
 	{0xC0, 0x2C, 0x10, 0x06, 0x00},	// LO Frequency Range: 1300 to 3300 MHz
@@ -1815,9 +1805,8 @@ static int32_t ad9361_get_split_table_gain(struct ad9361_rf_phy *phy,
 
 	rx_gain->tia_index = ad9361_spi_readf(spi, REG_GAIN_TABLE_READ_DATA2, TIA_GAIN);
 
-	rx_gain->lmt_gain = lna_table[ad9361_gt(phy)][rx_gain->lna_index] +
-			    mixer_table[ad9361_gt(phy)][rx_gain->mixer_index] +
-			    tia_table[rx_gain->tia_index];
+	rx_gain->lmt_gain = phy->gt_info[ad9361_gt(
+			phy)].abs_gain_tbl[rx_gain->fgt_lmt_index];
 
 	ad9361_spi_write(spi, REG_GAIN_TABLE_ADDRESS, tbl_addr);
 


codypiersall:
Note that this also effects the [kernel version](https://github.com/analogdevicesinc/linux/blob/eadedb6d5a4c235caa1edddca4a209c96698fc3f/drivers/iio/adc/ad9361.c#L1515-L1550)!  But I won't be putting in a PR for that one, since I can't test it.

amiclaus:
Have a look at https://github.com/analogdevicesinc/no-OS/wiki/Code-Style-guidelines to fix the Travis build please.

amiclaus:
Also, a commit description and "signed-off by" is required, in case of a merge on the master branch.

buha:
@codypiersall could you please do a `git commit --amend -s` on this commit ?
This will add a "Signed-off by" tag in the description.

dbogdan:
https://github.com/analogdevicesinc/no-OS/commit/004e8c1d419da913b66a00cda9a710e232ef2ca9 is fixing the issue.

PR #398, opened 2019-12-17
platform: i2c: update

amiclaus:
With #285, i2c R/W function headers changed.

This patch replaces `stop_bit` parameter with the `option` parameter
for all i2c source files.


amiclaus:
V2:

- improve consistency: replace `stop_bit` with `option` also for `xilinx/platform_drivers`

PR #405, opened 2019-12-19
drivers: platform: change init_param variable

amiclaus:
Change `init_param` to `param` matching all init function headers from `/include` folder and other platform drivers implementations.

Fixes unmatching parameter name warnings generated by Doxygen integration.


dbogdan:
I feel that "param" is too generic. I would better call it init instead.

amiclaus:
V2:
- update commit message

PR #429, opened 2020-01-14
includes: i2c: Modify name and type of options

mchindri:
Modify options name to transfer_mode.
Modify transfer_mode type to enum i2c_transfer_mode.
Add filed i2c_slave_monitor to transfer_mode necesary on xilinx.
implementation.
Make necesary modifications in platform driveres implementtions.
This changes make easy to use and understand the functionality of
transfer_mode parameter

### DIFF
--- a/drivers/platform/altera/i2c.c
+++ b/drivers/platform/altera/i2c.c
@@ -89,15 +89,13 @@ int32_t i2c_remove(struct i2c_desc *desc)
  * @param desc - The I2C descriptor.
  * @param data - Buffer that stores the transmission data.
  * @param bytes_number - Number of bytes to write.
- * @param option - Stop condition control.
- *                   Example: 0 - A stop condition will not be generated;
- *                            1 - A stop condition will be generated.
+ * @param transfer_mode - Transfer mode type
  * @return SUCCESS in case of success, FAILURE otherwise.
  */
 int32_t i2c_write(struct i2c_desc *desc,
 		  uint8_t *data,
 		  uint8_t bytes_number,
-		  uint8_t option)
+		  enum i2c_transfer_mode transfer_mode)
 {
 	if (desc) {
 		// Unused variable - fix compiler warning
@@ -111,7 +109,7 @@ int32_t i2c_write(struct i2c_desc *desc,
 		// Unused variable - fix compiler warning
 	}
 
-	if (option) {
+	if (transfer_mode) {
 		// Unused variable - fix compiler warning
 	}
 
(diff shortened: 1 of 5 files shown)


mchindri:
V2:
 - Add missing ,

mchindri:
I didn't check if the code run on xilinx because I don't know how.
Only the xilinx implementations uses the modified parameter so in the others implementation shouldn't be problems.
Sergiu said that the xilinx i2c driver it is not used in any project yet, so there is no need for project modifications.

mchindri:
V3:
 - Fix some indentations

mchindri:
v3
 - Change  wrong i2c_10_bit_transfer to i2c_repeated_start

include/i2c.h
```diff
@@ -56,11 +56,13 @@
...
 typedef enum i2c_transfer_mode {
 	/** Address every device connected */
-	i2c_general_call =	0x01,
+	i2c_general_call = 0x01,
 	/** Send multiple start conditions */
-	i2c_repeated_start =	0x02,
+	i2c_repeated_start = 0x02,
 	/** Use 10-bit address scheme */
-	i2c_10_bit_transfer =	0x04
+	i2c_10_bit_transfer = 0x04,
```
amiclaus: I'd use uppercase letters for enum's constants. It's easier to differeniate between them and other variables.

https://github.com/analogdevicesinc/no-OS/blob/b1cfb291db2e735fe61ad464ed573f475790d5ac/include/spi.h#L64-L73

https://github.com/analogdevicesinc/no-OS/blob/b1cfb291db2e735fe61ad464ed573f475790d5ac/drivers/adc/ad717x/ad717x.h#L56-L60

mchindri: True, I know something was wrong with the names but I was missing what was :)

mchindri:
v4:
 - fix astyle
 - add enum value I2C_NORMAL_TRANSFER
 - make enum values with caps

drivers/platform/altera/i2c.c
```diff
@@ -89,15 +89,13 @@ int32_t i2c_remove(struct i2c_desc *desc)
...
- *                   Example: 0 - A stop condition will not be generated;
- *                            1 - A stop condition will be generated.
+ * @param transfer_mode - Transfer mode type
  * @return SUCCESS in case of success, FAILURE otherwise.
  */
 int32_t i2c_write(struct i2c_desc *desc,
 		  uint8_t *data,
 		  uint8_t bytes_number,
-		  uint8_t option)
+		  enum i2c_transfer_mode transfer_mode)
```
buha: I think that because i2c_transfer_mode is defined as a typedef, you can skip the `enum` qualifier completely so your parameter could be just:
```
...,
i2c_transfer_mode transfer_mode)
```

mchindri: According to linux coding style we mustn't use typedefs. I wanted to remove the typedef but I'm not sure if some other file use the structure names without the struct keyword.

PR #437, opened 2020-01-16
drivers: axi: Add platform agnostic spi engine

scuciurean:
Added platform-agnostic spi engine inside xilinx platform drivers.

### DIFF
--- a/drivers/axi_core/spi_engine/spi_engine.c
+++ b/drivers/axi_core/spi_engine/spi_engine.c
@@ -1,0 +1,841 @@
+/*******************************************************************************
+ *   @file   spi_engine.c
+ *   @brief  Core implementation of the SPI Engine Driver.
+ *   @author Sergiu Cuciurean (sergiu.cuciurean@analog.com)
+********************************************************************************
+ * Copyright 2019(c) Analog Devices, Inc.
+ *
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *  - Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ *  - Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in
+ *    the documentation and/or other materials provided with the
+ *    distribution.
+ *  - Neither the name of Analog Devices, Inc. nor the names of its
+ *    contributors may be used to endorse or promote products derived
+ *    from this software without specific prior written permission.
+ *  - The use of this software may or may not infringe the patent rights
+ *    of one or more patent holders.  This license does not release you
+ *    from the requirement that you obtain separate licenses from these
+ *    patent holders to use this software.
+ *  - Use of the software either in source or binary form, must be run
+ *    on or directly connected to an Analog Devices Inc. component.
+ *
+ * THIS SOFTWARE IS PROVIDED BY ANALOG DEVICES "AS IS" AND ANY EXPRESS OR
+ * IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, NON-INFRINGEMENT,
+ * MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED.
+ * IN NO EVENT SHALL ANALOG DEVICES BE LIABLE FOR ANY DIRECT, INDIRECT,
+ * INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, INTELLECTUAL PROPERTY RIGHTS, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ ******************************************************************************/
+
+/******************************************************************************/
+/***************************** Include Files **********************************/
+/******************************************************************************/
+
+/* In debug mode the printf function used in displaying the messages is causing
+significant delays */
+//#define DEBUG_LEVEL 2
+
+#include <stdbool.h>
+#include <stdint.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <sleep.h>
+
+#include "axi_dmac.h"
+#include "axi_io.h"
+#include "error.h"
+#include "spi_engine.h"
+
+/******************************************************************************/
+/***************************** Static variables *******************************/
+/******************************************************************************/
+
+/**
+ * @brief Static variable used to sync transfers
+ *
+ */
+static uint8_t _sync_id = 0x01;
+
+/******************************************************************************/
+/************************** Functions Implementation **************************/
+/******************************************************************************/
+
+/**
+ * @brief Write SPI Engine's axi registers
+ *
+ * @param desc Decriptor containing SPI Engine's parameters
+ * @param reg_addr The address of the SPI Engine's axi register where the data
+ * 	will be written
+ * @param reg_data Data that will be written
+ * @return int32_t This function allways returns SUCCESS
+ */
+static int32_t spi_engine_write(struct spi_engine_desc *desc,
+				uint32_t reg_addr,
+				uint32_t reg_data)
+{
+	axi_io_write(desc->spi_engine_baseaddr, reg_addr, reg_data);
+
+	return SUCCESS;
+}
+
+/**
+ * @brief Read SPI Engine's axi registers
+ *
+ * @param desc Decriptor containing SPI Engine's parameters
+ * @param reg_addr The address of the SPI Engine's axi register from where the
+ * 	data where the data will be read
+ * @param reg_data Pointer where the read that will be stored
+ * @return int32_t This function allways returns SUCCESS
+ */
+static int32_t spi_engine_read(struct spi_engine_desc *desc,
+			       uint32_t reg_addr,
+			       uint32_t *reg_data)
+{
+	axi_io_read(desc->spi_engine_baseaddr, reg_addr, reg_data);
+	return SUCCESS;
+}
+
+/**
+ * @brief Set width of the transfered word over SPI
+ *
+ * @param desc Decriptor containing SPI interface parameters
+ * @param data_wdith The desired data width
+ * 	The supported values are:
+ * 		- 8
+ * 		- 16
+ * 		- 24
+ * 		- 32
+ */
+static int32_t spi_engine_set_transfer_width(struct spi_desc *desc,
+		uint8_t data_wdith)
+{
+	struct spi_engine_desc	*desc_extra;
+
+	desc_extra = desc->extra;
+
+	if (data_wdith > desc_extra->max_data_width)
+		desc_extra->data_width = desc_extra->max_data_width;
+	else
+		desc_extra->data_width = data_wdith;
+
+	return SUCCESS;
+}
+
+/**
+ * @brief Set the number of words transfered in a single transaction
+ *
+ * @param desc Decriptor containing SPI Engine's parameters
+ * @param bytes_number The number of bytes to be converted
+ * @return uint8_t A number of words in which bytes_number can be grouped
+ */
+static uint8_t spi_get_words_number(struct spi_engine_desc *desc,
+				    uint8_t bytes_number)
+{
+	uint8_t xfer_word_len;
+	uint8_t words_number;
+
+	xfer_word_len = desc->data_width / 8;
+	words_number = bytes_number / xfer_word_len;
+
+	if ((bytes_number % xfer_word_len) != 0)
+		words_number++;
+
+	return words_number;
+}
+
+/**
+ * @brief Get the word length in bytes
+ *
+ * @param desc Decriptor containing SPI Engine's parameters
+ * @return uint8_t Number of bytes that fit in one word
+ */
+static uint8_t spi_get_word_lenght(struct spi_engine_desc *desc)
+{
+	uint8_t word_lenght;
+
+	word_lenght = desc->data_width / 8;
+
+	return word_lenght;
+}
+
+/**
+ * @brief Compute the prescaler used to set the sleep period.
+ *
+ * @param desc Decriptor containing SPI interface parameters
+ * @param sleep_time_ns The amount of time where the transfer hangs
+ * @param sleep_div Clock prescaler
+ * @return int32_t This function allways returns SUCCESS
+ */
+static int32_t spi_get_sleep_div(struct spi_desc *desc,
+				 uint32_t sleep_time_ns,
+				 uint32_t *sleep_div)
+{
+	struct spi_engine_desc	*eng_desc;
+
+	eng_desc = desc->extra;
+
+	/*
+	 * Engine Wiki:
+	 *
+	 * The frequency of the SCLK signal is derived from the
+	 * module clock frequency using the following formula:
+	 * f_sclk = f_clk / ((div + 1) * 2)
+	 */
+
+	*sleep_div = (desc->max_speed_hz / 1000000 * sleep_time_ns / 1000) /
+		     ((eng_desc->clk_div + 1) * 2) - 1;
+
+	return SUCCESS;
+}
+
+/**
+ * @brief Create a new commands queue used in a spi transfer
+ *
+ * @param fifo Command buffer, usualy used in fifo mode
+ * @param cmd Command with wich the buffer will be initiated
+ * @return int32_t Failure if the memory allocation failed
+ */
+static int32_t spi_engine_queue_new_cmd(struct spi_engine_cmd_queue **fifo,
+					uint32_t cmd)
+{
+	struct spi_engine_cmd_queue *local_fifo;
+
+	local_fifo = (spi_engine_cmd_queue*)malloc(sizeof(*local_fifo));
+
+	if(!local_fifo)
+		return FAILURE;
+
+	local_fifo->cmd = cmd;
+	local_fifo->next = NULL;
+	*fifo = local_fifo;
+
+	return SUCCESS;
+}
+
+/**
+ * @brief Add a command at the end of an existing queue
+ *
+ * @param fifo Command buffer, usualy used in fifo mode
+ * @param cmd Command to be added
+ */
+static void spi_engine_queue_add_cmd(struct spi_engine_cmd_queue **fifo,
+				     uint32_t cmd)
+{
+	struct spi_engine_cmd_queue *to_add = NULL;
+	struct spi_engine_cmd_queue *local_fifo;
+
+	local_fifo = *fifo;
+	while (local_fifo->next != NULL) {
+		// Get the last element
+		local_fifo = local_fifo->next;
+	}
+
+	// Create a new element
+	spi_engine_queue_new_cmd(&to_add, cmd);
+	// Add as the next element
+	local_fifo->next = to_add;
+}
+
+/**
+ * @brief Add a command at the beginning of an existing queue
+ *
+ * @param fifo Command buffer, usualy used in fifo mode
+ * @param cmd Command to be added
+ */
+static void spi_engine_queue_append_cmd(struct spi_engine_cmd_queue **fifo,
+					uint32_t cmd)
+{
+	struct spi_engine_cmd_queue *to_add = NULL;
+
+	// Create a new element
+	spi_engine_queue_new_cmd(&to_add, cmd);
+	// Interchange the addresses
+	to_add->next = *fifo;
+	*fifo = to_add;
+}
+
+
+/**
+ * @brief Get the command from the beginning of the queue
+ *
+ * @param fifo Command buffer, usualy used in fifo mode
+ * @param cmd Value of the command that was extracted
+ * @return int32_t Return FAILURE if the queue is empty
+ */
+static int32_t spi_engine_queue_get_cmd(struct spi_engine_cmd_queue **fifo,
+					uint32_t *cmd)
+{
+	struct spi_engine_cmd_queue *local_fifo;
+
+	if (!*fifo)
+		return FAILURE;
+
+	local_fifo = *fifo;
+	*cmd = local_fifo->cmd;
+	if ((*fifo)->next) {
+		*fifo = local_fifo->next;
+		free(local_fifo);
+	} else {
+		free(*fifo);
+		*fifo = NULL;
+	}
+
+	return SUCCESS;
+}
+
+/**
+ * @brief Free the memory allocated by a queue
+ *
+ * @param fifo The queue that needs it's memory freed
+ * @return int32_t This function allways return SUCCESS
+ */
+static int32_t spi_engine_queue_free(struct spi_engine_cmd_queue **fifo)
+{
+	if(*fifo && (*fifo)->next)
+		spi_engine_queue_free(&(*fifo)->next);
+	if((*fifo) != NULL) {
+		free(*fifo);
+		*fifo = NULL;
+	}
+
+	return SUCCESS;
+}
+
+/**
+ * @brief Write the SPI engine's command fifo
+ *
+ * @param desc Decriptor containing SPI Engine's parameters
+ * @param cmd The command that will be written
+ * @return int32_t Returns FAILURE if the the axi transfer failed
+ */
+static int32_t spi_engine_write_cmd_reg(struct spi_engine_desc *desc,
+					uint32_t cmd)
+{
+	int32_t ret;
+
+	/* Check if offload is enabled */
+	if(desc->offload_config & (OFFLOAD_TX_EN | OFFLOAD_RX_EN)) {
+		ret = spi_engine_write(desc,
+				       SPI_ENGINE_REG_OFFLOAD_CMD_MEM(0),
+				       cmd);
+
+	} else {
+		ret = spi_engine_write(desc,
+				       SPI_ENGINE_REG_CMD_FIFO,
+				       cmd);
+	}
+
+	return ret;
+}
+
+/**
+ * @brief Write a transfer command to the SPI engine's command fifo
+ *
+ * @param desc Decriptor containing SPI Engine's parameters
+ * @param read_write Read/Write operation flag
+ * @param bytes_number Number of bytes to transfer
+ * @return int32_t This function allways returns SUCCESS
+ */
+static int32_t spi_engine_transfer(struct spi_engine_desc *desc,
+				   uint8_t read_write,
+				   uint8_t bytes_number)
+{
+	uint8_t words_number;
+
+	words_number = spi_get_words_number(desc, bytes_number);
+
+	desc->offload_tx_len += words_number;
+
+	/*
+	 * Engine Wiki:
+	 *
+	 * https://wiki.analog.com/resources/fpga/peripherals/spi_engine
+	 *
+	 * The words number is zero based
+	 */
+
+	spi_engine_write_cmd_reg(desc,
+				 SPI_ENGINE_CMD_TRANSFER(read_write,
+						 words_number  - 1));
+
+	return SUCCESS;
+}
+
+/**
+ * @brief Change the state of the chip select port
+ *
+ * @param desc Decriptor containing SPI interface parameters
+ * @param assert Chip select state.
+ * 		 The supported values are :
+ * 			-true (HIGH)
+ * 			-false (LOW)
+ */
+static void spi_engine_set_cs(struct spi_desc *desc,
+			      bool assert)
+{
+	uint8_t			mask;
+	struct spi_engine_desc	*eng_desc;
+
+	eng_desc = desc->extra;
+
+	mask = 0xFF;
+	/* Switch the state only of the selected chip select */
+	if (!assert)
+		mask ^= BIT(desc->chip_select);
+
+	spi_engine_write_cmd_reg(eng_desc,
+				 SPI_ENGINE_CMD_ASSERT(eng_desc->cs_delay,
+						 mask));
+}
+
+/**
+ * @brief Add a delay bewtheen the engine commands
+ *
+ * @param desc Decriptor containing SPI interface parameters
+ * @param sleep_time_ns Number of nanoseconds to sleep between commands
+ */
+static void spi_gen_sleep_ns(struct spi_desc *desc,
+			     uint32_t sleep_time_ns)
+{
+	uint32_t 		sleep_div;
+	struct spi_engine_desc	*eng_desc;
+
+	eng_desc = desc->extra;
+
+	spi_get_sleep_div(desc, sleep_time_ns, &sleep_div);
+
+	spi_engine_write_cmd_reg(eng_desc, SPI_ENGINE_CMD_SLEEP(sleep_div));
+}
+
+/**
+ * @brief Spi engine command interpreter
+ *
+ * @param desc Decriptor containing SPI interface parameters
+ * @param cmd Command to send to the engine
+ * @return int32_t - SUCCESS if the command is transfered
+ *		   - FAILURE if the command format is invalid
+ */
+static int32_t spi_engine_write_cmd(struct spi_desc *desc,
+				    uint32_t cmd)
+{
+	uint8_t				engine_command;
+	uint8_t				parameter;
+	uint8_t				modifier;
+	struct spi_engine_desc		*desc_extra;
+
+	desc_extra = desc->extra;
+
+	engine_command = (cmd >> 12) & 0x0F;
+	modifier = (cmd >> 8) & 0x0F;
+	parameter = cmd & 0xFF;
+
+	switch(engine_command) {
+	case SPI_ENGINE_INST_TRANSFER:
+		spi_engine_transfer(desc_extra, modifier, parameter);
+		break;
+
+	case SPI_ENGINE_INST_ASSERT:
+		if(parameter == 0xFF) {
+			/* Set the CS HIGH */
+			spi_engine_set_cs(desc, true);
+		} else if(parameter == 0x00) {
+			/* Set the CS LOW */
+			spi_engine_set_cs(desc, false);
+		}
+		break;
+
+	/* The SYNC and SLEEP commands got the same value but different
+	modifier */
+	case SPI_ENGINE_INST_SYNC_SLEEP:
+		/* SYNC instruction */
+		if(modifier == 0x00) {
+			spi_engine_write_cmd_reg(desc_extra, cmd);
+		} else if(modifier == 0x01) {
+			spi_gen_sleep_ns(desc, parameter);
+		}
+		break;
+	case SPI_ENGINE_INST_CONFIG:
+		spi_engine_write_cmd_reg(desc_extra, cmd);
+
+		break;
+
+	default:
+
+		return FAILURE;
+		break;
+	}
+
+	return SUCCESS;
+}
+
+/**
+ * @brief Prepare the command queue before sending it to the engine
+ *
+ * @param desc Decriptor containing SPI interface parameters
+ * @param msg Structure used to store the transfer messages
+ * @return int32_t This function allways returns SUCCESS
+ */
+static int32_t spi_engine_compile_message(struct spi_desc *desc,
+		struct spi_engine_msg *msg)
+{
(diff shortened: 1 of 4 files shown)


drivers/axi_core/spi_engine/spi_engine.c
```diff
@@ -0,0 +1,958 @@
...
+	local_fifo->next = to_add;
+}
+
+/**
+ * @brief Add a command at the beginning of an existing queue
+ *
+ * @param fifo Command buffer, ussualy used in fifo mode
+ * @param cmd Command to be added
+ */
+void spi_engine_queue_append_cmd(struct spi_engine_cmd_queue **fifo,
```
cristiantpop: Add static to all local functions.

drivers/axi_core/spi_engine/spi_engine.c
```diff
@@ -0,0 +1,958 @@
...
+	spi_engine_queue_new_cmd(&to_add, cmd);
+	// Interchange the addresses
+	to_add->next = *fifo;
+	*fifo = to_add;
+}
+
+/**
+ * @brief Get the command from the end of the queue
+ *
+ * @param fifo Command buffer, ussualy used in fifo mode
```
buha: usually

drivers/axi_core/spi_engine/spi_engine.c
```diff
@@ -0,0 +1,958 @@
...
+ * @param cmd The command that will be written
+ * @return int32_t Returns FAILURE if the the axi transfer failed
+ */
+int32_t spi_engine_write_cmd_reg(struct spi_engine_desc *desc,
+				 uint32_t cmd)
+{
+	int32_t ret;
+
+	/* Check if offload is enabled */
+	if(desc->offload_config & (OFFLOAD_TX_EN | OFFLOAD_RX_EN)) {
```
buha: `desc` should be tested before dereferencing. This applies to all subsequent functions that use desc as parameter.

drivers/axi_core/spi_engine/spi_engine.h
```diff
@@ -0,0 +1,188 @@
...
+/* Write and read data over SPI using the SPI engine */
+int32_t spi_engine_write_and_read(spi_desc *desc,
+				  uint8_t *data,
+				  uint8_t bytes_number);
+
+/* Free the resources used by the SPI engine device */
+int32_t spi_engine_remove(spi_desc *desc);
+
+/* Initialize the SPI engine offload module */
+int32_t spi_engine_offload_init(struct spi_desc *desc,
```
mchindri: Could this be inside spi_engine_init and pass `spi_engine_offload_init_param` as a field of spi_engine_init_param?
In the case the field is null make no offload init.
I'am asking this because there is no spi_engine_offload_remove and the remove for this is done in spi_engine_remove.
Also it seems the spi_engine_offload_transfer uses the spi_desc not an descriptor from offload_init. It looks a little puzzling to me.

drivers/axi_core/spi_engine/spi_engine.c
```diff
@@ -0,0 +1,958 @@
...
+}
+
+/**
+ * @brief Set the number of words transfered in a single transaction
+ *
+ * @param desc Decriptor containing SPI Engine's parameters
+ * @param bytes_number The number of bytes to be cnoverted
+ * @return uint8_t A number of words in which bytes_number can be grouped
+ */
+uint8_t spi_get_words_number(struct spi_engine_desc *desc,
```
mchindri: could this be an static inline function?

drivers/axi_core/spi_engine/spi_engine.c
```diff
@@ -0,0 +1,958 @@
...
+			     uint32_t cmd)
+{
+	uint8_t				engine_command;
+	uint8_t				parameter;
+	uint8_t				modifier;
+	struct spi_engine_desc		*desc_extra;
+
+	desc_extra = desc->extra;
+
+	engine_command = (cmd >> 12) & 0x0F;
```
buha: As somebody who is not familiar with the code, this looks obscure. Please document why this shifting and masking is done.

drivers/axi_core/spi_engine/spi_engine.c
```diff
@@ -0,0 +1,958 @@
...
+ */
+int32_t spi_engine_compile_message(struct spi_desc *desc,
+				   struct spi_engine_msg *msg)
+{
+	struct spi_engine_desc	*desc_extra;
+
+	desc_extra = desc->extra;
+
+	/* Set the data transfer length */
+	spi_engine_queue_append_cmd(&msg->cmds,
```
buha: `msg` needs to be tested before dereferencing

drivers/axi_core/spi_engine/spi_engine.c
```diff
@@ -0,0 +1,958 @@
...
+	spi_engine_compile_message(desc, msg);
+
+	offload_en = (desc_extra->offload_config & OFFLOAD_TX_EN) |
+		     (desc_extra->offload_config & OFFLOAD_RX_EN);
+
+	/* Write the command fifo buffer */
+	do {
+		spi_engine_queue_lifo_get_cmd(&msg->cmds, &data);
+		spi_engine_write_cmd(desc, data);
+		usleep(1000);
```
buha: This looks like a hack to make it work. Is this really needed ? If yes, please document why is it needed.

drivers/axi_core/spi_engine/spi_engine.c
```diff
@@ -0,0 +1,958 @@
...
+	struct spi_engine_desc		*eng_desc;
+	struct spi_engine_init_param	*spi_engine_init;
+
+	local_desc = (struct spi_desc *)malloc(sizeof(*local_desc));
+	eng_desc = (struct spi_engine_desc*)malloc(sizeof(*eng_desc));
+
+	if (!desc || !eng_desc)
+		return FAILURE;
+
+	spi_engine_init = param->extra;
```
buha: `param` needs to be tested before dereferencing

drivers/axi_core/spi_engine/spi_engine.c
```diff
@@ -0,0 +1,958 @@
...
+	/* Make sure the CS is HIGH before starting a transaction */
+	msg->cmds->next = NULL;
+	msg->cmds->cmd = CS_HIGH;
+	spi_engine_queue_add_cmd(&msg->cmds, CS_LOW);
+	spi_engine_queue_add_cmd(&msg->cmds, WRITE_READ(bytes_number));
+	spi_engine_queue_add_cmd(&msg->cmds, CS_HIGH);
+
+	/* Pack the bytes into engine WORDS */
+	for (i = 0; i < bytes_number; i++)
+		msg->tx_buf[i / word_len] |= data[i] << (desc_extra->data_width-
```
buha: `desc_extra` needs to be tested before dereferencing

drivers/axi_core/spi_engine/spi_engine.c
```diff
@@ -0,0 +1,958 @@
...
+	msg->cmds = (spi_engine_cmd_queue*)malloc(sizeof(*msg->cmds));
+	if (!msg->cmds || !msg)
+		return FAILURE;
+
+	msg->tx_buf =(uint32_t*)malloc(words_number * sizeof(msg->tx_buf[0]));
+	msg->rx_buf =(uint32_t*)malloc(words_number * sizeof(msg->rx_buf[0]));
+	msg->length = words_number;
+
+	/* Init the rx and tx buffers with 0s */
+	for (i = 0; i < words_number; i++) {
```
buha: you could use bzero() for this

drivers/axi_core/spi_engine/spi_engine.c
```diff
@@ -0,0 +1,958 @@
...
+ */
+int32_t spi_engine_write_and_read(struct spi_desc *desc,
+				  uint8_t *data,
+				  uint8_t bytes_number)
+{
+	uint8_t 		i;
+	uint8_t 		word_len;
+	uint8_t 		words_number;
+	int32_t 		ret;
+	struct spi_engine_msg	*msg;
```
mchindri: this haven't to be a pointer. This way there will not be need for  malloc and free

drivers/axi_core/spi_engine/spi_engine.c
```diff
@@ -0,0 +1,958 @@
...
+	struct spi_engine_desc	*desc_extra;
+
+	desc_extra = desc->extra;
+
+	words_number = spi_get_words_number(desc_extra, bytes_number);
+
+	msg = (spi_engine_msg *)malloc(sizeof(*msg));
+	msg->cmds = (spi_engine_cmd_queue*)malloc(sizeof(*msg->cmds));
```
mchindri: instead of this will be better something like 
spi_engine_queue_init(&msg->cmds);

drivers/axi_core/spi_engine/spi_engine.c
```diff
@@ -0,0 +1,958 @@
...
+	if (!msg->cmds || !msg)
+		return FAILURE;
+
+	msg->tx_buf =(uint32_t*)malloc(words_number * sizeof(msg->tx_buf[0]));
```
mchindri: you can use calloc and it will fill the buffers with 0.
Also calloc it is prefered in all the allocations.

scuciurean:
Changelog -> V2:

1) Add static specifier to all local functions.

drivers/axi_core/spi_engine/spi_engine.c
```diff
@@ -0,0 +1,960 @@
...
+	} while(msg->cmds->next != NULL);
+	/* Write the last command */
+	spi_engine_queue_fifo_get_cmd(&msg->cmds, &data);
+	spi_engine_write_cmd(desc, data);
+
+	/* Write a number of tx_length WORDS on the SDO line */
+
+	if(offload_en) {
+		for(i = 0; i < desc_extra->offload_tx_len; i++) {
+			spi_engine_write(desc_extra,
```
mchindri: there is no need for {}. The same for others calls to spi_engine_write

drivers/axi_core/spi_engine/spi_engine.c
mchindri: It is an internal function. I think here msg must always come well configure.

drivers/axi_core/spi_engine/spi_engine.c
mchindri: same for spi_engine_offload_transfer

drivers/axi_core/spi_engine/spi_engine.c
mchindri: same for spi_engine_offload_transfer

drivers/axi_core/spi_engine/spi_engine.c
```diff
@@ -0,0 +1,960 @@
...
+	while(i < msg.no_commands) {
+		spi_engine_queue_add_cmd(&transfer->cmds, msg.commands[i++]);
+
+	}
+
+	spi_engine_transfer_message(desc, transfer);
+
+	word_length = spi_get_word_lenght(eng_desc);
+	if(eng_desc->offload_config & OFFLOAD_TX_EN) {
+		axi_dmac_transfer(eng_desc->offload_tx_dma,
```
mchindri: no need for {}. Same for next if

drivers/platform/xilinx/spi.c
```diff
@@ -70,116 +157,79 @@ int32_t spi_init(struct spi_desc **desc,
...
+	struct xil_spi_desc 		*xdesc;
 	struct xil_spi_init_param	*xinit;
+	struct spi_engine_desc 		*edesc;
+	enum xil_spi_type		*spi_type;
 
-	sdesc = (struct spi_desc *)malloc(sizeof(*sdesc));
-	xdesc = (struct xil_spi_desc *)malloc(sizeof(*xdesc));
-	if(!sdesc || !xdesc) {
+	sdesc = malloc(sizeof(*sdesc));
+	if(!sdesc) {
```
mchindri: you must return only FAILURE, no free needed

drivers/platform/xilinx/spi.c
```diff
@@ -70,116 +157,79 @@ int32_t spi_init(struct spi_desc **desc,
...
+	 * have the fisrt member enum xil_spi_type so in this case
+	 * we can read the first member without casting the pointer to
+	 * a certain structure type.
+	*/
+	spi_type = param->extra;
+
+	if (*spi_type == SPI_ENGINE) {
+		edesc = malloc(sizeof(spi_engine_desc));
+		if(!edesc) {
+			free(edesc);
```
mchindri: free(sdesc). And the next one is not ok either. Check the goto failure pattern here https://github.com/analogdevicesinc/no-OS/blob/master/drivers/platform/aducm3029/spi.c#L105

drivers/platform/xilinx/spi.c
```diff
@@ -54,11 +54,98 @@
...
+
+pl_error:
+	free(xdesc->instance);
+#endif
+	free(xdesc);
+
+	return FAILURE;
+}
+
+int32_t spi_init_ps(struct xil_spi_desc *xdesc,
```
cristiantpop: Missing function comment.

drivers/platform/xilinx/spi.c
```diff
@@ -54,11 +54,98 @@
 #include "error.h"
 #include "spi.h"
 #include "spi_extra.h"
+#include "spi_engine.h"
 
 /******************************************************************************/
 /************************ Functions Definitions *******************************/
 /******************************************************************************/
 
+int32_t spi_init_pl(struct xil_spi_desc *xdesc,
```
cristiantpop: Missing function comment.

drivers/axi_core/spi_engine/spi_engine.h
```diff
@@ -0,0 +1,188 @@
...
+	/** The address where the data that will be received is situated */
+	uint32_t rx_addr;
+} spi_engine_offload_message;
+
+/******************************************************************************/
+/************************ Functions Declarations ******************************/
+/******************************************************************************/
+
+/* Initialize the SPI engine device */
+int32_t spi_engine_init(struct spi_desc **desc,
```
mchindri: I think a struct spi_engine_desc must be initialized which will be the extra field of the spi_desc. Review the init and remove flow from spi.c and you will see the problem.

drivers/platform/xilinx/spi.c
```diff
@@ -54,11 +54,98 @@
...
+				     XSPIPS_CLK_PRESCALE_64);
+	if(ret != SUCCESS)
+		goto ps_error;
+
+	return SUCCESS;
+
+ps_error:
+	free(xdesc->instance);
+#endif
+	free(xdesc);
```
mchindri: this free must not be done here. It is done in the outside function. Same for init_pl

mchindri:
f53eaba18faf9bda0648ebfaa4d5345d42355592 this comment should be splited in 2. First create the functions spi_init_ps and spi_init_pl and the support for spi_engine.

drivers/axi_core/spi_engine/spi_engine.c
```diff
@@ -0,0 +1,960 @@
...
+ *
+ * @param desc Decriptor containing SPI interface parameters
+ * @param data_wdith The desired data width
+ * 	The supported values are:
+ * 		- 8
+ * 		- 16
+ * 		- 24
+ * 		- 32
+ */
+static int32_t spi_engine_set_transfer_wdith(struct spi_desc *desc,
```
adrimbarean: nitpick: wdith -> width

drivers/axi_core/spi_engine/spi_engine.c
```diff
@@ -0,0 +1,960 @@
...
+		uint8_t data_wdith)
+{
+	struct spi_engine_desc	*desc_extra;
+
+	desc_extra = desc->extra;
+
+	if (data_wdith > desc_extra->max_data_width) {
+		desc_extra->data_width = desc_extra->max_data_width;
+		dev_dbg(desc, "Data width (%d) exceeds the maximum of %d bits",
```
adrimbarean: Maybe the message here should specifically state that the data width has been changed  to the maximum possible. The user might assume nothing changed based on the error message.

drivers/axi_core/spi_engine/spi_engine.c
```diff
@@ -0,0 +1,960 @@
...
+	return SUCCESS;
+}
+
+/**
+ * @brief Add a command at the end of an existing queue
+ *
+ * @param fifo Command buffer, ussualy used in fifo mode
+ * @param cmd Command to be added
+ */
+static void spi_engine_queue_add_cmd(struct spi_engine_cmd_queue **fifo,
```
adrimbarean: I don't think double pointer is necessary here. `fifo` should not change.

drivers/axi_core/spi_engine/spi_engine.c
```diff
@@ -0,0 +1,960 @@
...
+	struct spi_engine_cmd_queue *last_fifo = NULL;
+
+	local_fifo = *fifo;
+	while (local_fifo->next != NULL) {
+		// Get the last element
+		last_fifo = local_fifo;
+		local_fifo = local_fifo->next;
+	}
+	*cmd = local_fifo->cmd;
+	if(last_fifo->next != NULL) {
```
adrimbarean: Is this deferencing a NULL pointer if `fifo` has only one element?

drivers/axi_core/spi_engine/spi_engine.c
```diff
@@ -0,0 +1,960 @@
...
+}
+
+/**
+ * @brief Create a new commands queue used in a spi transfer
+ *
+ * @param fifo Command buffer, ussualy used in fifo mode
+ * @param cmd Command with wich the buffer will be initiated
+ * @return int32_t Failure if the memory allocation failed
+ */
+static int32_t spi_engine_queue_new_cmd(struct spi_engine_cmd_queue **fifo,
```
adrimbarean: nitpick: the command queue here behaves more like a directed graph than a FIFO, so probably another parameter name would be appropriate.

drivers/axi_core/spi_engine/spi_engine.c
adrimbarean: Also it seems like `last_fifo` will always be the second to last element in the FIFO, so `last_fifo->next` would never equal NULL.

drivers/axi_core/spi_engine/spi_engine.c
```diff
@@ -0,0 +1,960 @@
...
+ * @return int32_t Return FAILURE if the queue is empty
+ */
+static int32_t spi_engine_queue_lifo_get_cmd(struct spi_engine_cmd_queue **fifo,
+		uint32_t *cmd)
+{
+	struct spi_engine_cmd_queue *local_fifo;
+
+
+	if ((*fifo)->next) {
+		local_fifo = (*fifo);
```
adrimbarean: nitpick: No need for parenthesis.

drivers/platform/xilinx/spi.c
```diff
@@ -192,26 +242,35 @@ int32_t spi_remove(struct spi_desc *desc)
...
+
 		ret = XSpi_Stop((XSpi *)(xdesc->instance));
 		if(ret != SUCCESS)
 			goto error;
 #endif
 		break;
 	case SPI_PS:
 #ifdef XSPIPS_H
-
+		xdesc = desc->extra;
```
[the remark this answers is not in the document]
scuciurean: It's indented by the case above

drivers/platform/xilinx/spi.c
```diff
@@ -192,26 +242,35 @@ int32_t spi_remove(struct spi_desc *desc)
...
 #endif
 		break;
 	case SPI_ENGINE:
 #ifdef SPI_ENGINE_H
-
+		spi_engine_remove(desc);
```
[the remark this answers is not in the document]
scuciurean: It's indented by the case above

drivers/axi_core/spi_engine/spi_engine.c
```diff
@@ -0,0 +1,958 @@
...
+ * @param reg_data Data that will be written
+ * @return int32_t This function allways returns SUCCESS
+ */
+int32_t spi_engine_write(struct spi_engine_desc *desc,
+			 uint32_t reg_addr,
+			 uint32_t reg_data)
+{
+	dev_dbg_reg(desc, "reg_addr = 0x%X; reg_data = 0x%X", reg_addr,
+		    reg_data);
+	axi_io_write(desc->spi_engine_baseaddr, reg_addr, reg_data);
```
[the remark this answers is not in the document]
scuciurean: Since the descriptor is tested in the init function , there is no need to verify it each time

drivers/axi_core/spi_engine/spi_engine.c
```diff
@@ -0,0 +1,958 @@
...
+	struct spi_engine_cmd_queue *local_fifo;
+
+	local_fifo = (spi_engine_cmd_queue*)malloc(sizeof(*local_fifo));
+
+	if(!local_fifo)
+		return FAILURE;
+
+	local_fifo->cmd = cmd;
+	local_fifo->next = NULL;
+	*fifo = local_fifo;
```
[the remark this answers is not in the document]
scuciurean: I do this in the spi_engine_write_and_read once and the other cases in which fifo may be null is when a new queue is allocated. In that case, in spi_engine_queue_new_cmd the new fifo is verified

drivers/axi_core/spi_engine/spi_engine.c
```diff
@@ -0,0 +1,958 @@
...
+	return SUCCESS;
+}
+
+/**
+ * @brief Add a command at the end of an existing queue
+ *
+ * @param fifo Command buffer, ussualy used in fifo mode
+ * @param cmd Command to be added
+ */
+void spi_engine_queue_add_cmd(struct spi_engine_cmd_queue **fifo,
```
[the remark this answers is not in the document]
scuciurean: spi_engine_queue is used only in spi_engine.c and there is no need for a separate file

drivers/axi_core/spi_engine/spi_engine.c
scuciurean: The command structure is documented in the spi engine's wiki. I'll add a link in the coment above this command.

drivers/axi_core/spi_engine/spi_engine.c
scuciurean: This was forgotten after some tests and will be removed.

include/util.h
```diff
@@ -85,12 +85,7 @@
 
 #define BITS_PER_LONG 32
 
-#define GENMASK(h, l) ({ 										\
-		uint32_t t = (~0UL);									\
-		t = t << (BITS_PER_LONG - (h - l + 1));					\
-		t = t >> (BITS_PER_LONG - (h + 1));						\
-		t;														\
-})
+#define GENMASK(h, l)	(((~0UL) << (l)) & (~0UL >> (32 - 1 - (h))))
```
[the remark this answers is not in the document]
scuciurean: It's the same as include/linux/bits.h

drivers/axi_core/spi_engine/spi_engine_private.h
```diff
@@ -0,0 +1,201 @@
...
+	SPI_ENGINE_CMD(SPI_ENGINE_INST_MISC, 				\
+			SPI_ENGINE_MISC_SYNC, 				\
+			(id))
+
+/******************************************************************************/
+/******************************* Debug macros *********************************/
+/******************************************************************************/
+
+#ifdef DEBUG_LEVEL
+#define dev_dbg_reg(dev, format, ...){			\
```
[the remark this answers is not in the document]
scuciurean: Good point

drivers/axi_core/spi_engine/spi_engine_private.h
```diff
@@ -0,0 +1,201 @@
...
+#define dev_info(dev, format, ...){			\
+	printf(format, ## __VA_ARGS__);			\
+	printf("\n");					\
+	}
+
+/******************************************************************************/
+/*************************** Types Declarations *******************************/
+/******************************************************************************/
+
+typedef struct spi_engine_cmd_queue {
```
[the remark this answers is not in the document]
scuciurean: spi_engine_queue is used only in spi_engine.c and there is no need for a separate file

drivers/axi_core/spi_engine/spi_engine.c
```diff
@@ -0,0 +1,958 @@
...
+ * IN NO EVENT SHALL ANALOG DEVICES BE LIABLE FOR ANY DIRECT, INDIRECT,
+ * INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, INTELLECTUAL PROPERTY RIGHTS, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ ******************************************************************************/
+
+#ifndef _SPI_ENGINE_
```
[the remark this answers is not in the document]
scuciurean: This was copied by mistake among with the copyrights

drivers/axi_core/spi_engine/spi_engine.c
```diff
@@ -0,0 +1,958 @@
...
+#include "error.h"
+#include "spi_engine.h"
+
+/******************************************************************************/
+/***************************** Static variables *******************************/
+/******************************************************************************/
+
+/**
+ * @brief Static variable used to sync transfers
+ *
```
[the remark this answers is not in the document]
scuciurean: This was generated by a doxygen plugin that i use. Maybe it's a different format.

drivers/axi_core/spi_engine/spi_engine.c
```diff
@@ -0,0 +1,958 @@
...
+	uint32_t			spi_engine_version;
+	struct spi_desc			*local_desc;
+	struct spi_engine_desc		*eng_desc;
+	struct spi_engine_init_param	*spi_engine_init;
+
+	local_desc = (struct spi_desc *)malloc(sizeof(*local_desc));
+	eng_desc = (struct spi_engine_desc*)malloc(sizeof(*eng_desc));
+
+	if (!desc || !eng_desc)
+		return FAILURE;
```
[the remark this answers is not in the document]
scuciurean: In my opinion, the check after each allocation seems to be redundant since in the application you free (remove) the descriptor in case of init failure.

drivers/axi_core/spi_engine/spi_engine.c
```diff
@@ -0,0 +1,958 @@
...
+	uint32_t			data_width;
+	uint32_t			spi_engine_version;
+	struct spi_desc			*local_desc;
+	struct spi_engine_desc		*eng_desc;
+	struct spi_engine_init_param	*spi_engine_init;
+
+	local_desc = (struct spi_desc *)malloc(sizeof(*local_desc));
+	eng_desc = (struct spi_engine_desc*)malloc(sizeof(*eng_desc));
+
+	if (!desc || !eng_desc)
```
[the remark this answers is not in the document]
scuciurean: Right

drivers/axi_core/spi_engine/spi_engine.c
```diff
@@ -0,0 +1,958 @@
...
+
+	/* Perform a reset */
+	spi_engine_write(eng_desc, SPI_ENGINE_REG_RESET, 0x01);
+	usleep(1000);
+	spi_engine_write(eng_desc, SPI_ENGINE_REG_RESET, 0x00);
+
+	/* Get current data width */
+	spi_engine_read(eng_desc, SPI_ENGINE_REG_DATA_WIDTH, &data_width);
+	eng_desc->max_data_width = data_width;
+	dev_dbg(desc, "Data with is 8 bits (max %d)", eng_desc->max_data_width);
```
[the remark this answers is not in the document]
scuciurean: Yes indeed, the 8 was left behind from a previous implementation.

drivers/axi_core/spi_engine/spi_engine.c
```diff
@@ -0,0 +1,958 @@
...
+ *
+ * @param desc Decriptor containing SPI interface parameters
+ * @param param Structure containing the offload init parameters
+ * @return int32_t This function allways returns SUCCESS
+ */
+int32_t spi_engine_offload_init(struct spi_desc *desc,
+				const struct spi_engine_offload_init_param *param)
+{
+	struct spi_engine_desc	*eng_desc;
+	struct axi_dmac_init	*dmac_init;
```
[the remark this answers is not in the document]
scuciurean: I would rather pass them as pointers and allocate the memory inside the spi engine driver instead of constraining the user to allocate memory in his application.

drivers/axi_core/spi_engine/spi_engine.h
scuciurean: There are two structures (spi_engine and spi_engine_offload) because there are two independent modes in which the engine can work.

drivers/axi_core/spi_engine/spi_engine.c
scuciurean: I would rather pass them as pointers and allocate the memory inside the spi engine driver instead of constraining the user to allocate memory in his application.

drivers/axi_core/spi_engine/spi_engine.c
scuciurean: This was discussed in the previous comment.

drivers/axi_core/spi_engine/spi_engine.h
scuciurean: In this function, I need to access members from the spi_init_param so, I need a spi_desc to be passed.

drivers/platform/xilinx/spi.c
scuciurean: I believe you are suggesting that the user should check the init and call the remove function himself in case of error. In this way, again, we are forcing the user to call additional functions.

drivers/axi_core/spi_engine/spi_engine.c
scuciurean: If I would use spi_engine_queue_append_cmd, the new element would be added before the first one in the queue (fifo) so the address of the resulting queue would change to the newly added element.

drivers/axi_core/spi_engine/spi_engine.c
scuciurean: This queue has the possibility to add and remove items in both fifo and lifo modes. The name was chosen because in most cases is used as a fifo.

drivers/axi_core/spi_engine/spi_engine.c
scuciurean: After reading your comment I realized that I'm using a weird last->next state instead of just using the current address.

scuciurean:
Changelog -> V3:

1. Added function comments in spi.c
2. Removed the delay between the command writes
3. Now buffer zeroing is done with calloc.
4. Enclosed debug messages inside a while loop
5. Check pointers before dereferencing
6. Removed unnecessary memory allocations.

drivers/axi_core/spi_engine/spi_engine.c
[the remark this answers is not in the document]
scuciurean: Sorry, I've missed that init

scuciurean:
Changelog -> V4:

1. Applied astyle

drivers/platform/xilinx/spi.c
```diff
@@ -54,11 +54,112 @@
...
+}
+
+/**
+ * @brief Initialize the hardware SPI peripherial
+ *
+ * @param xdesc Platform specific SPI descriptor
+ * @param xinit Platform specific SPI init param
+ * @return int32_t FAILURE if the device couldn't be found or configured
+ */
+int32_t spi_init_ps(struct xil_spi_desc *xdesc,
```
cristiantpop: add static

scuciurean:
Changelog -> V5:

1. Add more function comments.
2. Add static specifier to internal functions.

scuciurean:
Changelog -> V6:

1. Removed mask macro from the read/write instructions in order to make possible the initialization of the user commands inside an array (arrays can't be initialized with values containing arithmetic operators)

scuciurean:
Changelog -> V6:

1. Remove debug macros
2. Assign dev extra to the descriptor in xil_spi.c

scuciurean:
Changelog -> V6:

scuciurean:
Changelog -> V8:

1. Add pointer differencing in spi_remove.

scuciurean:
Changelog -> V9:

1. Cleanup: removed some intermediary pointers.

scuciurean:
Changelog -> V10:

1. Add include check in spi.c in case spi_engine is not used.

scuciurean:
Changelog -> V11:

1. Removed extra paranthesis in cloc_div computing.

drivers/platform/xilinx/spi.c
```diff
@@ -54,166 +54,235 @@
...
 		break;
-ps_error:
-		free(xdesc->instance);
-#endif
-		goto error;
 	case SPI_ENGINE:
 #ifdef SPI_ENGINE_H
-		//TODO: Implement SPI engine feature
-
+		ret = spi_engine_init(desc, param);
```
mchindri: makes more sense if the header will be 
`int32_t spi_engine_init(struct spi_desc *desc, const struct spi_init_param *param)`
because the general descriptor is not allocated inside of these inits.
And the call to this functions will be:
```
ret = spi_init_pl(*desc, param);
ret = spi_init_ps(*desc, param);
ret = spi_engine_init(*desc, param);
```

drivers/axi_core/spi_engine/spi_engine.h
```diff
@@ -0,0 +1,192 @@
...
+struct spi_engine_init_param {
+	/** Type of implementation */
+	enum xil_spi_type	type;
+	/** Base address where the HDL core is situated */
+	uint32_t 		spi_engine_baseaddr;
+	/** Delay between the CS toggle and the start of SCLK */
+	uint32_t		cs_delay;
+	/** Data with of one SPI transfer ( in bits ) */
+	uint8_t			data_width;
+} spi_engine_init_param;
```
mchindri: Just as reminder: to remove `spi_engine_init_param` before } . Same for next

drivers/axi_core/spi_engine/spi_engine.h
```diff
@@ -0,0 +1,192 @@
...
+int32_t spi_engine_init(struct spi_desc **desc,
+			const struct spi_init_param *param);
+
+/* Write and read data over SPI using the SPI engine */
+int32_t spi_engine_write_and_read(spi_desc *desc,
+				  uint8_t *data,
+				  uint8_t bytes_number);
+
+/* Free the resources used by the SPI engine device */
+int32_t spi_engine_remove(spi_desc *desc);
```
mchindri: Just as reminder: use `struct spi_desc`

drivers/axi_core/spi_engine/spi_engine.c
```diff
@@ -0,0 +1,838 @@
...
+	struct spi_engine_init_param	*spi_engine_init;
+
+	eng_desc = (struct spi_engine_desc*)malloc(sizeof(*eng_desc));
+
+	if (!eng_desc)
+		return FAILURE;
+
+	spi_engine_init = param->extra;
+
+	(*desc)->max_speed_hz = param->max_speed_hz;
```
adrimbarean: I think *desc needs to be checked here. Usually memory allocation is done in the init and I assume that spi_engine_init should be called within or after the spi_ini method. But since it's an API, I think we should guard against it.

scuciurean:
Changelog -> V12:

1. Remove struct initializers.
2. Check descriptor pointer inside spi_engine_init

drivers/platform/xilinx/spi.c
```diff
@@ -54,166 +54,235 @@
...
+		goto init_error;
 		break;
 	}
 
-	*desc = sdesc;
-
-	return SUCCESS;
+	if (ret != SUCCESS) {
+init_error:
+		free(desc);
```
mchindri: desc is of type struct spi_desc **
you must free(*desc)

scuciurean: I missed that two times somehow.

scuciurean:
Changelog -> V13:

1. Correct memory freeing.

drivers/axi_core/spi_engine/spi_engine.h
mchindri: you forgot this.

mchindri:
https://github.com/analogdevicesinc/no-OS/pull/437#discussion_r378774660
You forgot this

drivers/axi_core/spi_engine/spi_engine.c
buha: Well, in this case the function should be documeted accordingly, something like "this function assumes non-NULL input msg parameter". At some point during development, it is possible that an internal function becomes external, and guess what, the developer won't check the contents and how the pointers are not dereferenced, it will simply remove 'static' qualifier, and you end up with a bug.

PR #439, opened 2020-01-19
drivers: platform: linux: Add axi_io implementation

dbogdan:
Implementation of AXI IO through UIO.

### DIFF
--- a/drivers/platform/linux/axi_io.c
+++ b/drivers/platform/linux/axi_io.c
@@ -1,0 +1,135 @@
+/***************************************************************************//**
+ *   @file   linux/axi_io.c
+ *   @brief  Implementation of AXI IO through UIO.
+ *   @author Dragos Bogdan (dragos.bogdan@analog.com)
+********************************************************************************
+ * Copyright 2020(c) Analog Devices, Inc.
+ *
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *  - Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ *  - Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in
+ *    the documentation and/or other materials provided with the
+ *    distribution.
+ *  - Neither the name of Analog Devices, Inc. nor the names of its
+ *    contributors may be used to endorse or promote products derived
+ *    from this software without specific prior written permission.
+ *  - The use of this software may or may not infringe the patent rights
+ *    of one or more patent holders.  This license does not release you
+ *    from the requirement that you obtain separate licenses from these
+ *    patent holders to use this software.
+ *  - Use of the software either in source or binary form, must be run
+ *    on or directly connected to an Analog Devices Inc. component.
+ *
+ * THIS SOFTWARE IS PROVIDED BY ANALOG DEVICES "AS IS" AND ANY EXPRESS OR
+ * IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, NON-INFRINGEMENT,
+ * MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED.
+ * IN NO EVENT SHALL ANALOG DEVICES BE LIABLE FOR ANY DIRECT, INDIRECT,
+ * INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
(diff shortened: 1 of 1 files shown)


dbogdan:
v2:
   - if (uio_fd < 1) -> if (uio_fd < 0)

drivers/platform/linux/axi_io.c
```diff
@@ -0,0 +1,128 @@
...
+ * @return SUCCESS in case of success, FAILURE otherwise.
+ */
+int32_t axi_io_read(uint32_t base, uint32_t offset, uint32_t *data)
+{
+	char buf[16];
+	int ret;
+	int uio_fd;
+	void *uio_addr;
+
+	sprintf(buf, "/dev/uio%d", base);
```
mchindri: the max uint32_t is 10 character, if this number is given will be a buffer overflow. Is the input is from some of our code then maybe it is ok if it is not given a big number. Same for write function.
Also, I think is good practice to use %u for unsigned. On small number is ok, but if the most significant bit is set %d will give a negative number.

drivers/platform/linux/axi_io.c
```diff
@@ -0,0 +1,128 @@
...
+
+	uio_fd = open(buf, O_RDWR);
+	if (uio_fd < 0) {
+		printf("%s: Can't open %s\n\r", __func__, buf);
+		return FAILURE;
+	}
+
+	uio_addr = mmap(NULL,
```
mchindri: shouldn't the returned value be checked?

drivers/platform/linux/axi_io.c
```diff
@@ -0,0 +1,128 @@
...
+			0x1000,
```
mchindri: I think if you map from the offset address you can map only the needed bytes

drivers/platform/linux/axi_io.c
```diff
@@ -0,0 +1,128 @@
...
+			PROT_READ|PROT_WRITE,
+			MAP_SHARED,
+			uio_fd,
+			0);
+
+	*data = (*((unsigned *) (uio_addr + offset)));
```
adrimbarean: Changing this to
`*data = (*((uint32_t *)uio_addr + offset));`
should solve the Codancy issue.

drivers/platform/linux/axi_io.c
```diff
@@ -0,0 +1,128 @@
...
+	}
+
+	uio_addr = mmap(NULL,
+			0x1000,
+			PROT_READ|PROT_WRITE,
+			MAP_SHARED,
+			uio_fd,
+			0);
+
+	*((unsigned *) (uio_addr + offset)) = data;
```
adrimbarean: Changing this to
`*((uint32_t *)uio_addr + offset) = data;`
should solve the Codancy issue.

drivers/platform/linux/axi_io.c
```diff
@@ -0,0 +1,128 @@
...
+ * @return SUCCESS in case of success, FAILURE otherwise.
+ */
+int32_t axi_io_write(uint32_t base, uint32_t offset, uint32_t data)
+{
+	char buf[16];
+	int ret;
+	int uio_fd;
+	void *uio_addr;
+
+	sprintf(buf, "/dev/uio%d", base);
```
adrimbarean: Changing this to
`sprintf(buf, "/dev/uio%u", base);`
should solve the Codancy issue.

drivers/platform/linux/axi_io.c
dbogdan: Fixed that.

dbogdan:
v2:
   - use %"PRIu32" instead of %d for printing uint32_t base;
   - check mmap() return value;
   - cast data to uint32_t instead of unsigned.

dbogdan:
v4:
   - create axi_io_read_write() in order to remove the code duplication;
   - use munmap() for removing the mappings created by mmap();
   - increase buf[] length to 32.

dbogdan:
v5:
   - Fixed Codacy complaints.

dbogdan:
v6:
   - Fix mmap length value (it should take in consideration the offset too).

dbogdan:
v7:
   - fix Codacy complaints.

drivers/platform/linux/axi_io.c
```diff
@@ -0,0 +1,131 @@
...
+/***************************** Include Files **********************************/
+/******************************************************************************/
+#include <fcntl.h>
+#include <inttypes.h>
+#include <stdio.h>
+#include <unistd.h>
+#include <sys/mman.h>
+#include "error.h"
+#include "axi_io.h"
+
```
mchindri: You forgot the function definition header

dbogdan: Done.

PR #443, opened 2020-01-21
Adrv9009 iio make

cristiantpop:
adrv9009 src.mk: Fix TINYIIOD build option.

Build was failing with: "TINYIIOD = n".


cristiantpop:
V2:

- Fix indentations.

projects/adrv9009/src.mk
```diff
@@ -119,20 +119,20 @@ INCS += $(DRIVERS)/axi_core/clk_altera_a10_fpll/clk_altera_a10_fpll.h	\
 	$(DRIVERS)/axi_core/jesd204/altera_adxcvr.h
 endif
 INCS +=	$(PLATFORM_DRIVERS)/spi_extra.h					\
-	$(PLATFORM_DRIVERS)/gpio_extra.h				\
-	$(PLATFORM_DRIVERS)/irq_extra.h					\
-	$(PLATFORM_DRIVERS)/uart_extra.h
```
mirceacaprioru: could you give more details why you are moving these ?

cristiantpop: This are files used only for IIO_EXAMPE, at least for the moment, so it is not necessary to include them when IIO is not used.

PR #444, opened 2020-01-21
uart: Fix uart_write function.

cristiantpop:
It was working fine for platforms: ZC702, ZCU102, ZC706, but not working
on ZedBoard.
Send buffer in multiple "XUartPs_Send" calls.

### DIFF
--- a/drivers/platform/xilinx/uart.c
+++ b/drivers/platform/xilinx/uart.c
@@ -46,6 +46,12 @@
 #endif
 
 /******************************************************************************/
+/********************** Macros and Constants Definitions **********************/
+/******************************************************************************/
+
+#define WRITE_SIZE 255
+
+/******************************************************************************/
 /************************ Functions Definitions *******************************/
 /******************************************************************************/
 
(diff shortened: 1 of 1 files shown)


drivers/platform/xilinx/uart.c
```diff
@@ -149,17 +149,19 @@ int32_t uart_read(struct uart_desc *desc, uint8_t *data, uint32_t bytes_number)
 int32_t uart_write(struct uart_desc *desc, const uint8_t *data,
 		   uint32_t bytes_number)
 {
+#define WRITE_SIZE 255
```
mchindri: Shouldn't this be at the beginning of the file?

drivers/platform/xilinx/uart.c
```diff
@@ -149,17 +149,19 @@ int32_t uart_read(struct uart_desc *desc, uint8_t *data, uint32_t bytes_number)
...
 	switch(xil_uart_desc->type) {
 	case UART_PS:
 #ifdef XUARTPS_H
-		total_sent = XUartPs_Send(xil_uart_desc->instance, (u8*)data,
-					  bytes_number);
-		while (XUartPs_IsSending(xil_uart_desc->instance));
-		if (total_sent < bytes_number)
-			return FAILURE;
+		while (offset < bytes_number) {
+			len = offset + WRITE_SIZE > bytes_number ? (bytes_number - offset) : WRITE_SIZE;
```
mchindri: I think is more understandable if you write it this way
`len =  ((bytes_number - offset) < WRITE_SIZE) ? (bytes_number - offset) : WRITE_SIZE;`

drivers/platform/xilinx/uart.c
cristiantpop: Moved.

drivers/platform/xilinx/uart.c
cristiantpop: Good suggestion! Done.

cristiantpop:
V2:

- Rewrite expression to be more understandable: 
len = ((bytes_number - offset) < WRITE_SIZE) ? (bytes_number - offset) : WRITE_SIZE;

- Move define at the beginning of file: #define WRITE_SIZE 255.

PR #447, opened 2020-01-22
Configure hmc7044 devices on ZU11EG + ADRV2CRR-FMC

buha:
An HMC7044 device resides on the ZU11EG SOM and is the one that feeds the FPGA and TALISE devices with required clocks.

This HMC7044 has it's inputs fed by another HMC7044 residing on the carrier board.

This pull request configures both to output the desired clocks.

The XCVR in the FPGA seems to initialize well hence this pull request.


projects/adrv9009/src/devices/adi_hal/parameters.h
```diff
@@ -102,41 +102,102 @@
 #else
 #define GPIO_DEVICE_ID			XPAR_PS7_GPIO_0_DEVICE_ID
 #endif
-#ifdef XPS_BOARD_ZCU102
```
mirceacaprioru: is the GPIO_OFFSET no longer used ?

mirceacaprioru: it seems unrelated with this commit

projects/adrv9009/src/app/app_config.h
```diff
@@ -42,6 +42,7 @@
 
 //#define XILINX_PLATFORM
 //#define ALTERA_PLATFORM
+//#define ZU11EG
```
mirceacaprioru: i don't know if adding another define is the best approach. We are trying to limit this to a minimum, if we keep adding them we get back to where we started in the first place.

buha: I agree too many of these defines make the code hard to read and thought of what the alternative would be. One I can think of would be a separate project but this would lead to too much code duplication.

projects/adrv9009/src/devices/adi_hal/parameters.h
```diff
@@ -151,7 +213,4 @@
 #define INTC_DEVICE_ID			XPAR_SCUGIC_SINGLE_DEVICE_ID
 #endif
 
-#define CLK_CS			0
-#define ADRV_CS			1
-
-#endif
+#endif
```
amiclaus: nitpick: add new line at end of file

buha:
Previous force push adds a newline at end of parameters.h

projects/adrv9009/src/devices/adi_hal/parameters.h
buha: GPIO_OFFSET will likely disappear from all projects when it will be implemented in gpio driver.

PR #451, opened 2020-01-27
drivers: platform:timer and delay

mchindri:
Add both timer and delay support for aducm 3029.
I put both in the same pull request because the delay uses the timer

### DIFF
--- a/drivers/platform/aducm3029/timer.c
+++ b/drivers/platform/aducm3029/timer.c
@@ -1,0 +1,339 @@
+/***************************************************************************//**
+ *   @file   aducm3029/timer.c
+ *   @brief  Implementation of TIMER driver for ADuCM302x.
+ *
+ *   This driver enables the user to create multiple instance of a
+ *   timer using only one hardware timer of the ADuCM3029.
+ *
+ *   @author Mihail Chindris (mihail.chindris@analog.com)
+********************************************************************************
+ * Copyright 2019(c) Analog Devices, Inc.
+ *
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *  - Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ *  - Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in
+ *    the documentation and/or other materials provided with the
+ *    distribution.
+ *  - Neither the name of Analog Devices, Inc. nor the names of its
+ *    contributors may be used to endorse or promote products derived
+ *    from this software without specific prior written permission.
+ *  - The use of this software may or may not infringe the patent rights
+ *    of one or more patent holders.  This license does not release you
+ *    from the requirement that you obtain separate licenses from these
+ *    patent holders to use this software.
+ *  - Use of the software either in source or binary form, must be run
+ *    on or directly connected to an Analog Devices Inc. component.
+ *
+ * THIS SOFTWARE IS PROVIDED BY ANALOG DEVICES "AS IS" AND ANY EXPRESS OR
+ * IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, NON-INFRINGEMENT,
+ * MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED.
+ * IN NO EVENT SHALL ANALOG DEVICES BE LIABLE FOR ANY DIRECT, INDIRECT,
+ * INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, INTELLECTUAL PROPERTY RIGHTS, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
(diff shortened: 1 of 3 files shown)


drivers/platform/aducm3029/timer_extra.h
```diff
@@ -0,0 +1,68 @@
...
+
+/******************************************************************************/
+/********************** Macros and Constants Definitions **********************/
+/******************************************************************************/
+
+/**
+ * This macro is used to select the desired hardware timer to use by the timer
+ * driver. For the ADuCM302x it can be 0, 1 or 2
+ */
+#define HW_TIMER_ID	0u
```
dbogdan: Can only one timer be enabled at a time?

mchindri: The driver use only hardware timer and the timer instances are based on it. There can be created any number of instances.
This define is to select which hardware timer to be used by the driver. For example if another driver use the timer 0, this can be set to use timer 1 or 2

drivers/platform/aducm3029/timer_extra.h
```diff
@@ -0,0 +1,68 @@
...
+ * INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, INTELLECTUAL PROPERTY RIGHTS, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+*******************************************************************************/
+
+#ifndef TIMER_H
+# define TIMER_H
```
amiclaus: nitpick: extra space between `#` and `define`

drivers/platform/aducm3029/timer.c
```diff
@@ -0,0 +1,339 @@
...
+#include "timer_extra.h"
+#include "timer.h"
+#include "error.h"
+
+/******************************************************************************/
+/********************** Macros and Constants Definitions **********************/
+/******************************************************************************/
+
+#if !defined(HW_TIMER_ID) || !(HW_TIMER_ID < 3 && HW_TIMER_ID >= 0)
+# error HW_TIMER_ID must exist and be 0, 1 or 2
```
cristiantpop: Remove extra space, here.

drivers/platform/aducm3029/timer.c
```diff
@@ -0,0 +1,339 @@
...
+	struct aducm_timer_desc *tmr_desc;
+
+	if (!desc)
+		return FAILURE;
+
+	tmr_desc = desc->extra;
+	if (!tmr_desc->started)
+		return FAILURE;
+
+	timer_counter_get(desc,  &counter);
```
adrimbarean: nitpick: extra space on second parameter.

drivers/platform/aducm3029/timer_extra.h
mchindri: The space is used to indent the defines

drivers/platform/aducm3029/timer.c
mchindri: removed all

drivers/platform/aducm3029/timer.c
mchindri: done

drivers/platform/aducm3029/timer.c
mchindri: removed all because HW_TIMER_ID not needed

mchindri:
v2
 - refactor for the timer.h modifications (remove  #define HW_TIMER_ID 0u and use timer_desc.id)
 - Remove extra space

mchindri:
rebase to fix doxygen

drivers/platform/aducm3029/timer_extra.h
amiclaus: Your point is valid. But according to all other header files in this repository, the indentation for defines is not applied. This is how some IDEs generate the defines when creating new header file. We should keep the consistency.
https://github.com/analogdevicesinc/no-OS/blob/7c2fb4c532324da088dc6a38cd23d7253f7db645/include/delay.h#L40-L41
https://github.com/analogdevicesinc/no-OS/blob/7c2fb4c532324da088dc6a38cd23d7253f7db645/include/error.h#L40-L41
https://github.com/analogdevicesinc/no-OS/blob/7c2fb4c532324da088dc6a38cd23d7253f7db645/include/fifo.h#L39-L40
https://github.com/analogdevicesinc/no-OS/blob/7c2fb4c532324da088dc6a38cd23d7253f7db645/include/gpio.h#L40-L41
https://github.com/analogdevicesinc/no-OS/blob/7c2fb4c532324da088dc6a38cd23d7253f7db645/include/i2c.h#L40-L41
https://github.com/analogdevicesinc/no-OS/blob/7c2fb4c532324da088dc6a38cd23d7253f7db645/include/irq.h#L40-L41
https://github.com/analogdevicesinc/no-OS/blob/7c2fb4c532324da088dc6a38cd23d7253f7db645/include/spi.h#L40-L41

mchindri: done

mchindri:
v4
 - Remove space befor define

mchindri:
Will merge today if no more comments.

PR #453, opened 2020-01-28
drivers: platform: Add IRQ support for ADuCM3029

mchindri:
This driver enables the user to create instances for the hardware
interrupts in ADuCM3029 and associate a callback to them.

### DIFF
--- a/drivers/platform/aducm3029/irq.c
+++ b/drivers/platform/aducm3029/irq.c
@@ -1,0 +1,286 @@
+/***************************************************************************//**
+ *   @file   aducm3029/irq.c
+ *   @brief  Implementation of External IRQ driver for ADuCM302x.
+ *   @author Mihail Chindris (mihail.chindris@analog.com)
+********************************************************************************
+ * Copyright 2019(c) Analog Devices, Inc.
+ *
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *  - Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ *  - Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in
+ *    the documentation and/or other materials provided with the
+ *    distribution.
+ *  - Neither the name of Analog Devices, Inc. nor the names of its
+ *    contributors may be used to endorse or promote products derived
+ *    from this software without specific prior written permission.
+ *  - The use of this software may or may not infringe the patent rights
+ *    of one or more patent holders.  This license does not release you
+ *    from the requirement that you obtain separate licenses from these
+ *    patent holders to use this software.
+ *  - Use of the software either in source or binary form, must be run
+ *    on or directly connected to an Analog Devices Inc. component.
+ *
+ * THIS SOFTWARE IS PROVIDED BY ANALOG DEVICES "AS IS" AND ANY EXPRESS OR
+ * IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, NON-INFRINGEMENT,
+ * MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED.
+ * IN NO EVENT SHALL ANALOG DEVICES BE LIABLE FOR ANY DIRECT, INDIRECT,
+ * INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, INTELLECTUAL PROPERTY RIGHTS, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+*******************************************************************************/
+
+/******************************************************************************/
+/************************* Include Files **************************************/
+/******************************************************************************/
+
+#include "irq.h"
+#include "irq_extra.h"
+#include "error.h"
+#include <stdlib.h>
+
+/******************************************************************************/
+/********************** Macros and Constants Definitions **********************/
+/******************************************************************************/
+
+/** The number of the first external interrupt, used by NVIC */
+#define BASE_XINT_NB			(XINT_EVT0_IRQn)
+
+/** Number of interrupts controllers available */
+#define NB_INTERRUPT_CONTROLLERS	1u
+
+/** Map the interrupt ID to the ADI_XINT_EVENT associated event */
+static const uint32_t id_map_event[NB_EXT_INTERRUPTS] = {
+	ADI_XINT_EVENT_INT0, // ID 0
+	ADI_XINT_EVENT_INT1, // ID 1
+	ADI_XINT_EVENT_INT2, // ID 2
+	ADI_XINT_EVENT_INT3  // ID 3
+};
+
+/******************************************************************************/
+/***************************** Global Variables *******************************/
+/******************************************************************************/
+
+/**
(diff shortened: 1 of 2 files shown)


mchindri:
v2
 - fix codacy

drivers/platform/aducm3029/irq.c
```diff
@@ -0,0 +1,298 @@
...
+#include "irq_extra.h"
+#include "gpio.h"
+#include <stdlib.h>
+
+/******************************************************************************/
+/********************** Macros and Constants Definitions **********************/
+/******************************************************************************/
+
+/** Get the number resulting by shifting 1 with n positions */
+#define BIT(n)		(1 << (n))
```
cristiantpop: already defined in util.h

mchindri:
TODO: Change commit description

mchindri:
v3
 - Change commit description
 - Remove gpio dependency

dbogdan:
It looks good.

mchindri:
v4
- Add error.h

mchindri:
Will merge today if no more comments.

drivers/platform/aducm3029/irq.c
```diff
@@ -0,0 +1,302 @@
...
+	struct aducm_irq_desc *aducm_desc;
+	if (!desc || !desc->extra || !initialized || !irq_handler ||
+	    irq_id >= NB_EXT_INTERRUPTS)
+		return FAILURE;
+
+	aducm_desc = desc->extra;
+	aducm_desc->irq_handler[irq_id] = irq_handler;
+	aducm_desc->mode[irq_id] = (enum irq_mode)dev_instance;
+
+	adi_xint_RegisterCallback(id_map_event[irq_id], internal_callback,
```
adrimbarean: Should this be done in init only once rather than every time when registering a new callback?

drivers/platform/aducm3029/irq.c
```diff
@@ -0,0 +1,302 @@
...
+ * @return \ref SUCCESS in case of success, \ref FAILURE otherwise.
+ */
+int32_t irq_unregister(struct irq_desc *desc, uint32_t irq_id)
+{
+	if (!desc || !desc->extra || !initialized ||
+	    irq_id >= NB_EXT_INTERRUPTS)
+		return FAILURE;
+	(void)irq_id;
+
+	irq_source_disable(desc, irq_id);
```
adrimbarean: I don't think this should also disable the interrupt. Maybe the user just wants to change the callback without disabling.

drivers/platform/aducm3029/irq.c
```diff
@@ -0,0 +1,302 @@
...
+ * @return \ref SUCCESS
+ */
+int32_t irq_global_enable(struct irq_desc *desc)
+{
+	struct aducm_irq_desc *aducm_desc;
+	if (!desc || !desc->extra || !initialized)
+		return FAILURE;
+
+	aducm_desc = desc->extra;
+	for (uint32_t i = 0; i < NB_EXT_INTERRUPTS; i++) {
```
adrimbarean: nitpick: I don't think this needs brackets.

drivers/platform/aducm3029/irq.c
```diff
@@ -0,0 +1,302 @@
...
+ * @param desc - Interrupt controller descriptor.
+ * @param irq_id - Id of the interrupt
+ * @return \ref SUCCESS in case of success, \ref FAILURE otherwise.
+ */
+int32_t irq_unregister(struct irq_desc *desc, uint32_t irq_id)
+{
+	if (!desc || !desc->extra || !initialized ||
+	    irq_id >= NB_EXT_INTERRUPTS)
+		return FAILURE;
+	(void)irq_id;
```
adrimbarean: Do we need this?

drivers/platform/aducm3029/irq.c
mchindri: https://github.com/analogdevicesinc/no-OS/blob/201833ce661b164904d7a2de9b39dbe64946566f/include/irq.h#L94
It the implementation for the header function.

drivers/platform/aducm3029/irq.c
mchindri: It for unused parameter warning. Even if I don't get one on my IDE :)

drivers/platform/aducm3029/irq.c
mchindri: true

drivers/platform/aducm3029/irq.c
mchindri: solved

drivers/platform/aducm3029/irq.c
mchindri: true

drivers/platform/aducm3029/irq.c
mchindri: I modify the comment of the function, maybe that was inaccurate.

mchindri:
v5:
- delet uneede irq_source_disable
- remove brackets
- adapt to with new modifications to irq.h

drivers/platform/aducm3029/irq.c
```diff
@@ -0,0 +1,298 @@
...
+ * @param event - Event that generated the callback
+ * @param arg - Unused
+ */
+static void internal_callback(void *aducm_desc, uint32_t event, void *arg)
+{
+	struct aducm_irq_desc *desc = aducm_desc;
+
+	(void)arg;
+
+	switch (event) {
```
scuciurean: Can't this switch be replaced by an if statement?
For example:
```
if (event < ADI_XINT_MAX_INT_NUM && desc->irq_handler[event])
    desc->irq_handler[event]((void *)event);
```

drivers/platform/aducm3029/irq.c
```diff
@@ -0,0 +1,298 @@
...
+ */
+int32_t irq_ctrl_init(struct irq_ctrl_desc **desc,
+		      const struct irq_init_param *param)
+{
+	struct aducm_irq_desc		*aducm_desc;
+
+	if (!desc || !param || initialized)
+		return FAILURE;
+
+	if (!(*desc = calloc(1, sizeof(**desc))))
```
scuciurean: It may be possible for the compiler to interpret this statement differently, depending on the used c language standard.

drivers/platform/aducm3029/irq.c
```diff
@@ -0,0 +1,298 @@
...
+	ADI_XINT_EVENT_INT2, // ID 2
+	ADI_XINT_EVENT_INT3  // ID 3
+};
+
+/******************************************************************************/
+/***************************** Global Variables *******************************/
+/******************************************************************************/
+
+/** Bitmap with initialized interrupt */
+static uint32_t		initialized;
```
scuciurean: Maybe move this inside the descriptor?

mchindri: It is used to store if the irq controller was already initialized and to not let the user create multiple instances of the controller.

drivers/platform/aducm3029/irq.c
mchindri: solved

drivers/platform/aducm3029/irq.c
mchindri: done

mchindri:
v6
 - replace switch with if
 - change the check for *desc
 - move irq_register to init

mchindri:
TODO: Add ADI_XINT_EVENT_UART_RX to the available events

drivers/platform/aducm3029/irq.c
```diff
@@ -0,0 +1,287 @@
...
+ */
+int32_t irq_ctrl_init(struct irq_ctrl_desc **desc,
+		      const struct irq_init_param *param)
+{
+	struct aducm_irq_desc		*aducm_desc;
+
+	if (!desc || !param || initialized)
+		return FAILURE;
+
+	*desc = calloc(1, sizeof(**desc));
```
adrimbarean: nitpick: we should be doing casts for calloc pointers from now on.

mchindri: image

mchindri: Even considering this should we use the cast?

mchindri:
v7
 - Remove unnecessary {}

mchindri:

drivers/platform/aducm3029/irq.c
adrimbarean: Yes, because we are also trying to make the code compatible in a C++ environment too, and C++ compilers require explicit cast for void pointers to another type of pointer. Bumped into this one too.

mchindri:
v8
 - Add cast for calloc

drivers/platform/aducm3029/irq.c
mchindri: solve

PR #457, opened 2020-01-29
xilinx spi prescaler based on max_speed_hz

buha:
max_speed_hz field of spi_desc was not taken into account and the SPI prescaler value was hardcoded to 64.
This pull request handles the max_speed_hz field and computes the closest possible prescaler that yields a clock speed as close as posible to max_speed_hz.


buha:
Previous force push added some code comments for the somewhat cryptic calculation of prescaler register values.

PR #458, opened 2020-01-30
includes:error modify SUCCESS -> NO_OS_SUCCESS

mchindri:
SUCCESS -> NO_OS_SUCCESS
FAILURE -> NO_OS_FAILURE

This change is needed to remove multiple definition in projects
that will use aducm3029 drivers

Makes the following replaces (Only if space or ( befor SUCCESS or FAILURE): 
 - " SUCCESS" -> " NO_OS_SUCCESS"
 - "SUCCESS" -> "(NO_OS_SUCCESS" 
 - "-SUCCESS" -> "-NO_OS_SUCCESS" Only used this way here https://github.com/analogdevicesinc/no-OS/blob/635c6b9eda56a9eff95e9c025757c8eb4fd28a96/drivers/frequency/hmc7044/hmc7044.c#L478
 - " FAILURE" -> " NO_OS_FAILURE"
 - "(FAILURE" -> "(NO_OS_FAILURE"


dbogdan:
I wouldn't do this... I would rather find a solution for the ADUCM3029.

PR #464, opened 2020-01-31
iio: Merge app with corresponding device.

cristiantpop:
This will result in a simpler system design, and better user experience.
Update projects accordingly.


cristiantpop:
Retrigger build with a force push

PR #469, opened 2020-02-03
axi_dac_core: use appropriate datasel function

amiclaus:
Use `axi_dac_sel_datasel` to select internal data sources for customized dac
channels.

### DIFF
--- a/drivers/axi_core/axi_dac_core/axi_dac_core.c
+++ b/drivers/axi_core/axi_dac_core/axi_dac_core.c
@@ -894,7 +894,7 @@ int32_t axi_dac_data_setup(struct axi_dac *dac)
 				}
 			}
 			axi_dac_write(dac, DAC_REG_DATA_PATTERN(i), chan->pat_data);
-			axi_dac_datasel(dac, i, chan->sel);
+			axi_dac_set_datasel(dac, i, chan->sel);
 		}
 	} else {
 		for (i = 0; i < dac->num_channels; i++) {


amiclaus:
> 
> 
> Can you give a bit more context on why do we need this change?

Sure. There are two similar functions implemented in the axi_dac_core:


https://github.com/analogdevicesinc/no-OS/blob/59e8bf67837104ecaa0b4fc1377d8dec568bee0f/drivers/axi_core/axi_dac_core/axi_dac_core.c#L364-L381

AND

https://github.com/analogdevicesinc/no-OS/blob/59e8bf67837104ecaa0b4fc1377d8dec568bee0f/drivers/axi_core/axi_dac_core/axi_dac_core.c#L746-L762

As you can see, the functions are pretty similar, only the register addresses are different.
AXI_DAC_REG_CHAN_CNTRL_7 
AXI_DAC_REG_CHAN_CNTRL_8
Following the guide here: https://wiki.analog.com/resources/fpga/docs/axi_dac_ip we may notice that the AXI_DAC_REG_CHAN_CNTRL_7 handles the internal data sources selection, while AXI_DAC_REG_CHAN_CNTRL_8 handles the IQ correction coefficient.

P.S. The `axi_dac_datasel` function should be renamed, since it is creating confusion.

buha:
I suggest 1 single function with proper naming that takes an extra parameter to specify which register to change:
AXI_DAC_REG_CHAN_CNTRL_7
AXI_DAC_REG_CHAN_CNTRL_8

dbogdan:
Sorry, I've missed the comments. @amiclaus , can you create a new commit that takes care of all the remarks? Thanks.

PR #478, opened 2020-02-07
Includes:irq.h Proposal for irq_register declaration

mchindri:
This is how I think it should look irq.h
I write a draft of how register will look in every platform to be better understood the approach.
In the function_header.c file are the function declaration for each platform to be easy to look at.
PS: Only look at the propose irq.h commit.

### DIFF
--- a/drivers/platform/xilinx/irq.c
+++ b/drivers/platform/xilinx/irq.c
@@ -229,6 +229,24 @@ int32_t irq_source_disable(struct irq_ctrl_desc *desc, uint32_t irq_id)
 	return SUCCESS;
 }
 
+status = irq_register(xil_uart_desc->irq_desc, xil_uart_desc->irq_id,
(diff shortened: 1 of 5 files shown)


drivers/platform/xilinx/irq.c
```diff
@@ -238,15 +256,27 @@ int32_t irq_source_disable(struct irq_ctrl_desc *desc, uint32_t irq_id)
  * @return SUCCESS in case of success, FAILURE otherwise.
  */
 int32_t irq_register(struct irq_ctrl_desc *desc, uint32_t irq_id,
-		     void (*irq_handler)(void *data), void *dev_instance)
+		     void (*irq_handler)(void *context, void *extra),
```
scuciurean: If you need multiple parameters (eg context and extra for the irq handler) you can pack them in a structure and pass it as reference (*data).

PR #492, opened 2020-02-21
lookup table based crc-8 implementation

buha:
This is code compatible with linux crc8.h although the implementation is
somewhat different.

This implementation is supposed to replace all crc-8 implementations in no-OS
drivers which are all duplicates of each other.

By comparing the performance of this new implementation with the old one on
an x86 machine and on large sample sizes, it seems the new implementation is
faster by a factor of 13.

I suggest to review but not merge this until we have a working projects/ad7768-1 to test it on. I did test it on x86 and gives the same result as our existing algorithm.

### DIFF
--- a/util/crc8.c
+++ b/util/crc8.c
@@ -1,0 +1,100 @@
+/***************************************************************************//**
+ *   @file   crc8.c
+ *   @brief  Source file of CRC-8 computation.
+ *   @author Darius Berghe (darius.berghe@analog.com)
+********************************************************************************
+ * Copyright 2020(c) Analog Devices, Inc.
+ *
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *  - Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ *  - Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in
+ *    the documentation and/or other materials provided with the
+ *    distribution.
+ *  - Neither the name of Analog Devices, Inc. nor the names of its
+ *    contributors may be used to endorse or promote products derived
+ *    from this software without specific prior written permission.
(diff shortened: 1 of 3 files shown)


util/crc8.c
```diff
@@ -0,0 +1,101 @@
...
+ * @param table     - Pointer to a CRC-8 lookup table for the desired polynomial.
+ * @param pdata     - Pointer to 8-bit data buffer.
+ * @param nbytes    - Number of bytes to compute the CRC-8 over.
+ * @param crc       - Initial value for the CRC-8 computation. Can be used to
+ *                    cascade calls to this function.
+ *
+ * @return crc      - Computed CRC-8 value.
+*******************************************************************************/
+uint8_t crc8(const uint8_t * table, const uint8_t *pdata, size_t nbytes,
+	     uint8_t crc)
```
scuciurean: Since the description says the last parameter (crc) represents the initial value, wouldn't it be more clear to rename it this way?

util/crc8.c
```diff
@@ -0,0 +1,101 @@
+/***************************************************************************//**
+ *   @file   crc8.c
+ *   @brief  Source file of CRC-8 computation.
+ *   @author Darius Berghe (darius.berghe@analog.com)
+********************************************************************************
+ * Copyright 2012(c) Analog Devices, Inc.
```
dbogdan: 2012 -> 2020.

util/crc8.c
```diff
@@ -0,0 +1,101 @@
...
+
+	while (nbytes--) {
+		idx = (crc ^ *pdata);
+		crc = (table[idx]) & 0xff;
+		pdata++;
+	}
+
+	return crc;
+}
+
```
amiclaus: nitpick: extra line :D

buha:
Following previous comments I:
- fixed file header with year
- removed extra lines
- ellaborated a bit more the doxy description of crc8()

util/crc8.c
buha: i added more explanation to the function documentation. basically, the value you put there is typically a crc so i think it makes sense the stay that way

PR #493, opened 2020-02-24
 util: Add list library

mchindri:
This library offer support for store data. This will be useful
in the future to remove the complexity of drivers/applications
to handle data structures

You can also check the documentation at:
https://mchindri.github.io/no-OS/list_8h.html

### DIFF
--- a/include/list.h
+++ b/include/list.h
@@ -1,0 +1,317 @@
+/***************************************************************************//**
+ *   @file   list.h
+ *   @brief  List library header
+ *   @author Mihail Chindris (mihail.chindris@analog.com)
+********************************************************************************
+ *   @copyright
+ * Copyright 2020(c) Analog Devices, Inc.
+ *
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *  - Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ *  - Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in
+ *    the documentation and/or other materials provided with the
+ *    distribution.
+ *  - Neither the name of Analog Devices, Inc. nor the names of its
+ *    contributors may be used to endorse or promote products derived
+ *    from this software without specific prior written permission.
+ *  - The use of this software may or may not infringe the patent rights
+ *    of one or more patent holders.  This license does not release you
+ *    from the requirement that you obtain separate licenses from these
+ *    patent holders to use this software.
+ *  - Use of the software either in source or binary form, must be run
+ *    on or directly connected to an Analog Devices Inc. component.
+ *
+ * THIS SOFTWARE IS PROVIDED BY ANALOG DEVICES "AS IS" AND ANY EXPRESS OR
+ * IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, NON-INFRINGEMENT,
+ * MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED.
+ * IN NO EVENT SHALL ANALOG DEVICES BE LIABLE FOR ANY DIRECT, INDIRECT,
+ * INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, INTELLECTUAL PROPERTY RIGHTS, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+********************************************************************************
+ *
+ *  @section list_details Library description
+ *   This library handles double linked lists and it expose inseart,
+ *   read, get and delete functions. \n
+ *   It also can be accesed using it member functions which wrapp function for
+ *   usual list types.\n
+ *  @subsection example Sample code
+ *   @code{.c}
+ *	// -- Use a generic list
+ *	struct list_desc *list1;
+ *	struct iterator  *it;
+ *	uint32_t	 a;
+ *	// Create list list1
+ *	list_init(&list1, LIST_DEFAULT, NULL);
+ *	// Add items to the list
+ *	list_add_last(list1, 1);
+ *	list_add_last(list1, 2);
+ *	list_add_last(list1, 3);
+ *	// Here the list will be: 1 -> 2 - > 3
+ *
+ *	list_read_last(list1, &a);
+ *	printf("Last: %d\n", a);
+ *	// 3 will be printed
+ *	// Create an iterator on the end of the list
+ *	iterator_init(&it, list1, 0);
+ *	// Move the iterator backword with one position
+ *	iterator_move(it, -1);
+ *	// Read the data at the current position
+ *	iterator_read(it, &a);
+ *	printf("Current: %d\n", a);
+ *	// 2 will be printed
+ *	iterator_remove(it);
+ *	list_remove(list1);
+ *
+ *	// -- Use a popular list
+ *	struct list_desc *stack;
+ *	// Create a FIFO list
+ *	list_init(&stack, LIST_STACK, NULL);
+ *	// Put elements in the list
+ *	stack->push(stack, 1);
+ *	stack->push(stack, 2);
+ *	stack->push(stack, 3);
+ *	// Read from the stack
+ *	stack->pop(stack, &a);
+ *	printf("Last: %d\n", a);
+ *	// 3 will be printed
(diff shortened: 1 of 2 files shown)


mchindri:
v2:
- Fix cla

include/list.h
```diff
@@ -0,0 +1,406 @@
...
+
+#include <stdint.h>
+#include <stdbool.h>
+
+/******************************************************************************/
+/*************************** Types Declarations *******************************/
+/******************************************************************************/
+
+/** @brief Reference of an iterator on a list */
+typedef void* Iterator;
```
mirceacaprioru: why create this rename for void*. It could be used as is.

include/list.h
```diff
@@ -0,0 +1,406 @@
...
+ * @brief Add an element in the list.
+ *
+ * The element of the list is created and the data field is stored in it.
+ * @param list_desc - Reference to the list. Created by \ref list_init
+ * @param data - Data to store in a list element
+ * @return
+ *  - \ref SUCCESS : On success
+ *  - \ref FAILURE : Otherwise
+ */
+typedef int32_t	(*f_add)(List_desc list_desc, void *data);
```
mirceacaprioru: i don't quite get this extra layer of abstraction.

util/list.c
```diff
@@ -0,0 +1,744 @@
...
+	struct list_adapter	adapter;
+};
+
+/** @brief Default function used to compare element in the list ( \ref f_cmp) */
+static int32_t default_comparator(void *data1, void *data2)
+{
+	int64_t d1 = (uintptr_t)data1;
+	int64_t d2 = (uintptr_t)data2;
+
+	switch (sizeof(void *)) {
```
mirceacaprioru: this will just tell you the size of a void*, no information here about the data inside the void*.  I think you need the sizeof the data to do the following operations correctly.

util/list.c
```diff
@@ -0,0 +1,744 @@
...
+		return FAILURE;
+
+	list->l_it.elem = list->first;
+	if (SUCCESS != iterator_find(&(list->l_it), cmp_data))
+		return FAILURE;
+
+	return iterator_edit(&(list->l_it), new_data);
+}
+
+void*	list_read_first(List_desc list_desc, int32_t *result)
```
mirceacaprioru: usually there is space between return type and function name not tab

include/list.h
mchindri: This are used in list_adapter: To be easier to use a queue or stack for example. See the example https://github.com/analogdevicesinc/no-OS/blob/42c195741abb015c18c379d95dd7bc626de32069/include/list.h#L24

util/list.c
mchindri: Will change this

util/list.c
mchindri: image
When you collapse the functions or in the header file this way is more readable.
Should I replaece all with space?

include/list.h
mchindri: Acording to linux coding style this if fine. It is the exeception a)
Maybe a better naming will be iterator_t
image

include/list.h
```diff
@@ -0,0 +1,406 @@
...
+ *	iterator_init(&it, list1, 0);
+ *	iterator_move(it, -1);
+ *	printf("Last: %d\n", iterator_read(it, NULL));
+ *
+ *	iterator_remove(it);
+ *	list_remove(list1);
+ *	list_remove(list2);
+ *    \endcode
+*****************************************************************************//*
+ * Copyright 2019(c) Analog Devices, Inc.
```
amiclaus: nitpick: 2020

util/list.c
```diff
@@ -0,0 +1,744 @@
+/***************************************************************************//**
+ *   @file   list.c
+ *   @brief  List library implementation
+ *   @author Mihail Chindris (mihail.chindris@analog.com)
+********************************************************************************
+ * Copyright 2019(c) Analog Devices, Inc.
```
amiclaus: nitpick: 2020

include/list.h
mchindri: I think this is a type c exeption for typedefs. 
I do not use the typedefs to hide something, they are use to make a difference between to types of functions that have the same prototype.
image

adrimbarean: I would argue that typedef here is a good practice because of readability too. Declarations would get really complex if we had a function that returns this type of function pointer, of if we need an array of these kind of function pointers.

include/list.h
```diff
@@ -0,0 +1,406 @@
...
+	LIST_DEFAULT,
+	/**
+	 * Adapter for a FIFO list (Last-in first-out). Elements are inserted
+	 * and extracted only from one end
+	 *  - \e Push: Insert element
+	 *  - \e Pop: Get next element (Read and remove)
+	 *  - \e Top_next: Read next element
+	 *  - \e Back: Read first element
+	 *  - \e Swap: Edit the content of the next element
+	 */
```
adrimbarean: Comment is a bit misleading. FIFO = first in first out. This would be the correct implementation of a queue as the name suggests, but elements are inserted through one end and extracted from the other.

mchindri:
v3:
- Change copyrigth to 2020
- Move example code before copyright: see(https://mchindri.github.io/no-OS/list_8h.html#details)
- Change default_cmp function
- Change adapter comment

include/list.h
mchindri: solved

util/list.c
mchindri: solved

util/list.c
mchindri: solved

include/list.h
adrimbarean: I would argue that a `void *` doesn't have much to hide and can only be used with a proper cast. So, in my opinion, you could `typedef` the actual pointer to the struct you want to use to hide it's contents, or just use the `void *` as is. Doing both seems redundant to me.

adrimbarean: As a side note I would lean towards the former solution because that is how you make the object inaccessible to outside use. The latter is more of a way to have a pointer that can point to anything when you are not sure what you want to use.

mchindri: changed to struct iterator

include/list.h
mchindri: Keep this define this way

mchindri:
v4:
- Replace typdef with struct
- Remove adapter and include wrapper function in the list_desc structure
- Modify example

mchindri:
TODO: Remove tab before functions

mchindri:
v5
 - Replace tab with space before function name

mchindri:
v6 
 - astyle

include/list.h
```diff
@@ -0,0 +1,431 @@
...
+ * @param result - If not null, result is filled with:
+ *  - \ref SUCCESS : On success
+ *  - \ref FAILURE : Otherwise
+ * @return \n
+ *  - Content of the list element
+ *  - NULL : If some error occure
+ * @note If the content of an element can be 0 then the result must be checked
+ * to see if the functions has succeded
+ */
+typedef void* (*f_read)(struct list_desc *list_desc, int32_t *result);
```
adrimbarean: Usually we use the return of the function to check the status and an output parameter to get something out of the function. Maybe we should keep the consistency on this matter.

util/list.c
```diff
@@ -0,0 +1,757 @@
...
+	return iterator_edit(&(list->l_it), new_data);
+}
+
+void* list_read_first(struct list_desc *list_desc, int32_t *result)
+{
+	struct _list_desc	*list;
+
+	set_result(result, FAILURE);
+	if (!list_desc)
+		return FAILURE;
```
adrimbarean: Return integer type on a `void *` return. This ties to my earlier point that we would benefit by keeping consistency and returning a status and having output parameters to get data.

include/list.h
mchindri: First I tough to write them as you are saying. Then I tough it will be more practical this way.
you can use it as function parameter for example:
`printf("%d", queue->pop(queue, NULL));`
Assuming the function will fail only I the list is not handled properly, which can be check during development using the result param.
If after this explanation you think I should change it, I will change it.

util/list.c
mchindri: I forgot to modify this when I changed to the approach of returning a void *

include/list.h
buha: If the function can fail at any point, the return value should be an error code and any output parameter should be passed back by pointers. Rephrased, the only situation where the return value could be a meaningful output parameter is when there is no way the function can fail.

I missed this during my review, but I think what @adrimbarean says is the correct approach here since this function has failure points.

mchindri:
v7:
 - return int32_t and void * move as parameter.

mchindri:
v8:
 - Fix doxygen

util/list.c
```diff
@@ -0,0 +1,739 @@
...
+
+	if (!list_desc)
+		return FAILURE;
+	list = list_desc->priv_desc;
+
+
+	/* Based on place iterator */
+	elem = list->first;
+	while (elem) {
+		if (0 > list->comparator(data, elem->data))
```
adrimbarean: Here the order of the comparator function elements matters because some people might want to use the comparator method with heterogeneous elements like a struc and a number that might be a member of that struct. If that is the case the user needs to treat the elements in order since they can't be discriminated by their type. The problem is that here the elements are reversed: https://github.com/analogdevicesinc/no-OS/pull/493/commits/178ac39fbc0d3ca935b99ba18601bad4c76439cf#diff-520e93493ec866ff3fded06f02f051fcR651

mchindri:
v9
- swap parameters at line 348

mchindri:

mchindri:
TODO: Move comments to .c

mchindri:
v10: 
 move comments to .c file

dbogdan:
Some code style issues were found.

mchindri:
v11:
 - Fix astyle

PR #494, opened 2020-02-25
Iio demo

cristiantpop:
iio_demo: Add iio demo module and application

Registers a iio_demo_interfcae, independent of a physical device and should
be used as reference, when creating a new iio application for a new iio
device.


iio/iio_demo/iio_demo.c
```diff
@@ -0,0 +1,434 @@
...
+		.get_xml = iio_demo_get_xml,
+		.transfer_dev_to_mem = iio_demo_transfer_dev_to_mem,
+		.read_data = iio_demo_read_dev,
+		.transfer_mem_to_dev = iio_demo_transfer_mem_to_dev,
+		.write_data = iio_demo_write_dev,
+	};
+
+	status = iio_register(iio_interface);
+	if (status < 0)
+		goto error_free_iio_demo_delete_dev;
```
mchindri: It this fails you also have to free iio_interface

cristiantpop:
V2:

- Add parentheses to macros;

- Add one line comments;

- Free iio_interface if "iio_register" fails.

iio/iio_types.h
```diff
@@ -45,6 +45,7 @@
 
 #include <stdbool.h>
 #include <stdint.h>
+#include <stdio.h>
```
buha: What is it needed for ? I don't see any other change in iio_types.h, so I suspect that it is needed only for something in a .c file, in which case it is more suitable to move it to that .c file

cristiantpop: Actually the file needed here is  "types.h".

cristiantpop:
V3:

- Include missing "types.h" header file, for "ssize_t" type.

iio/iio_types.h
```diff
@@ -45,6 +45,7 @@
```
adrimbarean: Could you explain in the commit body why it is needed?

cristiantpop: Sure

cristiantpop:
V4:

- Explain in the commit body why "types.h", needs to be included.

cristiantpop:
V5:

- Add explicit casts to be compatible with C++ compilers

iio/iio_demo/iio_demo.c
```diff
@@ -0,0 +1,436 @@
...
+	.ch_out = false,
+};
+
+static struct iio_channel *iio_demo_channels[] = {
+	&iio_demo_channel_voltage0_in,
+	NULL,
+};
+
+static struct iio_attribute *iio_demo_global_attributes[] = {
+	&iio_attr_demo,
```
dbogdan: Let's create an independent iio_attribute for channels.

cristiantpop:
V6:

- Create an independent iio_attribute for channels.

PR #499, opened 2020-02-27
projects: fmcdaq2: initial version

amiclaus:
Update FMCDAQ2 project to use latest platform drivers and AXI core implementations.
Folder structure is kept similar to the updated projects.

Add makefile support for FMCDAQ2 project (Makefile, src.mk).

NOTE: Current implementation supports the following carrier boards: KCU105, VC707, KC705


amiclaus:
Close due to TravisCI issues.

PR #502, opened 2020-03-05
util: Add circular buffer library

mchindri:
This library will be used in the wifi driver.
It is useful to write data into the buffer from one task and read
it from another task.
This will also be useful for other future applications or drivers

### DIFF
--- a/util/circular_buffer.c
+++ b/util/circular_buffer.c
@@ -1,0 +1,245 @@
+/***************************************************************************//**
+ *   @file   circular_buffer.h
+ *   @brief  Circular buffer implementation
+ *   @author Mihail Chindris (mihail.chindris@analog.com)
+********************************************************************************
+ *   @copyright
+ * Copyright 2020(c) Analog Devices, Inc.
+ *
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *  - Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ *  - Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in
+ *    the documentation and/or other materials provided with the
+ *    distribution.
+ *  - Neither the name of Analog Devices, Inc. nor the names of its
+ *    contributors may be used to endorse or promote products derived
+ *    from this software without specific prior written permission.
+ *  - The use of this software may or may not infringe the patent rights
+ *    of one or more patent holders.  This license does not release you
+ *    from the requirement that you obtain separate licenses from these
+ *    patent holders to use this software.
+ *  - Use of the software either in source or binary form, must be run
+ *    on or directly connected to an Analog Devices Inc. component.
+ *
+ * THIS SOFTWARE IS PROVIDED BY ANALOG DEVICES "AS IS" AND ANY EXPRESS OR
+ * IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, NON-INFRINGEMENT,
+ * MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED.
+ * IN NO EVENT SHALL ANALOG DEVICES BE LIABLE FOR ANY DIRECT, INDIRECT,
+ * INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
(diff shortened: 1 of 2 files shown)


mchindri:
v2:
 - fix codacy

mchindri:
v3:
- fix codacy again

util/circular_buffer.c
```diff
@@ -0,0 +1,246 @@
...
+{
+	struct circular_buffer	*ldesc;
+	uint32_t		size;
+
+	if (!desc || !element_size || !nb_elements)
+		return FAILURE;
+
+	size = nb_elements * element_size;
+	if (size < nb_elements || size < element_size)
+		return FAILURE; //Integer overflow
```
buha: I am not convinced that this catches all scenarios. Somehow I have a gut feeling there is a counterexample that may get undetected by this check.
Worth checking: https://stackoverflow.com/questions/1815367/catch-and-compute-overflow-during-multiplication-of-two-large-integers

util/circular_buffer.c
```diff
@@ -0,0 +1,246 @@
...
+	int8_t		*write_ptr;
+	/** Set when the buffer is full */
+	bool		full;
+};
+
+/******************************************************************************/
+/************************ Functions Definitions *******************************/
+/******************************************************************************/
+
+int32_t circular_buffer_init(circular_buffer_t *desc, uint32_t nb_elements,
```
buha: too much to type, use `cb_init` or something like that which is shorter

util/circular_buffer.c
```diff
@@ -0,0 +1,246 @@
...
+
+	if (!ldesc)
+		return FAILURE;
+
+	*nb_elements =  size_to_elements(ldesc, get_unread_size(ldesc));
+
+	return SUCCESS;
+}
+
+int32_t circular_buffer_write(circular_buffer_t *desc, void *data,
```
buha: const void * data, to show it is an input buffer that is not edited

util/circular_buffer.c
mchindri: nb_elemnts and elemnt_size are unsigned so I think there is no counterexample. If one of them is 0 is checked in the previous if.

mchindri:
v4:
- Replace typedef void *circular_buffer -> struct circular_buffer
- Rename functions from circular_buffer -> cb
- In cb_write replace void *data -> const void *data

mchindri:
v5: 
 - Remove unnecessary functions. (size_to_element, element_to_size)

util/circular_buffer.c
```diff
@@ -0,0 +1,228 @@
...
+		return FAILURE;
+	desc->nb_elem += nb_elements;
+
+	size_to_end = desc->size - ptr_to_offset(desc, desc->write_ptr);
+	if (to_write_size > size_to_end) {
+		memcpy(desc->write_ptr, data, size_to_end);
+		to_write_size -= size_to_end;
+		memcpy(desc->buff, (uint8_t *)data + size_to_end,
+		       to_write_size);
+		desc->write_ptr = desc->buff + to_write_size;
```
adrimbarean: What if the write would pass the read pointer? Do we lose that data or guard against that?

mchindri:
Todo: overwrite when buffer is full

util/circular_buffer.c
```diff
@@ -0,0 +1,228 @@
...
+	*nb_elements =  desc->nb_elem;
+
+	return SUCCESS;
+}
+
+int32_t cb_write(struct circular_buffer *desc, const void *data,
+		 uint32_t nb_elements)
+{
+	uint32_t		to_write_size;
+	uint32_t		size_to_end;
```
mchindri: remove extra tabs

mchindri:
v6
 - Remove extra tabs
 - Enable overwritting

mchindri:
v7 
 - Fix coday

util/circular_buffer.c
```diff
@@ -0,0 +1,204 @@
...
+	int8_t		*read_ptr;
+	/** Pointer to writing location */
+	int8_t		*write_ptr;
+};
+
+/******************************************************************************/
+/************************ Functions Definitions *******************************/
+/******************************************************************************/
+
+int32_t cb_init(struct circular_buffer **desc, uint32_t nb_elements,
```
adrimbarean: Could we get just the short description from `@brief` here in the .c file too without breaking doxygen? I am accustomed to have a short description wherever I look at the function, but maybe that is just me.

mchindri: I think will be a good practice to do not write documentation in 2 places, just for maintenance reasons. If something changes, the documentation has to be modified in 2 places.
Maybe for review will be useful to be in .c file. But if you want just to use the functions, you will look at the .h file. And the review is don once.

mchindri:
v8
 - Move comments to .c file

mchindri:
will merge if no more comments

PR #503, opened 2020-03-09
spi ps clock/prescaler fixes

buha:
- the driver assumed 100MHz speed for the prescaler calculation which may not be the case
- fix corner case in prescaler calculation

### DIFF
--- a/drivers/platform/xilinx/spi.c
+++ b/drivers/platform/xilinx/spi.c
@@ -182,6 +182,7 @@ static int32_t spi_init_ps(struct spi_desc *desc,
 	if (desc->max_speed_hz != 0u) {
 		uint32_t div = input_clock / desc->max_speed_hz;
 		uint32_t rem = input_clock % desc->max_speed_hz;
+		uint32_t po2 = !(div & (div - 1)) && !rem;
 
 		// find the power of two just higher than div and
 		// store the exponent in prescaler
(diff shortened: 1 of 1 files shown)


dbogdan:
Please fix the Codacy complaints.

buha:
Fixed codacy complaints

drivers/platform/xilinx/spi.c
```diff
@@ -179,9 +183,21 @@ static int32_t spi_init_ps(struct spi_desc *desc,
 	if(ret != SUCCESS)
 		goto ps_error;
 
+	switch (xinit->device_id) {
+	case 0:
+		input_clock = SPI_CLK_FREQ_HZ(0);
+		break;
+	case 1:
+		input_clock = SPI_CLK_FREQ_HZ(1);
+		break;
```
dbogdan: I would do:
case 0:
case 1:
   input_clock = SPI_CLK_FREQ_HZ(xinit->device_id);
   break;
However, is there any reason for limiting this to 0 and 1?

buha: It won't work like that, because of what SPI_CLK_FREQ_HZ does. It uses the preprocessor directive ## to literally combine strings.

I checked the PS7 and PSU architecture manual and both are limited only to SPI0 and SPI1 so this switch should be ok. In case we wlil support a new architecture that will have SPI2, SPI3 etc. we will need to extend the switch.

dbogdan: OK. Thanks.

PR #507, opened 2020-03-12
Small fixes mixed in one pull request

mchindri:
Each fix description if found in its own commit.

### DIFF
--- a/drivers/platform/aducm3029/timer.c
+++ b/drivers/platform/aducm3029/timer.c
@@ -69,13 +69,13 @@
 /******************************************************************************/
 
 /** Incremented each millisecond by \ref tmr_callback() */
-static uint64_t g_count;
+static volatile uint64_t	g_count;
 /** Counts the number of instances created */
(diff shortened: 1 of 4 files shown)


include/error.h
```diff
@@ -51,4 +51,7 @@
 #define FAILURE		-1
 #endif
 
+#define RET_OK(x)	((x) == SUCCESS)
+#define RET_FAIL(x)	((x) != SUCCESS)
```
adrimbarean: Does it make it more readable? I feel like more often than not abstracting things can create confusion. That said I do not oppose the change, I would just wish to know if the majority of us see improvement here.

mchindri: I will put this in other pull request

cristiantpop: I wouldn't define this macros. They might create confusion.

mchindri:
V2
- Remove error.h commit and move to https://github.com/analogdevicesinc/no-OS/pull/509

PR #509, opened 2020-03-16
include:error.h Create define to check function result

mchindri:
With this macros, checking the result of no-os functions will be more readable:
Old:
    if (SUCCESS != spi_init())
         goto handle_fail;
New:
   if (RET_FAIL(spi_init())
        goto handle_fail;

### DIFF
--- a/include/error.h
+++ b/include/error.h
@@ -51,4 +51,6 @@
 #define FAILURE		-1
 #endif
 
+#define IS_ERR_VALUE(x)	((x) < 0)
+
 #endif // ERROR_H_


mchindri:
This idea is based on windows drivers return status:
https://docs.microsoft.com/en-us/windows-hardware/drivers/kernel/using-ntstatus-values
The use an NT_SUCCESS macro to test the result of functions.
First opinions start here: https://github.com/analogdevicesinc/no-OS/pull/507#discussion_r392930575

include/error.h
```diff
@@ -51,4 +51,7 @@
 #define FAILURE		-1
 #endif
 
+#define RET_OK(x)	((x) == SUCCESS)
+#define RET_FAIL(x)	((x) != SUCCESS)
+
```
scuciurean: Usually, the return code is checked if less than 0. In our case, we got only 2 return codes so we can use '!=' but maybe we should add our own return codes later on.

mchindri:
v2:
 - Compare with 0

include/error.h
```diff
@@ -51,4 +51,7 @@
 #define FAILURE		-1
 #endif
 
+#define RET_OK(x)	((x) >= 0)
+#define RET_FAIL(x)	((x) < 0)
```
dbogdan: I would only define an IS_ERR_VALUE(), similar to Linux.

mchindri:
v3
 - define IS_ERR_VALUE

mchindri:
Will merge if no more comments today.

PR #510, opened 2020-03-17
drivers:platform:xilinx: update I2C driver for multiple instances

adrimbarean:
The I2C driver needs to accustom devices on the same I2C bus. Add support
to init and remove multiple instances of the driver on the same bus without
re-initializing the core or BSP driver. Add support to change transmission
parameters like bit rate and slave address at every transmission, if
necessary.

### DIFF
--- a/drivers/platform/xilinx/i2c.c
+++ b/drivers/platform/xilinx/i2c.c
@@ -56,8 +56,122 @@
 #include "i2c_extra.h"
 
 /******************************************************************************/
+/*************************** Types Declarations *******************************/
+/******************************************************************************/
+
+/** Table of existing BSP driver handlers */
+#ifdef XPAR_XIIC_NUM_INSTANCES
+/**
+ * @struct pl_table_item
+ * @brief Table item for PL I2C cores already started.
+ */
+struct pl_table_item {
+	/** Instance the BSP driver uses for this core. */
+	XIic *instance;
+	/** Core ID in BSP. */
+	uint32_t device_id;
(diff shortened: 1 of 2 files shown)


adrimbarean:
V0 -> V1:
 - fix Codacy error.

drivers/platform/xilinx/i2c.c
```diff
@@ -55,10 +55,124 @@
...
+struct pl_table_item {
+	/** Instance the BSP driver uses for this core. */
+	XIic *instance;
+	/** Core ID in BSP. */
+	uint32_t device_id;
+	/** Number of ADI driver instances running on this core. */
+	uint32_t inst_no;
+};
+/** Dynamically allocated table for PL I2C BSP driver instances. */
+static struct pl_table_item *pl_table = NULL;
```
mchindri: I think you can use a list and the associated functions to handle this.
Also using a list functions used for find and for allocation will not be needed

adrimbarean: Do you mean the FIFO utils? Or which list?

mchindri: https://github.com/analogdevicesinc/no-OS/pull/493#pullrequestreview-368703841

adrimbarean: I would like to do that, but I will try and do it after it has been merged to master.

mchindri: Fine for me

PR #513, opened 2020-03-20
drivers: freq: ad9528: check if hw reset enabled

amiclaus:
Not all projects that use ad9528 driver have a gpio connected to the
reset pin of the device.

This patch adds a `hw_reset_en` parameter to the device initialization
structure that is used to verify if hardware reset can be performed.

Update adrv9009 project to use the latest feature of the driver.

Fixes: a223f817f40aed7a33ade8b472855f0632171f73

### DIFF
--- a/drivers/frequency/ad9528/ad9528.c
+++ b/drivers/frequency/ad9528/ad9528.c
@@ -374,18 +374,18 @@ int32_t ad9528_setup(struct ad9528_dev **device,
 
 	dev->pdata = init_param.pdata;
 
-	/* GPIO */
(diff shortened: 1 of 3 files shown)


dbogdan:
What's the project you are talking about? For many components, the RESET is more than recommended.

amiclaus:
> 
> 
> What's the project you are talking about? For many components, the RESET is more than recommended.
FMCDAQ3 - the RESET_N pin of the clock chip is floating, as it can be seen at page 4 from the schematic.
https://wiki.analog.com/_media/resources/eval/user-guides/ad-fmcdaq3-ebz/fmcdaq3_revc.pdf

PR #514, opened 2020-03-23
include:uart Add nonblocking functions + support for aducm3029

mchindri:
This changes where requested during the meeting about the irq_register_callback.
Only when using this function a callback will be generated for the uart.
A pull request for the irq will come soon.

### DIFF
--- a/drivers/platform/aducm3029/uart.c
+++ b/drivers/platform/aducm3029/uart.c
@@ -45,6 +45,10 @@
 #include "uart_extra.h"
 #include <stdlib.h>
 
+#ifndef min
+#define min(x, y) (((x) > (y)) ? (y) : (x))
+#endif
+
 /******************************************************************************/
 /*************************** Types Declarations *******************************/
 /******************************************************************************/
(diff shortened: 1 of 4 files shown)


mchindri:
v2:
 - Fix astyle

dbogdan:
Implementing uart_read_nonblocking()/uart_write_nonblocking() but keeping the same struct uart_desc (only expanding its *extra member) is a bad thing - makes the code too hard to understand.
By the way, how knows the user that the data is available/was sent? How can the data be read by the user?
We definitely need some nonblocking functions, but I'm not sure if they should be part of uart.c/uart.h...

mchindri:
The user will know when the transaction is done after registering a callback. This will be a new pull request that joins this one with the modification in the irq. I did it this way to avoid making too much changes in one pull request.
I don't really understand way is a bad think to keep the uart_desc. This way you can read blocking and nonblocking with the same descriptor, depending on the needs. For example, you can submit a buffer to read a character (nonblocking) and when you read it, switch to blocking mode.

dbogdan:
*extra member should be used only for the platform specific requirements/settings. If a callback is required for all platforms, if should be part of the the generic uart_desc structure.

mchindri:
true. I didn't think about that. I will move what I added to the extra to the uart_desc.

drivers/platform/aducm3029/uart.c
```diff
@@ -42,9 +42,14 @@
 /******************************************************************************/
 
 #include "uart.h"
+#include "irq.h"
```
mchindri: not needed

mchindri:
v3:
 - Remove changes unrelated with this commit:
    -Remove irq.h
    -remove callback form uart_extra.h

PR #518, opened 2020-03-23
drivers: ad5770r: remove ch_enable register

mike-bradley:
The CH_ENABLE register is not implemented and is not documented in the
data sheet. This removes the #defines and associated functions.

Signed-off-by: Michael Bradley <Michael.Bradley@analog.com>


mirceacaprioru:
This change will have to be ported to Linux too. Would you like to do it or should we tackle this ?

mike-bradley:
if you can take on the linux porting that'd be great, I'm not setup to build/test on that platform.

mirceacaprioru:
ok got it.

Thank you for the fix.

PR #519, opened 2020-03-24
xilinx: gpio: add get_value implementation

amiclaus:
Add implementation for the `gpio_get_value` function in the Xilinx GPIO
driver.

### DIFF
--- a/drivers/platform/xilinx/gpio.c
+++ b/drivers/platform/xilinx/gpio.c
@@ -382,12 +382,36 @@ int32_t gpio_set_value(struct gpio_desc *desc,
 int32_t gpio_get_value(struct gpio_desc *desc,
 		       uint8_t *value)
 {
-	if (desc) {
-		// Unused variable - fix compiler warning
-	}
+	if (!desc)
+		return FAILURE;
 
-	if (value) {
-		// Unused variable - fix compiler warning
+	struct xil_gpio_desc	*extra;
+	uint8_t pin = desc->number;
+
+#ifdef XGPIO_H
+	uint8_t channel;
+#endif
+	extra = desc->extra;
+
+	switch (extra->type) {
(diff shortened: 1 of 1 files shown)


drivers/platform/xilinx/gpio.c
```diff
@@ -382,12 +382,35 @@ int32_t gpio_set_value(struct gpio_desc *desc,
...
+			channel = 2;
+			pin -= 32;
+		} else
+			channel = 1;
+		*value = (XGpio_DiscreteRead(extra->instance, channel) >> pin) & 0x01;
+#endif
+		break;
+	case GPIO_PS:
+#ifdef XGPIOPS_H
+		*value = XGpioPs_ReadPin(extra->instance, pin);
```
mchindri: I see that you use pin here but is declared only when XGPIO_H is defined.
I think you can declare it outside the define.

drivers/platform/xilinx/gpio.c
```diff
@@ -382,12 +382,35 @@ int32_t gpio_set_value(struct gpio_desc *desc,
...
-		// Unused variable - fix compiler warning
-	}
+	if (!desc)
+		return FAILURE;
 
-	if (value) {
-		// Unused variable - fix compiler warning
+	struct xil_gpio_desc	*extra;
+#ifdef XGPIO_H
+	uint8_t pin = desc->number;
```
mchindri: Nitpick: I think is better if we  declare the variables at the begging of the function

drivers/platform/xilinx/gpio.c
amiclaus: You are correct. This is modified now.

amiclaus:
V2:
- move `pin` variable declaration at the start of the function

PR #520, opened 2020-03-24
drivers:platform:aducm: Add irq support for uart_read/write_nonblocking (DRAFT)

mchindri:
The first 6  commits are the ones existing in the others pull request.
The last 3 will be a new pull request when the first two are merged.

I put these changes here as a draft to view the final result only.


include/irq.h
```diff
@@ -50,6 +50,19 @@
 /*************************** Types Declarations *******************************/
 /******************************************************************************/
 
+/**
+ * @enum irq_uart_event_e
+ * @brief Possible events for uart interrupt
+ */
+enum irq_uart_event_e {
```
mchindri: If we didn't need dependency on irq from uart.h, should this be here? 
(To put the events here was decided in a previous meeting, but in the current one, it was specified that we shouldn't have irq in uart)

PR #521, opened 2020-03-24
projects: fmcadc5 initial version

amiclaus:
Update FMCADC5 project to use latest platform drivers and AXI core implementations.
Folder structure is kept similar to the updated projects.
Add makefile support for FMCADC5 project (Makefile, src.mk).

**NOTE**: Requires #519


amiclaus:
V2:
- fix remaining code style issues.

PR #525, opened 2020-03-26
drivers:platform:xilinx: use list util for I2C driver

adrimbarean:
The list utility can be used for the I2C driver to manage the initialized
cores table easier.

### DIFF
--- a/drivers/platform/xilinx/i2c.c
+++ b/drivers/platform/xilinx/i2c.c
@@ -54,47 +54,31 @@
 #include "error.h"
 #include "i2c.h"
 #include "i2c_extra.h"
+#include "list.h"
 
 /******************************************************************************/
 /*************************** Types Declarations *******************************/
 /******************************************************************************/
 
-/** Table of existing BSP driver handlers */
-#ifdef XPAR_XIIC_NUM_INSTANCES
+static struct list_desc *pl_list = NULL;
+static struct iterator  *pl_it = NULL;
+static struct list_desc *ps_list = NULL;
+static struct iterator  *ps_it = NULL;
+
 /**
- * @struct pl_table_item
- * @brief Table item for PL I2C cores already started.
+ * @struct inst_table_item
+ * @brief Table item for I2C cores already started.
  */
-struct pl_table_item {
+struct inst_table_item {
 	/** Instance the BSP driver uses for this core. */
-	XIic *instance;
+	void *instance;
 	/** Core ID in BSP. */
 	uint32_t device_id;
 	/** Number of ADI driver instances running on this core. */
 	uint32_t inst_no;
 };
-/** Dynamically allocated table for PL I2C BSP driver instances. */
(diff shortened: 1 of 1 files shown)


drivers/platform/xilinx/i2c.c
```diff
@@ -205,11 +167,20 @@ int32_t i2c_init(struct i2c_desc **desc,
...
+		struct inst_table_item tab_check_pl;
+		struct inst_table_item *temp_el_pl;
+
+		if (!pl_list)
+			list_init(&pl_list, LIST_DEFAULT, i2c_cmp);
+		if (!pl_it)
+			iterator_init(&pl_it, pl_list, true);
+
+		tab_check_pl.device_id = xdesc->device_id;
+		ret = iterator_find(pl_it, &tab_check_pl);
```
mchindri:
```suggestion
		if (IS_ERR_VALUE(list_read_find(pl_list, (void **)&temp_el_pl, &tab_check_pl))
		      goto pl_error;
                 else
                      //Item found and stored in temp_el_pl
```

mchindri:
```suggestion
		if (IS_ERR_VALUE(list_read_find(pl_list, (void **)&temp_el_pl, &tab_check_pl))
		      goto pl_error; //Item not found
                 else
                      //Item found and stored in temp_el_pl
```
You don't need to use an iterator. Same ps_it

adrimbarean: Will change.

adrimbarean: I am still using the iterator for the remove function. I can avoid using the iterator there too using the same principle, but it would mean searching for an object twice, once to read and modify it and once to erase it if necessary. I think that keeping the iterator for the remove function is worthwhile, but I'm not sure. Am I wrong in this?

adrimbarean:
V1:
 - update the init function to not need the iterator.

drivers/platform/xilinx/i2c.c
mchindri: I am agree

drivers/platform/xilinx/i2c.c
```diff
@@ -338,70 +302,50 @@ int32_t i2c_remove(struct i2c_desc *desc)
...
-				else
-					memcpy(&tab_check_pl[i], &pl_table[(i + 1)],
-					       sizeof(*pl_table));
-			}
-			free(pl_table);
-			pl_table = tab_check_pl;
-			pl_core_no--;
-		}
+
+		iterator_get(pl_it, (void **)&temp_el_pl);
```
mchindri: Nitpick: You can add a comment saying that you just delete the item here.

mchindri: LGTM

adrimbarean:
V2:
 - add comment on list element erase;
 - fix astyle.

PR #529, opened 2020-03-27
altera: gpio: update driver

amiclaus:
Add implementation for `gpio_remove` and `gpio_get_value`.

Move implementation of `gpio_direction_output` to `gpio_set_value`. The
current implementation only sets the gpio value, while the information
regarding the direction of the gpio is contained in the `base_address`.

The `gpio_set_value` is called inside `gpio_direction_output` to avoid
errors within the projects that are using these functions.

### DIFF
--- a/drivers/platform/altera/gpio.c
+++ b/drivers/platform/altera/gpio.c
@@ -99,7 +99,8 @@ int32_t gpio_get(struct gpio_desc **desc,
 int32_t gpio_remove(struct gpio_desc *desc)
 {
 	if (desc) {
-		// Unused variable - fix compiler warning
+		free(desc->extra);
+		free(desc);
 	}
 
 	return SUCCESS;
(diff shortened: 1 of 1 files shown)


drivers/platform/altera/gpio.c
```diff
@@ -130,31 +131,7 @@ int32_t gpio_direction_input(struct gpio_desc *desc)
 int32_t gpio_direction_output(struct gpio_desc *desc,
```
mchindri: So this function is used in some drivers to set the gpio value?

amiclaus: Generally speaking, the aim of this function is to set the direction of a gpio as output, and also the value of the gpio (high/low).

Since in this particular case there is no need to set the direction, the function will only set the value.

As a reference for a complete implementation for `gpio_direction_output`, you can have a look here: https://github.com/analogdevicesinc/no-OS/blob/5c6135e9d50b61527e184c429566c6488acc246c/drivers/platform/xilinx/gpio.c#L268-L279

`XGpio_SetDataDirection` sets the gpio direction, while  `XGpio_DiscreteWrite` sets the output value, both contained in the same function `gpio_direction_output`.

mchindri: I understand, I got confused the function with gpio_set_direction.

dbogdan:
This is good for now. However, the implementation assumes that the output direction is already enabled when gpio_set_value() is called. Maybe this constraint can be removed by a future commit.

PR #530, opened 2020-03-27
drivers: uart + irq: Add uart_read/write_nonblocking functions

mchindri:


scuciurean:
#012c204290c6 is not signed.

mchindri:
V2:
 - Sign: https://github.com/analogdevicesinc/no-OS/pull/530/commits/fa6fa5b94e2877287740fdfdc5398f2d13466653 .

drivers/platform/aducm3029/uart.c
```diff
@@ -45,6 +45,10 @@
 #include "uart_extra.h"
 #include <stdlib.h>
 
+#ifndef min
+#define min(x, y) (((x) > (y)) ? (y) : (x))
```
dbogdan: min() is defined in util.h. We should include this header instead of redefining it.

dbogdan:
@mchindri , can you show us the commit that updates drivers/platform/aducm3029/irq.c too? It will help us a lot to understand the rest of the changes.

dbogdan:
Sorry, I missed it. It's here as well. :)

dbogdan:
In this implementation, the interrupts seem to be enabled by the BSP even for the blocking situation. That means that the irq_global_disable() can't do what it's expected to do.
Aren't other functions available that can be used in the blocking situations for removing this constraint?

drivers/platform/aducm3029/uart.c
mchindri: I have some conflicts with util.h. I will solve the problem in an other way if defining min is not ok

dbogdan: Let's fix util.h then.

mchindri:
There aren't other functions. The irq are enabled by the BSP in the init function. But the irq_global_disable will deactivate the user callback. 
image

I can disable them, but I don't know if it is a good idea.

dbogdan:
Can't we just directly access TX/RX registers (hDevice->pUARTRegs->TX/hDevice->pUARTRegs->RX) in blocking mode?

mchindri:
v3:
 - Add new commit with changes in irq_gloabal_disable/enable

mchindri:
Remove last commit and fix util

mchindri:
v4: 
 - Remove https://github.com/analogdevicesinc/no-OS/pull/530/commits/8227c492c42bc860c95b1cc54f0563f3e1cdf673
 - Replace #define min with min from utils

drivers/platform/aducm3029/uart.c
mchindri: I solved by including util.h after <stdlib.h>

drivers/platform/aducm3029/irq.c
```diff
@@ -142,34 +127,64 @@ int32_t irq_ctrl_init(struct irq_ctrl_desc **desc,
...
+
+	aducm_desc = desc->extra;
+
+	if (irq_id < NB_EXT_INTERRUPTS) {
+		aducm_desc->conf[irq_id].xint_conf =
+			(enum irq_mode)callback_desc->config;
+		adi_xint_RegisterCallback(id_map_event[irq_id],
+					  callback_desc->callback,
+					  callback_desc->ctx);
+	} else { //if (irq_id == ADUCM_UART_INT_ID) {
```
adrimbarean, line 177: We probably need this as an `else if` if we want to add more interrupt options to this. Also, this `if else` clause might get bigger if it will support all other drivers. Should we consider making it a `switch case` statement?

drivers/platform/aducm3029/irq.c
```diff
@@ -181,14 +196,28 @@ int32_t irq_register_callback(struct irq_ctrl_desc *desc, uint32_t irq_id,
...
 		return FAILURE;
 
-	((struct aducm_irq_ctrl_desc *)desc->extra)->
-			callbacks[irq_id].callback = 0;
+	aducm_desc = desc->extra;
 
-	return SUCCESS;
+	if (irq_id < NB_EXT_INTERRUPTS) {
+		adi_xint_RegisterCallback(id_map_event[irq_id], 0, 0);
+	} else if (aducm_desc->conf[irq_id].uart_conf) { //irq_id = UART_ID
```
adrimbarean, line 210: nitpick: we probably want to use C style comments to keep the consistency.

drivers/platform/aducm3029/irq.c
mchindri: Yes, a switch will be better

drivers/platform/aducm3029/irq.c
mchindri: A switch will be used here too

mchindri:
v5:
- Replace if else with switch case

PR #532, opened 2020-03-29
Implement a generic clock layer

dbogdan:


dbogdan:
v2:
 - Fix astyle complaints.

dbogdan:
v3:
 - Fix run_build complaints.

include/clk.h
```diff
@@ -0,0 +1,88 @@
...
+/***************************** Include Files **********************************/
+/******************************************************************************/
+#include <stdint.h>
+
+/******************************************************************************/
+/************************* Structure Declarations *****************************/
+/******************************************************************************/
+struct clk_hw {
+	void	*dev;
+	int32_t (*dev_clk_enable)();
```
mchindri: If no parameters are defined, it means it can receive any number of parameters?

dbogdan: Normally, yes, but the this approach was chosen because the parameter's type is unknown. Two examples:
 - axi_jesd204_tx_lane_clk_enable(struct axi_jesd204_tx *jesd)
 - axi_jesd204_rx_lane_clk_enable(struct axi_jesd204_tx *jesd)

dbogdan:
v4:
 - rename hw_chan_no to hw_ch_num;
 - remove hw_ch_num parameter from dev_clk_enable()/dev_clk_disable().

PR #533, opened 2020-03-30
RTC platform driver

adrimbarean:
This pull request adds the RTC header to define the RTC driver API and adds RTC support for the ADuCM3029 platform. Add driver implementation as well as interrupt support for the driver.

### DIFF
--- a/drivers/platform/aducm3029/irq_extra.h
+++ b/drivers/platform/aducm3029/irq_extra.h
@@ -53,7 +53,11 @@
 /******************************************************************************/
 
 /** Number of available interrupts */
-#define NB_INTERRUPTS		5u
+#define NB_INTERRUPTS		6u
+
+/** RTC interrupt defines */
+#define RTC_COUNT_INT		ADI_RTC_COUNT_INT
+#define RTC_COUNT_ROLLOVER_INT	ADI_RTC_COUNT_ROLLOVER_INT
 
 /******************************************************************************/
 /*************************** Types Declarations *******************************/
@@ -73,7 +77,9 @@ enum irq_id {
 	/** External interrupt 0, on GPIO 33 */
 	ADUCM_EXTERNAL_INT3_ID,
 	/** UART interrupt ID*/
-	ADUCM_UART_INT_ID
+	ADUCM_UART_INT_ID,
+	/** RTC interrupt ID*/
+	ADUCM_RTC_INT_ID
 };
 
 /**
@@ -94,6 +100,17 @@ enum irq_mode {
 };
 
 /**
+ * @struct rtc_irq_config
+ * @brief RTC interrupt configuration routine.
+ */
+struct rtc_irq_config {
+	/** RTC driver handler */
+	struct rtc_desc *rtc_handler;
+	/** Active interrupts OR'ed together */
+	uint32_t active_interrupts;
+};
+
+/**
  * @union irq_config
  * @brief Configuration for the callback
  */
@@ -102,6 +119,8 @@ union irq_config {
 	struct uart_desc	*uart_conf;
 	/** Trigger condition for the external interrupt */
 	enum irq_mode		xint_conf;
+	/** RTC interrupt config */
+	struct rtc_irq_config	*rtc_conf;
 };
 
 /**
--- a/drivers/platform/aducm3029/rtc.c
+++ b/drivers/platform/aducm3029/rtc.c
@@ -1,0 +1,193 @@
+/***************************************************************************//**
+ *   @file   aducm3029/rtc.c
+ *   @brief  Implementation of RTC driver for ADuCM302x
+ *   @author Andrei Drimbarean (andrei.drimbarean@analog.com)
+********************************************************************************
+ * Copyright 2020(c) Analog Devices, Inc.
+ *
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *  - Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ *  - Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in
+ *    the documentation and/or other materials provided with the
+ *    distribution.
+ *  - Neither the name of Analog Devices, Inc. nor the names of its
+ *    contributors may be used to endorse or promote products derived
+ *    from this software without specific prior written permission.
+ *  - The use of this software may or may not infringe the patent rights
+ *    of one or more patent holders.  This license does not release you
+ *    from the requirement that you obtain separate licenses from these
+ *    patent holders to use this software.
+ *  - Use of the software either in source or binary form, must be run
+ *    on or directly connected to an Analog Devices Inc. component.
+ *
+ * THIS SOFTWARE IS PROVIDED BY ANALOG DEVICES "AS IS" AND ANY EXPRESS OR
+ * IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, NON-INFRINGEMENT,
+ * MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED.
+ * IN NO EVENT SHALL ANALOG DEVICES BE LIABLE FOR ANY DIRECT, INDIRECT,
+ * INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, INTELLECTUAL PROPERTY RIGHTS, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+*******************************************************************************/
+
+/******************************************************************************/
+/************************* Include Files **************************************/
+/******************************************************************************/
+
+#include <stdlib.h>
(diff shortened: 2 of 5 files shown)


drivers/platform/aducm3029/rtc.c
```diff
@@ -0,0 +1,186 @@
...
+int32_t rtc_init(struct rtc_dev **device, struct rtc_init_param *init_param)
+{
+	int32_t ret;
+	struct rtc_dev *dev;
+	struct aducm_rtc_dev *adev;
+	struct aducm_rtc_init_param *ainit = init_param->extra;
+
+	dev = (struct rtc_dev *)calloc(1, sizeof(*dev));
+	if (!dev)
+		return FAILURE;
```
scuciurean: I believe the goto statements are shifted.
Instead of returning failure, you should use goto error_dev( same for the other).

scuciurean:
Fix the astyle issue, rest looks good.

include/rtc.h
```diff
@@ -0,0 +1,97 @@
...
+	void *extra;
+};
+
+/**
+ * @struct rtc_init_param
+ * @brief Structure holding RTC initialization parameters.
+ */
+struct rtc_init_param {
+	/** ID of the real time clock core. */
+	uint8_t id;
```
mchindri, line 70: I think we should add a load_value here

include/rtc.h
```diff
@@ -0,0 +1,97 @@
...
+
+/******************************************************************************/
+/*************************** Types Declarations *******************************/
+/******************************************************************************/
+
+/**
+ * @struct rtc_dev
+ * @brief Structure holding RTC descriptor.
+ */
+struct rtc_dev {
```
mchindri, line 57: I think you should rename it to rtc_desc for consistency with the other drivers.

include/rtc.h
```diff
@@ -0,0 +1,97 @@
...
+/******************************************************************************/
+
+/** Initialize the RTC peripheral. */
+int32_t rtc_init(struct rtc_dev **device, struct rtc_init_param *init_param);
+
+/** Free the resources allocated by rtc_init(). */
+int32_t rtc_remove(struct rtc_dev *dev);
+
+/** Start the real time clock. */
+int32_t rtc_start(struct rtc_dev *dev);
```
mchindri, line 86: I don't think if someone will need to start and stop an rtc, they can be put in the init and remove. For start and stop functionalities I think a timer should be used. Inspired from https://os.mbed.com/docs/mbed-os/v5.15/mbed-os-api-doxy/mbed__rtc__time_8h_source.html

drivers/platform/aducm3029/rtc_extra.h
```diff
@@ -0,0 +1,75 @@
...
+	uint8_t prescaler;
+};
+
+/**
+ * @struct aducm_rtc_init_param
+ * @brief Structure holding ADuCM3029 specific RTC initialization parameters.
+ */
+struct aducm_rtc_init_param {
+	/** Input clock prescaler. */
+	uint8_t prescaler;
```
mchindri, line 72: I think you can use an enum with the available frequencies and calculate the prescaler inside the driver.
Also specify somewhere that only for the RTC1 a frequency can be set and for RTC0 is 1Hz. 
Maybe move this to generic file as frequency

include/rtc.h
```diff
@@ -0,0 +1,97 @@
...
+
+/** Stop the real time clock. */
+int32_t rtc_stop(struct rtc_dev *dev);
+
+/** Get the current count for the real time clock. */
+int32_t rtc_get_cnt(struct rtc_dev *dev, uint32_t *tmr_cnt);
```
mchindri, line 92: I whould use a uint64_t tmr_cnt. This is because if we want to count the seconds since January 1, 1970, the uin32_t will overflow soon.

drivers/platform/aducm3029/rtc.c
```diff
@@ -0,0 +1,186 @@
...
+ */
+int32_t rtc_stop(struct rtc_dev *dev)
+{
+	struct aducm_rtc_dev *adev;
+
+	if (!dev)
+		return FAILURE;
+	adev = dev->extra;
+
+	return adi_rtc_Enable(adev->instance, false);
```
mirceacaprioru, line 161: are these error codes compatibe to our api ?

drivers/platform/aducm3029/rtc_int.h
```diff
@@ -0,0 +1,59 @@
...
+#include <stdint.h>
+#include <drivers/rtc/adi_rtc.h>
+#include "rtc.h"
+
+/******************************************************************************/
+/************************ Functions Declarations ******************************/
+/******************************************************************************/
+
+/** Attach a callback routine to the DFP provided interrupt handler. */
+int32_t rtc_int_register_callback(struct rtc_dev *desc, ADI_CALLBACK cb,
```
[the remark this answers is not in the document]
adrimbarean: Will do this as soon as possible.

include/rtc.h
adrimbarean: Will add.

include/rtc.h
adrimbarean: Will do.

include/rtc.h
adrimbarean: I would argue that it is better to be able to stop and start the RTC in case we need to adjust it's mode of operation, or maybe re-calibrate it.

drivers/platform/aducm3029/rtc_extra.h
adrimbarean: Will do.

drivers/platform/aducm3029/rtc.c
adrimbarean: At the moment, yes.

include/rtc.h
adrimbarean: The count register for the ADuCM3029 RTC is a 32 bit register. More accurately it is 2 16 bit registers that function together. It does contain another 15 bit register, but that can only be read as a fractional part to the current value. It is the duty of the application to provide time measurement using this function an interrupts.

mchindri: I know. I'am not strongly want this change, but I still think that doing this inside the driver will ease the application development

adrimbarean:
Changed pull request from "draft" to "ready".
V1:

- Add irq support for the RTC driver;
- Change the handler struct name to rtc_desc;
- Add load and frequency members to the handler struct as well as a frequency options enum based on the possible prescaler values as well as nominal input clock.

drivers/platform/aducm3029/rtc.c
```diff
@@ -0,0 +1,210 @@
...
+	dev->freq = init_param->freq;
+	dev->load = init_param->load;
+	dev->extra = adev;
+
+	ret = adi_rtc_Open(dev->id, adev->memory, ADI_RTC_MEMORY_SIZE,
+			   &adev->instance);
+	if(ret != SUCCESS)
+		goto error_mem;
+
+	switch (dev->freq) {
```
mchindri, line 88:
```suggestion
	if (dev->freq < AUDCM_32768HZ || dev->freq > AUDCM_1HZ)
	    ret = adi_rtc_SetPreScale(adev->instance, dev->freq);
	    if (ret != ADI_RTC_SUCCESS )
```

drivers/platform/aducm3029/rtc_extra.h
```diff
@@ -0,0 +1,97 @@
...
+
+/******************************************************************************/
+/********************** Macros and Constants Definitions **********************/
+/******************************************************************************/
+
+/** If more interrupts are needed they need to be enabled here. To do this, OR
+ *  them with each other.
+ *  Example: if rollover and alarm interrupts are needed write
+ *  #define RTC_USED_INTERRUPTS (ADI_RTC_COUNT_ROLLOVER_INT | ADI_RTC_ALARM_INT) */
+#define RTC_USED_INTERRUPTS (ADI_RTC_COUNT_ROLLOVER_INT)
```
mchindri, line 57: I think this should go in the next commit, an instead of define it here I would use them in the config param from
irq_register_callback

drivers/platform/aducm3029/rtc.c
adrimbarean: Is the `if` not redundant if I have a `switch case` with `default`? This should cover all possible values.

drivers/platform/aducm3029/rtc_extra.h
adrimbarean: I can move it in the next commit, but where should I define it if I also have to use it in `irq_enable`. That is where I enable the specific interrupts I need.

mchindri: you can save it in `aducm_irq_ctrl_desc.conf[ADUCM_RTC_INT_ID]` when registering the callback

drivers/platform/aducm3029/rtc.c
mchindri: I just suggested a shorter alternative for the `switch`. In the `else `of the `if `will be the what is in the `default `now.

drivers/platform/aducm3029/irq_extra.h
```diff
@@ -93,6 +95,17 @@ enum irq_mode {
...
 
+/**
+ * @struct rtc_irq_config
+ * @brief RTC interrupt configuration routine.
+ */
+struct rtc_irq_config {
+	/** RTC driver handler */
+	struct rtc_desc *rtc_handler;
+	/** Active interrupts OR'ed together */
+	uint32_t active_interrupts;
```
mchindri, line 106: I think you should define what values can be used for this parameter: 
Like 
```
#define RTC_ALARM                ADI_RTC_ALARM_INT
#define RTC_MOD60ALM        ADI_RTC_MOD60ALM_INT
#define RTC_ISO_DONE_INT    ADI_RTC_ISO_DONE_INT
```
Defining just what can be used with this driver

adrimbarean: Would this add anything or just be a translation from the DFP driver to this one?

mchindri: In order to not include the DFP from the appilcation and include only irq_extra.h

mchindri: In order to use no-os or dfp driver, not both

adrimbarean:
V2:
 - replace `switch case` with `if` whe setting RTC frequency;
 - add the RTC used interrupts in the callback config to make them user accessible;
 - add RTC supported interrupts as defines in the irq_extra.h header.

PR #538, opened 2020-03-31
GPIO updates

amiclaus:
- add `gpio_get_optional` function
- ad9528: update gpio reset handling

### DIFF
--- a/drivers/frequency/ad9528/ad9528.c
+++ b/drivers/frequency/ad9528/ad9528.c
@@ -380,12 +380,12 @@ int32_t ad9528_setup(struct ad9528_dev **device,
 		return ret;
 
 	/* GPIO */
-	if(init_param.hw_reset_en) {
-		ret = gpio_get(&dev->gpio_resetb, &init_param.gpio_resetb);
-		if (ret < 0)
-			return ret;
-		ad9528_reset(dev);
-	}
+	ret = gpio_get_optional(&dev->gpio_resetb, init_param.gpio_resetb);
+	if (ret < 0)
+		return ret;
+
+	ad9528_reset(dev);
+
 	ret = ad9528_spi_write_n(dev,
 				 AD9528_SERIAL_PORT_CONFIG,
 				 AD9528_SER_CONF_SOFT_RESET |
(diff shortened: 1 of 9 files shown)


drivers/frequency/ad9528/ad9528.c
```diff
@@ -380,12 +380,13 @@ int32_t ad9528_setup(struct ad9528_dev **device,
...
-	if(init_param.hw_reset_en) {
-		ret = gpio_get(&dev->gpio_resetb, &init_param.gpio_resetb);
-		if (ret < 0)
-			return ret;
+	ret = gpio_get_optional(&dev->gpio_resetb, init_param.gpio_resetb);
+	if (ret < 0)
+		return ret;
+
+	if (dev->gpio_resetb)
 		ad9528_reset(dev);
```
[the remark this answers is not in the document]
amiclaus: I think it is done in the next line:
https://github.com/analogdevicesinc/no-OS/blob/37df58822266ea2aea17e7477ab1efb9bb984a69/drivers/frequency/ad9528/ad9528.c#L390-L394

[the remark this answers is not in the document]
amiclaus: But indeed, it is done also in the `ad9528_reset` function.

drivers/frequency/ad9528/ad9528.c
```diff
@@ -380,12 +380,13 @@ int32_t ad9528_setup(struct ad9528_dev **device,
...
-	}
+
 	ret = ad9528_spi_write_n(dev,
```
buha, line 390: I like the new gpio_get_optional !
However, this soft reset (command over SPI) should be done only if no HW reset was done. It doesn't harm to also do it when a hard reset was done, but it doesn't make sense. So I simply suggest to move it in an else branch of the if (dev->gpio_resetb) condition. ;)

drivers/frequency/ad9528/ad9528.c
```diff
@@ -380,12 +380,13 @@ int32_t ad9528_setup(struct ad9528_dev **device,
 		return ret;
 
 	/* GPIO */
-	if(init_param.hw_reset_en) {
-		ret = gpio_get(&dev->gpio_resetb, &init_param.gpio_resetb);
-		if (ret < 0)
-			return ret;
+	ret = gpio_get_optional(&dev->gpio_resetb, init_param.gpio_resetb);
+	if (ret < 0)
```
adrimbarean, line 384: Should the driver use the error.h definitions?

amiclaus:
V2:
- ad9528: move soft reset from `ad9528_setup` to `ad9528_reset`

drivers/frequency/ad9528/ad9528.c
dbogdan: Anyhow, I guess this is the subject of a different commit/pull-request.

drivers/frequency/ad9528/ad9528.c
dbogdan: Without some additional testing, I would still keep the software reset too.

PR #540, opened 2020-04-02
Ad713X driver update

adrimbarean:
Update the AD713X driver to new silicon status.


adrimbarean:
V1:
 - remove redundant include file;
 - fix doxygen complaints.

drivers/adc/ad713x/ad713x.c
```diff
@@ -0,0 +1,561 @@
...
+ */
+static int32_t ad713x_init_chan_bw(struct ad713x_dev *dev)
+{
+	int8_t i;
+	int32_t ret;
+
+	for (i = CH3; i >= 0; i--) {
+		ret = ad713x_wideband_bw_sel(dev, i, 0);
+		if (IS_ERR_VALUE(ret))
+			break;
```
mirceacaprioru, line 434: should this be return ret?

drivers/adc/ad713x/ad713x.c
```diff
@@ -0,0 +1,561 @@
...
+	gpio_set_value(dev->gpio_resetn, true);
+
+	mdelay(100);
+
+	dev->dev_id = init_param->dev_id;
+
+	ret = ad713x_spi_reg_read(dev, AD713X_REG_DEVICE_CONFIG, &data);
+	if (IS_ERR_VALUE(ret))
+		goto error_gpio;
+	data |= AD713X_DEV_CONFIG_PWR_MODE(HIGH_POWER);
```
mirceacaprioru, line 501: nitpick: for cosmetic reasons i would add some extra lines between these it makes it easier to follow. The same i would do for other places in the code where things are very tight together and a bit harder to follow where one ends.

drivers/adc/ad713x/ad713x.c
```diff
@@ -0,0 +1,561 @@
...
+	ret = ad713x_mag_phase_clk_delay(dev, init_param->clk_delay_en);
+	if (IS_ERR_VALUE(ret))
+		goto error_gpio;
+	ret = ad713x_init_chan_bw(dev);
+	if (IS_ERR_VALUE(ret))
+		goto error_gpio;
+
+	*device = dev;
+
+	mdelay(1000);
```
mirceacaprioru, line 527: is this delay necessary after setting up the device ? maybe a comment here specifying this ?

drivers/adc/ad713x/ad713x.h
```diff
@@ -0,0 +1,878 @@
...
+
+/******************************************************************************/
+/********************** Macros and Constants Definitions **********************/
+/******************************************************************************/
+/*
+ * Create a contiguous bitmask starting at bit position @l and ending at
+ * position @h.
+ */
+#ifndef GENMASK
+#define GENMASK(h, l) (((~0UL) - (1UL << (l)) + 1) & (~0UL >> (31 - (h))))
```
mirceacaprioru, line 61: these are already define in https://github.com/analogdevicesinc/no-OS/blob/master/include/util.h
should just #include "util.h"

drivers/adc/ad713x/ad713x.h
```diff
@@ -0,0 +1,878 @@
...
+#define AD713X_REG_DVDD1V8_VALUE			0x4E
+#define AD713X_REG_CLKVDD_VALUE				0x4F
+#define AD713X_REG_IOVDD_VALUE				0x50
+#define AD713X_REG_TEMPERATURE_DATA			0x51
+
+/*
+ * AD713X_REG_INTERFACE_CONFIG_A
+ */
+#define AD713X_INT_CONFIG_A_SOFT_RESET_MSK			BIT(7)
+#define AD713X_INT_CONFIG_A_SOFT_RESET_MODE(x)		(((x) & 0x1) << 7)
```
mirceacaprioru, line 153: if a define is limited to a bit i would no do a MSK and MODE just 
<code>
#define AD713X_INT_CONFIG_A_SOFT_RESET			BIT(7)

val |= AD713X_INT_CONFIG_A_SOFT_RESET;
val &= ~AD713X_INT_CONFIG_A_SOFT_RESET;
</code>

adrimbarean:
V2:
 - introduce spaces between code blocks by function;
 - return error instead of breaking when operation fails in loop;
 - remove unnecessary delay;
 - include `util.h` and remove redundant defines;
 - remove mode define for one bit register fields.

drivers/adc/ad713x/ad713x.c
```diff
@@ -258,137 +300,278 @@ int32_t ad713x_mag_phase_clk_delay(ad713x_dev *dev,
...
+	ret = ad713x_init_gpio(dev, init_param);
+	if(IS_ERR_VALUE(ret))
+		goto error_gpio;
 	/*
 	 * Tie this pin to IOVDD for master mode operation,
 	 * tie this pin to IOGND for slave mode operation.
 	 */
-	gpio_set_value(dev->gpio_mode1, GPIO_HIGH);
-	gpio_set_value(dev->gpio_mode1, GPIO_LOW);
+	gpio_direction_output(dev->gpio_mode, init_param->mode_master_nslave);
```
mchindri, line 489: Why don't you put this also in gpio_init?

drivers/adc/ad713x/ad713x.h
```diff
@@ -41,247 +42,292 @@
...
-#define AD713X_REG_CHIP_TYPE			0x03
-#define AD713X_REG_PRODUCT_ID_LSB		0x04
-#define AD713X_REG_PRODUCT_ID_MSB		0x05
-#define AD713X_REG_CHIP_GRADE			0x06
-#define AD713X_REG_SCTATCH_PAD			0x0A
-#define AD713X_REG_SPI_REVISION			0x0B
-#define AD713X_REG_VENDOR_ID_LSB		0x0C
-#define AD713X_REG_VENDOR_ID_MSB		0x0D
-#define AD713X_REG_STREAM_MODE			0x0E
+#define AD713X_REG_DEVICE_CONFIG			0x02
```
mchindri, line 62: It looks like there is a problem with indentation you use in your editor size of 4 spaces: 
size =4 -> https://github.com/analogdevicesinc/no-OS/blob/a6c6afb83168a0c32763397b8106376ecad64584/drivers/adc/ad713x/ad713x.h?ts=4
 
size =8 -> https://github.com/analogdevicesinc/no-OS/blob/a6c6afb83168a0c32763397b8106376ecad64584/drivers/adc/ad713x/ad713x.h?ts=8

adrimbarean:
V3:
 - fix astyle complaints.

drivers/adc/ad713x/ad713x.c
adrimbarean: Good point. I will make this change.

adrimbarean:
V4:
 - move all GPIO initialization routines in one place;
 - adjust define indentations for 8-space tab size.

PR #541, opened 2020-04-02
Draft commit about wifi structure

mchindri:
The architecture will be similar to the one used by Mbed.
image
image
https://os.mbed.com/docs/mbed-os/v5.15/reference/ip-networking.html
https://github.com/ARMmbed/mbed-os/tree/master/features/netsocket

Doing it this way, we can add later more connectivity support (Like Ethernet, UDP socket ...)

### DIFF
--- a/drivers/wifi/example.c
+++ b/drivers/wifi/example.c
@@ -1,0 +1,31 @@
+#include "wifi.h"
+#include "tcp_socket.h"
+
+int example_main() {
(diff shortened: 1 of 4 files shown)


mchindri:
I put it here for you to take and look and suggest changes and in the next day I will submit the implementation

mirceacaprioru:
I like the structure of the pack. It will be a lot easier to add functionalities.

PR #543, opened 2020-04-02
jesd204

dbogdan:
Bring the axi_jesd204_[rx|tx] no-OS drivers in sync with the Linux versions:
https://github.com/analogdevicesinc/linux/blob/master/drivers/iio/jesd204/axi_jesd204_rx.c
https://github.com/analogdevicesinc/linux/blob/master/drivers/iio/jesd204/axi_jesd204_tx.c

### DIFF
--- a/drivers/axi_core/jesd204/axi_jesd204_rx.c
+++ b/drivers/axi_core/jesd204/axi_jesd204_rx.c
@@ -58,6 +58,10 @@
 #define JESD204_RX_REG_SYNTH_NUM_LANES			0x10
 #define JESD204_RX_REG_SYNTH_DATA_PATH_WIDTH	0x14
 
+#define JESD204_RX_REG_SYNTH_REG_1		0x18
+#define JESD204_RX_ENCODER_MASK			GENMASK(9, 8)
+#define JESD204_RX_ENCODER_GET(x)	field_get(JESD204_RX_ENCODER_MASK, x)
+
 #define JESD204_RX_REG_LINK_DISABLE		0xc0
 #define JESD204_RX_REG_LINK_STATE		0xc4
 #define JESD204_RX_REG_LINK_CLK_RATIO	0xc8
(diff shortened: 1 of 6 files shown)


dbogdan:
v2:
 - Fixed Coday complaint.

dbogdan:
v3:
 - align "SYNC~: asserted" message.

dbogdan:
v4:
 - Remove unimplemented axi_jesd204_rx_lane_clk_set_rate().

drivers/axi_core/jesd204/axi_jesd204_rx.c
```diff
@@ -58,6 +58,10 @@
 #define JESD204_RX_REG_SYNTH_NUM_LANES			0x10
 #define JESD204_RX_REG_SYNTH_DATA_PATH_WIDTH	0x14
 
+#define JESD204_RX_REG_SYNTH_REG_1		0x18
+#define JESD204_RX_ENCODER_MASK			GENMASK(9, 8)
+#define JESD204_RX_ENCODER_GET(x)	field_get(JESD204_RX_ENCODER_MASK, x)
```
mirceacaprioru, line 63: if you have this utilities as functions should we just use them as they are ? I don't particularly like masking functions with macros.

mirceacaprioru:
looks ok overall, just one comment inline.

drivers/axi_core/jesd204/axi_jesd204_rx.c
dbogdan: This seems to be a common approach in Linux kernel - however, it's true that most of those functions are GCC built-in ones.
In this particular case, I just want to use JESD204_RX_ENCODER_GET() the same way as it is used in the Linux kernel.

drivers/axi_core/jesd204/axi_jesd204_rx.c
```diff
@@ -311,6 +350,48 @@ int32_t axi_jesd204_rx_laneinfo_read(struct axi_jesd204_rx *jesd, uint32_t lane)
...
+ */
+static int32_t axi_jesd204_rx_laneinfo_64b66b_read(struct axi_jesd204_rx *jesd,
+		uint32_t lane,
+		uint32_t lane_status)
+{
+	uint8_t extend_multiblock;
+
+	extend_multiblock = JESD204_EMB_STATE_GET(lane_status);
+
+	printf("\tState of Extended multiblock alignment: %s\n",
```
adrimbarean, line 364: Is it OK for drivers to try to print to the output stream?

dbogdan: On the platforms where JESD is used, printf is redirected to UART.

PR #549, opened 2020-04-07
Wifi pack with TCP connection

mchindri:
This is the implementation for the proposed interface at https://github.com/analogdevicesinc/no-OS/pull/541 . Only little changes have been made.

This is a visual overview of the pack:
image

### DIFF
--- a/network/data_link/at_params.h
+++ b/network/data_link/at_params.h
@@ -1,0 +1,265 @@
+/***************************************************************************//**
+ *   @file   at_params.h
+ *   @brief  Structures used by at_parser
+ *   @author Mihail Chindris (mihail.chindris@analog.com)
+********************************************************************************
+ * Copyright 2020(c) Analog Devices, Inc.
+ *
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *  - Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ *  - Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in
+ *    the documentation and/or other materials provided with the
+ *    distribution.
+ *  - Neither the name of Analog Devices, Inc. nor the names of its
+ *    contributors may be used to endorse or promote products derived
+ *    from this software without specific prior written permission.
+ *  - The use of this software may or may not infringe the patent rights
+ *    of one or more patent holders.  This license does not release you
+ *    from the requirement that you obtain separate licenses from these
+ *    patent holders to use this software.
+ *  - Use of the software either in source or binary form, must be run
+ *    on or directly connected to an Analog Devices Inc. component.
+ *
+ * THIS SOFTWARE IS PROVIDED BY ANALOG DEVICES "AS IS" AND ANY EXPRESS OR
+ * IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, NON-INFRINGEMENT,
+ * MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED.
+ * IN NO EVENT SHALL ANALOG DEVICES BE LIABLE FOR ANY DIRECT, INDIRECT,
+ * INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, INTELLECTUAL PROPERTY RIGHTS, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+*******************************************************************************/
+
+#ifndef AT_PARAMS_H
+#define AT_PARAMS_H
+
+/******************************************************************************/
+/*************************** Types Declarations *******************************/
+/******************************************************************************/
+
+/**
+ * @struct at_buff
+ * @brief Structure used in the at_parser to work with buffers
+ *
+ * It can be initialized from a null terminated string using \ref str_to_at
+ * and vice versa with \ref at_to_str .
+ */
+struct at_buff {
+	/** Address of the buffer */
+	uint8_t		*buff;
+	/** Size of data written in buffer */
+	uint32_t	len;
+};
+
+/**
+ * @enum encryption_type
+ * @brief Type of wifi security
+ */
+enum encryption_type {
+	/** Open */
+	OPEN		= 0u,
+	/** WEP */
+	WEP		= 1u,
+	/** WPA_PSK */
+	WPA_PSK		= 2u,
+	/** WPA2_PSK */
+	WPA2_PSK	= 3u,
+	/** WPA_WPA2_PSK */
+	WPA_WPA2_PSK	= 4u
+};
+
+/**
+ * @enum socket_type
+ * @brief Type of a connection
+ */
+enum socket_type {
+	/** TCP Connection */
+	SOCKET_TCP,
+	/** UDP Connection */
+	SOCKET_UDP
+};
+
+/**
+ * @enum cwmod_param
+ * @brief Type of operation for module
+ */
+enum cwmod_param {
+	/** Operate as client */
+	CLIENT = 1,
+	/** Operate as access point */
+	ACCESS_POINT,
+	/** Operate as both */
+	CLIENT_AND_ACCESS_POINT
+};
+
+/**
+ * @struct cwjap_param
+ * @brief Parameter used to connect to a wifi network
+ */
+struct cwjap_param {
+	/** Ssid */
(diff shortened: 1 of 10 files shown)


mchindri:
v2:
 - Fix checks

mchindri:
v2 bis:
 - Fix checks (without example.c which will be removed )

drivers/wifi/at_parser.c
```diff
@@ -0,0 +1,1181 @@
...
+
+	/* State of the callback */
+	volatile enum {
+		/* Normal mode. Read each char and interpret the result */
+		READING_RESPONSES,
+		/* When an +IPD is received, callback enter in this mode */
+		READING_PAYLOAD
+	}			callback_operation;
+	/* Used for AT_SEND */
+	volatile bool		waiting_send;
```
buha, line 190: I don't think it's a good idea to have a combination of a state machine and some other variables around it that are also some soft of state machines in themselves.
Consider creating more states in `callback_operation` in such a way as to eliminate the need for the `waiting_send` boolean.
This way there is only 1 clear state machine, easier to understand, code, debug.

drivers/wifi/at_parser.c
```diff
@@ -0,0 +1,1181 @@
...
+static int32_t send_cmd(struct at_desc *desc, enum at_cmd cmd,
+			union in_param *in_param)
+{
+	uart_write(desc->uart_desc, desc->cmd.buff, desc->cmd.len);
+	if (cmd == AT_SEND) {
+		/* Waiting for ok */
+		if (SUCCESS != wait_for_response(desc))
+			return FAILURE;
+		/* Wait unitil '>' is received */
+		while (desc->waiting_send)
```
buha, line 503: Hmmm, this could be a locking point.

drivers/wifi/tcp_socket.c
```diff
@@ -0,0 +1,181 @@
...
+int32_t socket_remove(struct tcp_socket_desc *desc)
+{
+	int32_t ret;
+
+	if (!desc)
+		return FAILURE;
+
+	ret = desc->net->socket_close(desc->net->net, desc->id);
+	if (IS_ERR_VALUE(ret))
+		return FAILURE;
```
buha, line 115: We don't have good status/error code support in no-OS (yet), but I think for the future it would be better in these situations to forward `ret` back to caller. Sometimes it is indeed needed to return a specific error code, but many times, including here, it's ok to just pass `ret`. And When we will have more than SUCCESS and FAILURE you won't need to update the code here.

drivers/wifi/at_parser.c
```diff
@@ -0,0 +1,1181 @@
...
+	{{(uint8_t *)"+CIFSR", 6}, AT_EXECUTE_OP},
+	{{(uint8_t *)"+CIPMUX", 7}, AT_QUERY_OP | AT_SET_OP},
+	{{(uint8_t *)"+CIPSERVER", 10}, AT_SET_OP},
+	{{(uint8_t *)"+CIPMODE", 8}, AT_QUERY_OP | AT_SET_OP},
+	{{(uint8_t *)"+CIPSTO", 7}, AT_QUERY_OP | AT_SET_OP},
+	{{(uint8_t *)"+PING", 5}, AT_SET_OP}
+};
+
+/* Structure storing a circular buffer for a connection */
+struct at_circular_buff {
```
cristiantpop, line 111: We already have a circular buffer implementation in util folder. Why not use that?

drivers/wifi/at_parser.c
```diff
@@ -0,0 +1,1181 @@
...
+				/* Return the index of the match message */
+				return count;
+		} else if (ch == msg->buff[0]) {
+			*idx = 1;
+		} else {
+			*idx = 0;
+		}
+	}
+
+	return -1;
```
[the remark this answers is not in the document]
mchindri: I removed the if that returns FAILURE, that was used for debugging only.

drivers/wifi/at_parser.c
```diff
@@ -0,0 +1,1181 @@
...
+	if (desc->ipd_idx < at_ipd.len) {
+		match_messages(desc, &at_ipd, &desc->ipd_idx, 1, ch);
+		return false;
+	}
+	/* Update ipd_idx until ':' is found */
+	if (ch == ':')
+		goto process_len;
+	else if ((!(ch >= '0' && ch <= '9') && ch != ',') ||
+		 desc->ipd_idx > max_ch_search)
+		goto reset_idx;
```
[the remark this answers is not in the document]
mchindri: Done

drivers/wifi/at_parser.c
mchindri: added timeout

drivers/wifi/tcp_socket.c
mchindri: returned ret

mchindri:
v3:
 - remove extra new lines
 - move waiting_send and is_reset variables in the state machine
 - wrap sscanf instead of implementing it
 - refactor is_payload_message
 - add timeout when waiting for WAITING_SEND
 - return the result from the of low level calls if they failed

drivers/wifi/at_parser.c
mchindri: I didn't find a way to use it without many modification to the cb from util.
Here the buffer is used to be written with DMA and I can't tell the DMA to write to an other address when the buffer is full, using the functions from circular_buffer.h

mchindri:
v3:
- remove extra new lines
- move waiting_send and is_reset variables in the state machine
- wrap sscanf instead of implementing it
- refactor is_payload_message
- add timeout when waiting for WAITING_SEND
- return the result from the of low level calls if they failed

mchindri:
v4:
 - Small fix discovered during MQTT developement
 - Move files in the decided folder structure
 - Add network.h in includes in the first commit

network/network_interface/network_interface.h
```diff
@@ -0,0 +1,181 @@
...
+	 * @param data - Destination buffer for received data
+	 * @param size - Maximum data to read
+	 * @param from - Destination for the source address or NULL
+	 * @return
+	 *  - Number of sent bytes : On success
+	 *  - \ref FAILURE : Otherwise
+	 */
+	int32_t (*socket_recvfrom)(void *net, uint32_t sock_id,
+				   void *data, uint32_t size,
+				   struct socket_address *from);
```
adrimbarean, line 178: These can be named `socket_sendtcp`, `socket_recvtcp`, `spcket_sendudp` and `socket_recvudp`. This could improve clarity.

network/data_link/at_params.h
```diff
@@ -0,0 +1,265 @@
...
+	/** ssid*/
+	struct at_buff		ssid;
+	/** mac */
+	struct at_buff		mac;
+	/** channel */
+	uint8_t			channel;
+};
+
+#if defined(PARSE_RESULT)
+/* Search networks. Not used in driver*/
```
adrimbarean, line 221: What is the purpose of the things not used? Do they need to be implemented?

network/data_link/at_parser.c
```diff
@@ -0,0 +1,1158 @@
...
+	/* Size of the buffer */
+	uint32_t	size;
+	/* Address to the buffer */
+	uint8_t		*buff;
+	/* Read index */
+	uint32_t	ri;
+	/* Write index */
+	uint32_t	wi;
+	/* Write index */
+	uint32_t	count;
```
adrimbarean, line 121: Maybe the comments can be more specific how the `wi` and `count` differ fr each other, or why do we need them both.

network/data_link/at_parser.c
```diff
@@ -0,0 +1,1158 @@
...
+	uint32_t				to_read;
+	/* True if connection established */
+	bool					active;
+	/* Type of connection */
+	enum socket_type			type;
+};
+
+/* Structure storing the status of the parser */
+struct at_desc {
+	// Uart related fields
```
adrimbarean, line 144: nitpick: I'd say to keep consistency and only use C style comments.

network/data_link/at_parser.c
```diff
@@ -0,0 +1,1158 @@
...
+ * 		if (index == strlen(message))
+ * 			return message_by_id
+ * 		}
+ * 	else
+ * 		index = 0;
+ *
+ * 	return -1;
+ * }
+ */
+static inline int32_t match_messages(struct at_desc *desc,
```
adrimbarean, line 222: Is the `desc` parameter necessary?

network/network_interface/network_interface.h
mchindri: I tried to keep the same naming as in <sys/socket.h> https://pubs.opengroup.org/onlinepubs/007908799/xns/syssocket.h.html
I am thinking to make them `send `, `recv `, `sendto `and `recvfrom`

network/data_link/at_params.h
mchindri: Nop, I left them for an eventual further developement. Maybe I will need for example to get the status of a connection.

network/data_link/at_parser.c
mchindri: Copy paste error :) will fix

network/data_link/at_parser.c
mchindri: will remove

network/data_link/at_parser.c
mchindri: fixed

network/data_link/at_parser.c
mchindri: no, removed.

network/data_link/at_parser.c
mchindri: I tried to use it this way as a section title. Changed.

mchindri:
v5:
- fix doxygen
- fix comment for at_circular_buff.count
- use c style comments in at_desc
- remove desc param from match_message function

network/data_link/at_params.h
mchindri: Or for scan wifi network if it will be needed

PR #557, opened 2020-04-10
spi generic layer + adc1 example

amiclaus:
This a proposal for the spi generic layer implementation.

And example of usage is provided using the fmcjesdadc1 project with the
Xilinx platform specific driver.

An additional layer will be added afterwards between spi generic layer
and xilinx spi layer for the CPLD demux spi driver.

### DIFF
--- a/drivers/spi/spi.c
+++ b/drivers/spi/spi.c
@@ -1,0 +1,41 @@
+/*
+ * spi.c
+ *
+ *  Created on: Apr 10, 2020
+ *      Author: amiclaus
+ */
+#include <inttypes.h>
+#include "spi.h"
+#include <stdlib.h>
+#include "error.h"
+
+int32_t spi_init(struct spi_desc **desc,
+		 const struct spi_init_param *param){
+	(*desc)->dev = calloc(1, sizeof(struct spi_dev));
+	if(!((*desc)->dev))
+		return FAILURE;
+
+	(*desc)->dev->spi_dev_init = param->dev.spi_dev_init;
+	(*desc)->dev->spi_dev_remove = param->dev.spi_dev_remove;
+	(*desc)->dev->spi_dev_write_and_read = param->dev.spi_dev_write_and_read;
+
+	return ((*desc)->dev->spi_dev_init(desc, param));
+}
+
+
+/* Free the resources allocated by spi_init(). */
+int32_t spi_remove(struct spi_desc *desc){
+	if(desc->dev->spi_dev_remove(desc))
+		return FAILURE;
+
+	free(desc->dev);
+
+	return SUCCESS;
+}
+
+/* Write and read data to/from SPI. */
+int32_t spi_write_and_read(struct spi_desc *desc,
+			   uint8_t *data,
+			   uint16_t bytes_number){
+	return (desc->dev->spi_dev_write_and_read(desc, data, bytes_number));
+}
--- a/include/spi.h
+++ b/include/spi.h
@@ -72,6 +72,11 @@ typedef enum spi_mode {
 	SPI_MODE_3 = (SPI_CPOL | SPI_CPHA)
 } spi_mode;
 
+struct spi_dev {
+	int32_t (*spi_dev_init)();
+	int32_t (*spi_dev_write_and_read)();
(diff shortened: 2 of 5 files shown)


amiclaus:
Appropiate indentation and comments will be added when the pull request is made.

mchindri:
[spi_proposal.txt](https://github.com/analogdevicesinc/no-OS/files/4461942/spi_proposal.txt)

mchindri:
```
//spi.h
struct spi_interface {
	int32_t (*spi_init)(struct spi_desc **, const struct spi_init_param *);
	int32_t (*spi_write_and_read)(struct spi_desc *, uint8_t *, uint16_t);
	int32_t (*spi_remove)(struct spi_desc *);
}

struct spi_init_param {
	struct spi_interface *interface;
	//...
}

struct spi_desc {
	struct spi_interface *interface;
	//...
}

//xilinx/spi_extra.h
int32_t xil_configure_interface(struct spi_interface *interface);

//xilinx/spi.c
int32_t xil_configure_interface(struct spi_interface *interface) {
	interface->spi_init = xil_spi_init;
	interface->spi_write_and_read = xil_spi_write_and_read;
	interface->spi_remove = xili_spi_remove;
}

//spi.c
int32_t spi_init(struct spi_desc **desc,
		 const struct spi_init_param *param){

	int32_t ret;

	ret = param->interface->spi_init(desc, param);
	if (IS_ERR_VALUE(ret))
		return (ret);
	//Here de descriptor is allocated by the driver
	desc->interface = param->interface;

	return SUCCESS;
}

//fmcjesdadc1.c
struct spi_desc *ad9250;
struct spi_init_param ad9250_spi_param = {
	//...
}
xil_configure_interface(&ad9250_spi_param.spi_interface);
spi_init(&ad9250, &ad9250_spi_param);

```

dbogdan:
I like the idea of having an interface, but having to call an additional function (configure_interface()) is not ideal... Maybe each driver can directly expose the interface to be used by spi_init().

include/spi.h
```diff
@@ -72,6 +72,11 @@ typedef enum spi_mode {
 	SPI_MODE_3 = (SPI_CPOL | SPI_CPHA)
 } spi_mode;
 
+struct spi_dev {
```
scuciurean, line 75: Maybe find a better fitting naming? Like spi_ops or spi_actions.

mchindri:
@dbogdan You mean an exposed interface like this? It seems better to me too.
`
//xilinx/spi_extra.h
constr struct spi_interface xil_spi_interface = {
	.spi_init = xil_spi_init,
	.spi_write_and_read = xil_spi_init,
	.spi_remove = xili_spi_remove
}
`

dbogdan:
@mchindri , yes, something like that.

drivers/spi/spi.c
```diff
@@ -0,0 +1,41 @@
...
+	free(desc->dev);
+
+	return SUCCESS;
+}
+
+/* Write and read data to/from SPI. */
+int32_t spi_write_and_read(struct spi_desc *desc,
+			   uint8_t *data,
+			   uint16_t bytes_number){
+	return (desc->dev->spi_dev_write_and_read(desc, data, bytes_number));
```
buha, line 40: no-OS is a library, test for NULL function pointer

buha: or maybe even better, return some failure code in `spi_init`, the driver doesn't make sense to lack any of these pointers (maybe some drivers could lack the `remove`)

this way you dont need to test for NULL here

adrimbarean:
How many projects do have this problem? I could argue that the SPI driver and the device drivers are fine as it is, but it's actually the applications responsibility here to provide correct SPI implementation that will replace the one in the driver. I don't think we need to change too much xilinx spi driver. Just enough to make it accept custom implementation if necessary.

dbogdan:
Besides the SPI expander used on ADC1, all the SPI engine projects are affected by this limitation. What we did until now (considering the SPI engine as a Xilinx SPI type) was only a workaround. SPI engine should be treated as an independent SPI controller since it's available on Intel platforms too.

amiclaus:
Closing this draft.
PR added with #577

PR #564, opened 2020-04-15
Network interface and tcp socket implementation

mchindri:
Splitted from Splitted from https://github.com/analogdevicesinc/no-OS/pull/549

### DIFF
--- a/include/tcp_socket.h
+++ b/include/tcp_socket.h
@@ -1,0 +1,94 @@
+/***************************************************************************//**
+ *   @file   tcp_socket.h
+ *   @brief  TCP Socket functions
+ *   @author Mihail Chindris (mihail.chindris@analog.com)
+********************************************************************************
+ *   @copyright
+ * Copyright 2020(c) Analog Devices, Inc.
+ *
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *  - Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ *  - Redistributions in binary form must reproduce the above copyright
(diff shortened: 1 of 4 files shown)


include/tcp_socket.h
```diff
@@ -0,0 +1,94 @@
+/***************************************************************************//**
```
adrimbarean, line 1: A short description would help in the commit body. For example what do the files added do and how do they relate to each other. Maybe how they help in building a WiFi connection.

network/transport/tcp_socket.c
```diff
@@ -0,0 +1,158 @@
+/***************************************************************************//**
```
adrimbarean, line 1: I think we should almost never have a commit body empty. Just a short meta description of the file like what it does and why it needs to do it.

network/transport/tcp_socket.c
```diff
@@ -0,0 +1,158 @@
...
+#include <stdlib.h>
+#include "error.h"
+#include "tcp_socket.h"
+
+/******************************************************************************/
+/*************************** Types Declarations *******************************/
+/******************************************************************************/
+
+/* Socket descriptor */
+struct tcp_socket_desc {
```
adrimbarean, line 54: Why is the `tcp_socket_desc` defined here and not in the header?

mchindri: Because the application shouldn't access it's fields. It is enough to have a reference to it.
That's way in the header is declared as an abstract structure:
`struct tcp_socket_desc;`

include/tcp_socket.h
mchindri: done

network/transport/tcp_socket.c
mchindri: done

mchindri:
v2:
 - add description in the commit messages

PR #567, opened 2020-04-16
util: Add find_closest() and field_prep()

dbogdan:
find_closest() - Locate the closest element in an array.
field_prep() - Shift the value and apply the specified mask.

### DIFF
--- a/util/util.c
+++ b/util/util.c
@@ -84,6 +84,35 @@ uint32_t find_last_set_bit(uint32_t word)
 }
 
 /**
+ * Locate the closest element in an array.
+ */
+uint32_t find_closest(int32_t val,
+		      const int32_t *array,
+		      uint32_t size)
+{
+	int32_t diff = abs(array[0] - val);
+	uint32_t ret = 0;
+	uint32_t i;
+
+	for (i = 1; i < size; i++) {
+		if (abs(array[i] - val) < diff) {
+			diff = abs(array[i] - val);
+			ret = i;
+		}
+	}
+
+	return ret;
+}
+
+/**
+ * Shift the value and apply the specified mask.
+ */
+uint32_t field_prep(uint32_t mask, uint32_t val)
+{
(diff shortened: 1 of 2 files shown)


util/util.c
```diff
@@ -83,6 +83,35 @@ uint32_t find_last_set_bit(uint32_t word)
...
+		}
+	}
+
+	return ret;
+}
+
+/**
+ * Shift the value and apply the specified mask.
+ */
+uint32_t field_prep(uint32_t mask, uint32_t val)
```
mchindri, line 110: For performance, I think it will be better if this function will be a define or at least inline if type check is needed

util/util.c
```diff
@@ -83,6 +83,35 @@ uint32_t find_last_set_bit(uint32_t word)
...
 }
 
+/**
+ * Locate the closest element in an array.
+ */
+uint32_t find_closest(int32_t val,
+		      const int32_t *array,
+		      uint32_t size)
+{
+	uint32_t diff = abs((abs(array[0]) - abs(val)));
```
mchindri, line 93: Is the abs needed for each element?
If you have array[0] = 1 and val = -1 => diff will be 0
I think 
```suggestion
	uint32_t diff = abs(array[0] - val));
```
is better. Same for below

util/util.c
buha: @dbogdan Do we really need it as a function or as a macro ? I can see the register bit definitions using an equivalent macro FIELD_PREP like on linux, but what is the usecase where we'd need a function ?

util/util.c
dbogdan: Thanks, if fixed that.

dbogdan:
v2:
 - Fixed find_closest().

util/util.c
dbogdan: For consistency with the rest of the file, I would keep it like this. Another commit/pull-request can implement these improvements for all the functionts.

PR #569, opened 2020-04-22
drivers:adc:ad713x make GPIOs optional

adrimbarean:
In certain applications the GPIOs are not going to be software controlled.
Make the GPIO usage optional for this use case.


adrimbarean:
Hello. I am also changing the header file so that the initialization structure contains pointers to GPIO, rather than actual structures: https://github.com/analogdevicesinc/no-OS/pull/569/files#diff-ef09541308853347b68dbc19ee50e052R705

This way, if you do not want to initialize a GPIO just give it NULL at the pointer to the initialization structure.

mphalke:
You are correct, sorry I missed those changes from the header file. 
Thanks a lot.

PR #570, opened 2020-04-23
Add HMC7043 support

dbogdan:
Corresponding Linux version: https://github.com/analogdevicesinc/linux/blob/master/drivers/iio/frequency/hmc7044.c

### DIFF
--- a/drivers/frequency/hmc7044/hmc7044.c
+++ b/drivers/frequency/hmc7044/hmc7044.c
@@ -506,7 +506,7 @@ static int32_t hmc7044_setup(struct hmc7044_dev *dev)
 
 	/* Select the VCO range */
 	hmc7044_write(dev, HMC7044_REG_EN_CTRL_0,
-		      HMC7044_RF_RESEEDER_EN |
+		      (dev->rf_reseeder_en ? HMC7044_RF_RESEEDER_EN : 0) |
 		      HMC7044_VCO_SEL(high_vco_en ?
 				      HMC7044_VCO_HIGH :
 				      HMC7044_VCO_LOW) |
@@ -685,6 +685,7 @@ int32_t hmc7044_init(struct hmc7044_dev **device,
 		init_param->high_performance_mode_clock_dist_en;
 	dev->high_performance_mode_pll_vco_en =
 		init_param->high_performance_mode_pll_vco_en;
+	dev->rf_reseeder_en = !init_param->rf_reseeder_disable;
 	dev->sync_pin_mode = init_param->sync_pin_mode;
 	dev->pulse_gen_mode = init_param->pulse_gen_mode;
 
--- a/drivers/frequency/hmc7044/hmc7044.c
+++ b/drivers/frequency/hmc7044/hmc7044.c
@@ -1,9 +1,9 @@
 /***************************************************************************//**
  *   @file   hmc7044.c
- *   @brief  Implementation of HMC7044 Driver.
+ *   @brief  Implementation of HMC7044, HMC7043 Driver.
  *   @author DBogdan (dragos.bogdan@analog.com)
 ********************************************************************************
- * Copyright 2018(c) Analog Devices, Inc.
+ * Copyright 2018-2020(c) Analog Devices, Inc.
  *
  * All rights reserved.
  *
@@ -41,6 +41,7 @@
 /***************************** Include Files **********************************/
 /******************************************************************************/
 #include <stdlib.h>
+#include <stdio.h>
 #include "error.h"
 #include "util.h"
 #include "hmc7044.h"
@@ -167,6 +168,10 @@
 
 #define HMC7044_REG_SYSREF_TIMER_MSB	0x005D
 #define HMC7044_SYSREF_TIMER_MSB(x)	(((x) & 0xf00) >> 8)
+
+#define HMC7044_CLK_INPUT_CTRL		0x0064
+#define HMC7044_LOW_FREQ_INPUT_MODE	BIT(0)
+#define HMC7044_DIV_2_INPUT_MODE	BIT(1)
 
 /* Status and Alarm readback */
 #define HMC7044_REG_ALARM_READBACK	0x007D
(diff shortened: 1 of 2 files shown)


drivers/frequency/hmc7044/hmc7044.h
```diff
@@ -103,6 +104,7 @@ struct hmc7044_init_param {
 	bool		clkin1_vcoin_en;
 	bool		high_performance_mode_clock_dist_en;
 	bool		high_performance_mode_pll_vco_en;
+	bool		rf_reseeder_disable;
```
mchindri, line 109: Shoulden't be named the same? both disable or enable?

drivers/frequency/hmc7044/hmc7044.c
```diff
@@ -640,6 +645,122 @@ static int32_t hmc7044_setup(struct hmc7044_dev *dev)
...
+	struct hmc7044_chan_spec *chan;
+	uint32_t i;
+
+	if (dev->clkin_freq_ccf[0])
+		dev->pll2_freq = dev->clkin_freq_ccf[0];
+	else
+		dev->pll2_freq  = dev->clkin_freq[0];
+
+	if (!dev->pll2_freq) {
+		printf("%s: Failed to get valid parent rate\n", __func__);
```
mchindri, line 664: Is it ok to have printf in drivers?

drivers/frequency/hmc7044/hmc7044.c
```diff
@@ -640,6 +645,122 @@ static int32_t hmc7044_setup(struct hmc7044_dev *dev)
...
+
+	/* Load the configuration updates (provided by Analog Devices) */
+	hmc7044_write(dev, HMC7044_REG_CLK_OUT_DRV_LOW_PW, 0x4d);
+	hmc7044_write(dev, HMC7044_REG_CLK_OUT_DRV_HIGH_PW, 0xdf);
+
+	/* Disable all channels */
+	for (i = 0; i < HMC7044_NUM_CHAN; i++)
+		hmc7044_write(dev, HMC7044_REG_CH_OUT_CRTL_0(i), 0);
+
+	if (dev->pll2_freq < 1000000000U)
```
mchindri, line 682: could 1000000000U be a define value?

drivers/frequency/hmc7044/hmc7044.c
```diff
@@ -640,6 +645,122 @@ static int32_t hmc7044_setup(struct hmc7044_dev *dev)
...
+		hmc7044_write(dev, HMC7044_REG_CH_OUT_CRTL_8(chan->num),
+			      HMC7044_DRIVER_MODE(chan->driver_mode) |
+			      HMC7044_DRIVER_Z_MODE(chan->driver_impedance) |
+			      (chan->dynamic_driver_enable ?
+			       HMC7044_DYN_DRIVER_EN : 0) |
+			      (chan->force_mute_enable ?
+			       HMC7044_FORCE_MUTE_EN : 0));
+
+		hmc7044_write(dev, HMC7044_REG_CH_OUT_CRTL_3(chan->num),
+			      chan->fine_delay & 0x1F);
```
mchindri, line 735: 0x1f can be defined as mask

mchindri: same for 0x3 and BIT(4) with a define

drivers/frequency/hmc7044/hmc7044.c
scuciurean: Usualy it helps to get some feedback in the setup phase.

drivers/frequency/hmc7044/hmc7044.h
```diff
@@ -92,6 +93,7 @@ struct hmc7044_dev {
 
 struct hmc7044_init_param {
 	spi_init_param	*spi_init;
+	bool		is_hmc7043;
```
scuciurean, line 96: Both 7043 and 7044 got device id registers (0x78 0x79 0x7A). It is the same device id? If not, this member can be dropped and get the device type from the registers

drivers/frequency/hmc7044/hmc7044.h
scuciurean: I think it doesn't matter, only the init_param member is accessible to the user

buha: Functionally it's the same but I agree it should be enable, otherwise it all gets rather confusing.

drivers/frequency/hmc7044/hmc7044.c
```diff
@@ -640,6 +645,122 @@ static int32_t hmc7044_setup(struct hmc7044_dev *dev)
...
+
+	if (dev->pll2_freq < 1000000000U)
+		hmc7044_write(dev, HMC7044_CLK_INPUT_CTRL,
+			      HMC7044_LOW_FREQ_INPUT_MODE);
+
+	hmc7044_write(dev, HMC7044_REG_EN_CTRL_0,
+		      (dev->rf_reseeder_en ? HMC7044_RF_RESEEDER_EN : 0) |
+		      HMC7044_SYSREF_TIMER_EN);
+
+	/* Program the SYSREF timer */
```
adrimbarean, line 690: nitpick: is this comment from this driver? I feel like it's leftover from porting.

drivers/frequency/hmc7044/hmc7044.c
```diff
@@ -640,6 +645,122 @@ static int32_t hmc7044_setup(struct hmc7044_dev *dev)
...
+	else
+		dev->pll2_freq  = dev->clkin_freq[0];
+
+	if (!dev->pll2_freq) {
+		printf("%s: Failed to get valid parent rate\n", __func__);
+		return FAILURE;
+	}
+
+	/* Resets all registers to default values */
+	hmc7044_write(dev, HMC7044_REG_SOFT_RESET, HMC7044_SOFT_RESET);
```
cristiantpop, line 669: Should we check the return value?

dbogdan:
I would keep the no-OS version in sync with the Linux one.

PR #574, opened 2020-04-23
 drivers: frequency: Add ADF4371 support 

dbogdan:
I would keep the no-OS version as close as possible to the Linux one: https://github.com/analogdevicesinc/linux/blob/master/drivers/iio/frequency/adf4371.c

### DIFF
--- a/drivers/frequency/adf4371/adf4371.c
+++ b/drivers/frequency/adf4371/adf4371.c
@@ -1,0 +1,820 @@
+/***************************************************************************//**
+ *   @file   adf4371.c
+ *   @brief  Implementation of ADF4371 Driver.
+ *   @author DBogdan (dragos.bogdan@analog.com)
+********************************************************************************
+ * Copyright 2020(c) Analog Devices, Inc.
+ *
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *  - Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ *  - Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in
+ *    the documentation and/or other materials provided with the
+ *    distribution.
+ *  - Neither the name of Analog Devices, Inc. nor the names of its
+ *    contributors may be used to endorse or promote products derived
+ *    from this software without specific prior written permission.
+ *  - The use of this software may or may not infringe the patent rights
+ *    of one or more patent holders.  This license does not release you
+ *    from the requirement that you obtain separate licenses from these
+ *    patent holders to use this software.
+ *  - Use of the software either in source or binary form, must be run
+ *    on or directly connected to an Analog Devices Inc. component.
+ *
+ * THIS SOFTWARE IS PROVIDED BY ANALOG DEVICES "AS IS" AND ANY EXPRESS OR
+ * IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, NON-INFRINGEMENT,
+ * MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED.
+ * IN NO EVENT SHALL ANALOG DEVICES BE LIABLE FOR ANY DIRECT, INDIRECT,
+ * INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, INTELLECTUAL PROPERTY RIGHTS, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+*******************************************************************************/
+
+/******************************************************************************/
+/***************************** Include Files **********************************/
+/******************************************************************************/
+#include <stdlib.h>
+#include <stdbool.h>
+#include <inttypes.h>
+#include <stdio.h>
+#include "error.h"
+#include "util.h"
+#include "adf4371.h"
+
+/******************************************************************************/
+/********************** Macros and Constants Definitions **********************/
+/******************************************************************************/
+#define ADF4371_WRITE			(0 << 15)
+#define ADF4371_READ			(1 << 15)
+#define ADF4371_ADDR(x)			((x) & 0x7FFF)
+
+/* Registers address macro */
+#define ADF4371_REG(x)			(x)
+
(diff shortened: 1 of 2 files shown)


drivers/frequency/adf4371/adf4371.c
```diff
@@ -0,0 +1,820 @@
...
+	return adf4371_write(dev, ADF4371_REG(0x10), dev->integer & 0xFF);
+}
+
+/**
+ * Power down one channel.
+ * @param dev - The device structure.
+ * @return SUCCESS in case of success, negative error code otherwise.
+ */
+static int32_t adf4371_channel_power_down(struct adf4371_dev *dev,
+		uint32_t channel,
```
buha, line 512: doxygen will complain about 2 undocumented parameters here

drivers/frequency/adf4371/adf4371.c
```diff
@@ -0,0 +1,820 @@
...
+	if (dev->spi_3wire_en)
+		en = false;
+
+	ret = adf4371_update(dev, ADF4371_REG(0x0),
+			     ADF4371_SDO_ACT_MSK | ADF4371_SDO_ACT_R_MSK,
+			     ADF4371_SDO_ACT(en) | ADF4371_SDO_ACT_R(en));
+	if (ret < 0)
+		return ret;
+
+	for (i = 0; i < ARRAY_SIZE(adf4371_reg_defaults); i++) {
```
buha, line 596: I understand the need of adf4371_reg_defaults but I think the original linux driver and this driver might not use it properly. The situation where you'd need it is when you keep a cache of the device settings and sometimes you may need to read values from that cache. At driver initialization you'd need to initialize the cache with settings from adf4371_reg_defaults. But here, there is no cache and adf4371_reg_defaults is only used to write these values once, at initialization to the device. I didn't check the datasheet but I believe these are default register values after reset so this look won't do anything useful. 
It doesnt harm, but it wont help either.

drivers/frequency/adf4371/adf4371.c
```diff
@@ -0,0 +1,820 @@
...
+	dev->differential_ref_clk = init_param->differential_ref_clock;
+
+	if (init_param->loop_filter_inverting)
+		dev->pd_pol = 0;
+	else
+		dev->pd_pol = 1;
+
+	if (init_param->charge_pump_microamp == 0) {
+		/* Default charge pump current is 1.75mA */
+		dev->cp_settings.icp = 1750;
```
buha, line 772: these values look a bit cryptic to me, maybe comment them more to see what they correspond to in datasheet or defined them as macros

drivers/frequency/adf4371/adf4371.c
```diff
@@ -0,0 +1,820 @@
...
+	uint16_t cmd;
+	int ret;
+
+	cmd = ADF4371_READ | ADF4371_ADDR(reg);
+	buf[0] = cmd >> 8;
+	buf[1] = cmd & 0xFF;
+	buf[2] = 0;
+
+	ret = spi_write_and_read(dev->spi_desc, buf, ARRAY_SIZE(buf));
+	if (ret < 0)
```
adrimbarean, line 299: Maybe we should use the `error.h` define to check for error.

drivers/frequency/adf4371/adf4371.c
```diff
@@ -0,0 +1,820 @@
...
+			      uint8_t mask,
+			      uint8_t val)
+{
+	uint8_t read_val;
+	int32_t ret = 0;
+
+	ret = adf4371_read(dev, reg, &read_val);
+	read_val &= ~mask;
+	read_val |= val;
+	ret |= adf4371_write(dev, reg, read_val);
```
adrimbarean, line 326: It would be better for the principle of failing fast to check the status of the function immediately after return. My suggestion for the code would be:
```suggestion
	ret = adf4371_read(dev, reg, &read_val);
	if (IS_ERR_VALUE(ret))
		return FAILURE; //or return ret;
	read_val &= ~mask;
	read_val |= val;
	return adf4371_write(dev, reg, read_val);
```

drivers/frequency/adf4371/adf4371.c
```diff
@@ -0,0 +1,820 @@
...
+ * Calculate closest possible rate
+ * @param dev - The device structure
+ * @param rate - The desired rate.
+ * @param rounded_rate - The closest possible rate of desired rate.
+ * @return SUCCESS in case of success, negative error code otherwise.
+ */
+int32_t adf4371_clk_round_rate(struct adf4371_dev *dev, uint64_t rate,
+			       uint64_t *rounded_rate)
+{
+	*rounded_rate = rate;
```
adrimbarean, line 716: This API is a bit odd and it seems like it does not do much. Maybe this could be implemented by first saving `dev->integer`, `dev->fract1`, `dev->fract2` and `dev->mod2` in temporary variables and then doing: `adf4371_pll_fract_n_compute(rate, dev->fpfd, &dev->integer, &dev->fract1, &dev->fract2, &dev->mod2);
*rounded_rate = adf4371_pll_fract_n_get_rate(dev, channel);`
Followed by restoring the `dev` members to the original levels. But this would mean that the API should depend on the channel in some way.

adrimbarean: My question is: Is this API good as it is or does it need more work?

drivers/frequency/adf4371/adf4371.c
```diff
@@ -0,0 +1,820 @@
...
+
+	for (i = 0; i < init_param->num_channels; i++) {
+		dev->channel_cfg[init_param->channels[i].num].enable = true;
+		dev->channel_cfg[init_param->channels[i].num].freq =
+			init_param->channels[i].power_up_frequency;
+	}
+
+	*device = dev;
+
+	return adf4371_setup(dev);
```
adrimbarean, line 804: Should the device and spi memories be freed in case this fails?

drivers/frequency/adf4371/adf4371.c
```diff
@@ -0,0 +1,820 @@
...
+
+/**
+ * Remove the device - release resources.
+ * @param device - The device structure.
+ * @return SUCCESS in case of success, negative error code otherwise.
+ */
+int32_t adf4371_remove(struct adf4371_dev *device)
+{
+	int32_t ret;
+
```
adrimbarean, line 815: It should be checked if device is NULL or not.

drivers/frequency/adf4371/adf4371.h
```diff
@@ -0,0 +1,125 @@
...
+/******************************************************************************/
+/***************************** Include Files **********************************/
+/******************************************************************************/
+#include <stdint.h>
+#include "spi.h"
+
+/******************************************************************************/
+/********************** Macros and Types Declarations *************************/
+/******************************************************************************/
+struct adf4371_channel_config {
```
adrimbarean, line 51: I think these need Doxygen style comments.

dbogdan:
v2:
 - fix adf4371_channel_power_down() comment.

drivers/frequency/adf4371/adf4371.c
dbogdan: Right, it's better. I've fixed this.

drivers/frequency/adf4371/adf4371.c
dbogdan: I've fixed this too.

drivers/frequency/adf4371/adf4371.c
dbogdan: Right, I've fixed this.

dbogdan:
v3:
 - free resources if adf4371_setup() fails;
 - check if device->spi_desc is available before calling spi_remove().

PR #578, opened 2020-04-29
True Random Number Generator support

mchindri:
Add interface for TRNG and ADUCM implementation. 
This will be needed by TLS socket.

### DIFF
--- a/drivers/platform/aducm3029/trng.c
+++ b/drivers/platform/aducm3029/trng.c
@@ -1,0 +1,175 @@
+/***************************************************************************//**
+ *   @file   trng.c
+ *   @brief  Implementation of true random number generator
+ *   @author Mihail Chindris (mihail.chindris@analog.com)
+********************************************************************************
+ *   @copyright
+ * Copyright 2020(c) Analog Devices, Inc.
+ *
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *  - Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ *  - Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in
+ *    the documentation and/or other materials provided with the
+ *    distribution.
+ *  - Neither the name of Analog Devices, Inc. nor the names of its
+ *    contributors may be used to endorse or promote products derived
+ *    from this software without specific prior written permission.
+ *  - The use of this software may or may not infringe the patent rights
+ *    of one or more patent holders.  This license does not release you
+ *    from the requirement that you obtain separate licenses from these
+ *    patent holders to use this software.
+ *  - Use of the software either in source or binary form, must be run
+ *    on or directly connected to an Analog Devices Inc. component.
+ *
+ * THIS SOFTWARE IS PROVIDED BY ANALOG DEVICES "AS IS" AND ANY EXPRESS OR
+ * IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, NON-INFRINGEMENT,
+ * MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED.
+ * IN NO EVENT SHALL ANALOG DEVICES BE LIABLE FOR ANY DIRECT, INDIRECT,
+ * INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, INTELLECTUAL PROPERTY RIGHTS, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+*******************************************************************************/
+
+/******************************************************************************/
+/***************************** Include Files **********************************/
+/******************************************************************************/
+
+#include <stdlib.h>
+#include <drivers/rng/adi_rng.h>
+#include "trng.h"
+#include "util.h"
+#include "error.h"
+
+/******************************************************************************/
+/********************** Macros and Constants Definitions **********************/
+/******************************************************************************/
+
+/* LenReload: 0 - 4095 */
+#define TRNG_CNT_VAL		4095
+/* Prescaler: 0 - 10 */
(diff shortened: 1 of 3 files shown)


adrimbarean:
Commits missing sign

drivers/platform/aducm3029/trng.c
```diff
@@ -0,0 +1,169 @@
...
+/*************************** Types Declarations *******************************/
+/******************************************************************************/
+
+/**
+ * @struct trng_desc
+ * @brief Stucture holding the TRNG descriptor.
+ */
+struct trng_desc {
+	/** Memory used by the DFP */
+	uint32_t	dev_mem[(ADI_RNG_MEMORY_SIZE + 3)/4]; //Aligned memory
```
cristiantpop, line 70: Non standard comment

drivers/platform/aducm3029/trng.c
```diff
@@ -0,0 +1,169 @@
...
+static struct trng_desc	g_desc;
+
+/******************************************************************************/
+/************************ Functions Definitions *******************************/
+/******************************************************************************/
+
+/* Initialize the TRNG device and the global descriptor */
+static void gdesc_init()
+{
+	adi_rng_Open(0, g_desc.dev_mem, ADI_RNG_MEMORY_SIZE, &g_desc.dev);
```
cristiantpop, line 91: Check return values for all functions

drivers/platform/aducm3029/trng.c
```diff
@@ -0,0 +1,169 @@
...
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+*******************************************************************************/
+
+/******************************************************************************/
+/***************************** Include Files **********************************/
+/******************************************************************************/
+
+#include <stdlib.h>
+#include <drivers/rng/adi_rng.h>
```
buha, line 46: I think there will be build issues with this kind of paths. It should instead be:
```suggestion
#include "adi_rng.h"
```

drivers/platform/aducm3029/trng.c
buha, line 70: where is ADI_RNG_MEMORY_SIZE coming from ? I couldn't find it.
Also, perhaps an explanation of this arithmetic would be good here as a comment.

drivers/platform/aducm3029/trng.c
```diff
@@ -0,0 +1,169 @@
...
+}
+
+/**
+ * @brief Initialize descriptor
+ * @param desc - Where to store the TRNG Descriptor
+ * @return \ref SUCCESS or \ref FAILURE if desc is null
+ */
+int32_t trng_init(struct trng_desc **desc)
+{
+
```
buha, line 114: extra space here

drivers/platform/aducm3029/trng.c
```diff
@@ -0,0 +1,169 @@
...
+	if (!desc)
+		return FAILURE;
+
+	if (!nb_references)
```
buha, line 118: Could nb_references be moves into some sort of a descriptor instead of being global ?

drivers/platform/aducm3029/trng.c
mchindri: No need. Allays returns success.

drivers/platform/aducm3029/trng.c
mchindri: This way is how DFP drivers are included in aducm

drivers/platform/aducm3029/trng.c
mchindri: From  #include <drivers/rng/adi_rng.h>

drivers/platform/aducm3029/trng.c
mchindri: No, nb_reference is used to count the references to the global descriptor

mchindri:
v2:
- sign commits
- remove extra space
- add comment to dev_mem

drivers/platform/aducm3029/trng.c
mchindri: solved

drivers/platform/aducm3029/trng.c
mchindri: solved

mchindri:
add trng.c in generic

include/trng.h
```diff
@@ -0,0 +1,70 @@
...
+/******************************************************************************/
+
+#include <stdint.h>
+
+/******************************************************************************/
+/*************************** Types Declarations *******************************/
+/******************************************************************************/
+
+/* TRNG Descriptor */
+struct rng_desc;
```
buha, line 55: should be `extern` in header
see this https://stackoverflow.com/questions/1433204/how-do-i-use-extern-to-share-variables-between-source-files

mchindri, line 55: trng_desc

include/trng.h
```diff
@@ -0,0 +1,70 @@
...
+
```
mchindri, line 56: add init_param

include/trng.h
mchindri: struct rng_desc is a structure declaration. It is not a variable.

mchindri:
v3:
 - Add trng.c in generic (modification in first commit message too)
 - Add init_param

include/trng.h
buha: oh, my bad
Then why isn't it properly declared here ? I see the actual declaration in `platform/aducm3029/trng.c`.

buha: Do you expect different trng structure forms based on platform ?

mchindri: Here is a declaration and in .c file is the definition. Yes, each platform can define its own structure this way.
In aducm are used only platform specific fields

buha: Ok, then two comments here:
- the buffer doesn't seem to be platform specific so you would have to duplicate it in every platform
- if you move the buffer in the generic one, you would need to to something similar to the `void *extra` thing that we do elsewhere

mchindri: The buffer is platform specific. It is the memory used by the BSP driver to hold the descriptor to don't allocate memory.
Inside adi_rng_Open is something like
`g_desc.dev = g_desc.dev_mem`

drivers/platform/aducm3029/trng.c
```diff
@@ -0,0 +1,171 @@
...
+static struct trng_desc	g_desc;
+
+/******************************************************************************/
+/************************ Functions Definitions *******************************/
+/******************************************************************************/
+
+/* Initialize the TRNG device and the global descriptor */
+static void gdesc_init()
+{
+	adi_rng_Open(0, g_desc.dev_mem, ADI_RNG_MEMORY_SIZE, &g_desc.dev);
```
adrimbarean, line 92: Since the dev_id is defined in trng_init_param why not use it here?

mchindri: because only device 0 is supported in aducm. dev_id was added to be used by other platforms

adrimbarean: OK. I would say to use a define then so that it looks better and is more descriptive of what the 0 means.

mchindri:
v4:
- replace 0 with ADUCM_TRNG_DEVICE_ID

mchindri:
v5:
- Fix doxygen

PR #583, opened 2020-05-04
AD9250 fix & update

amiclaus:
Add `register_data` pointer as parameter to the ad9250_read function.

This holds the information about the values read from the device.

Until now, the function was returning the value read, while the checks
through the entire driver were expecting a return code.
(SUCCESS/FAILURE).

This patch updates the read function and remove redundant checks based
on the previous implementation.

Assign valid addresses to the LID config registers.

For more information please check the datasheet of the device.
https://www.analog.com/media/en/technical-documentation/data-sheets/AD9250.pdf (Page 40 of 45)


drivers/adc/ad9250/ad9250.c
```diff
@@ -413,28 +414,18 @@ int32_t ad9250_is_shadow_register(int32_t register_address)
...
-		}
-		return (ret & AD9250_PDWN_CHIP(0x3)) >> 0;
-	}
-
-	return ret;
+		return (ad9250_set_bits_to_reg(dev,
+					       AD9250_REG_PDWN,
+					       AD9250_PDWN_CHIP(mode),
+					       AD9250_PDWN_CHIP(0x3)));
+	} else
```
adrimbarean, line 427: nitpick: else still needs brackets here

adrimbarean: Better to remove brackets from if actually.

drivers/adc/ad9250/ad9250.c
```diff
@@ -808,13 +808,13 @@ int32_t ad9250_jesd204b_setup(struct ad9250_dev *dev)
 		return ret;
```
adrimbarean, line 808: nitpick: maybe the link to the datasheet would be useful in the commit description.

amiclaus: done. thanks.

drivers/adc/ad9250/ad9250.c
amiclaus: Done

drivers/adc/ad9250/ad9250.c
```diff
@@ -228,14 +229,15 @@ int32_t ad9250_read(struct ad9250_dev *dev,
 					 buffer,
 					 3);
 		if(ret != SUCCESS) {
-			return ret;
+			return FAILURE;
 		}
 		reg_address--;
 		reg_value <<= 8;
 		reg_value |= buffer[2];
+		*register_data = reg_value;
```
buha, line 237: register_data could be `NULL`. I know it sucks checking all parameters but no-OS is a library and there is no guarantee users won't pass NULL as parameter.

drivers/adc/ad9250/ad9250.c
```diff
@@ -364,31 +369,27 @@ int32_t ad9250_set_bits_to_reg(struct ad9250_dev *dev,
...
+		reg_value = dev->shadow_regs[sr_cnt];
+	else
+		ret = ad9250_read(dev, register_address, &reg_value);
+
 	reg_value &= (~mask);
 	reg_value |= bits_value;
-	ret = ad9250_write(dev,
-			   register_address,
-			   reg_value);
+	ret |= ad9250_write(dev,
```
buha, line 385: avoid doint this, perform the check after every call:
```
if (ret)
    return ret;
```

drivers/adc/ad9250/ad9250.c
```diff
@@ -516,23 +487,17 @@ int32_t ad9250_test_mode(struct ad9250_dev *dev,
...
-		ret = ad9250_write(dev,
-				   AD9250_REG_OFFSET,
-				   AD9250_REG_OFFSET_ADJUST(adj));
-	} else {
-		return (ad9250_read(dev,
-				    AD9250_REG_OFFSET));
-	}
-
-	return ret;
+	if((adj >= -32) && (adj <= 31))
```
buha, line 495: This change is a little bit suspicios because it changes more than what this commit is about.
The commit should be only about addint the extra parameter to the ad9250_read.
I'm also not even sure if this is functionally correct, removing the ad9250_read call from that else branch.

drivers/adc/ad9250/ad9250.c
```diff
@@ -546,28 +511,18 @@ int32_t ad9250_offset_adj(struct ad9250_dev *dev,
...
-				  AD9250_REG_OUT_MODE);
-		if(ret != SUCCESS) {
-			return ret;
-		}
-		return (ret & AD9250_OUT_MODE_DISABLE) != 0;
-	}
-
-	return ret;
+	if((en == 0) || (en == 1))
+		return ad9250_set_bits_to_reg(dev,
```
buha, line 520: Same as above, and I see this repeats below. I think this commit does a bit too much and could be split, it would be easier to track changes and avoid mixing them up.

drivers/adc/ad9250/ad9250.c
amiclaus: It is stated in the commit message:
`Until now, the function was returning the value read, while the checks
through the entire driver were expecting a return code.
(SUCCESS/FAILURE).
`

drivers/adc/ad9250/ad9250.c
amiclaus: If i split the commit, adding the `register_data` parameter, i would have to modify all the read functions that are used here by adding an extra variable to each function in the driver and changing all headers to support also the data read.

All of this just to delete them in another commit.

I know it's a bit messy, but this driver was already broken.

drivers/adc/ad9250/ad9250.c
```diff
@@ -616,28 +561,18 @@ int32_t ad9250_output_invert(struct ad9250_dev *dev,
...
-		return (ret & AD9250_OUT_MODE_DATA_FORMAT(1)) >> 0;
-	}
-
-	return ret;
+	if( (format == 0) || (format == 1))
+		return ad9250_set_bits_to_reg(dev,
+					      AD9250_REG_OUT_MODE,
+					      AD9250_OUT_MODE_DATA_FORMAT(format),
+					      AD9250_OUT_MODE_DATA_FORMAT(1));
+	else
```
mchindri, line 574: else clause is not needed you can just return FAILURE

mchindri: or a better approach will be:
```suggestion
if(!(format == 0 || format == 1))
     return FAILURE;
     
return ad9250_set_bits_to_reg(...)

```

amiclaus:
Closing this. Making a new PR with minimal changes of the driver.

PR #585, opened 2020-05-05
Add EAGAIN error to error.h

mchindri:
- This is needed in order to implement secure tcp socket with mbedtls.
- EAGAIN will be returned when socket_recv is called but there is no data available. If 0 is returned doesn't work with mbedtls.
- An other other error name can be used, depending what noos team decides.
- Changes have been made in mqtt_noos_read and in at_read_buffer to adapt them to the new change.

### DIFF
--- a/include/error.h
+++ b/include/error.h
@@ -53,4 +53,6 @@
 
 #define IS_ERR_VALUE(x)	((x) < 0)
 
+#define	EAGAIN		11
+
 #endif // ERROR_H_
--- a/network/application/MQTT_eclipse_paho/mqtt_noos_support.c
+++ b/network/application/MQTT_eclipse_paho/mqtt_noos_support.c
@@ -47,6 +47,7 @@
 #include "error.h"
 #include "util.h"
 #include "delay.h"
+#include "error.h"
 
 /******************************************************************************/
 /**************************** Global Variables ********************************/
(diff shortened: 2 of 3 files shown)


network/application/MQTT_eclipse_paho/mqtt_noos_support.c
```diff
@@ -152,16 +153,22 @@ int mqtt_noos_read(Network* net, unsigned char* buff, int len, int timeout)
 	uint32_t	sent;
 	int32_t		rc;
 
+	if (!len)
+		return 0;
```
cristiantpop, line 157: Add new line after single line body

include/error.h
```diff
@@ -53,4 +53,8 @@
 
 #define IS_ERR_VALUE(x)	((x) < 0)
 
+//Where to put this and at_parser errors?
+//OR EWOULDREAD
```
buha, line 57: I think you meant EWOULDBLOCK not EWOULDREAD.
EAGAIN is shorter so let's go with it.
Regarding the question, I think this is the best spot for the errors codes to be defined.

network/application/MQTT_eclipse_paho/mqtt_noos_support.c
```diff
@@ -152,16 +153,22 @@ int mqtt_noos_read(Network* net, unsigned char* buff, int len, int timeout)
...
+ 			if (IS_ERR_VALUE(rc))
+ 				return rc;
+ 			sent += rc;
+ 			if (sent >= len)
+ 				return sent;
+ 		}
+
 		mdelay(1);
-	}
+	} while (--timeout);
```
buha, line 171: I think you should decrement your timeout only in case of error.

With the current approach:
Assume you have a timeout of 5 and first 3 loops you get an error. 
Also assume your message is split across 3 messages.
So you get 3 times error + 2 valid messages which hits the timeout. 
If you had done 1 more loop you'd have gotten the full message.

mchindri: If the functions returns 10 bytes but you requested 100, if you call the function again will not return more, that's way delay should be on success too.

include/error.h
mchindri: Yes I meant EWOULDBLOCK

dbogdan: I vote for EAGAIN too. However, I would define it similar as on Linux (11).

mchindri:
v2:
- change EAGAIN to 11 as in https://github.com/analogdevicesinc/linux/blob/d1873cb62263704a93d904420daf5941f38599b4/include/uapi/asm-generic/errno-base.h
- add new line after single line body

include/error.h
mchindri: done

network/application/MQTT_eclipse_paho/mqtt_noos_support.c
mchindri: done

PR #590, opened 2020-05-06
SPI Demux

amiclaus:
1. Add an intermediary Demux SPI layer to configure device selection via SPI data transfer.
2. fmcjesdadc1: add support for the demux SPI layer.

### DIFF
--- a/drivers/io-expander/demux_spi/demux_spi.c
+++ b/drivers/io-expander/demux_spi/demux_spi.c
@@ -1,0 +1,158 @@
+/***************************************************************************//**
+ *   @file   demux_spi.c
+ *   @brief  Implementation of the SPI Demux Interface
+ *   @author Antoniu Miclaus (antoniu.miclaus@analog.com)
+********************************************************************************
+ * Copyright 2020(c) Analog Devices, Inc.
+ *
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *  - Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ *  - Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in
+ *    the documentation and/or other materials provided with the
+ *    distribution.
+ *  - Neither the name of Analog Devices, Inc. nor the names of its
+ *    contributors may be used to endorse or promote products derived
+ *    from this software without specific prior written permission.
+ *  - The use of this software may or may not infringe the patent rights
+ *    of one or more patent holders.  This license does not release you
+ *    from the requirement that you obtain separate licenses from these
+ *    patent holders to use this software.
+ *  - Use of the software either in source or binary form, must be run
+ *    on or directly connected to an Analog Devices Inc. component.
+ *
+ * THIS SOFTWARE IS PROVIDED BY ANALOG DEVICES "AS IS" AND ANY EXPRESS OR
+ * IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, NON-INFRINGEMENT,
+ * MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED.
+ * IN NO EVENT SHALL ANALOG DEVICES BE LIABLE FOR ANY DIRECT, INDIRECT,
+ * INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, INTELLECTUAL PROPERTY RIGHTS, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+*******************************************************************************/
+
+/******************************************************************************/
+/***************************** Include Files **********************************/
+/******************************************************************************/
+
+#include "demux_spi.h"
+#include "error.h"
+#include <stdlib.h>
+#include <string.h>
+
+/******************************************************************************/
+/**************************** Types Definitions *******************************/
+/******************************************************************************/
+
+/**
+ * @brief Demux specific SPI platform ops structure
+ */
+const struct spi_platform_ops demux_spi_platform_ops = {
+	.spi_ops_init = demux_spi_init,
+	.spi_ops_remove = demux_spi_remove,
+	.spi_ops_write_and_read = demux_spi_write_and_read
+};
+
+/**
+ * @brief Initialize the SPI demux layer.
+ * @param desc - The SPI descriptor.
+ * @param param - The structure that contains the SPI parameters.
+ * @return SUCCESS in case of success, FAILURE otherwise.
+ */
+int32_t demux_spi_init(struct spi_desc **desc,
+		       const struct spi_init_param *param)
+{
+	int32_t ret;
+
+	struct spi_desc *descriptor;
+	struct spi_desc *spi_dev_desc;
+	struct spi_init_param *spi_dev_param;
+
+	if (!param)
+		return FAILURE;
+
+	descriptor = (struct spi_desc *)calloc(1, sizeof(*descriptor));
+	if (!descriptor)
+		return FAILURE;
+
+	descriptor->chip_select = param->chip_select;
+	descriptor->max_speed_hz = param->max_speed_hz;
+	descriptor->mode = param->mode;
+
+	spi_dev_param = param->extra;
+
+	ret = spi_init(&spi_dev_desc, spi_dev_param);
+	if (ret != SUCCESS) {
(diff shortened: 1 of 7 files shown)


drivers/io-expander/demux_spi/demux_spi.c
```diff
@@ -0,0 +1,148 @@
...
+
+	descriptor->chip_select = param->chip_select;
+	descriptor->max_speed_hz = param->max_speed_hz;
+	descriptor->mode = param->mode;
+
+	spi_dev_param = param->extra;
+
+	ret = spi_init(&spi_dev_desc, spi_dev_param);
+
+	(descriptor->extra) = spi_dev_desc;
```
mchindri, line 92:
```suggestion
	struct dmux_desc *dmux = calloc(1, sizeof(struct dmux_desc));
	descriptor->extra = dmux;
	
	switch(dmux_param->operation) {
	case INIT_DEMUX:
		dmux->is_dmux_instance = 1;
		//Init next device (either platform device or other dmux)
		spi_init(&dmux->next_dev, dmux_param->param.next_param); //here is platform initialization
		break;
	case ATTACH_SPI_DEVICE:
		dmux->is_dmux_instance = 0;
		descriptor->chip_select = param->chip_select;
		descriptor->max_speed_hz = param->max_speed_hz;
		descriptor->mode = param->mode;

		struct dmux_desc *mux_ref =  dmux_param->param.dmux->extra;
		dmux->next_dev = mux_ref->next_dev;
		break;
	}
```

drivers/io-expander/demux_spi/demux_spi.c
```diff
@@ -0,0 +1,148 @@
...
+ * @return SUCCESS in case of success, FAILURE otherwise.
+ */
+int32_t demux_spi_init(struct spi_desc **desc,
+		       const struct spi_init_param *param)
+{
+	int32_t ret;
+
+	struct spi_desc *descriptor;
+	struct spi_desc *spi_dev_desc;
+	struct spi_init_param *spi_dev_param;
```
mchindri, line 75:
```suggestion
	struct dmux_init_param *dmux_param  = param->extra;
```

drivers/io-expander/demux_spi/demux_spi.h
```diff
@@ -0,0 +1,80 @@
...
+/******************************************************************************/
+/*************************** Types Declarations *******************************/
+/******************************************************************************/
+
+/**
+ * @struct spi_desc
+ * @brief Structure initialization with the platform specific SPI functions
+ */
+extern const struct spi_platform_ops demux_spi_platform_ops;
+
```
mchindri, line 64:
```suggestion

struct dmux_init_param {
	enum {
		INIT_DEMUX,
		ATTACH_SPI_DEVICE
	} operation;
	union {
		spi_init_param  *next_param;
		struct spi_desc *dmux;
	} param;
};

struct dmux_desc {
	/* Used to know is it is a mux instance or a device instance.
	 No operation should be done in the mux instance. It is only used to
	 pass the next_dev instance to the other devices attached to the mux */
	bool		is_dmux_instance;
	/* this is the reference the mux will write to */
	spi_desc	*next_dev;
};
```

projects/fmcjesdadc1/src/app/fmcjesdadc1.c
```diff
@@ -89,8 +104,11 @@ int main(void)
 #endif
 		.device_id = SPI_DEVICE_ID
 	};
-	ad9250_spi_param.extra = &xil_spi_param;
-	ad9517_spi_param.extra = &xil_spi_param;
+
+	demux_spi_param.extra = &xil_spi_param;
```
mchindri, line 108:
```suggestion
	struct spi_desc *dmux_desc;

	struct spi_init_param demux_xil_parm = {
		.max_speed_hz = 2000000u,
		.chip_select = 0,
		.mode = SPI_MODE_0,
		.spi_ops = SPI_PLATFORM_OPS,
		.extra = &xil_spi_param,
	};

	struct dmux_init_param dmux_init_param = {
		.operation = INIT_DEMUX,
		.pram.next_param = &demux_xil_parm;
	};

	struct spi_init_param demux_spi_param = {
		.spi_ops = SPI_MUX_OPS,
		.extra = &dmux_init_param;
	};
	
	spi_init(&dmux_desc, demux_spi_param);

	struct dmux_init_param dmux_attach_param = {
		.operation = ATTACH_SPI_DEVICE;
		.param.dmux = dmux_desc;
	}
	demux_spi_param.extra = &dmux_attach_param;
```

mchindri:
M-am mai gandit dupa, si cred ca mai bine de cat cum am sugerat sa se faca init la o instanta de mux, cred ca ar fi mai bine ca instantele sa fie handl-uite cumva global

mirceacaprioru:
> M-am mai gandit dupa, si cred ca mai bine de cat cum am sugerat sa se faca init la o instanta de mux, cred ca ar fi mai bine ca instantele sa fie handl-uite cumva global

Messages should be posted in English.

mchindri:
> > M-am mai gandit dupa, si cred ca mai bine de cat cum am sugerat sa se faca init la o instanta de mux, cred ca ar fi mai bine ca instantele sa fie handl-uite cumva global
> 
> Messages should be posted in English.

I edited it

amiclaus:
V2:

- resolve newly appeared conflicts

amiclaus:
V3:

- perform rebase to get latest Travis build fixes.

drivers/io-expander/demux_spi/demux_spi.c
```diff
@@ -0,0 +1,148 @@
...
+	if (!descriptor)
+		return FAILURE;
+
+	descriptor->chip_select = param->chip_select;
+	descriptor->max_speed_hz = param->max_speed_hz;
+	descriptor->mode = param->mode;
+
+	spi_dev_param = param->extra;
+
+	ret = spi_init(&spi_dev_desc, spi_dev_param);
```
adrimbarean, line 90: I would say that it would be better to test `ret` now with an `if` and exit the method early in case of error.

drivers/io-expander/demux_spi/demux_spi.c
```diff
@@ -0,0 +1,148 @@
...
+	return ret;
+}
+
+/**
+ * @brief Free the resources allocated by demux_spi_init().
+ * @param desc - The SPI descriptor.
+ * @return SUCCESS in case of success, FAILURE otherwise.
+ */
+int32_t demux_spi_remove(struct spi_desc *desc)
+{
```
adrimbarean, line 105: Should `desc` is `NULL` first.

drivers/io-expander/demux_spi/demux_spi.c
buha, line 90: Which spi_init() is called here exactly ? 
The way I understand, we got here due to the project application calling [for example] ad9517_setup() which calls the generic layer spi_init() which maps the call onto a platform specific [for example] demux_spi_init(). And here, you call a spi_init() and I don't get how it's mapped onto physical platform xil_spi_init() [let's say].

projects/fmcjesdadc1/src/app/fmcjesdadc1.c
```diff
@@ -67,20 +68,33 @@ int main(void)
...
+	struct spi_init_param demux_spi_param = {
 		.max_speed_hz = 2000000u,
 		.chip_select = 0,
 		.mode = SPI_MODE_0,
 		.platform_ops = &xil_platform_ops
 	};
 
-	struct spi_init_param ad9517_spi_param = {
+	struct spi_init_param ad9250_0_spi_param = {
 		.max_speed_hz = 2000000u,
```
buha, line 80: I think the `.max_speed_hz` and `.mode` are only relevant in the `demux_spi_param` and not for the devices themselves. You could remove them from here to avoid confusion.

buha: One other idea would be to force the user, with some checks in demux code, to put identical values for `.mode` and `.max_speed_hz` because the demux that we implement here only supports slave devices with different chip selects but with identical mode and speed.

drivers/io-expander/demux_spi/demux_spi.c
buha: I actually agree with the need for a change like this.
The problem that we have with the current implementation is that we perform the platform's `spi_init` once for each added device. And of course it works as-is because the only difference between various device settings is the chip select number (not the speed, not the mode). Specifically, in projects/fmcjesdadc1, xil_spi_init is called at the initialization of each of the mux driven devices plus the mux itself -> 4 times. The prescaler and SPI mode being identical the code works. But we should avoid this multiple platform `spi_init` call.

projects/fmcjesdadc1/src/app/fmcjesdadc1.c
amiclaus: > 
> 
> I think the `.max_speed_hz` and `.mode` are only relevant in the `demux_spi_param` and not for the devices driven by demux themselves. You could remove the initialization from here to avoid confusion.

I agree that these parameters are irrelevant for the devices, but I would not let them uninitialized.
For a better understanding and avoiding confusion, I would initialize them with the parameters provided by `demux_spi_init`.

Something like this:
`struct spi_init_param ad9250_0_spi_param = { .max_speed_hz = demux_spi_param.max_speed_hz, .mode = demux_spi_param.mode, `

amiclaus:
V4:
- add checks for NULL descriptors and return value from `spi_init` function.
- pass spi `.mode` and `max_speed_hz` parameters from the `demux_spi_param` initialization structure to the device specific spi structures.

amiclaus:
V5:
- fix Codacy complains

projects/fmcjesdadc1/src/app/fmcjesdadc1.c
buha: Makes sense, it shows that device clocks and mode cannot be anything else than demux clock /mode.

PR #592, opened 2020-05-06
SPI generic layer

amiclaus:
This patch contains the following updates regarding the SPI structure and implementation:

1. add generic SPI interface
2. update Xilinx and Altera platform drivers to support the new generic layer.
3. update the related projects from the projects folder.

### DIFF
--- a/drivers/platform/xilinx/spi_extra.h
+++ b/drivers/platform/xilinx/spi_extra.h
@@ -98,4 +98,24 @@ typedef struct xil_spi_desc {
 	void			*instance;
 } xil_spi_desc;
 
+/**
+ * @brief Xilinx platform specific SPI platform ops structure
+ */
+extern const struct spi_platform_ops xil_platform_ops;
+
+/******************************************************************************/
+/************************ Functions Declarations ******************************/
+/******************************************************************************/
+
+/* Initialize the SPI communication peripheral. */
+int32_t xil_spi_init(struct spi_desc **desc,
+		     const struct spi_init_param *param);
+
+/* Free the resources allocated by spi_init(). */
+int32_t xil_spi_remove(struct spi_desc *desc);
+
+/* Write and read data to/from SPI. */
+int32_t xil_spi_write_and_read(struct spi_desc *desc, uint8_t *data,
+			       uint16_t bytes_number);
+
 #endif // SPI_EXTRA_H_
--- a/drivers/spi/spi.c
+++ b/drivers/spi/spi.c
@@ -1,0 +1,87 @@
+/***************************************************************************//**
+ *   @file   spi.c
+ *   @brief  Implementation of the SPI Interface
+ *   @author Antoniu Miclaus (antoniu.miclaus@analog.com)
+********************************************************************************
+ * Copyright 2020(c) Analog Devices, Inc.
+ *
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *  - Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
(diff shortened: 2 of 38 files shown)


drivers/platform/xilinx/spi_extra.h
```diff
@@ -98,4 +98,24 @@ typedef struct xil_spi_desc {
 	void			*instance;
 } xil_spi_desc;
 
-#endif // SPI_EXTRA_H_
\ No newline at end of file
+/**
+ * @brief Xilinx platform specific SPI platform ops structure
+ */
+extern const struct spi_platform_ops xil_platform_ops;
```
mchindri, line 104: I think we can do this to avoid the stuff with extern which maybe is a little messy:
```suggestion
#define SPI_PLATFORM_OPS (&(struct spi_platform_ops) {\
               .spi_ops_init = xil_spi_init,\
               .spi_ops_write_and_read = xil_spi_write_and_read,\
               .spi_ops_remove = xil_spi_remove\
}))
```

mchindri: And in spi_init_param will be directly:
`.platform_ops = SPI_PLATFORM_OPS`

include/spi.h
```diff
@@ -72,6 +72,13 @@ typedef enum spi_mode {
 	SPI_MODE_3 = (SPI_CPOL | SPI_CPHA)
 } spi_mode;
 
+/**
+ * @struct spi_platform_ops
+ * @brief Structure holding SPI function pointers that point to the platform
+ * specific function
+ */
+struct spi_platform_ops ;
```
mchindri, line 80: Shouldn't the name be spi_ops only? It is not platform dependent.
And instead of xil_ops and altera_ops have only one name: platform_ops.

include/spi.h
```diff
@@ -83,6 +90,7 @@ typedef struct spi_init_param {
 	uint8_t		chip_select;
 	/** SPI mode */
 	enum spi_mode	mode;
+	const struct spi_platform_ops *platform_ops;
```
mchindri, line 93:
```suggestion
	const struct spi_ops *spi_ops;
```

mchindri: same for descriptor

drivers/platform/xilinx/spi_extra.h
amiclaus: I think this is a matter of preferences.

include/spi.h
amiclaus: The `spi_platform_ops` holds platform specific information therefore the naming explicit enough. Moreover, the naming was already discussed in the meeting and sent via email.

drivers/spi/spi.c
```diff
@@ -0,0 +1,88 @@
...
+/**
+ * @brief Initialize the SPI communication peripheral.
+ * @param desc - The SPI descriptor.
+ * @param param - The structure that contains the SPI parameters.
+ * @return SUCCESS in case of success, FAILURE otherwise.
+ */
+int32_t spi_init(struct spi_desc **desc,
+		 const struct spi_init_param *param)
+{
+
```
mirceacaprioru, line 54: nitpick extra line

mirceacaprioru:
looks good overall.

drivers/platform/xilinx/spi_extra.h
mchindri: ok, this way it would be in just one place, but I don't really want this change.  It was just a suggestion.

amiclaus:
V2: 

- remove extra line

amiclaus:
V3:

- remove unnecessary brackets at the return of `spi_write_and_read` function

mirceacaprioru:
i think this could be good to merge.

amiclaus:
V4:

- resolve newly appeared conflicts.

PR #594, opened 2020-05-07
Ad5592r improvements

endakilgarriff:
- Remove unused gpio_map device struct variable
- Allow for assignment of GPI and GPO separately in ad5592r_set_channel_modes and clear variables to resolve issues with overwriting other modes.
- Add ldac_mode device struct variable and write value to ldac mode register to maintain setting
- Add stop bits to I2C reads/writes 
- Add function to handle data from ADC sequence
- Ignore failure generated from NAK on last frame of software reset - NAK expected as per datasheet
- Add macros and bit masks
- Add read/write device struct declarations to driver header


drivers/adc-dac/ad5592r/ad5593r.c
```diff
@@ -122,6 +123,40 @@ int32_t ad5593r_read_adc(struct ad5592r_dev *dev, uint8_t chan,
 	return 0;
 }
 
+/**
+ * Read ADC sequence.
+ *
+ * @param dev - The device structure.
+ * @return 0 in case of success, negative error code otherwise
+ */
+int32_t ad5593r_read_adc_sequence(struct ad5592r_dev *dev,
```
mirceacaprioru, line 132: this does not look different than just reading each individual channel with ad5593r_read_adc

If a sequence of adc would want to be read together then the actual i2c transmission would be extended with the number of channels. The transfer would be 2 * NO_CHANNELS not just 2 and using a for loop.

Check Figure 37. I2c Block Read page 23 of the datasheet.

drivers/adc-dac/ad5592r/ad5592r.c
```diff
@@ -131,6 +132,34 @@ int32_t ad5592r_read_adc(struct ad5592r_dev *dev, uint8_t chan,
...
+int32_t ad5592r_read_adc_sequence(struct ad5592r_dev *dev,
+				  uint16_t in_adc_seq,
+				  uint16_t *adc_values)
+{
+	int32_t ret;
+
+	if (!dev)
+		return FAILURE;
+
+	for (uint8_t i = 0; i <= dev->num_channels; i++) {
```
mirceacaprioru, line 150: the same would be here also similar to the i2c part. Create a sequence read no just multiple single reads.

Figure 44. page 33 of the datasheet.

drivers/adc-dac/ad5592r/ad5592r-base.h
```diff
@@ -77,11 +77,42 @@ enum ad5592r_registers {
 	AD5592R_REG_RESET		= 0xF,
 };
 
+#define AD5592R_REG_PD_PD_ALL		BIT(10)
+#define AD5592R_REG_PD_PD_ALL_RD(x)		(((x) & AD5592R_REG_PD_PD_ALL) >> 10)
```
mirceacaprioru, line 81: i don't see the need for a read macro. Where would this be used? If the reason would be just to check if it is set this ca be done with:

```
if (reg & AD5592R_REG_PD_PD_ALL)
{
      prinft("Powerdown is set!\r\n");
}

File: drivers/adc-dac/ad5592r/ad5592r-base.h [hunk 4]
```diff
@@ -77,11 +77,42 @@ enum ad5592r_registers {
...
 #define AD5592R_REG_PD_EN_REF		BIT(9)
-#define AD5592R_REG_CTRL_ADC_RANGE	BIT(5)
```
mirceacaprioru, line 80: the commit refers to adding macros, why are these removed ?

drivers/adc-dac/ad5592r/ad5592r.h
```diff
@@ -49,6 +49,8 @@
 	((((x) & 0x00ff) << 8) | \
 	 (((x) & 0xff00) >> 8))
 
+extern const struct ad5592r_rw_ops ad5592r_rw_ops;
```
mirceacaprioru, line 52: this is already declared in ad5592r-base.h

mirceacaprioru, line 52: why would this need to be shared with anything else? The driver is the only one that would use these operations.

drivers/adc-dac/ad5592r/ad5592r-base.h
endakilgarriff: The macros from ADC and DAC range have just been moved. You can see them on line 95 and 97

drivers/adc-dac/ad5592r/ad5592r-base.h
endakilgarriff: You are correct. I was using them to shift to the LSB but that is not necessary I will remove these

drivers/adc-dac/ad5592r/ad5593r.c
endakilgarriff: In this case I am initiating the ADC sequence at the application level and then the ad5593r_read_adc_seqeunce function gathers the I2C frames (or SPI frames for the AD5592R) for the number ADC channels that were included in the sequence. These are added to the adc_values array returning only the adc_code. This function was added to get the data from each channel in a sequence. I could only get the data from the last channel read in the sequence with the ad5593r_read_adc function, unless there is another way to do so that I didn't think of?

mirceacaprioru: my suggestion here was to increase the transmission size to more than 2 bytes. That way you eliminate the need for the for loop. something like:
```
uint8_t data[2 * adc_seq_channels];

ret = i2c_read(dev->i2c, data, ARRAY_SIZE(data), STOP_BIT);
if (ret < 0)
      return ret;
```
This way you get all the channels data that are enabled in the sequence in one transmission. 
Then you would just use a loop to interpret the data
```
for (uint8_t i = 0; i <= adc_seq_channels; i++) 
         adc_values[i] = ((uint16_t)(((data[2 * i] & 0x0F) << 8) + data[2 * i + 1]) & 0xFFF);
```

drivers/adc-dac/ad5592r/ad5592r-base.c
```diff
@@ -231,9 +234,12 @@ int32_t ad5592r_set_channel_modes(struct ad5592r_dev *dev)
 			adc |= BIT(i);
 			break;
 
-		case CH_MODE_GPIO:
-			dev->gpio_map |= BIT(i);
-			dev->gpio_in |= BIT(i); /* Default to input */
+		case CH_MODE_GPI:
+			dev->gpio_in |= BIT(i);
+			break;
+
```
amiclaus, line 240: nitpick: remove please extra line. no needed

drivers/adc-dac/ad5592r/ad5593r.c
endakilgarriff: Looking into the above solution, it will work to read back all of the channels in a single ADC read. However, I was looking to try and enable the repeat bit in the ADC sequencer and capture the data coming from the bus indefinitely. This is what the original solution was trying to accomplish, having set n channels, read back n frames, loop at the application layer and repeat reading data. In the meantime, another solution came to mind running sections depending on seq_status, this would be an alternative to ad5593r_read_adc: Thoughts?

`int32_t ad5593r_read_adc(struct ad5592r_dev *dev,
				  uint16_t *adc_values, int32_t seq_status)
{
	uint16_t temp;
	int32_t ret;
	uint8_t i, num_ch_in_seq = 0, data[MAX_ARRAY_SIZE];

	if (!dev)
		return FAILURE;
	
	for (i = 0; i <= dev->num_channels; i++)
	{
		if (AD5592R_REG_ADC_SEQ_INCL(i) & dev->adc_sequence_reg)
			num_ch_in_seq++; 	
	}
	
	switch (seq_status)
	{
	case REPEAT_SEQUENCE:
		ret = i2c_read(dev->i2c, data, (2 * num_ch_in_seq), RESTART_BIT);
		if (ret < 0)
			return ret;
		break;
		
	case STOP_SEQUENCE:
		ret = i2c_read(dev->i2c, data, (2 * num_ch_in_seq), STOP_BIT);
		if (ret < 0)
			return ret;
		break;
		
	case START_SEQUENCE:
	default:
		/* fall-through */
		temp = dev->adc_sequence_reg;

		data[0] = AD5593R_MODE_CONF | AD5592R_REG_ADC_SEQ | (AD5592R_REG_ADC_SEQ_REP & dev->adc_sequence_reg);
		data[1] = (dev->adc_sequence_reg & 0xFF);
		data[2] = temp & 0xFF;

		ret = i2c_write(dev->i2c, data, 3, STOP_BIT);
		if (ret < 0)
			return ret;

		data[0] = AD5593R_MODE_ADC_READBACK;
		ret = i2c_write(dev->i2c, data, 1, RESTART_BIT);
		if (ret < 0)
			return ret;
		
		ret = i2c_read(dev->i2c, data, (2 * num_ch_in_seq), 
			(dev->adc_sequence_reg & AD5592R_REG_ADC_SEQ_REP) ? RESTART_BIT : STOP_BIT );
		if (ret < 0)
			return ret;
		break;
	}
	
	for (i = 0; i < num_ch_in_seq; i++)
	{
		adc_values[i] = (uint16_t)(((data[2 * i] & 0xFF) << 8) + data[(2 * i) + 1]);	
	}
	
	return 0;
}`

drivers/adc-dac/ad5592r/ad5592r-base.c
```diff
@@ -3,7 +3,7 @@
  *   @brief  Implementation of AD5592R Base Driver.
```
adrimbarean, line 3: Could you explain in the commit message, briefly, why the `gpio_map` was removed?

drivers/adc-dac/ad5592r/ad5592r-base.h
```diff
@@ -112,6 +112,7 @@ struct ad5592r_dev {
 	uint8_t gpio_out;
```
adrimbarean, line 146: Can you please add a short description in the commit about what is LDAC mode and why we need it?

drivers/adc-dac/ad5592r/ad5593r.c
```diff
@@ -3,7 +3,7 @@
  *   @brief  Implementation of AD5593R driver.
```
adrimbarean, line 3: Can you mention in the commit message that this is done to conform with the new platform driver?

drivers/adc-dac/ad5592r/ad5592r-base.c
endakilgarriff: Making these corrections now, just to make sure - you would like the end of the CH_MODE_GPI case to be directly above the CH_MODE_GPO case with no space? Like this:
`		case CH_MODE_GPI:
			dev->gpio_in |= BIT(i);
			break;
		case CH_MODE_GPO:
			dev->gpio_out |= BIT(i);
			break;`

amiclaus: Yes, indeed. You can use as a reference other drivers that use the switch statement,

endakilgarriff: The original implementation had spaces between each, you're sure you want me to remove them all from this switch/case statement?

amiclaus: If the original implementation for all cases has spaces in between, then keep it like that. I did not have a look at the entire driver, but only on your modifications.
Then the spacing is subject to another commit.

endakilgarriff: Perfect, I will leave the spacing as is for the moment

amiclaus:
Hello @endakilgarriff . Can you please perform a rebase to sync this branch with the no-OS `master` branch? There's a strange issue in the TravisCI build. 

Also, if you force-push, please add here a comment with the changes that you made so we can have a look again on the new version of the PR.
Example:
https://github.com/analogdevicesinc/no-OS/pull/604#issuecomment-630262807

endakilgarriff:
V2: 
- Generally improve commit messages 
- Remove read_adc_sequence funciton compared with last PR
- Add functionality to read and return data for an entire adc sequence to read_adc
- Correct params in comments for doxygen

PR #595, opened 2020-05-07
ad9250: update device init struct

amiclaus:
Updated device initialization structure by adding `ad9250_st_init`
structure to the device initialization parameters.

The user will now configure the device settings inside the
main application.

### DIFF
--- a/drivers/adc/ad9250/ad9250.h
+++ b/drivers/adc/ad9250/ad9250.h
@@ -513,6 +513,8 @@ struct ad9250_dev {
 struct ad9250_init_param {
 	/* SPI */
 	spi_init_param	spi_init;
+	/* Device Settings */
+	struct ad9250_state ad9250_st_init;
 };
 
 /******************************************************************************/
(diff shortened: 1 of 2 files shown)


drivers/adc/ad9250/ad9250.h
```diff
@@ -513,6 +513,8 @@ struct ad9250_dev {
 struct ad9250_init_param {
 	/* SPI */
 	spi_init_param	spi_init;
+	/* Device Settings */
+	struct ad9250_state ad9250_st_init;
```
buha, line 517: i would just name the structure member `state` or something generic. `ad9250_` is redundant since it is part of an `ad9250_init_param` struct.

amiclaus: I agree. I just wanted keep consistency within the driver.
https://github.com/analogdevicesinc/no-OS/blob/7d37ec72a03168914854d309816894b984580938/drivers/adc/ad9250/ad9250.h#L509

I think renaming all structure parameters should be part of another commit.

buha: that could do it then

PR #596, opened 2020-05-11
Ad9517 update

amiclaus:
1. Update device initialization
2. Fix write operation for LVPECL

### DIFF
--- a/drivers/frequency/ad9517/ad9517.c
+++ b/drivers/frequency/ad9517/ad9517.c
@@ -42,7 +42,6 @@
 /******************************************************************************/
 #include <stdlib.h>
 #include "ad9517.h"
-#include "ad9517_cfg.h"
 
 /***************************************************************************//**
  * @brief Initializes the AD9517.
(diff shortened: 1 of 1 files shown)


drivers/frequency/ad9517/ad9517.c
```diff
@@ -115,7 +115,11 @@ int32_t ad9517_setup(struct ad9517_dev **device,
 	/* Update the device with user settings for the LVPECL output
 	 * channels. */
 	for(index = 0; index < 4; index++) {
-		reg_address = AD9517_REG_LVPECL_OUT0 + index;
+		if (index < 2)
```
buha, line 118: This code seems correct but if for me a bit cryptic to understand and probably will be cryptic for others reading it in the future. One other approach would be:
```
static uint16_t lvepcl_out_ch[4] = {
    AD9517_REG_LVPECL_OUT0,
    AD9517_REG_LVPECL_OUT1,
    AD9517_REG_LVPECL_OUT2,
    AD9517_REG_LVPECL_OUT3
};
```
And directly use `lvepcl_out_ch[3]` to access channel 3 for example, so there is no need for arithmetic.

amiclaus: Yes, this approach is more readable. Will do.

amiclaus:
V2:

- add `lvepcl_out_ch` array for easier readability of the LVPECL channels.

PR #602, opened 2020-05-15
AD9517: Activate PLL and calibrate when using VCO

Jargendas:
I am currently working with the AD9467 eval board and wanted to use an external 10 MHz reference to generate the ADC sampling clock using the on-board AD9517-4. It didn't work out of the box by just setting the correct values in ad9517_cfg.h, and after a lot of digging in the datasheet, I found that the PLL is not enabled by default or in the driver. This makes no sense in my opinion, as using the VCO without the PLL enabled is not possible as far as I can tell. So it should be enabled by the driver when vco_clk_sel is enabled I think.

Also, the VCO must be calibrated in order for the PLL to lock. That should definitely also be kicked off by the driver.

### DIFF
--- a/drivers/frequency/ad9517/ad9517.c
+++ b/drivers/frequency/ad9517/ad9517.c
@@ -153,10 +153,25 @@ int32_t ad9517_setup(struct ad9517_dev **device,
 			return ret;
 	}
 	/* Check if VCO is selected as input. */
-	if(dev->ad9517_st.pdata->vco_clk_sel)
+	if(dev->ad9517_st.pdata->vco_clk_sel) {
 		/* Sets the VCO frequency. */
 		ad9517_vco_frequency(dev,
 				     dev->ad9517_st.pdata->int_vco_freq);
+
+		/* Activate PLL */
+		reg_value = AD9517_PLL_POWER_DOWN(0x0) |
+			    AD9517_CP_MODE (0x3) |
+			    AD9517_CP_CURRENT (0x7) |
+			    0 * AD9517_PFD_POLARITY;
+		ret = ad9517_write(dev, AD9517_REG_PFD_CHARGE_PUMP, reg_value);
+		if(ret < 0)
+			return ret;
+
+		/* Start VCO Calibration */
+		ret = ad9517_write(dev, AD9517_REG_PLL_CTRL_3, AD9517_VCO_CAL_NOW);
+		if(ret < 0)
+			return ret;
+	}
 
 	*device = dev;
 


drivers/frequency/ad9517/ad9517.c
```diff
@@ -153,11 +153,23 @@ int32_t ad9517_setup(struct ad9517_dev **device,
...
 	}
 	/* Check if VCO is selected as input. */
-	if(dev->ad9517_st.pdata->vco_clk_sel)
+	if(dev->ad9517_st.pdata->vco_clk_sel) {
 		/* Sets the VCO frequency. */
 		ad9517_vco_frequency(dev,
 				     dev->ad9517_st.pdata->int_vco_freq);
 
+		/* Activate PLL */
+		reg_value =
```
buha, line 162: for styling consistency, let's do something like this:
```suggestion
		reg_value = AD9517_PLL_POWER_DOWN(0) |
				    AD9517_CP_MODE (3) |
				    AD9517_CP_CURRENT (5) |
				    0 * AD9517_PFD_POLARITY;
```

buha:
Thanks for your contribution.

For this to be merged you would need to fix the github warnings above with this CLA contribution signing (I think you need an account) and also doing `git commit --amend -s` which will append a line to your commit description.

Styling-wise, it is always a good idea to execute `ci/travis/astyle.sh` to format the style of the last commit.

drivers/frequency/ad9517/ad9517.c
```diff
@@ -153,11 +153,26 @@ int32_t ad9517_setup(struct ad9517_dev **device,
...
-	if(dev->ad9517_st.pdata->vco_clk_sel)
+	if(dev->ad9517_st.pdata->vco_clk_sel) {
 		/* Sets the VCO frequency. */
 		ad9517_vco_frequency(dev,
 				     dev->ad9517_st.pdata->int_vco_freq);
 
+		/* Activate PLL */
+		reg_value = AD9517_PLL_POWER_DOWN(0) |
+			    AD9517_CP_MODE (3) |
+			    AD9517_CP_CURRENT (5) |
```
buha, line 164: I understand that PLL_POWER_DOWN must be zero because you want to enable PLL, that's clear.
But if you compare the CP_MORE to the datasheet, it matches the default value in the register.
Yet the CP_CURRENT is 0x5 and in the datasheet is 0x7 by default.
We should stick to default values unless we add a way to configure this setting or there is a reasonable argument to go with different setting by default.

Jargendas: Yes, CP_CURRENT should be 0x7, I misread it from the datasheet.

buha: Ok, just change that and you have an approve from me, thanks for contributing !

buha: Just make sure the astyle.sh is performed, you probably saw there is an automatic test for that that currently fails.

Jargendas: Yes, that's very weird, it succeeded before I executed astyle.sh, now I did it and it fails...

buha: I think it might have to do with branch not being in sync with master ? Try rebasing your branch on top of latest analogdevicesinc/master, I think astyle rule will pass after that.

PR #607, opened 2020-05-19
util: add UART-based CLI [WIP]

adrimbarean:
Add Command Line Interface implementation for UART enabled projects. This
is to help building and interactive CLI with two sets of commands for any
no-OS project.

### DIFF
--- a/include/cli.h
+++ b/include/cli.h
@@ -1,0 +1,139 @@
+/***************************************************************************//**
+*   @file   cli.h
+*   @brief  CLI header.
+*   @author Andrei Drimbarean (andrei.drimbarean@analog.com)
+********************************************************************************
+* Copyright 2018(c) Analog Devices, Inc.
+*
+* All rights reserved.
+*
+* Redistribution and use in source and binary forms, with or without
+* modification, are permitted provided that the following conditions are met:
+*  - Redistributions of source code must retain the above copyright
+*    notice, this list of conditions and the following disclaimer.
+*  - Redistributions in binary form must reproduce the above copyright
+*    notice, this list of conditions and the following disclaimer in
+*    the documentation and/or other materials provided with the
+*    distribution.
+*  - Neither the name of Analog Devices, Inc. nor the names of its
+*    contributors may be used to endorse or promote products derived
+*    from this software without specific prior written permission.
+*  - The use of this software may or may not infringe the patent rights
+*    of one or more patent holders.  This license does not release you
+*    from the requirement that you obtain separate licenses from these
+*    patent holders to use this software.
+*  - Use of the software either in source or binary form, must be run
+*    on or directly connected to an Analog Devices Inc. component.
+*
+* THIS SOFTWARE IS PROVIDED BY ANALOG DEVICES "AS IS" AND ANY EXPRESS OR IMPLIED
+* WARRANTIES, INCLUDING, BUT NOT LIMITED TO, NON-INFRINGEMENT, MERCHANTABILITY
+* AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED.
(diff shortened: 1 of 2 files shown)


adrimbarean:
V1:
 - fix Astyle.

include/cli.h
```diff
@@ -0,0 +1,140 @@
...
+/***************************** Include Files **********************************/
+/******************************************************************************/
+
+#include "uart.h"
+
+/******************************************************************************/
+/********************** Macros and Constants Definitions **********************/
+/******************************************************************************/
+
+#define _NC 0
```
buha, line 53: Some of these characters are not used, are you sure they are necessary ?

include/cli.h
```diff
@@ -0,0 +1,140 @@
...
+/******************************************************************************/
+
+/** CLI command type. All CLI commands must abide to this format. */
+typedef int32_t (*cmd_func)(void *, uint8_t *);
+
+/**
+ * @struct cli_init_param
+ * @brief Initialization structure for CLI.
+ */
+struct cli_init_param {
```
buha, line 73: do you expect to expand this structure with more members ?

util/cli.c
```diff
@@ -0,0 +1,430 @@
...
+#include "cli.h"
+#include "error.h"
+#include "uart_extra.h"
+#include "irq.h"
+
+/******************************************************************************/
+/************************** Variable Definitions ******************************/
+/******************************************************************************/
+
+static uint8_t uart_current_line[100];
```
buha, line 55: how long is the max command you expect to send/receive ?

util/cli.c
```diff
@@ -0,0 +1,430 @@
...
+			uart_current_line[i] = uart_previous_line[i];
+		} while(uart_previous_line[i++] != '\0');
+		cli_write_string(dev, uart_current_line);
+		uart_line_index = --i;
+		break;
+	case _NC:
+		break;
+	default:
+		uart_current_line[uart_line_index++] = c;
+		if(uart_line_index == 256) {
```
buha, line 215: are you sure this is the correct condition ? uart_line_index is uint8_t

buha:
and maybe one more general comment, aren't there any open source customizable command parsers ? this kind of mapping commands to functions is done pretty often, so maybe we could save the effort of doing it from scratch and testing it by using something proven

eggcar:
> 
> 
> and maybe one more general comment, aren't there any open source customizable command parsers ? this kind of mapping commands to functions is done pretty often, so maybe we could save the effort of doing it from scratch and testing it by using something that exists and is proven to work

Happened to see this comment in my mail subscription. 
Try [this](https://github.com/eggcar/ECShell)  parsing commands with hash map. related implementation in [ecshell_exec.c](https://github.com/eggcar/ECShell/blob/master/ecshell_exec.c)

include/cli.h
adrimbarean: I will remove those unnecessary.

include/cli.h
adrimbarean: I don't at the moment. The CLI could work on some other physical layer, but that will be implemented when necessary.

util/cli.c
adrimbarean: Some commands can get pretty long, such in the case of send random ASCII string command. Do you have a good suggestion for this?

util/cli.c
adrimbarean: I think this is a mistake. Will fix.

util/cli.c
buha: I don't have a suggestion, i just asked because it's good to follow some sort of a spec where it says what the max could be and not use an arbitrary size. This VT100 80 size seems ok as a start, maybe (optionally) even put a comment next to it as to where it comes from.

adrimbarean:
V2:
 - remove unnecessary defines;
 - fix bug in parser default;
 - change size of the line to 80 to match legacy terminals.

PR #609, opened 2020-05-21
Aducm makefile

mchindri:
This PR will be a draft until libraries support is added.
Because the makefile can be run both on windows and linux a separated file was created to don't duplicate the code.
There is also a proposed src.mk in the project directory with an approach that can be used in further projects.

### DIFF
--- a/projects/aducm_blinky_example/Makefile
+++ b/projects/aducm_blinky_example/Makefile
@@ -1,0 +1,1 @@
+include ../../tools/scripts/aducm.mk
--- a/projects/aducm_blinky_example/src.mk
+++ b/projects/aducm_blinky_example/src.mk
@@ -1,0 +1,31 @@
+################################################################################
+#									       #
+#     Shared variables:							       #
+#	- PROJECT							       #
+#	- DRIVERS							       #
+#	- INCLUDE							       #
+#	- PLATFORM_DRIVERS						       #
+#	- NO-OS								       #
+#									       #
+################################################################################
+
+#In aducm projects the name of the source dir should be different from src
+#Direcory where app srcs are stored
+
+SRC_DIRS +=	$(PROJECT)/app_src		\
+		$(PLATFORM_DRIVERS)		\
+		$(INCLUDE)			\
+		$(NO-OS)/util			\
+
+#Include makefiles from each source directory if they exist
+SUB_MAKES= $(wildcard $(addsuffix /src.mk, $(SRC_DIRS)))
+include $(SUB_MAKES)
+
+DIRECORIES_WITH_MAKEFILES = $(patsubst %/src.mk, %,$(SUB_MAKES))
+REMAINING_DIRECORIES = $(filter-out $(DIRECORIES_WITH_MAKEFILES), $(SRC_DIRS))
+
+# recursive wildcard
+rwildcard = $(wildcard $1$2) $(foreach d,$(wildcard $1*),$(call rwildcard,$d/,$2))
+
+SRCS += $(foreach dir, $(REMAINING_DIRECORIES), $(call rwildcard, $(dir),*.c))
+INCS += $(foreach dir, $(REMAINING_DIRECORIES), $(call rwildcard, $(dir),*.h))
--- a/tools/scripts/aducm.mk
+++ b/tools/scripts/aducm.mk
@@ -1,0 +1,468 @@
+#------------------------------------------------------------------------------
+#                             EXPORTED VARIABLES                               
+#------------------------------------------------------------------------------
+# Used by nested Makefils (mbedtls, fatfs, iio)
+export CFLAGS
+export CC
+export AR
+
+#------------------------------------------------------------------------------
+#                     PLATFORM SPECIFIC INITIALIZATION                               
+#------------------------------------------------------------------------------
+
+# New line variable
+define ENDL
+
+
+endef
+
+# Initialize copy_fun and remove_fun
+# Initialize CCES_HOME to default, if directory not found show error
+#	WINDOWS
+ifeq ($(OS), Windows_NT)
+SHELL=cmd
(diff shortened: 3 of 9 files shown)


projects/aducm_blinky_example/src.mk
```diff
@@ -0,0 +1,28 @@
...
+#Direcory where app srcs are stored
+APP_SRCS = app_src
+
+SRC_FOLDERS +=	$(DRIVERS)/accel/adxl345
+#		$(NO-OS)/network 	 \
+		$(DRIVERS)/sd-card	 \
+		$(NO-OS)/libraries/fatfs
+
+#Include makefiles from each source directory if they exist
+DIRECORIES_WITH_MAKEFILES = $(wildcard $(foreach dir,$(SRC_FOLDERS),$(dir)\src.mk))
```
scuciurean, line 22: Only the projects will have 'src.mk' files

projects/aducm_blinky_example/src.mk
```diff
@@ -0,0 +1,28 @@
...
+include $(DIRECORIES_WITH_MAKEFILES)
+
+#For the folders with no makefiles include all *.c and *.h
+REMAINING_DIRECORIES = $(filter-out, $(DIRECORIES_WITH_MAKEFILES), $(SRC_FOLDERS))
+SRCS += $(foreach dir, $(REMAINING_DIRECORIES), $(wildcard $(dir)/*.c))
```
scuciurean, line 27: Importing wildcards is a bad ideea

tools/scripts/aducm.mk
```diff
@@ -0,0 +1,179 @@
...
+#                           RULES                              
+#------------------------------------------------------------------------------
+
+# Build project Release Configuration
+PHONY := all
+all: update
+	cces -nosplash -application com.analog.crosscore.headlesstools \
+	-data $(WORKSPACE) \
+	-project $(PROJECT_PATH) \
+	-build Release
```
scuciurean, line 88: I don't see a way to add compiler or linker flags

projects/aducm_blinky_example/src.mk
mchindri: The idea behind this is: 
If it is a folder with sources than should not be imported it can be used a src.mk in that folder with somethin like:
if (PLATFORM = xilinx)
   SRCS += ...
else
  SRCS += ....
This way you don't have to know in each project which files are needed for a driver.
Also, If we will agree with this aproach, I think this should be moved inside the makefile and let the user define SRC_FOLDERS only.

tools/scripts/aducm.mk
mchindri: This command just builds the projects.
Flags are added in the update target like:
-append-switch compiler -D=SOMETHING.
for example $(INCLUDES) is a variable with all the -I flags

projects/aducm_blinky_example/src.mk
mchindri: Explained in the next comment

projects/aducm_blinky_example/src.mk
scuciurean: Here https://github.com/analogdevicesinc/no-OS/tree/master/projects/adrv9009/profiles are the same sources for different profiles

mchindri: In that case src.mk can look like:
SRC_FOLDERS +=	$(PROJECT)/src\
                                $(PROJECT)/profiles/desired_profiles

mchindri:
v2:
 - Some code refactoring
 - Add makfile for fatfs
 - Integrate fatfs and mbedtls libraries (libtinyiiod will be added when it will be used on aducm)

projects/aducm_blinky_example/Makefile
```diff
@@ -0,0 +1 @@
+include ../../tools/scripts/aducm.mk
```
adrimbarean, line 1: nitpick: extra line needed

mchindri: done

mchindri:
v3:
 - Now with the makefile a project can be build without creating a project. The project build system don't allways works well. This make should do that job.
 - Add new line

mchindri:
v4:
- Remove autogenerated srcs
- Make also creates a project and then builds it
- Small modification in fatfs makefile

mchindri:
v5:
 - Small fixes, now works on linux too
 - Add check for path with space and show error if so
 - Init mbedtls submodule if empty

mchindri:
v6:
- Add to gitignore folders created by makefile that are usefull to be kept in the local computer
- Add fix to make it work on Windows with cygwin (no work on Andrei's desktop :) )
- Fix asytle on blinky_example.c

tools/scripts/aducm.mk
mchindri: In the new version is more visible

mchindri:
v7:
 - Add new code on startup patch.
 - Create separated commit with the startup patch to be easier to the modification

mchindri:
v8:
- Do not remove workspace metadata on clean_all rule (was used only for development. In project development that workspace will be used and it is not good idea to clean all the configurations)
 - system.rteconfig will is keept as it is (patching will not solve the version issues)
 - startup_aducm3029_patch.c will be keept as source to remove the need for patching tools (git format-patch  or patch.exe for example when downloading a repo as zip file) . If I misunderstood the meeting decision please correct me.

mchindri:
v9:
 - rename build directory to builld_aducm to be enable to create both aducm and xilinx projects
 - set NO-OS and WORKSPACE variables depending if is no-os repo or aducm repo
 - set NO-OS and WORKSPACE with ?= . This will enable the user to set other path in the project makefile and do not brake the make.

mchindri:

PR #610, opened 2020-05-22
adrv9009 profile fixes

buha:
This PR intends to improve on profile compatibility with various boards and setups.
Prior to this, only bw 200 profile worked with zu11eg, now we also have a working bw400 and bw100 profile.

Also, zc706 + adrv9009 should fail using bw200 and bw400 because of CPLL limits. Prior to this, it was being used _overclocked_, this commit disallows usage of those profiles. bw100 still works well on this platform.

Also added some documentation in src.mk and app_config.h on what boards are supported and what profiles are compatible with what boards.


projects/adrv9009/profiles/tx_bw100_ir122p88_rx_bw100_or122p88_orx_bw100_or122p88_dc122p88/talise_config.c
```diff
@@ -209,7 +209,7 @@ taliseInit_t talInit = {
 	/* Digital Clock Settings */
 	.clocks =
 	{
-		.deviceClock_kHz = 122880,            /* CLKPLL and device reference clock frequency in kHz */
+		.deviceClock_kHz = 245760,            /* CLKPLL and device reference clock frequency in kHz */
```
buha, line 212: i think i should better try to come up with a fully working profile instead of just partially improving on this one.
will replace this commit with equivalent of https://github.com/analogdevicesinc/linux/commit/bd552e615421c58e96a71a7fe7014abd7578954a

dbogdan: OK, we'll wait merging these changes until you do this.

buha:
V2: fully working profiles on both Talis SOM and zc706 + adrv9009.

include/util.h
```diff
@@ -85,11 +85,11 @@
 
 #define BITS_PER_LONG 32
 
-#define GENMASK(h, l) ({ 										\
-		uint32_t t = (~0UL);									\
-		t = t << (BITS_PER_LONG - (h - l + 1));					\
-		t = t >> (BITS_PER_LONG - (h + 1));						\
-		t;														\
+#define GENMASK(h, l) ({ 					\
+		uint32_t t = (uint32_t)(~0UL);			\
```
mchindri, line 89: I'am not sure, but I think you can just remove remove the L to avoid the warning.

buha: I think some platforms could define U as a 16-bit and UL as 32-bit in which case this would only work up to 16-bits. Making it UL and then explicitly casting to 32-bit is safest in my opinion.

mchindri: true

PR #611, opened 2020-05-22
Ad9517 fixes

amiclaus:
This PR contains multiple fixes for the AD9517 driver:

1. Increase maximum allowed VCO frequency.
2. Fix int types
3. Fix VCO calibration
4. Fix divider 0 usage.


drivers/frequency/ad9517/ad9517.h
```diff
@@ -296,7 +296,7 @@
 #define AD9517_UPDATE_ALL_REGS			(1 << 0)
 
 #define AD9517_MIN_VCO_FREQ			1450000000
-#define AD9517_MAX_VCO_FREQ			1800000000
+#define AD9517_MAX_VCO_FREQ			3000000000
```
buha, line 299: is this true for all types of ad9517 (-1, -3, -4) ?if they have different max vco freq, it would be nice to define the max based on chip type

amiclaus: Yes, I think it would be nice to have multiple types defined in the driver. This will imply changing the driver structure adding a type member to identify the chip version. 
These macros are used only in one place, and it is the first check from the setup function.
https://github.com/analogdevicesinc/no-OS/blob/3e3dc89785d1f2225e97869be914328bdaf6f78b/drivers/frequency/ad9517/ad9517.c#L335-L337
From the point of view of the users, it will be indeed more intelligible by adding device types.

dbogdan:
I will merge these changes to make the driver working on JESDADC1, but a commit that adds support for the AD9517 type selection will be nice.

PR #612, opened 2020-05-25
Ad5592r changes

endakilgarriff:
- Remove unused gpio_map variable
- Split GPIO channel mode
- Add STOP/RESTART bits to conform with updated platform drivers
- Ignore NAK on I2C restart - NAK expected
- Add ldac_mode device variable 
- Initialize device struct with read/write ops in init funciton
- Update ADC function to allow multi channel readback
- Add macros and bit masks, organize them by register

This is a clean version of pull request #594


drivers/adc-dac/ad5592r/ad5592r-base.c
```diff
@@ -3,7 +3,7 @@
  *   @brief  Implementation of AD5592R Base Driver.
  *   @author Mircea Caprioru (mircea.caprioru@analog.com)
 ********************************************************************************
- * Copyright 2018(c) Analog Devices, Inc.
+ * Copyright 2018, 2020(c) Analog Devices, Inc.
```
buha, line 6: i would put these date changes in a separate commit

drivers/adc-dac/ad5592r/ad5593r.c
```diff
@@ -144,7 +145,13 @@ int32_t ad5593r_reg_write(struct ad5592r_dev *dev, uint8_t reg,
 	data[1] = value >> 8;
 	data[2] = value;
 
-	return i2c_write(dev->i2c, data,sizeof(data), STOP_BIT);
+	ret = i2c_write(dev->i2c, data,sizeof(data), STOP_BIT);
+
+	if (reg == AD5592R_REG_RESET && ret < 0) {
```
buha, line 157: Is this `ret < 0` indented ? I tried to find in the datasheet but failed to find how the chip is supposed to behave when a software reset is issued. If this condition is empyrical I think we should remove it.

drivers/adc-dac/ad5592r/ad5592r.c
```diff
@@ -3,7 +3,7 @@
  *   @brief  Implementation of AD5592R driver.
  *   @author Mircea Caprioru (mircea.caprioru@analog.com)
 ********************************************************************************
- * Copyright 2018(c) Analog Devices, Inc.
+ * Copyright 2018, 2020(c) Analog Devices, Inc.
```
buha, line 6: I have the same comment as in the other commit regarding the date.
This should be a separate commit changing all dates, otherwise suppose there is a commit revert due to a bug, and the reverse operation will reverse more than what you intend... :)

drivers/adc-dac/ad5592r/ad5593r.c
```diff
@@ -50,6 +50,7 @@
 
 #define STOP_BIT	1
```
buha, line 51: we use (or at least try to) a bunch of tabs to align the define values, it would be nice if they'd be aligned

drivers/adc-dac/ad5592r/ad5593r.c
```diff
@@ -83,30 +84,33 @@ int32_t ad5593r_write_dac(struct ad5592r_dev *dev, uint8_t chan,
...
+	uint8_t data[AD5593R_ADC_VALUES_BUFF_SIZE], i, num_chans = 0;
 	int32_t ret;
-	uint8_t data[3];
-	uint16_t temp;
 
 	if (!dev)
 		return FAILURE;
 
-	temp = BIT(chan);
+	for (i = 0; i <= dev->num_channels; i++) {
```
buha, line 103: we use a thing called `hweight8` for achieving this, you would simplify the code a bit by getting rid of the loop and the `AD5592R_REG_ADC_SEQ_INCL` define.

drivers/adc-dac/ad5592r/ad5592r-base.h
```diff
@@ -77,12 +77,31 @@ enum ad5592r_registers {
...
+#define AD5592R_REG_PD_CHANNEL(x)		BIT(x)
 
+#define AD5592R_REG_CTRL_ADC_PC_BUFF		BIT(9)
+#define AD5592R_REG_CTRL_ADC_BUFF_EN		BIT(8)
+#define AD5592R_REG_CTRL_CONFIG_LOCK		BIT(7)
+#define AD5592R_REG_CTRL_W_ALL_DACS		BIT(6)
+#define AD5592R_REG_CTRL_ADC_RANGE		BIT(5)
+#define AD5592R_REG_CTRL_DAC_RANGE		BIT(4)
+
+#define AD5592R_REG_ADC_SEQ_REP					BIT(9)
```
buha, line 91: You might not have known this, but we use 8 space tabs, try aligning these values to match the others around.

https://github.com/analogdevicesinc/no-OS/blob/bad1c1f2e02d2cb1bfcafc3c1c465819a842b9cd/drivers/adc-dac/ad5592r/ad5592r-base.h?ts=8

drivers/adc-dac/ad5592r/ad5593r.c
endakilgarriff: It is intended, if the check fails but a reset is what is being performed the error should be ignored. I have attached an excerpt from the datasheet (pg.30 - AD5593R). 

ad5593r_reset

buha: exactly what i was looking for but missed it, thanks

drivers/adc-dac/ad5592r/ad5592r-base.c
endakilgarriff: I'll put these into a new commit

drivers/adc-dac/ad5592r/ad5593r.c
endakilgarriff: Good idea, this will require including the util library, is that okay?

buha: I would say yes, the stuff we have in util is pretty minimal, it shouldn't frighten anybody from using it. I also did a recursive grep in drivers after `util.h` and got 29 results.

drivers/adc-dac/ad5592r/ad5592r.c
```diff
@@ -91,23 +91,29 @@ int32_t ad5592r_write_dac(struct ad5592r_dev *dev, uint8_t chan,
...
  * @param dev - The device structure.
- * @param chan - The channel number.
- * @param value - ADC value
+ * @param chans - The ADC channels to be readback
+ * @param values - ADC value array
  * @return 0 in case of success, negative error code otherwise
  */
-int32_t ad5592r_read_adc(struct ad5592r_dev *dev, uint8_t chan,
-			 uint16_t *value)
+int32_t ad5592r_read_adc(struct ad5592r_dev *dev, uint16_t chans,
```
mirceacaprioru, line 101: i would leave this function to only read one channel. To approach that was used before to add another function for multi channel read is the way to go. The only thing that was wrong about that was that it had multiple spi_reads instead of only one.

drivers/adc-dac/ad5592r/ad5592r.c
```diff
@@ -122,12 +128,12 @@ int32_t ad5592r_read_adc(struct ad5592r_dev *dev, uint8_t chan,
...
 		return ret;
 
-	ret = ad5592r_spi_wnop_r16(dev, &dev->spi_msg);
-	if (ret < 0)
-		return ret;
-
-	*value = dev->spi_msg;
-
+	for (i = 0; i < num_chans; i++) {
+		ret = ad5592r_spi_wnop_r16(dev, &dev->spi_msg);
```
mirceacaprioru, line 132: one single spi_write_and_read call should be used for multiple channel read. No need for the for loop.

endakilgarriff: I'm not sure how that would work, would it not be necessary to call multiple SPI frame reads to capture the data being generated? For instance, if 5 channels were to be read back, 5 SPI reads would need to be occur for that data to be captured and then stored in 5 elements of the array. One SPI read call would only read back a single channel. See datasheet excerpt for 2 channel readback on the AD5592R (page 33).

AD5592R_Multichannel_noRepeat

mirceacaprioru: This is the function:
```
int32_t spi_write_and_read(struct spi_desc *desc,
			   uint8_t *data,
			   uint16_t bytes_number)
```
You can send the number of bytes you want to fulfill your purpose for the spi transfer. In your case 5 channels (16 bits each) -> bytes_number = 5 * 2 = 10 bytes. Just make sure our data array is large enough to store all that data.

drivers/adc-dac/ad5592r/ad5592r-base.c
buha: I talked to the team and it seems we're very flexible about these dates in the header commentary. Consider it a matter of preferences and leave it as-is if you so wish.

endakilgarriff:
V2:
- Returned read ADC function to single ADC channel reads; Added multi channel read function
- Correct alignment of macros
-  Include util library - used in multiple ADC channel read function (hweight8) to check number of set bits
- Remove BIT(x) macro as defined in util.h

endakilgarriff:
V3:
- Change variable type that checks number of set bits

drivers/adc-dac/ad5592r/ad5592r.c
```diff
@@ -131,6 +132,51 @@ int32_t ad5592r_read_adc(struct ad5592r_dev *dev, uint8_t chan,
...
+ */
+int32_t ad5592r_multi_read_adc(struct ad5592r_dev *dev, uint16_t chans,
+			       uint16_t *values)
+{
+	int32_t ret;
+	uint8_t samples;
+	uint8_t i;
+
+	if (!dev)
+		return FAILURE;
```
amiclaus, line 151: If you decided to use `error` defines, I would use them through the entire function.

drivers/adc-dac/ad5592r/ad5593r.c
```diff
@@ -121,6 +126,50 @@ int32_t ad5593r_read_adc(struct ad5592r_dev *dev, uint8_t chan,
...
+ */
+int32_t ad5593r_multi_read_adc(struct ad5592r_dev *dev, uint16_t chans,
+			       uint16_t *values)
+{
+	uint8_t data[AD5593R_ADC_VALUES_BUFF_SIZE], i;
+	int32_t ret;
+	uint8_t samples;
+
+	if (!dev)
+		return FAILURE;
```
amiclaus, line 145: Same as above, either you use error codes through the entire function, either `return -1`

drivers/adc-dac/ad5592r/ad5592r.c
endakilgarriff: This is how errors were handled in other functions. I can change them, but that would require changing functions that I haven't changed too to keep things consistent. You're suggesting return SUCCESS rather than 0 or  alternatively changing FAILURE to -1?

amiclaus: This driver is older, that's why the error codes are not used. These were introduced later on this repository. You can just return -1 here so you don't have to change the entire driver.

Same for `ad5593r_multi_read_adc`

endakilgarriff: Return FAILURE is used elsewhere in the driver as well, in areas that I haven't altered. Should those be reverted too?

amiclaus: I had a look on the initial commit of the driver, and indeed a combination of defines raw values was used from the start.

Leave it as it is if the driver was implemented in this manner.

endakilgarriff: Perfect, will leave as is.

drivers/adc-dac/ad5592r/ad5592r.c
endakilgarriff: I don't think that will work after all. The conversations for the AD5593/2R only occur on the falling edge of the CS pin. Using a single spi_write_and_read will only cause the CS pin to fall once - meaning that even given the right size array to store the data, only data from one channel will be collected. 

spi_read_and_write_ADC_read

AD5592R_CS

This is a reupload - my original comment seems to have disappeared when I updated the pull request.

mirceacaprioru: You are correct. I missed this when checking the datasheet, it needs a SYNC to do the conversion for the next channel. Then you can do it with multiple spi_read_writes.

endakilgarriff: I have separated single and multiple channel reads into two functions, is there a need to make any other changes?

mirceacaprioru: I think this is ok. Will check the rest of the modifications.

PR #613, opened 2020-05-25
AD9517 update & fmcjesdadc1 driver integration

amiclaus:
1. Add part type check for ad9517.
2. Integrate ad9517 driver from `drivers/frequency` folder.

### DIFF
--- a/drivers/frequency/ad9517/ad9517.c
+++ b/drivers/frequency/ad9517/ad9517.c
@@ -73,6 +73,7 @@ int32_t ad9517_setup(struct ad9517_dev **device,
 		return -1;
 
 	dev->ad9517_st = init_param.ad9517_st;
+	dev->ad9517_type = init_param.ad9517_type;
 
 	/* Initializes the SPI peripheral */
 	ret = spi_init(&dev->spi_desc, &init_param.spi_init);
@@ -332,9 +333,30 @@ int64_t ad9517_vco_frequency(struct ad9517_dev *dev,
 	uint8_t prescaler = 0;
 	uint32_t reg_value = 0;
 
-	if((frequency < AD9517_MIN_VCO_FREQ) ||
-	    (frequency > AD9517_MAX_VCO_FREQ))
+	switch(dev->ad9517_type) {
+	case AD9517_1:
+		if((frequency < AD9517_1_MIN_VCO_FREQ) ||
+		    (frequency > AD9517_1_MAX_VCO_FREQ))
+			return -1;
+		break;
+	case AD9517_2:
+		if((frequency < AD9517_2_MIN_VCO_FREQ) ||
+		    (frequency > AD9517_2_MAX_VCO_FREQ))
+			return -1;
+		break;
+	case AD9517_3:
+		if((frequency < AD9517_3_MIN_VCO_FREQ) ||
+		    (frequency > AD9517_3_MAX_VCO_FREQ))
+			return -1;
+		break;
+	case AD9517_4:
+		if((frequency < AD9517_4_MIN_VCO_FREQ) ||
+		    (frequency > AD9517_4_MAX_VCO_FREQ))
+			return -1;
+		break;
+	default:
 		return -1;
+	}
 	if(dev->ad9517_st.pdata->ref_sel_pin_en)
 		/* Reference selection is made using REF_SEL pin. */
 		ref_freq = dev->ad9517_st.pdata->ref_sel_pin ?
--- a/drivers/frequency/ad9517/ad9517.h
+++ b/drivers/frequency/ad9517/ad9517.h
@@ -295,8 +295,14 @@
 /* AD9517_REG_UPDATE_ALL_REGS Definition */
 #define AD9517_UPDATE_ALL_REGS			(1 << 0)
 
-#define AD9517_MIN_VCO_FREQ			1450000000
-#define AD9517_MAX_VCO_FREQ			3000000000
+#define AD9517_1_MIN_VCO_FREQ			2300000000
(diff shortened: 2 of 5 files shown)


drivers/frequency/ad9517/ad9517.c
```diff
@@ -332,9 +333,30 @@ int64_t ad9517_vco_frequency(struct ad9517_dev *dev,
 	uint8_t prescaler = 0;
 	uint32_t reg_value = 0;
 
-	if((frequency < AD9517_MIN_VCO_FREQ) ||
-	    (frequency > AD9517_MAX_VCO_FREQ))
+	switch(dev->ad9517_type) {
+	case AD9517_1:
+		if((frequency < AD9517_1_MIN_VCO_FREQ) ||
+		    (frequency > AD9517_1_MAX_VCO_FREQ))
+			return -1;
```
buha, line 340: use `FAILURE`, it will be easier to search and replace it when we will have more detailed return error codes

amiclaus: The driver does not include `error.h`.
This change is related only to adding part types.
Including SUCCESS and FAILURE should be part of another commit.
Moreover, the current state of the driver includes functions with return values that are different from SUCCESS/FAILURE return codes.

buha: ok, sure ! will approve like this

drivers/frequency/ad9517/ad9517.c
```diff
@@ -332,9 +333,30 @@ int64_t ad9517_vco_frequency(struct ad9517_dev *dev,
...
+		break;
+	case AD9517_2:
+		if((frequency < AD9517_1_MIN_VCO_FREQ) ||
```
mchindri, line 343:
```suggestion
		if((frequency < AD9517_2_MIN_VCO_FREQ) ||
```

drivers/frequency/ad9517/ad9517.h
```diff
@@ -337,6 +343,13 @@ struct ad9517_lvpecl_channel_spec {
 	uint8_t name[16];	  // Optional descriptive channel name.
 };
 
+enum ad9517_type {
+	AD9517_1 = 0,
+	AD9517_2 = 1,
+	AD9517_3 = 2,
+	AD9517_4 = 3,
```
mchindri, line 350: Nitpick: extra , 
Also if you don't want specific values for an enum you can not initialize them. By default the will have have values from 0 to n

projects/fmcjesdadc1/src/app/fmcjesdadc1.c
```diff
@@ -280,8 +366,106 @@ int main(void)
...
+	if(status < 0) {
+		printf("Error channel ad9517_update()\n");
+		return FAILURE;
+	}
+
+	status = ad9517_read(ad9517_device, AD9517_REG_PLL_READBACK, &reg_value);
+	if (status < 0) {
+		printf("AD9517 PLL not locked!\n");
+		return FAILURE;
+	} else
```
adrimbarean, line 467: nitpick: need brackets at `else` because `if` has brackets.

amiclaus: Sure. Will do.

buha: actually `else` could be removed and code under `else` placed at the same level as the `if`

amiclaus: yes, you are right. I will remove the `else` statement.

adrimbarean: Also, I think that the register value is what validates the PLL lock here.
```suggestion
	} else if (reg_value & AD9517_DIGITAL_LOCK_DETECT) {
```
Only a successful read status does not guarantee the PLL lock happened.

amiclaus: You are right, I did check for the specific PLL lock bit at some point, but forgot to add it to the commit. Will add.

drivers/frequency/ad9517/ad9517.h
amiclaus: I know. I just wanted to make it clear enough for the user. Will remove the explicit initialization.

amiclaus:
V2:
- fixed part of the issues stated above; will push again when all of them are solved and state the modifications in the changelog

projects/fmcjesdadc1/src/app/fmcjesdadc1.c
```diff
@@ -280,8 +366,109 @@ int main(void)
...
+	if(status < 0){
+			printf("Error ad9517_update()\n");
+			return FAILURE;
+	}
+
+	status = ad9517_read(ad9517_device, AD9517_REG_PLL_READBACK, &reg_value);
+	if (status < 0) {
+		printf("Error ad9517_read()!\n");
+		return FAILURE;
+	} else if (reg_value & AD9517_DIGITAL_LOCK_DETECT) {
```
adrimbarean, line 467: So the condition here would be the opposite: `if (!(reg_value & AD9517_DIGITAL_LOCK_DETECT)) { PLL not locked }`. Sorry about the confusion. If the bit is 1 means the PLL is locked, if it is 0 it failed to lock.

amiclaus: You are right, I mentioned in the changelog that not everything was fixed since I found another issue in the meantime.

amiclaus:
V3:
- Fix min/max values for the AD9517-2 part.
- Fix PLL lock detection

The PR is ready for review.

PR #614, opened 2020-05-25
fix gpio_get_direction output when using PL

### LINKED ISSUE #608: GPIO get direction returns wrong value when using PL gpio
function int32_t gpio_get_direction(struct gpio_desc *desc, uint8_t *direction) in drivers/platform/xilinx/gpio.c returns wrong direction when using PL gpio. 

maybe should be like this: 
*direction = ((XGpio_GetDataDirection(extra->instance,channel) >> pin) & 0x01) ^ 0x1;

buha:
From XGpio_GetDataDirection documentation:
	Bits set to 0 are output and bits set to 1 are input.

no-OS (and also xilinx PS) convention is:
GPIO_IN = 0
GPIO_OUT = 1

Fixes #608

### DIFF
--- a/drivers/platform/xilinx/gpio.c
+++ b/drivers/platform/xilinx/gpio.c
@@ -344,7 +344,8 @@ int32_t gpio_get_direction(struct gpio_desc *desc,
 			pin -= 32;
 		} else
 			channel = 1;
-		*direction = (XGpio_GetDataDirection(extra->instance,channel) >> pin) & 0x01;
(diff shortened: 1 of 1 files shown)


buha:
@amiclaus your proposal won't work because `XGpioPs_GetDirectionPin` is used and returns inverted values for direction. So either we invert when calling `XGpio_GetDataDirection` and we leave the defines as they are or we invert when we call `XGpioPs_GetDirectionPin` use the defines you proposed.

buha:
Fixed astyle, complained about long line...

PR #621, opened 2020-05-29
Rename network folder to connectivity and restructure it

mchindri:
Changes:
 - Restructure folder to be easy to navigate through it
 - move network headers files from include next to .c files

I think this new folder structure is easier to understand than the previous one


buha:
I think renaming `network` to `connectivity` is extra, we settled on `network` so let's leave it as-is.

buha:
Why is `MQTT_paho_files` written with caps, while other files start with `mqtt*` ? Mixing up cases is annoying when using command line, let's stick to lower case if possible.

mchindri:
> I think renaming `network` to `connectivity` is extra, we settled on `network` so let's leave it as-is.

As I wrote in the commit message, I rename it in order to have inside a folder network. A suggestion for the inside folder is welcome

mchindri:
> Why is `MQTT_paho_files` written with caps, while other files start with `mqtt*` ? Mixing up cases is annoying when using command line, let's stick to lower case if possible.

MQTT_paho with caps are external, not written by us. I didn't rename the files in this pr. I just move them.

mchindri:
v2
 - Do not rename network and move wifi folder.
 - Final: 
image
 - MQTT paho will be moved in an other pull request

PR #622, opened 2020-05-29
drivers:platform:aducm3029: Uart read and write more than 1024

mchindri:
The dfp is able to read or write 1024 bytes in each transaction.
With this commit, uart_read and uart_write will run multiple transactions
if bytes_number is bigger than 1024

### DIFF
--- a/drivers/platform/aducm3029/uart.c
+++ b/drivers/platform/aducm3029/uart.c
@@ -271,15 +271,16 @@ static void uart_callback(void *ctx, uint32_t event, void *buff)
 int32_t uart_read(struct uart_desc *desc, uint8_t *data,
 		  uint32_t bytes_number)
 {
-	uint32_t errors;
-	struct aducm_uart_desc *extra;
+	struct aducm_uart_desc	*extra;
+	uint32_t		errors;
+	uint32_t		to_read;
+	uint32_t		idx;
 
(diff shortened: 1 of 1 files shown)


drivers/platform/aducm3029/uart.c
```diff
@@ -288,14 +289,19 @@ int32_t uart_read(struct uart_desc *desc, uint8_t *data,
...
-	if (ADI_UART_SUCCESS != adi_uart_Read(
-		    (ADI_UART_HANDLE const)extra->uart_handler,
-		    (void *const)data,
-		    (uint32_t const)bytes_number,
-		    bytes_number > 4 ? true : false,
-		    &errors))
-		goto failure;
-
+	idx = 0;
+	while (bytes_number) {
```
cristiantpop, line 293: It seems to me, that this function is becoming blocking now, regardless of `extra->read_desc.is_nonblocking` value

mchindri: This function is always blocking. There is uart_read/write_nonblocking for nonblocking operations.
The extra->read_desc.is_nonblocking is a flag used to know if a nonblocking is executed. And wait for it to start a new transaction.  A blocking transaction can't begin if a nonblocking is in process.

PR #630, opened 2020-06-04
projects: ad9801: change IIO define naming

amiclaus:
Use `IIO_EXAMPLE` naming for the iio application option of the ad9081
project.

This will allign the naming with the rest of the projects that support
IIO integration.

### DIFF
--- a/projects/ad9081/Makefile
+++ b/projects/ad9081/Makefile
@@ -7,7 +7,7 @@ else
 include ../../tools/scripts/linux.mk
 endif
 ifeq (y,$(strip $(TINYIIOD)))
-	CFLAGS += -D IIO_SUPPORT
+	CFLAGS += -D IIO_EXAMPLE
 endif
 ifeq (y,$(strip $(QUAD_MXFE)))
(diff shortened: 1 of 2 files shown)


dbogdan:
Since this is not only an example, but a different way of interfacing with the board, I think IIO_SUPPORT is more appropriate than IIO_EXAMPLE, so I would better rename IIO_EXAMPLE to IIO_SUPPORT...

buha:
@dbogdan but then we need to do it for all projects

amiclaus:
Since this one was the only project with IIO_SUPPORT naming, I prefferred to change it instead of changing all the other projects to IIO_SUPPORT.

But indeed, I find IIO_SUPPORT macro more suggestive, as @dbogdan  said.

I can do the changes through the entire projects folder if we decide to do so.

dbogdan:
If everyone agrees, let's use IIO_SUPPORT everywhere.

buha:
Ok, let's do that and replace this pull request with one that renames all occurences of IIO_EXAMPLE to IIO_SUPPORT.

PR #634, opened 2020-06-05
fmcadc2: add iio application

amiclaus:
Allow the user to read, write and edit axi_adc parameters via UART.

IIO-Oscilloscope application is now compatible with the fmcadc2
project.


projects/fmcadc2/src/app/app_iio.c
```diff
@@ -0,0 +1,122 @@
...
+
+	status = iio_app_init(&iio_app_desc, &iio_app_init_par);
+	if (status < 0)
+		return status;
+
+	status = iio_axi_adc_init(&iio_axi_adc_desc, adc_init);
+	if (status < 0)
+		return status;
+
+	return iio_app(iio_app_desc);
```
cristiantpop, line 121: `iio_server_init` is more than an init if `iio_app` is called inside

amiclaus: I agree with your statement.
I followed the same approach as in:
https://github.com/analogdevicesinc/no-OS/blob/aceedabbd08feee17085c9a83c4188c8beb886da/projects/ad9081/src/app_iio.c#L86-L126
https://github.com/analogdevicesinc/no-OS/blob/aceedabbd08feee17085c9a83c4188c8beb886da/projects/fmcjesdadc1/src/app/app_iio.c#L86-L125
If we decide to change it, we should do it in all projects that support IIO app.

projects/fmcadc2/src/app/app_config.h
```diff
@@ -43,4 +43,6 @@
 
 //#define XILINX_PLATFORM
 
+//#define IIO_SUPPORT
```
mchindri, line 46: Nitpcik: Typo in commit message. Used->user

amiclaus:
V2:
- fix commit description typo

PR #637, opened 2020-06-09
Ad5770r update

adrimbarean:
Add GPIO handling to the driver and correct a compilation error possibility.


drivers/dac/ad5770r/ad5770r.c
```diff
@@ -656,6 +684,17 @@ int32_t ad5770r_init(struct ad5770r_dev **device,
 	/* SPI */
 	ret = spi_init(&dev->spi_desc, &init_param->spi_init);
 
+	/* GPIO */
+	ret = gpio_get_optional(&dev->nalarm, init_param->nalarm);
```
buha, line 688: test for each `ret` and return if non zero

drivers/dac/ad5770r/ad5770r.h
```diff
@@ -319,7 +323,10 @@ struct ad5770r_dev {
 struct ad5770r_init_param {
 	/* SPI */
 	spi_init_param				spi_init;
-
+	/* GPIO */
+	struct gpio_init_param			*nalarm;
+	struct gpio_init_param			*nreset;
```
buha, line 328: This is a matter of preferences, but we also have some precedence in this regard, I would name these with gpio_ prefix like `gpio_reset` or `gpio_reset_n` in this case, given it's active low.

drivers/dac/ad5770r/ad5770r.c
```diff
@@ -631,6 +631,34 @@ int32_t ad5770r_set_monitor_setup(struct ad5770r_dev *dev,
 	return ret;
 };
 
+/**
+ * @brief Deallocate memory for the GPIOs assigned.
+ * @param dev - Device driver handler.
+ * @return SUCCESS in case of success, FAILURE otherwise.
+ */
+static int32_t ad5770r_remove_gpio(struct ad5770r_dev *dev)
```
mchindri, line 639: I think this function is not needed. You can do this inside of ad5770r_remove

drivers/dac/ad5770r/ad5770r.h
```diff
@@ -319,7 +323,10 @@ struct ad5770r_dev {
 struct ad5770r_init_param {
 	/* SPI */
 	spi_init_param				spi_init;
-
+	/* GPIO */
+	struct gpio_init_param			*nalarm;
```
mchindri, line 327: You should give a hint some where that these gpios are optional. At least in the naming, but I think is better in a commnet.

drivers/dac/ad5770r/ad5770r.c
mchindri: this will also fix  the codacy problem

adrimbarean: Will correct.

drivers/dac/ad5770r/ad5770r.h
adrimbarean: Will do this.

drivers/dac/ad5770r/ad5770r.c
```diff
@@ -632,6 +631,34 @@ int32_t ad5770r_set_monitor_setup(struct ad5770r_dev *dev,
...
+{
+	int32_t ret;
+
+	if(dev->nalarm) {
```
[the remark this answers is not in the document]
adrimbarean: Will correct regardless.

drivers/dac/ad5770r/ad5770r.c
adrimbarean: I initially wanted to include more changes in a future PR because no `ret` is actually checked in the initialization function. This function will help avoid code duplication and memory leaks when I start checking for fails in the init function. But I will add that now as well.

drivers/dac/ad5770r/ad5770r.h
adrimbarean: Will add a comment.

adrimbarean:
V1:
 - check for failure inside the initialization function;
 - change GPIO naming to be consistent with other drivers;
 - add note to emphasize that the GPIOs are optional;
 - fix `if` form.

PR #638, opened 2020-06-09
spi_engine: Remove SPI engine platform dependency.

cristiantpop:
This will allow using it on different platforms. Update related project.


drivers/platform/xilinx/spi_extra.h
```diff
@@ -102,7 +102,7 @@ typedef struct xil_spi_desc {
 /**
  * @brief Xilinx platform specific SPI platform ops structure
  */
-extern const struct spi_platform_ops xil_platform_ops;
+extern const struct spi_platform_ops spi_eng_platform_ops;
```
buha, line 105: This change breaks all projects.

PR #642, opened 2020-06-10
spi_engine: Remove SPI engine platform dependency.

cristiantpop:
This will allow using it on different platforms. Update related project.


buha:
This one looks good, it seems it's not breaking projects anymore. Tested some builds and they build ok.

PR #644, opened 2020-06-11
Aducm small makefile fixes

mchindri:
- Initialize mqtt repository if not initialized
- Add libs rule
- use variables instead of duplicated paths
- fix CLEAN_MBEDTLS for windows
- Add noos_mbedtls_config.h as dependency for mbedtls
- Refactor CFLAGS
- Add reset_ADuCM3029.s in build


mchindri:

PR #645, opened 2020-06-11
TLS Improvements

mchindri:
- Update tls for minimal configuration.
- Add support to verify server identity
- Add EINVAL error code  to error.h
- Add support  for secure client identification
- Enable PEM certificate format
- Add memory optimization options


network/tcp_socket.c
```diff
@@ -164,6 +172,28 @@ static int32_t stcp_socket_init(struct secure_socket_desc **desc,
 					  MBEDTLS_SSL_VERIFY_NONE);
 	}
 
+	if (param->cli_cert) {
+		if (!param->cli_pk) {
```
mirceacaprioru, line 188: this can be moved out of this if statement

mchindri: If a user set cli_cert then it should set also cli_pk. On the other side, if cli_cert is not set then the value of cli_pk doesn't matter. I think checking it outside wouldn't make sense in this context.

mirceacaprioru: understood, makes sense.

mchindri:
v2:
 Add missing  #include "noos_mbedtls_config.h" in tcp_socket.c

mchindri:
v3:
 - Add missing endif in aducm.mk

PR #646, opened 2020-06-11
fmcjesdadc1: add IIO support for PS

amiclaus:
Add IIO application support for PS based carriers.

Update Makefiles to support IIO integration.

Remove axi_dac from makefile.


amiclaus:
V2:

- add new line at the end of file

buha:
I tried building but there is a missing \ on line 26 of src.mk
For some reason this line was not changed in this commit.
If you change that it builds correctly.

amiclaus:
V3:
- build app_iio only if TINYIIOD is enabled.
- add missing `\` into the `src.mk` file

PR #649, opened 2020-06-12
Ad778 1 spi eng

cristiantpop:
ad7768-1: Update project with new SPI engine

A common SPI engine implementation is used now.


projects/ad7768-1fmcz/src/ad77681evb.c
```diff
@@ -1,5 +1,5 @@
 /***************************************************************************//**
-* @file ad77681_evb.c
+* @file ad77681evb.c
```
mchindri, line 2: nitpick: This is related with the previous commit

projects/ad7768-1fmcz/src/ad77681evb.c
```diff
@@ -133,31 +131,31 @@ int main()
...
+		ret = spi_engine_offload_transfer(spi_eng_desc, spi_engine_offload_message,
+						  AD77681_EVB_SAMPLE_NO);
+		if (ret != SUCCESS)
+			return ret;
 
-		data = (uint8_t*)adc_dev->spi_desc->rx_dma_startaddr;
+		mdelay(1000);
 
-		mdelay(10000);
+		Xil_DCacheInvalidateRange(spi_engine_offload_message.rx_addr,
```
mchindri, line 152: Shouldn't #ifdef XILINX_PLATFORM be used for this function and 	Xil_ICacheEnable();	Xil_DCacheEnable();
like in other projects?

projects/ad7768-1fmcz/src/ad77681evb.c
cristiantpop: It seems to me that this is the name of the file, otherwise doxygen will complain.

projects/ad7768-1fmcz/src/ad77681evb.c
cristiantpop: `XILINX_PLATFORM ` comes from an `app_config.h` in other projects. ad7768-1fmcz project, it doesn't have such a config file yet. Probably it should be added in a future commit.

projects/ad7768-1fmcz/src/ad77681evb.c
mchindri: I thought you have renamed it, it is ok then.

drivers/adc/ad7768-1/ad77681.c
```diff
@@ -512,7 +513,7 @@ int32_t ad77681_setup(struct ad77681_dev **device,
 	dev->crc_sel = init_param.crc_sel;
 	dev->status_bit = init_param.status_bit;
 
-	spi_init(&dev->spi_desc, init_param.spi_eng_dev_init);
+	spi_init(&dev->spi_desc, &init_param.spi_eng_dev_init);
```
buha, line 516: test status code

cristiantpop:
V2:

- Test return value for `spi_init`

cristiantpop:
I think that they are already included in here:
https://github.com/analogdevicesinc/no-OS/blob/8dd8f9e560911a29f449be3d33167f9e0e6da359/projects/ad7768-1fmcz/src.mk

cristiantpop:
V3:

- Include in the drivers makefile:
```
-I./axi_core/spi_engine \
-I./platform/xilinx
```

PR #651, opened 2020-06-12
fmcadc5: add iio application & minor fixes

amiclaus:
Allow the user to read, write and edit axi_adc parameters via UART.

IIO-Oscilloscope application is now compatible with the fmcadc5
project.

Fix minor build issues for the fmcadc5 project.


projects/fmcadc5/src/app/app_iio.c
```diff
@@ -0,0 +1,126 @@
...
+	struct iio_app_init_param iio_app_init_par = {
+		.iio_server_ops = &uart_iio_server_ops,
+	};
+	struct iio_app_desc *iio_app_desc;
+	struct iio_axi_adc_desc *iio_axi_adc_0_desc;
+	struct iio_axi_adc_desc *iio_axi_adc_1_desc;
+	int32_t status;
+
+	status = uart_init(&uart_desc, &uart_init_par);
+	if (status < 0)
```
adrimbarean, line 110: nitpick: This could be implemented by the "error.h" header's `IS_ERR_VALUE(x)`.

projects/fmcadc5/src/app/app_iio.c
```diff
@@ -0,0 +1,126 @@
...
+
+	status = iio_axi_adc_init(&iio_axi_adc_0_desc, adc_0_init);
+	if (status < 0)
+		return status;
+
+	status = iio_axi_adc_init(&iio_axi_adc_1_desc, adc_1_init);
+	if (status < 0)
+		return status;
+
+	return iio_app(iio_app_desc);
```
adrimbarean, line 125: Might be worthwhile to free memory by calling the remove functions after the iio_app ends. I am not sure about this, though. Any thoughts?

amiclaus: I agree on this. But this implies adapting also the rest of the projects that were updated and merged already into master.

Therefore, I would do a separate commit/PR updating all projects.

projects/fmcadc5/src/app/app_iio.c
amiclaus: Will add.

amiclaus:
V2:
- use `IS_ERR_VALUE` for error code check.

PR #654, opened 2020-06-16
daq2: fix device spi_param.

amiclaus:
Fix devices' spi_param structures for the daq2 project.
Issue was introduced when the spi generic layer support was added.


buha:
I think we need to work on a way to catch this kind of issues in Travis.

PR #657, opened 2020-06-19
1 - Move iio files to libraries

mchindri:
- Copy iio source files to libraries/iio
- Add libraries/iio/libtinyiiod as submodule

This change will be done in order to build iio as a library because tinyiiod is used only by iio so for me it makes sense both to be in one place.
More, with this approach new project will not depend on defined symbols needed by tinyiiod.
Finally, doing this will help speed up the developement of new changes that have been decide to be done on iio and there is no need to update all projects on each change which would slow the development. When all new changes are stable, the project can be modified to use the new iio library with small changes.
During development only iio_demo project will be modified in order for the changes to be better understood.


libraries/iio/Makefile
```diff
@@ -0,0 +1,56 @@
...
+
+OBJS = $(SRCS:.c=.o)
+
+CFLAGS += $(IIO_DEFS) \
+	  -I"$(TINYIIOD_DIR)"
+
+.PHONEY = all clean
+
+#Init tiniiod submodule if not initialized
+ifeq ($(wildcard $(TINYIIOD_DIR)/README.md),)
```
scuciurean, line 40: Wildcards are used for a specific pattern, not for a single file

libraries/iio/Makefile
```diff
@@ -0,0 +1,56 @@
...
+all: $(LIB_NAME)
+endif
+
+$(LIB_NAME): $(OBJS)
+	$(AR) $(ARFLAGS) $@ $(OBJS)
+
+clean:
+	-$(call remove_fun,$(OBJS) $(LIB_NAME))
+
+AUX = $(wildcard $(TINYIIOD_DIR)/README.md)
```
scuciurean, line 54: Same here

libraries/iio/iio.c
```diff
@@ -0,0 +1,843 @@
+/***************************************************************************//**
```
amiclaus, line 1: do we want to have duplicate files?
https://github.com/analogdevicesinc/no-OS/blob/master/iio/iio.c

libraries/iio/iio.h
```diff
@@ -0,0 +1,98 @@
+/***************************************************************************//**
```
amiclaus, line 1: do we want to have duplicate files?
https://github.com/analogdevicesinc/no-OS/blob/master/iio/iio.h

libraries/iio/iio_types.h
```diff
@@ -0,0 +1,119 @@
+/***************************************************************************//**
```
amiclaus, line 1: same as above

amiclaus:
With this PR you add: `libraries/iio/libtinyiiod` as submodule. 
The submodule already exists under `libraries` therefore we will have 2x libtinyiiod submodules in the same repo (no-OS).

libraries/iio/Makefile
mchindri: In this case wildcard is used to check if file exists.

libraries/iio/Makefile
mchindri: This was for testing, I will remove it

libraries/iio/Makefile
cuciureansergiu: > In this case wildcard is used to check if file exists.

You can use this as an example

FILE=/etc/resolv.conf
if [ -f "$FILE" ]; then
    echo "$FILE exists."
else 
    echo "$FILE does not exist."
fi

Also wouldn't assume if a file named README.md exists in a folder, it world be a repository

mchindri: I think is simpler this way.
I didn't find an other way to check if the submodule is initialized.

mchindri: Also it seems what you propose is a shell scrips, which I would rather avoid inside the makefile

libraries/iio/iio.c
mchindri: yes, the motivation is in the description

mchindri:
v2:
 - Remove last commit from this PR. Will create an other PR for it

libraries/iio/Makefile
adrimbarean: You could check for the .git folder. This might be more generic.

libraries/iio/Makefile
```diff
@@ -0,0 +1,56 @@
...
+ifeq ($(wildcard $(TINYIIOD_DIR)/README.md),)
+all:
+	git submodule update --init --remote -- $(TINYIIOD_DIR)
+	$(MAKE) $(LIB_NAME)
+else
+all: $(LIB_NAME)
+endif
+
+$(LIB_NAME): $(OBJS)
+	$(AR) $(ARFLAGS) $@ $(OBJS)
```
adrimbarean, line 49: nitpick: this could also be possible and a bit more concise
```suggestion
	$(AR) $(ARFLAGS) $@ $^
```

mchindri:
v3:
 - Move iio files to libraries/iio with git mv and then copy them back to iio. This way changes can be tracked in libraries/iio
 - Remove Makefile commit  from this pull request. Add it here https://github.com/analogdevicesinc/no-OS/pull/674

libraries/iio/Makefile
mchindri: will solve here: https://github.com/analogdevicesinc/no-OS/pull/674

libraries/iio/Makefile
mchindri: will solve here: https://github.com/analogdevicesinc/no-OS/pull/674

libraries/iio/Makefile
mchindri: will solve here: https://github.com/analogdevicesinc/no-OS/pull/674

PR #661, opened 2020-06-22
projects: ad9172: add iio support

amiclaus:
Add IIO support for ad9172 project.

Allow the user to read/write via UART using IIO-Oscilloscope app.

### DIFF
--- a/projects/ad9172/src/main.c
+++ b/projects/ad9172/src/main.c
@@ -51,6 +51,7 @@
 #include "inttypes.h"
 #include "error.h"
 #include <xparameters.h>
+#include <xil_cache.h>
 #include "app_config.h"
 #include "spi.h"
 #include "spi_extra.h"
@@ -60,6 +61,10 @@
 #ifdef DAC_DMA_EXAMPLE
 #include "axi_dmac.h"
 #endif /* DAC_DMA_EXAMPLE */
+
+#ifdef IIO_SUPPORT
+#include "app_iio.h"
+#endif
 
 int main(void)
 {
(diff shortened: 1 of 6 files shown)


projects/ad9172/src/main.c
```diff
@@ -291,6 +296,32 @@ int main(void)
 	axi_dac_set_datasel(tx_dac, -1, AXI_DAC_DATA_SEL_DDS);
 #endif /* DAC_DMA_EXAMPLE */
 
+#ifdef IIO_SUPPORT
+	printf("The board accepts libiio clients connections through the serial backend.\n");
```
buha, line 300: I think this message should be put under iio_server_init, otherwise we duplicate it in each project.

amiclaus: Indeed. But either it is duplicated here or duplicated in the app_iio.c file, since every project has it's own `iio_server_init` function (the number of function parameters and the type may vary from project to project).

projects/ad9172/src/main.c
```diff
@@ -291,6 +296,32 @@ int main(void)
...
+	struct axi_dmac *tx_dmac;
+
+	axi_dmac_init(&tx_dmac, &tx_dmac_init);
+
+	struct iio_axi_dac_init_param iio_axi_dac_init_par;
+	iio_axi_dac_init_par = (struct iio_axi_dac_init_param) {
+		.tx_dac = tx_dac,
+		.tx_dmac = tx_dmac,
+		.dac_ddr_base = DDR_MEM_BASEADDR,
+#ifndef PLATFORM_MB
```
dbogdan, line 317: According to https://github.com/analogdevicesinc/no-OS/commit/b438e7c075dd662925bfbf27123313ba371542c2, this should not be present.

amiclaus: You are right. Will change.

amiclaus:
V2:

- remove `PLATFORM_MB` define check from main app.

PR #667, opened 2020-06-23
2 - tools:scripts:aducm.mk: Add support for new iio library

mchindri:
Only last commit relevant.
Built on top of #657

### DIFF
--- a/tools/scripts/aducm.mk
+++ b/tools/scripts/aducm.mk
@@ -223,9 +223,35 @@ $(MQTT_LIB):
 
 endif
 
+#	IIO
+IIO_DIR = $(NO-OS)/libraries/iio
+#If iio is found in SRC_DIRS
+ifneq ($(if $(findstring $(IIO_DIR), $(SRC_DIRS)), 1),)
+#Remove iio from srcdirs because it is a library
+SRC_DIRS := $(filter-out $(IIO_DIR), $(SRC_DIRS))
+
+IIO_LIB		= $(IIO_DIR)/libiio.a
+
+LIBS_DIRS	+= "$(IIO_DIR)"
+LIBS		+= $(notdir $(IIO_DIR))
+INCLUDE_DIRS	+= $(IIO_DIR)
+
+### This shouldn't be needed when all iio changes have been made
+#This iio.h shuldn't include libtinyiiod.h
+INCLUDE_DIRS	+= $(IIO_DIR)/libtinyiiod
+CFLAGS		+= -D_USE_STD_INT_TYPES
+
+###
+
+CLEAN_IIO	= $(MAKE) -C $(IIO_DIR) clean
+$(IIO_LIB):
+	$(MAKE) -C $(IIO_DIR)
+
+endif
+
 LIB_FLAGS = $(addprefix -l,$(subst lib,,$(basename $(LIBS))))
 LIB_DIR_FLAGS = $(addprefix -L,$(LIBS_DIRS))
-LIB_TARGETS = $(MBEDTLS_TARGETS) $(FATFS_LIB) $(MQTT_LIB)
+LIB_TARGETS = $(MBEDTLS_TARGETS) $(FATFS_LIB) $(MQTT_LIB) $(IIO_LIB)
 
 #------------------------------------------------------------------------------
 #                           UTIL FUNCTIONS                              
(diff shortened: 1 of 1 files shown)


libraries/iio/iio.c
```diff
@@ -0,0 +1,843 @@
+/***************************************************************************//**
```
buha, line 1: Let's use `git mv` or something that keeps track of history when renaming (moving) files around.

tools/scripts/aducm.mk
```diff
@@ -223,9 +223,35 @@ $(MQTT_LIB):
 
 endif
 
+#	IIO
```
adrimbarean, line 226: Question: this assumes a project treats the IIO as sources instead of a library and changes that. If we pass this change should the projects treat iio as a library directly instead of sources?

libraries/iio/iio.c
mchindri: https://github.com/analogdevicesinc/no-OS/pull/673 how about this?

tools/scripts/aducm.mk
```diff
@@ -223,9 +223,35 @@ $(MQTT_LIB):
...
+INCLUDE_DIRS	+= $(IIO_DIR)
+
+### This shouldn't be needed when all iio changes have been made
+#This iio.h shuldn't include libtinyiiod.h
+INCLUDE_DIRS	+= $(IIO_DIR)/libtinyiiod
+CFLAGS		+= -D_USE_STD_INT_TYPES
+
+###
+
+CLEAN_MQTT	= $(MAKE) -C $(IIO_DIR) clean
```
buha, line 246: MQTT ?

tools/scripts/aducm.mk
```diff
@@ -223,9 +223,35 @@ $(MQTT_LIB):
 
 endif
 
+#	IIO
+IIO_DIR = $(NO-OS)/libraries/iio
+#If mqtt is found in SRC_DIRS
```
buha, line 228: mqtt ?
Perhaps a more general comment would be to make the comments in the makefile more homogenous in style and also make sure they are updated.

tools/scripts/aducm.mk
```diff
@@ -223,9 +223,35 @@ $(MQTT_LIB):
...
+ifneq ($(if $(findstring $(IIO_DIR), $(SRC_DIRS)), 1),)
+#Remove iio from srcdirs because it is a library
+SRC_DIRS := $(filter-out $(IIO_DIR), $(SRC_DIRS))
+
+IIO_LIB		= $(IIO_DIR)/libiio.a
+
+LIBS_DIRS	+= "$(IIO_DIR)"
+LIBS		+= $(notdir $(IIO_DIR))
+INCLUDE_DIRS	+= $(IIO_DIR)
```
buha, line 237: LIBS_DIRS has quotes. Is it for whitespaces in paths ? Then perhaps INCLUDE_DIRS should also have quotes.

libraries/iio/iio.c
mchindri: Solved here: https://github.com/analogdevicesinc/no-OS/pull/657

tools/scripts/aducm.mk
mchindri: copy paste miss :)

tools/scripts/aducm.mk
mchindri: copy paste miss :)

tools/scripts/aducm.mk
mchindri: If libraries/iio is found in src_dirs will use libraries/iio library. 
I am sure other aproach will fit better.
But for now, for consistency with the rest of the makefile I prefere to keep them this way and in an other PR  use an other approach for the rest of the libraries too.

tools/scripts/aducm.mk
mchindri: The quotes were needed to link the libraries to a project.
In this PR I will keep this way for consistency, but will fix them in an other PR with suggestions from Andrei too

tools/scripts/aducm.mk
mchindri: I will come with a generic approach to link libraries in an other PR and I will put a generic comment there

mchindri:
v2: 
 - Rebase on top of new updates on the other branches
 - Fix copy-paste errors
 - Add $(IIO_CLEAN) to clean rule

mchindri:
v3: 
 - Fix copy-paste errors
 - Add $(IIO_CLEAN) to clean rule

mchindri:

PR #671, opened 2020-06-23
tools:scripts:platform:xilinx: update windows script

adrimbarean:
Update windows script to not create duplicates PATH environment variables
and only affect the variable locally to avoid permanently corrupting it.
The way to use it is the same, only the extension has been changed.


mchindri:
whould .ps1 work if I try to run it from power shell?

PR #688, opened 2020-07-10
projects: ad9208: add iio support

amiclaus:
Add IIO support for ad9208 project.

Allow the user to read/write via UART using IIO oscilloscope app.

### DIFF
--- a/projects/ad9208/src/main.c
+++ b/projects/ad9208/src/main.c
@@ -61,6 +61,10 @@
 #include "gpio.h"
 #include "gpio_extra.h"
 
+#ifdef IIO_SUPPORT
+#include "app_iio.h"
+#endif
+
 int main(void)
 {
 	struct xil_spi_init_param xil_spi_param = {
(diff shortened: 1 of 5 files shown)


projects/ad9208/src/main.c
```diff
@@ -488,6 +492,26 @@ int main(void)
...
 		xil_printf("axi_dmac_transfer() error: %"PRIi32"\n", status);
 
+#ifdef IIO_SUPPORT
+	printf("The board accepts libiio clients connections through the serial backend.\n");
+
+	struct iio_axi_adc_init_param iio_axi_adc_0_init_par;
+	iio_axi_adc_0_init_par = (struct iio_axi_adc_init_param) {
+		.rx_adc = rx_0_adc,
+		.rx_dmac = rx_dmac,
+		.adc_ddr_base = DDR_MEM_BASEADDR + 0x800000,
```
adrimbarean, line 502: Will using the same `adc_ddr_base` for both ADCs not overwrite the data from one of them in case they sample simultaneously? In short, can they both sample at the same time in this configuration?

amiclaus: image

Red trace is channel 1 from the first adc
Orange trace is channel 1 from the second adc
Both adc are running in the ramp test mode.
As you may notice, the samples of the two channels are interspersed.

PR #689, opened 2020-07-13
platform: generic: gpio: update gpio_get

amiclaus:
Update `gpio_get` function in the GPIO generic driver to align it with
the gpio header file from `/include`.


amiclaus:
V2:
- fix TravisCI complaint

PR #670, opened 2020-06-23
6 - iio:libraries: Replace iio_interfaces with list

mchindri:
Only last commit is relevant
Built on top of: #666

PR #692, opened 2020-07-15
7 -  libraries:iio: Generate xml from iio_interface 

mchindri:
Built on top of: https://github.com/analogdevicesinc/no-OS/pull/670

Last 2 commits relevant

### DIFF
--- a/libraries/iio/iio.c
+++ b/libraries/iio/iio.c
@@ -55,6 +55,30 @@
 /*************************** Types Declarations *******************************/
 /******************************************************************************/
 
+static char header[] =
+	"<?xml version=\"1.0\" encoding=\"utf-8\"?>"
+	"<!DOCTYPE context ["
+	"<!ELEMENT context (device | context-attribute)*>"
+	"<!ELEMENT context-attribute EMPTY>"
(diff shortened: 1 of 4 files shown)


mchindri:
v2: 
 - Rebase onto master
 - apply astyle

mchindri:
v3:
 - Fix cppcheck
 - Keep only 3 values for iio_chan_type

mchindri:
v4:
 - add missing ;

libraries/iio/iio.c
```diff
@@ -742,6 +696,103 @@ ssize_t iio_step(struct iio_desc *desc)
...
+		n = 0;
+	else
+		n = buff_size;
+
+	if (buff == NULL)
+		/* Set dummy value for buff. It is used only for counting */
+		buff = ch_id;
+
+	i = 0;
+	i += snprintf(buff, max(n - i, 0),
```
buha, line 741: snprintf may return a negative number and you assume all returns are positive

mchindri: I researched a bit and from the documentation, negative error will be returned when a **encoding ** error occurs and encoding only occurs when using unicode strings with "%ls". I use only ascii string so no encoding happens.
I tested with all ascii characters and no error.

mchindri:

PR #693, opened 2020-07-15
includes:error.h: Include <errno.h>

mchindri:
EAGAIN and EINVAlL are inside errno.h and other error codes that can be used.
This way we don't need to redefine each error code.

### DIFF
--- a/include/error.h
+++ b/include/error.h
@@ -40,6 +40,8 @@
 #ifndef ERROR_H_
 #define ERROR_H_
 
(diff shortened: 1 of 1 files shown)


mirceacaprioru:
We don't have errno.h in our repo. Won't this give an error about the include ?

mchindri:
> We don't have errno.h in our repo. Won't this give an error about the include ?

It is a standard C header.

PR #694, opened 2020-07-15
iio_wifi_backend 1: Update circular buffer implementation

mchindri:
Circular buffer will be used for wifi implementation instead of the custom implementation that is used right now.
This change was done in order to implement server operations to be used in iio.
Wifi changes will be in a new PR

### DIFF
--- a/util/circular_buffer.c
+++ b/util/circular_buffer.c
@@ -312,6 +312,77 @@ static int32_t cb_operation(struct circular_buffer *desc,
 }
 
 /**
+ * @brief Prepare asynchronous write
+ *
+ * Get the inside raw buffer to be used in DMA transactions.
+ *
+ * @param desc - Circular buffer reference
+ * @param size_to_write - Number of bytes needed to write to the buffer.
+ * @param write_buff - Address where to store the buffer where to write to.
+ * @param size_avilable - min(size_to_write, size until end of allocated buffer)
+ * @return
+ *  - \ref SUCCESS   - No errors
+ *  - -EINVAL   - Wrong parameters used
+ *  - -EBUSY    - Asynchronous transaction already started
+ */
+int32_t cb_prepare_async_write(struct circular_buffer *desc,
+			       uint32_t size_to_write,
+			       void **write_buff,
+			       uint32_t *size_avilable)
+{
+	return cb_prepare_async_operation(desc, size_to_write, write_buff,
+					  size_avilable, 0);
+}
+
+/**
+ * @brief Prepare asynchronous read
+ *
+ * Get the inside raw buffer to be used in DMA transactions.
+ *
+ * @param desc - Circular buffer reference
+ * @param size_to_read - Number of bytes needed to write to the buffer.
+ * @param read_buff - Address where to store the buffer where data will be read.
+ * @param size_avilable - min(size_to_read, size until end of allocated buffer)
+ * @return
+ *  - \ref SUCCESS   - No errors
+ *  - -EAGAIN   - No data available at this moment
+ *  - -EINVAL   - Wrong parameters used
+ *  - -EBUSY    - Asynchronous transaction already started
+ *  - -EOVERRUN - An overrun occurred and some data have been overwritten
+ */
+int32_t cb_prepare_async_read(struct circular_buffer *desc,
+			      uint32_t size_to_read,
+			      void **read_buff,
+			      uint32_t *size_avilable)
+{
+	return cb_prepare_async_operation(desc, size_to_read, read_buff,
+					  size_avilable, 1);
+}
+
+/**
+ * \defgroup end_async_group End Ashyncronous functions
+ * @brief End asynchronous transaction
+ *
+ * @param desc - Circular buffer reference
+ * @return
+ *  - \ref SUCCESS   - No errors
+ *  - \ref FAILURE   - Asynchronous transaction not started
+ *  - -EINVAL        - Wrong parameters used
+ * @{
+ */
+int32_t cb_end_async_write(struct circular_buffer *desc)
+{
+	return cb_end_async_operation(desc, 0);
+}
+
+int32_t cb_end_async_read(struct circular_buffer *desc)
+{
+
+	return cb_end_async_operation(desc, 1);
+}
+/** @} */
+
+/**
  * @brief Write data to the buffer (Blocking)
  * @param desc - Circular buffer reference
  * @param data - Buffer from where data is copied to the circular buffer
--- a/util/circular_buffer.c
+++ b/util/circular_buffer.c
@@ -1,5 +1,5 @@
 /***************************************************************************//**
- *   @file   circular_buffer.h
+ *   @file   circular_buffer.c
  *   @brief  Circular buffer implementation
  *   @author Mihail Chindris (mihail.chindris@analog.com)
 ********************************************************************************
@@ -54,22 +54,33 @@
 /******************************************************************************/
 
 /**
+ * @struct cb_ptr
+ * @brief Circular buffer pointer
+ */
+struct cb_ptr {
+	/** Index of data in the buffer */
+	uint32_t	idx;
+	/** Counts the number of times idx exceeds the liniar buffer */
+	uint32_t	spin_count;
+	/** Set if async transaction is active */
+	bool		async_started;
+	/** Number of bytes to update after an async transaction is finished */
+	uint32_t	async_size;
+};
+
+/**
  * @struct circular_buffer
  * @brief Circular buffer descriptor
  */
 struct circular_buffer {
 	/** Size of the buffer in bytes */
 	uint32_t	size;
-	/** Number of user elements in the buffer */
-	uint32_t	nb_elem;
-	/** Size of an user element */
-	uint32_t	elem_size;
-	/** Address to the buffer */
+	/** Address of the buffer */
 	int8_t		*buff;
-	/** Pointer to reading location */
-	int8_t		*read_ptr;
-	/** Pointer to writing location */
-	int8_t		*write_ptr;
+	/** Write pointer */
+	struct cb_ptr	write;
+	/** Read pointer */
+	struct cb_ptr	read;
 };
 
 /******************************************************************************/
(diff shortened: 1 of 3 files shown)


mchindri:
v2: 
- Fix codacy

mchindri:
v3:
 - Add EOVERRUN error to error.h

util/circular_buffer.c
```diff
@@ -146,12 +141,148 @@ int32_t cb_remove(struct circular_buffer *desc)
...
+{
+	uint32_t overflows;
+	uint32_t overflowed_size;
+
+	if (!desc || !size)
+		return -EINVAL;
+
+	if (desc->write.nb_overflows < desc->read.nb_overflows)
+		overflows = UINT32_MAX - desc->read.nb_overflows +
+			    desc->write.nb_overflows;
```
adrimbarean, line 154: Would there be actually:
```suggestion
		overflows = UINT32_MAX - desc->read.nb_overflows +
			    desc->write.nb_overflows + 1;
```
To get the accurate number of overflows? Since `nb_overflows` starts at 0 after overflowing.

util/circular_buffer.c
```diff
@@ -146,12 +141,148 @@ int32_t cb_remove(struct circular_buffer *desc)
...
+	else
+		overflows = desc->write.nb_overflows - desc->read.nb_overflows;
+
+	/* Min function is used to avoid an integer overflow. */
+	overflowed_size = desc->size * min(2, overflows);
```
adrimbarean, line 159: Which integer's overflow does the min function prevent?

mchindri: overflow * desc->size can overflwo. But there is no need to consider 2 overflows in order to get the correct size.

util/circular_buffer.c
mchindri: nice catch

mchindri:
v4:
-  update cb_size function to be more clear
- rename cb_ptr.nb_overflows to  cb_ptr.spin_count and update it description
- in cb_size rename overflows to nb_spins;
- add +1 in counting nb_spins on integer overflow.
- remove extra line
- fix comments
- add missing comments
- Add note about concurency in cb_init comment
- edit 2nd commit

util/circular_buffer.c
mchindri: II updated the cb_size implementation

util/circular_buffer.c
mchindri: solved

util/circular_buffer.c
```diff
@@ -141,105 +147,183 @@ int32_t cb_remove(struct circular_buffer *desc)
...
+		}
+
+		/* We can only read available data */
+		requested_size = min(requested_size, available_size);
+		if (!requested_size)
+			return -EAGAIN;
 	}
 
+	/* Size to end of buffer */
+	ptr->async_size = min(requested_size, desc->size - ptr->idx);
```
adrimbarean, line 219: Should this take into consideration the spin_count?

util/circular_buffer.c
```diff
@@ -141,105 +147,183 @@ int32_t cb_remove(struct circular_buffer *desc)
...
- * @param desc - Circular buffer reference
- * @param data - Buffer from where data is copied to the circular buffer
- * @param nb_elements - Number of elements to be copied
- * @return
- *  - \ref SUCCESS : On success
- *  - \ref FAILURE : If overflow, if nb_elements == 0 or if desc is NULL
- */
-int32_t cb_write(struct circular_buffer *desc, const void *data,
-		 uint32_t nb_elements)
+static int32_t cb_prepare_async_operation(struct circular_buffer *desc,
```
adrimbarean, line 183: Maybe a short comment about why this function is needed will help with future reference.

util/circular_buffer.c
```diff
@@ -141,105 +147,183 @@ int32_t cb_remove(struct circular_buffer *desc)
...
- * @param desc - Circular buffer reference
- * @param data - Buffer where to data is copied from the circular buffer
- * @param nb_elements - Number of elements to be copied
- * @return
- *  - \ref SUCCESS : On success
- *  - \ref FAILURE : If not enaught data available in the buffer
- */
-int32_t cb_read(struct circular_buffer *desc, void *data,
-		uint32_t nb_elements)
+static int32_t cb_end_async_operation(struct circular_buffer *desc,
```
adrimbarean, line 231: A comment about why this function is needed might help future users find their way around easier.

util/circular_buffer.c
```diff
@@ -141,105 +147,183 @@ int32_t cb_remove(struct circular_buffer *desc)
...
+
+	sticky_overrun = 0;
+	i = 0;
+	while (i < size) {
+		do {
+			ret = cb_prepare_async_operation(desc, size - i,
+							 (void **)&buff,
+							 &available_size,
+							 is_read);
+		} while (ret == -EBUSY || ret == -EAGAIN);
```
adrimbarean, line 282: Blocking while data or space is not available. Is this intended?

util/circular_buffer.c
```diff
@@ -141,105 +147,183 @@ int32_t cb_remove(struct circular_buffer *desc)
...
+		new_val %= desc->size;
+	}
+	ptr->idx = new_val;
+	ptr->async_size = 0;
+	ptr->async_started = false;
+
+	return SUCCESS;
+}
+
+static int32_t cb_operation(struct circular_buffer *desc,
```
adrimbarean, line 261: I think a short comment about the purpose of the function will be beneficial.

util/circular_buffer.c
```diff
@@ -299,6 +299,76 @@ static int32_t cb_operation(struct circular_buffer *desc,
...
+ *  - \ref FAILURE   - Asynchronous transaction not started
+ *  - -EINVAL        - Wrong parameters used
+ *  @{
+ */
+int32_t cb_end_async_write(struct circular_buffer *desc)
+{
+	return cb_end_async_operation(desc, 0);
+}
+
+int32_t cb_end_async_read(struct circular_buffer *desc)
```
adrimbarean, line 366: Needs commenting.

util/circular_buffer.c
mchindri: No it shouldn't. If is a read operation requested_size is set on line 213 to read a maximum of available size. If you want to write, you can write any amount of memory you want, having as limit the end of the allocated buffer (desc->size - current_idx_location)

util/circular_buffer.c
mchindri: I thought a read operation should be blocking. So yes is intended.

util/circular_buffer.c
mchindri: done

util/circular_buffer.c
mchindri: done

util/circular_buffer.c
mchindri: done

util/circular_buffer.c
mchindri: done

mchindri:
v5:
 - Add missing comments
 - Fix group comment

util/circular_buffer.c
```diff
@@ -141,105 +147,266 @@ int32_t cb_remove(struct circular_buffer *desc)
...
+ * @param desc - Circular buffer reference
+ * @param data - Buffer from where data is copied to the circular buffer
+ * @param size - Size to write
+ * @return
+ *  - \ref SUCCESS - No errors
+ *  - -EINVAL      - Wrong parameters used
+ */
+int32_t cb_write(struct circular_buffer *desc, const void *data, uint32_t size)
+{
+	return cb_operation(desc, (void *)data, size, 0);
```
amiclaus, line 396: Are these wrappers `cb_write` and `cb_read` really necessary?

Can't we just define
`#define WRITE 0`
and
`#define READ 1`

And use directly `cb_operation(desc, data, size, WRITE)` and `cb_operation(desc, data, size, READ)` 

Just an opinion.

util/circular_buffer.c
```diff
@@ -141,105 +147,266 @@ int32_t cb_remove(struct circular_buffer *desc)
...
+ * @brief Write data to the buffer (Blocking)
+ * @param desc - Circular buffer reference
+ * @param data - Buffer from where data is copied to the circular buffer
+ * @param size - Size to write
+ * @return
+ *  - \ref SUCCESS - No errors
+ *  - -EINVAL      - Wrong parameters used
+ */
+int32_t cb_write(struct circular_buffer *desc, const void *data, uint32_t size)
+{
```
amiclaus, line 395: Are these wrappers `cb_write` and `cb_read` really necessary?

Can't we just define
`#define WRITE 0`
and
`#define READ 1`

And use directly `cb_operation(desc, data, size, WRITE)` and `cb_operation(desc, data, size, READ)` 

Just an opinion.

mchindri: I thought about that, but for readability in code is better to have:
cb_read(cb, buff, size) than cb_operation(desc, data, size, WRITE);

util/circular_buffer.c
```diff
@@ -141,105 +147,195 @@ int32_t cb_remove(struct circular_buffer *desc)
...
+			   desc->write.spin_count + 1;
+
+	if (nb_spins > 0)
+		*size = desc->size + desc->write.idx - desc->read.idx;
+	else
+		*size = desc->write.idx - desc->read.idx;
+
+	if (*size > desc->size) {
+		*size = desc->size;
+		return -EOVERRUN;
```
buha, line 177: I don't get how this is possible.
Maybe the term is wrong ? Overrun is reading past a buffer size. This happens in hardware FIFO (think of an SPI FIFO) but I don't understand how it's possible to have this error in a circular buffer implementation where you control how much the user reads. You can block reads of larger size than what is available in the circular buffer.

mchindri: Overrun on a circular buffer occurs when you write more data than the space available and the old data is overwritten.
So, this error is returned on a read operation in order to know that some data have been overwritten. 
In the app then, you can ignore this error if old data is useless (ex audio buffers) or you know that you should use a bigger buffer.

util/circular_buffer.c
amiclaus: Are there other operations that can be performed with cb_operation besides write and read?

mchindri: no

mchindri: I can add the prototypes of cb_operation and the other functions to .h file and create define for cb_read /write like:
#define cb_read(desc, data, size) cb_operation((desc), (data), (size), 1)
#define cb_write(desc, data, size) cb_operation((desc), (data), (size), 0)
In order to do not add wrappers. But maybe the compiler optimization will solve wrapper problems, if that is the problem

PR #698, opened 2020-07-16
iio_wifi_backend 3: Add function for use tcp_socket as server

mchindri:
This functions will be used to implement network backend for iio.

**Last 1 commits** are related to this PR. The first ones will are from #694  and #696

PR #699, opened 2020-07-17
wifi: small changes

mchindri:
These are some changes that are needed by #698 but can be merged separately

- Add wifi_get_ip function
- Fix AT_SET_SERVER command
- At wifi init set esp8266 to work as an client (hotspot is the default on some at_firmwares)


mchindri:
v2:
 - Fix commit typo

PR #705, opened 2020-07-22
drivers:ad7768-1: Added ad77681_programmable_filter function

jgarlitos:
Added ad77681_programmable_filter function. This function allows the user to upload its own set of filter coefficients.
Signed-off-by: JohnNeeko Garlitos <JohnNeeko.Garlitos@analog.com>

### DIFF
--- a/drivers/adc/ad7768-1/ad77681.c
+++ b/drivers/adc/ad7768-1/ad77681.c
@@ -46,6 +46,7 @@
 #include <string.h>
 #include "ad77681.h"
 #include "error.h"
+#include "delay.h"
 
 /******************************************************************************/
 /************************** Functions Implementation **************************/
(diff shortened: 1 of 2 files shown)


drivers/adc/ad7768-1/ad77681.c
```diff
@@ -1169,6 +1169,139 @@ int32_t ad77681_apply_gain(struct ad77681_dev *dev,
...
+	/* In case the number of active filter coefficient is less */
+	/* than 56, zeros will be padded before the desired coeff. */
+	for (i = 0; i < coeff_reg_length; i++) {
+		/* Set the coeff address */
+		ret |= ad77681_spi_reg_write(dev,
+					     AD77681_REG_COEFF_CONTROL,
+					     address);
+
+		/* Wait for Twait uSeconds*/
+		_delay_us(twait);
```
mchindri, line 1240: In noos we use `udelay()` :
https://github.com/analogdevicesinc/no-OS/blob/115ced3d31d78a705c242b1bab1a3420251047cc/include/delay.h#L54
You should include "delay.h" to use it and add the to the build delay.c from drivers/platform/

jgarlitos: Done fixing the udelay.

drivers/adc/ad7768-1/ad77681.c
```diff
@@ -1169,6 +1170,139 @@ int32_t ad77681_apply_gain(struct ad77681_dev *dev,
...
+
+
+	/* Set Filter to FIR */
+	ret = ad77681_spi_write_mask(dev,
+				     AD77681_REG_DIGITAL_FILTER,
+				     AD77681_DIGI_FILTER_FILTER_MSK,
+				     AD77681_DIGI_FILTER_FILTER(AD77681_FIR));
+
+	/* Write the first access key to the ACCESS_KEY register */
+	ret |= ad77681_spi_write_mask(dev,
```
buha, line 1205: We would like to avoid this way of reacting to error code. I know there are similar examples in no-OS code to what you did but we're working on improving things.

What you should do after each call is something like:
```
if (ret < 0)
    return ret;
```

jgarlitos: Can I do this? 

	if ((check_back != 1) | (ret < 0))
		return FAILURE;

I also need to check the key bit if it is equal to 1 before I could start the coefficient writes.
image

buha: Of course you can combine, but in this case I would do it like this:
```
if ((ret < 0) || (check_back != 1))
	return FAILURE;
```
Placing the ret test first, because it doesn't make sense to test the parameters that the function outputs through pointers if the return of the function is an error.

jgarlitos: Thanks. Done implementing changes.

drivers/adc/ad7768-1/ad77681.c
```diff
@@ -46,6 +46,7 @@
 #include <string.h>
```
adrimbarean, line 46: Could you add a short description to the commit about why this change is needed?

drivers/adc/ad7768-1/ad77681.c
```diff
@@ -1169,6 +1170,139 @@ int32_t ad77681_apply_gain(struct ad77681_dev *dev,
...
+	ret = ad77681_spi_write_mask(dev,
+				     AD77681_REG_DIGITAL_FILTER,
+				     AD77681_DIGI_FILTER_FILTER_MSK,
+				     AD77681_DIGI_FILTER_FILTER(AD77681_FIR));
+
+	/* Write the first access key to the ACCESS_KEY register */
+	ret |= ad77681_spi_write_mask(dev,
+				      AD77681_REG_ACCESS_KEY,
+				      AD77681_ACCESS_KEY_MSK,
+				      AD77681_ACCESS_KEY(key1));
```
mchindri, line 1208: I think Darius ment:
```suggestion
	ret = ad77681_spi_write_mask(dev,
				     AD77681_REG_DIGITAL_FILTER,
				     AD77681_DIGI_FILTER_FILTER_MSK,
				     AD77681_DIGI_FILTER_FILTER(AD77681_FIR));
        if (ret < 0)
           return ret;
	/* Write the first access key to the ACCESS_KEY register */
	ret = ad77681_spi_write_mask(dev,
				      AD77681_REG_ACCESS_KEY,
				      AD77681_ACCESS_KEY_MSK,
				      AD77681_ACCESS_KEY(key1));
	if (ret < 0)
	     return ret;
``` 
       and so on.

drivers/adc/ad7768-1/ad77681.c
jgarlitos: Done adding description.

drivers/adc/ad7768-1/ad77681.c
jgarlitos: @buha @mchindri  apologies for missing it. Done implementing the suggested implementation.

PR #707, opened 2020-07-23
drivers:photo-electronic: add ADPD410x driver

adrimbarean:
Add ADPD410x device driver for the no-OS platform.
More information on the part at:
https://www.analog.com/media/en/technical-documentation/data-sheets/adpd4100-4101.pdf

### DIFF
--- a/drivers/photo-electronic/adpd410x/adpd410x.c
+++ b/drivers/photo-electronic/adpd410x/adpd410x.c
@@ -1,0 +1,659 @@
+/***************************************************************************//**
+ *   @file   adpd410x.c
+ *   @author Andrei Drimbarean (andrei.drimbarean@analog.com)
+********************************************************************************
+ * Copyright 2020(c) Analog Devices, Inc.
+ *
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *  - Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ *  - Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in
+ *    the documentation and/or other materials provided with the
+ *    distribution.
+ *  - Neither the name of Analog Devices, Inc. nor the names of its
+ *    contributors may be used to endorse or promote products derived
+ *    from this software without specific prior written permission.
+ *  - The use of this software may or may not infringe the patent rights
+ *    of one or more patent holders.  This license does not release you
+ *    from the requirement that you obtain separate licenses from these
+ *    patent holders to use this software.
+ *  - Use of the software either in source or binary form, must be run
+ *    on or directly connected to an Analog Devices Inc. component.
+ *
+ * THIS SOFTWARE IS PROVIDED BY ANALOG DEVICES "AS IS" AND ANY EXPRESS OR
+ * IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, NON-INFRINGEMENT,
+ * MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED.
+ * IN NO EVENT SHALL ANALOG DEVICES BE LIABLE FOR ANY DIRECT, INDIRECT,
+ * INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, INTELLECTUAL PROPERTY RIGHTS, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+*******************************************************************************/
+
+/******************************************************************************/
+/***************************** Include Files **********************************/
+/******************************************************************************/
+
+#include <stdlib.h>
+#include <stdint.h>
+#include <string.h>
+#include "error.h"
+#include "util.h"
+#include "adpd410x.h"
+
+/******************************************************************************/
+/************************ Functions Definitions *******************************/
+/******************************************************************************/
+
+/**
+ * @brief Read device register.
+ * @param dev - Device handler.
+ * @param address - Register address.
+ * @param data - Pointer to the register value container.
+ * @return SUCCESS in case of success, FAILURE otherwise.
+ */
+int32_t adpd410x_reg_read(struct adpd410x_dev *dev, uint16_t address,
+			  uint16_t *data)
+{
+	int32_t ret;
+	uint8_t buff[] = {0, 0, 0, 0};
+
+	switch (dev->dev_type) {
+	case ADPD4100:
+		buff[0] = field_get(ADPD410X_UPPDER_BYTE_SPI_MASK, address);
+		buff[1] = (address << 1) & ADPD410X_LOWER_BYTE_SPI_MASK;
+
+		ret = spi_write_and_read(dev->dev_ops.spi_phy_dev, buff, 4);
+		if(ret != SUCCESS)
+			return FAILURE;
+		break;
+	case ADPD4101:
+		buff[0] = field_get(ADPD410X_UPPDER_BYTE_I2C_MASK, address);
+		buff[0] |= 0x80;
+		buff[1] = address & ADPD410X_LOWER_BYTE_I2C_MASK;
+
+		/* No stop bit */
+		ret = i2c_write(dev->dev_ops.i2c_phy_dev, buff, 2, 0);
+		if(ret != SUCCESS)
+			return FAILURE;
+		ret = i2c_read(dev->dev_ops.i2c_phy_dev, (buff + 2), 2, 1);
+		if(ret != SUCCESS)
+			return FAILURE;
+		break;
+	default:
+		return FAILURE;
+	}
+
+	*data = ((uint16_t)buff[2] << 8) & 0xff00;
+	*data |= buff[3] & 0xff;
+
+	return SUCCESS;
+}
+
+/**
+ * @brief Write device register.
+ * @param dev - Device handler.
+ * @param address - Register address.
+ * @param data - New register value.
+ * @return SUCCESS in case of success, FAILURE otherwise.
+ */
+int32_t adpd410x_reg_write(struct adpd410x_dev *dev, uint16_t address,
+			   uint16_t data)
+{
+	uint8_t buff[] = {0, 0, 0, 0};
+
+	switch (dev->dev_type) {
+	case ADPD4100:
+		buff[0] = field_get(ADPD410X_UPPDER_BYTE_SPI_MASK, address);
+		buff[1] = (address << 1) & ADPD410X_LOWER_BYTE_SPI_MASK;
+		buff[2] = field_get(0xff00, data);
+		buff[3] = data & 0xff;
+
+		return spi_write_and_read(dev->dev_ops.spi_phy_dev, buff, 4);
+	case ADPD4101:
+		buff[0] = field_get(ADPD410X_UPPDER_BYTE_I2C_MASK, address);
+		buff[0] |= 0x80;
+		buff[1] = address & ADPD410X_LOWER_BYTE_I2C_MASK;
+		buff[2] = field_get(0xff00, data);
+		buff[3] = data & 0xff;
+
+		return i2c_write(dev->dev_ops.i2c_phy_dev, buff, 4, 1);
+	default:
+		return FAILURE;
+	}
+}
+
+/**
+ * @brief Do a read and write of a register to update only part of a register.
+ * @param dev - Device handler.
+ * @param address - Address of the register.
+ * @param data - Value to be written to the device.
+ * @param mask - Mask of the bit field to update.
+ * @return SUCCESS in case of success, FAILURE otherwise.
+ */
+int32_t adpd410x_reg_write_mask(struct adpd410x_dev *dev, uint16_t address,
+				uint16_t data, uint16_t mask)
+{
+	int32_t ret;
+	uint16_t reg_val;
+	uint32_t bit_pos;
+
+	ret = adpd410x_reg_read(dev, address, &reg_val);
+	if (ret != SUCCESS)
+		return FAILURE;
+	reg_val &= ~mask;
+	bit_pos = find_first_set_bit((uint32_t)mask);
+	reg_val |= (data << bit_pos) & mask;
+
+	return adpd410x_reg_write(dev, address, reg_val);
+}
+
+/**
+ * @brief Update the clocking option of the device.
+ * @param dev - Device handler.
+ * @return SUCCESS in case of success, FAILURE otherwise.
+ */
+static int32_t adpd410x_get_clk_opt(struct adpd410x_dev *dev)
+{
+	int32_t ret;
+	uint16_t reg_temp;
+
+	ret = adpd410x_reg_read(dev, ADPD410X_REG_SYS_CTL, &reg_temp);
+	if(ret != SUCCESS)
+		return FAILURE;
+	dev->clk_opt = (reg_temp & BITM_SYS_CTL_ALT_CLOCKS) >>
+		       BITP_SYS_CTL_ALT_CLOCKS;
+
+	return SUCCESS;
+}
+
+/**
+ * @brief Do a software reset.
+ * @param dev - Device handler.
+ * @return SUCCESS in case of success, FAILURE otherwise.
+ */
+int32_t adpd410x_reset(struct adpd410x_dev *dev)
+{
+	int32_t ret;
+
+	ret = adpd410x_reg_write_mask(dev, ADPD410X_REG_SYS_CTL, 1,
+				      BITM_SYS_CTL_SW_RESET);
+	if(ret != SUCCESS)
+		return ret;
+
+	return adpd410x_get_clk_opt(dev);
+}
+
+/**
+ * @brief Set operation mode.
+ * @param dev - Device handler.
+ * @param mode - New operation mode.
+ * @return SUCCESS in case of success, FAILURE otherwise.
+ */
+int32_t adpd410x_set_opmode(struct adpd410x_dev *dev,
+			    enum adpd410x_opmode mode)
+{
+	return adpd410x_reg_write_mask(dev, ADPD410X_REG_OPMODE, mode,
+				       BITM_OPMODE_OP_MODE);
+}
+
+/**
+ * @brief Set number of active time slots.
+ * @param dev - Device handler.
+ * @param timeslot_no - Last time slot to be enabled.
+ * @return SUCCESS in case of success, FAILURE otherwise.
+ */
+int32_t adpd410x_set_last_timeslot(struct adpd410x_dev *dev,
+				   enum adpd410x_timeslots timeslot_no)
+{
+	return adpd410x_reg_write_mask(dev, ADPD410X_REG_OPMODE,
+				       timeslot_no, BITM_OPMODE_TIMESLOT_EN);
+}
+
+/**
+ * @brief Set device sampling frequency.
+ * @param dev - Device handler.
+ * @param sampling_freq - New sampling frequency.
+ * @return SUCCESS in case of success, FAILURE otherwise.
+ */
+int32_t adpd410x_set_sampling_freq(struct adpd410x_dev *dev,
+				   uint32_t sampling_freq)
+{
+	int32_t ret;
+	uint32_t reg_load;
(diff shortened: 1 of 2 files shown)


adrimbarean:
V1:
 - remove memory leak in adpd410x_get_data_packet() function.

drivers/photo-electronic/adpd410x/adpd410x.c
```diff
@@ -0,0 +1,660 @@
+/***************************************************************************//**
+ *   @file   adpd410x.c
+ *   @author Andrei Drimbarean (andrei.drimbarean@analog.com)
+********************************************************************************
+ * Copyright 2019(c) Analog Devices, Inc.
```
mirceacaprioru, line 5: 2020

drivers/photo-electronic/adpd410x/adpd410x.c
```diff
@@ -0,0 +1,660 @@
...
+ * @param data - Pointer to the register value container.
+ * @return SUCCESS in case of success, FAILURE otherwise.
+ */
+int32_t adpd410x_reg_read(struct adpd410x_dev *dev, uint16_t address,
+			  uint16_t *data)
+{
+	int32_t ret;
+	uint8_t buff[] = {0, 0, 0, 0};
+
+	if(dev->dev_type == ADPD4100) {
```
mirceacaprioru, line 67: maybe use a switch here in case you add future parts to this driver it would be a lot easier to include them.

drivers/photo-electronic/adpd410x/adpd410x.c
```diff
@@ -0,0 +1,660 @@
...
+		if(ret != SUCCESS)
+			goto error_data_byte;
+	}
+	for(i = 0; i < (expect_byte_count / 2); i++) {
+		data_byte_buff[(i * 2)] = (data_buff[i] & 0xff00) >> 8;
+		data_byte_buff[(i * 2 + 1)] = data_buff[i] & 0xff;
+	}
+
+	i = 0;
+	data_byte_count = 0;
```
mchindri, line 414: Is better if you initialize sample_index here

drivers/photo-electronic/adpd410x/adpd410x.c
```diff
@@ -0,0 +1,660 @@
...
+			data_byte_count++;
+			data[sample_index] |= data_byte_buff[data_byte_count] <<
+					      24;
+			data_byte_count++;
+			data[sample_index] |= data_byte_buff[data_byte_count] <<
+					      16;
+			data_byte_count++;
+			break;
+		}
+		sample_index++;
```
mchindri, line 454:
```suggestion
		case 0:
			continue;
		case 1:
			*data = *(data_byte_buff++);
			break;
		case 2:
			*data = ((*(data_byte_buff++)) << 8) |
				    (*(data_byte_buff++))
			break;
		case 3:
			*data = (*(data_byte_buff++) << 8) |
			 	    (*(data_byte_buff++)) |
				    (*(data_byte_buff++) << 16)
			break;
		case 4:
			*data = (*(data_byte_buff++) << 8) |
			 	    (*(data_byte_buff++)) |
				    (*(data_byte_buff++) << 24) |
				    (*(data_byte_buff++) << 16)
			break;
		}
		data++;
```

mchindri: You can increment and access directly the pointers. I thinks it improves readability this way

drivers/photo-electronic/adpd410x/adpd410x.c
adrimbarean: Will update.

drivers/photo-electronic/adpd410x/adpd410x.c
```diff
@@ -0,0 +1,660 @@
...
+ */
+int32_t adpd410x_reg_write(struct adpd410x_dev *dev, uint16_t address,
+			   uint16_t data)
+{
+	uint8_t buff[] = {0, 0, 0, 0};
+
+	if(dev->dev_type == ADPD4100) {
+		buff[0] = (address & 0x7f80) >> 7;
+		buff[1] = (address & 0x7f) << 1;
+		buff[1] &= 0xfe;
```
scuciurean, line 110: I would use
`buff[1] = (addres << 1) & 0xFE;`

drivers/photo-electronic/adpd410x/adpd410x.c
adrimbarean: I will do that.

drivers/photo-electronic/adpd410x/adpd410x.c
adrimbarean: Why is it better?

drivers/photo-electronic/adpd410x/adpd410x.c
adrimbarean: Will do.

drivers/photo-electronic/adpd410x/adpd410x.c
adrimbarean: Would it be worth it to do this since I would have to free(data_byte_buff) after?

drivers/photo-electronic/adpd410x/adpd410x.c
adrimbarean: Can do.

drivers/photo-electronic/adpd410x/adpd410x.c
adrimbarean: Will fix.

drivers/photo-electronic/adpd410x/adpd410x.c
adrimbarean: Can do. Why is it better?

drivers/photo-electronic/adpd410x/adpd410x.c
adrimbarean: Is it worth it to do this if I have to `free(data_byte_buff)` afterwards?

drivers/photo-electronic/adpd410x/adpd410x.c
adrimbarean: Can do.

drivers/photo-electronic/adpd410x/adpd410x.c
buha: I dont agree that it improves readability. Operator precedence in C is a mess and one-liners that do too much are more complicated to read.

drivers/photo-electronic/adpd410x/adpd410x.c
```diff
@@ -0,0 +1,660 @@
...
+ * @param data - New register value.
+ * @return SUCCESS in case of success, FAILURE otherwise.
+ */
+int32_t adpd410x_reg_write(struct adpd410x_dev *dev, uint16_t address,
+			   uint16_t data)
+{
+	uint8_t buff[] = {0, 0, 0, 0};
+
+	if(dev->dev_type == ADPD4100) {
+		buff[0] = (address & 0x7f80) >> 7;
```
buha, line 108: I think we should use a define for 0x7f80 as it is used in several places and also perhaps a macro that performs the shifting and masking. See field_prep() function.

drivers/photo-electronic/adpd410x/adpd410x.c
```diff
@@ -0,0 +1,660 @@
...
+		/* No stop bit */
+		ret = i2c_write(dev->dev_ops.i2c_phy_dev, buff, 2, 0);
+		if(ret != SUCCESS)
+			return FAILURE;
+		ret = i2c_read(dev->dev_ops.i2c_phy_dev, (buff + 2), 2, 1);
+		if(ret != SUCCESS)
+			return FAILURE;
+	}
+
+	*data = ((uint16_t)buff[2] << 8) & 0xff00;
```
buha, line 89: nitpick: I don't think 0xff00 and 0xff masks are needed, those bits will be zero after shifting.

drivers/photo-electronic/adpd410x/adpd410x.c
```diff
@@ -0,0 +1,660 @@
...
+		return i2c_write(dev->dev_ops.i2c_phy_dev, buff, 4, 1);
+	}
+}
+
+/**
+ * @brief Do a software reset.
+ * @param dev - Device handler.
+ * @return SUCCESS in case of success, FAILURE otherwise.
+ */
+int32_t adpd410x_reset(struct adpd410x_dev *dev)
```
buha, line 132: When you do a reset, members in your `dev->` will probably not correspond anymore to the value on the chip and your driver representation of the chip is not accurate. I think that after successful reset you should call a function (something like reset_defaults()) that sets all the parameters of dev-> to default state. There is a catch, some may be non-zero.

drivers/photo-electronic/adpd410x/adpd410x.c
```diff
@@ -0,0 +1,660 @@
...
+ * @param mode - New operation mode.
+ * @return SUCCESS in case of success, FAILURE otherwise.
+ */
+int32_t adpd410x_set_opmode(struct adpd410x_dev *dev,
+			    enum adpd410x_opmode mode)
+{
+	int32_t ret;
+	uint16_t data;
+
+	ret = adpd410x_reg_read(dev, ADPD410X_REG_OPMODE, &data);
```
buha, line 157: Several functions in your driver perform a read, apply a mask and then perform a write back to a register, which is significant code duplication. You could use a general function spi_write_mask() that does this and all these functions would be simpler.

drivers/photo-electronic/adpd410x/adpd410x.c
```diff
@@ -0,0 +1,660 @@
...
+			return FAILURE;
+		if(reg_temp & BITP_SYS_CTL_LFOSC_SEL)
+			reg_load = ADPD410X_LOW_FREQ_OSCILLATOR_FREQ1;
+		else
+			reg_load = ADPD410X_LOW_FREQ_OSCILLATOR_FREQ2;
+	} else {
+		reg_load = dev->ext_lfo_freq;
+	}
+
+	reg_load /= sampling_freq;
```
buha, line 213: I'm not sure about this division here, remainder is not checked. Is this ok ?

drivers/photo-electronic/adpd410x/adpd410x.c
```diff
@@ -0,0 +1,660 @@
...
+
+	/* Setup inputs */
+	data = init->ts_inputs.option << (init->ts_inputs.pair * 4);
+	ret = adpd410x_reg_write(dev, (ADPD410X_REG_INPUTS_A + timeslot_no * 0x20),
+				 data);
+	if(ret != SUCCESS)
+		return FAILURE;
+
+	/* Set precondition PD */
+	ret = adpd410x_reg_read(dev, (ADPD410X_REG_CATHODE_A + timeslot_no * 0x20),
```
buha, line 265: There are many occurences of this, it looks like timeslot_no * 0x20 could also be a register macro that takes the timeslot_no as argument.

drivers/photo-electronic/adpd410x/adpd410x.c
```diff
@@ -0,0 +1,660 @@
...
+ */
+int32_t adpd410x_get_fifo_bytecount(struct adpd410x_dev *dev, uint16_t *bytes)
+{
+	int32_t ret;
+
+	ret = adpd410x_reg_read(dev, ADPD410X_REG_FIFO_STATUS, bytes);
+	if(ret != SUCCESS)
+		return FAILURE;
+
+	*bytes &= BITM_INT_STATUS_FIFO_FIFO_BYTE_COUNT;
```
buha, line 352: I think bytes parameter could be better described. I don't get if it's a count (number) or a special bit field (because you do masking here).

drivers/photo-electronic/adpd410x/adpd410x.c
```diff
@@ -0,0 +1,660 @@
...
+	if(dev->dev_type == ADPD4100)
+		ret = spi_init(&dev->dev_ops.spi_phy_dev,
+			       &init_param->dev_ops_init.spi_phy_init);
+	else
+		ret = i2c_init(&dev->dev_ops.i2c_phy_dev,
+			       &init_param->dev_ops_init.i2c_phy_init);
+	if(ret != SUCCESS)
+		goto error_dev;
+
+	ret = gpio_get(&dev->gpio0, &init_param->gpio0);
```
buha, line 541: Do these GPIO's have a default state ? If so, you should set it here, before reset.

drivers/photo-electronic/adpd410x/adpd410x.h
```diff
@@ -0,0 +1,4116 @@
...
+int32_t adpd410x_get_data(struct adpd410x_dev *dev, uint32_t *data);
+
+/** Setup the device and the driver. */
+int32_t adpd410x_setup(struct adpd410x_dev **device,
+		       struct adpd410x_init_param *init_param);
+
+/** Free memory allocated by adpd400x_setup(). */
+int32_t adpd410x_remove(struct adpd410x_dev *dev);
+
+#endif /* ADPD410X_H_ */
```
buha, line 4116: wow, that's a long header :)

drivers/photo-electronic/adpd410x/adpd410x.c
mchindri: You started using it in data[sample_index] and I have to scroll up to find it's first value.
Also because of this, I think is a good practice to initialize the variables where used.

drivers/photo-electronic/adpd410x/adpd410x.c
```diff
@@ -0,0 +1,660 @@
...
+
+	i = 0;
+	data_byte_count = 0;
+	do {
+		switch(slot_bytecount_tab[i]) {
+		case 0:
+			continue;
+		case 1:
+			data[sample_index] = data_byte_buff[data_byte_count];
+			data_byte_count++;
```
mchindri, line 421:
```suggestion
			data[sample_index] = data_byte_buff[j++];
```
I think this is better than my other suggestion.
I think you can rename data_byte_count to j because it is just an index and increment it inside []

drivers/photo-electronic/adpd410x/adpd410x.c
mchindri: I suggested an other approach then

drivers/photo-electronic/adpd410x/adpd410x.c
```diff
@@ -0,0 +1,660 @@
...
+{
+	int32_t ret;
+	uint16_t data;
+
+	ret = adpd410x_reg_read(dev, ADPD410X_REG_SYS_CTL, &data);
+	if(ret != SUCCESS)
+		return ret;
+	data |= BITM_SYS_CTL_SW_RESET;
+
+	return adpd410x_reg_write(dev, ADPD410X_REG_SYS_CTL, data);
```
cristiantpop, line 142: Maybe a delay is necessary after a reset.

drivers/photo-electronic/adpd410x/adpd410x.c
adrimbarean: Yes, we do not use the remainder.

drivers/photo-electronic/adpd410x/adpd410x.c
adrimbarean: The GPIOs are tristated by default. I included them in case the user would need them.

drivers/photo-electronic/adpd410x/adpd410x.c
adrimbarean: The function is called in the init function. I did not find reference to a delay being needed.

drivers/photo-electronic/adpd410x/adpd410x.c
adrimbarean: I would keep this if it's not too bad. I learned at an university course about architectures that actually duplicated the last bit instead of filling with zero when shifting.

drivers/photo-electronic/adpd410x/adpd410x.c
adrimbarean: Should I name it differently? The comments should explain what it is.

adrimbarean:
V2 change log:
 - Update copyright notice year;
 - Use switch case for device type;
 - Initialize `sample_index` just before using it;
 - Update the read and write functions using `field_prep` and defines;
 - Add `write_mask` function for the read-modify-write routines in the driver;
 - Updated header defines to avoid code duplication;
 - Concatenated code when appropriate.

adrimbarean:
V3 changes:
 - fix write register bug.

drivers/photo-electronic/adpd410x/adpd410x.c
```diff
@@ -0,0 +1,659 @@
...
+	if(ret != SUCCESS)
+		goto error_gpio3;
+	ret = gpio_tristate(dev->gpio0);
+	if(ret != SUCCESS)
+		goto error_gpio3;
+
+	ret = adpd410x_reg_read(dev, ADPD410X_REG_CHIP_ID, &reg_temp);
+	if(ret != SUCCESS)
+		goto error_gpio3;
+	if(reg_temp != 0x00C2)
```
mchindri, line 575:
```suggestion
        #define ADPD410X_CHIP_ID 0xC2
	if((reg_temp & BITM_CHIP_ID_CHIP_ID) != ADPD410X_CHIP_ID )
```

mchindri: Also, can BITM_CHIP_ID_CHIP_ID have an other name instead of repeating CHIP_ID?

adrimbarean:
V4 changes:
 - Add CHIP_ID define and use it in the initialization routine along with the chip ID mask.

drivers/photo-electronic/adpd410x/adpd410x.c
mchindri: It looks better now :)

drivers/photo-electronic/adpd410x/adpd410x.h
```diff
@@ -0,0 +1,1244 @@
...
+#define BITM_INT_ACLEAR_INT_ACLEAR_DATA_J       	0x0200
+#define BITM_INT_ACLEAR_INT_ACLEAR_DATA_K       	0x0400
+#define BITM_INT_ACLEAR_INT_ACLEAR_DATA_L       	0x0800
+#define BITM_INT_ACLEAR_INT_ACLEAR_FIFO         	0x8000
+
+/* ADPD410X_REG_CHIP_ID */
+#define BITP_CHIP_ID_CHIP_ID                    	0
+#define BITP_CHIP_ID_VERSION                    	8
+#define BITM_CHIP_ID_CHIP_ID                    	0x00ff
+#define BITM_CHIP_ID_VERSION                    	0xff00
```
mchindri, line 254: Can BITM_CHIP_ID  and BITM_CHIP_VERSION be used?

adrimbarean: I was going for the naming scheme: REGISTER_NAME_FIELD_NAME. But in this case they coincide. I will change to avoid confusion.

adrimbarean:
V5 changes:
 - Update CHIP_ID register defines.

drivers/photo-electronic/adpd410x/adpd410x.c
buha: Ok, if they don't need a default state, it's fine.

drivers/photo-electronic/adpd410x/adpd410x.h
```diff
@@ -0,0 +1,1244 @@
...
+#define ADPD410X_REG_SIGNAL2_L(ts)     	(0x0032 + ts * 0x08)
+#define ADPD410X_REG_SIGNAL2_H(ts)     	(0x0033 + ts * 0x08)
+#define ADPD410X_REG_DARK1_L(ts)       	(0x0034 + ts * 0x08)
+#define ADPD410X_REG_DARK1_H(ts)       	(0x0035 + ts * 0x08)
+#define ADPD410X_REG_DARK2_L(ts)       	(0x0036 + ts * 0x08)
+#define ADPD410X_REG_DARK2_H(ts)       	(0x0037 + ts * 0x08)
+#define ADPD410X_REG_IO_ADJUST        	0x00B4
+#define ADPD410X_REG_I2C_KEY         	0x00B6
+#define ADPD410X_REG_I2C_ADDR        	0x00B7
+#define ADPD410X_REG_TS_CTRL(ts)       	(0x0100 + ts * 0x20)
```
buha, line 100: Always paranthesize macro parameters in C, this is a potential bug.
What if uses passes `ts` as arithmetic ? This gets expanded to `0x0100 + a + b * 0x20` which is not what you want. If you use parantheses `0x0100 + (a + b) * 0x20` would give correct result.

drivers/photo-electronic/adpd410x/adpd410x.c
buha: I had a look at your dev, if seems it doesn't contain any setting that resets to non-zero default value in the chip so this should be ok for now, until such setting is added to the driver.

drivers/photo-electronic/adpd410x/adpd410x.c
```diff
@@ -0,0 +1,659 @@
...
+	ret = gpio_direction_output(dev->gpio0, GPIO_HIGH);
+	if(ret != SUCCESS)
+		goto error_gpio3;
+	ret = adpd410x_reg_write(dev, 0xB5, 0x04);
+	if(ret != SUCCESS)
+		goto error_gpio3;
+	ret = adpd410x_reg_write(dev, 0xB5, 0x00);
+	if(ret != SUCCESS)
+		goto error_gpio3;
+	ret = gpio_tristate(dev->gpio0);
```
adrimbarean, line 568: API does not exist yet and this form is not agreed upon yet.

PR #711, opened 2020-07-27
9-Remaing iio_changes

mchindri:


mchindri:
v2: 
 - rebase onto master

libraries/iio/iio.c
```diff
@@ -102,6 +102,7 @@ struct element_info {
  * with a "iio_device *iio" that describes capabilities of the device.
  */
 struct iio_interface {
+	const char		dev_id[10];
```
cristiantpop, line 98: Is this long enough ?

libraries/iio/iio.c
```diff
@@ -866,11 +867,12 @@ ssize_t iio_register(struct iio_desc *desc, struct iio_device *dev_descriptor,
...
 	desc->xml_desc = aux;
 	/* Print the new device xml at the end of the xml */
-	iio_generate_device_xml(iio_interface->dev_descriptor, iio_interface->name,
+	iio_generate_device_xml(iio_interface->dev_descriptor,
+				iio_interface->name,
 				desc->dev_count,
 				desc->xml_desc + desc->xml_size_to_last_dev,
 				new_size - desc->xml_size_to_last_dev);
-
+	sprintf((char *)iio_interface->dev_id, "device%d", (int)desc->dev_count);
```
adrimbarean, line 797: If the device IDs are zero based dev_count here could end up being one more than we want it to. Is that the case?

libraries/iio/iio.c
```diff
@@ -311,132 +273,55 @@ static ssize_t iio_read_all_attr(void *device, char *buf, size_t len,
...
-	if (!iio_device)
+static ssize_t iio_rd_wr_attribute(struct attr_fun_params *params,
+				   struct iio_attribute **attributes,
+				   char *attr_name,
+				   bool is_write)
+{
+	struct iio_attribute *attr;
+
+	/* Search attribute */
+	for (attr = attributes[0]; attr; attr++)
```
adrimbarean, line 313: Are we sure that the `attributes` table will always end in NULL pointer?

adrimbarean:
Also see Codancy review. Seems to have some concerns.

libraries/iio/iio.c
mchindri: Will not be one more. If you register one device will have the id: "device0" and after desc->dev_count will increment to 1 if this is what you mean.

libraries/iio/iio.c
mchindri: Yes, in linux is this way so we will use the same pattern.

mchindri:
v3:
 - Add comment to device id to explain size of buffer
 - rebase onto master (this should fix codacy problems)
 - Last commit was actually a fix for the first commit, so I squashed it to the first commit

libraries/iio/iio.c
mchindri: Added comment to explain this size. Should be enough for 999 devices :)

PR #713, opened 2020-07-28
Ad400x fmcz

cristiantpop:
Update with new SPI engine


cristiantpop:
Changelog V2:

- Set value for `ref_clk_hz`, in SPI engine projects

drivers/adc/ad400x/ad400x.c
```diff
@@ -136,33 +128,32 @@ int32_t ad400x_spi_single_conversion(struct ad400x_dev *dev,
...
 
-	dev->dev_id = init_param.dev_id;
-	dev->spi_desc->cs_delay = init_param.spi_init.cs_delay;
+	dev->reg_access_speed = init_param->reg_access_speed;
+	dev->dev_id = init_param->dev_id;
 
-	dev->spi_desc->max_data_width = init_param.num_bits;
-
-	spi_set_transfer_length(dev->spi_desc, 16);
+	spi_engine_set_transfer_width(dev->spi_desc, 16);
```
mirceacaprioru, line 149: this is not 16, it depends on the number of bits of the ADC, with AD4020 it is 20.

drivers/axi_core/spi_engine/spi_engine.c
```diff
@@ -140,6 +140,26 @@ static int32_t spi_engine_set_transfer_width(struct spi_desc *desc,
 	return SUCCESS;
 }
 
+/**
+ * @brief Set SPI engine clock frequency
+ *
+ * @param desc Decriptor containing SPI Engine's parameters
+ * @param speed_hz SPI engine transfer speed
+ * @return uint32_t This function always returns SUCCESS
```
mchindri, line 148: General question: Shouldn't we set function that always return true to return void?

mchindri, line 148: General question: Shouldn't we set function that always return true to return void?

drivers/axi_core/spi_engine/spi_engine_private.h
```diff
@@ -50,8 +50,6 @@
 /************************** Spi Engine parameters *****************************/
 /******************************************************************************/
 
-#define SPI_ENGINE_AXI_SPEED_HZ			100000000UL
```
mchindri, line 53: Remove also the Spi Engine parameters comments

drivers/adc/ad400x/ad400x.h
```diff
@@ -64,16 +64,19 @@ enum ad400x_supported_dev_ids {
...
+	uint32_t reg_access_speed;
 	/* Device Settings */
 	enum ad400x_supported_dev_ids dev_id;
 };
 
 struct ad400x_init_param {
 	/* SPI */
 	spi_init_param spi_init;
+	/* Register access speed */
+	uint32_t reg_access_speed;
```
mchindri, line 77: Isn't the access speed specific to the chip? Can it be a define?
Also you didn't set it from the project main (at least in this commit)

drivers/adc/ad400x/ad400x.c
```diff
@@ -61,22 +61,17 @@ int32_t ad400x_spi_reg_read(struct ad400x_dev *dev,
...
 	// register access runs at a lower clock rate (~2MHz)
-	if (dev->spi_desc->spi_clk_hz > dev->spi_desc->spi_clk_hz_reg_access) {
-		spi_clk_hz_tmp = dev->spi_desc->spi_clk_hz;
-		dev->spi_desc->spi_clk_hz = dev->spi_desc->spi_clk_hz_reg_access;
-	}
+	spi_engine_set_speed(dev->spi_desc, dev->reg_access_speed);
 
 	ret = spi_write_and_read(dev->spi_desc, buf, 2);
-	*reg_data = buf[1];
+	*reg_data = buf[0];
```
mchindri, line 72: Why did you change this to 0? 
Can't you change then 
```
spi_write_and_read(dev->spi_desc, buf, 2);
``` 
to
```
uint8_t buff = AD400X_READ_COMMAND;
spi_write_and_read(dev->spi_desc, &buff , 1); 
*reg_data = buff;
```
?

drivers/adc/ad400x/ad400x.c
```diff
@@ -136,33 +128,32 @@ int32_t ad400x_spi_single_conversion(struct ad400x_dev *dev,
  * @return 0 in case of success, negative error code otherwise.
  */
 int32_t ad400x_init(struct ad400x_dev **device,
-		    struct ad400x_init_param init_param)
+		    struct ad400x_init_param *init_param)
```
mchindri, line 131: I think you should split this commit in 2:
 - ad400x Driver update
 - Update project ad400x

drivers/adc/ad400x/ad400x.c
```diff
@@ -136,33 +128,32 @@ int32_t ad400x_spi_single_conversion(struct ad400x_dev *dev,
...
 {
 	struct ad400x_dev *dev;
 	int32_t ret;
 	uint8_t data = 0;
+	struct spi_engine_init_param *spi_eng_init_param = init_param->spi_init.extra;
```
mchindri, line 136: Init_param should be checked before dereferencing

projects/ad400x-fmcz/src/ad400x_fmcz.c
```diff
@@ -0,0 +1,115 @@
...
+
+		mdelay(2000);
+		Xil_DCacheInvalidateRange(0x800000, AD400x_EVB_SAMPLE_NO * 4);
+		offload_data = (uint32_t *)msg.rx_addr;
+
+		for(i = 0; i < AD400x_EVB_SAMPLE_NO; i++) {
+			data = *offload_data & 0xFFFFF;
+			if (data > 524287)
+				data = data - 1048576;
+			printf("ADC%d: %d\n", i, data);
```
mchindri, line 106: change i to int32_t or %d->%u to fix codacy

drivers/axi_core/spi_engine/spi_engine.h
```diff
@@ -86,6 +86,8 @@ and will be used inside the function call */
  * @brief  Structure containing the init parameters needed by the SPI engine
  */
 struct spi_engine_init_param {
+	/** SPI engine reference clock */
+	uint32_t	ref_clk_hz;
```
adrimbarean, line 90: Is there a way to detect this from the platform instead/in addition to the user providing it?

adrimbarean:
Codacy has a small issue.

drivers/axi_core/spi_engine/spi_engine.h
cristiantpop: If we have the guarantee that this parameter is named "XPAR_PS7_SPI_0_SPI_CLK_FREQ_HZ" always, yes we can use this. Should we use it?

drivers/adc/ad400x/ad400x.c
cristiantpop: 2 represents the number of read/writes on SPI engine. Bytes are received starting with position 0 in new SPI engine, so this is correct.

cristiantpop:
Changelog V3:

- Apply mask on captured data, corresponding to convertor resolution.

- Remove also the Spi Engine parameters comments

- Explicitly initialize "reg_access_speed" attribute

- Check Init_param before dereferencing

- Change i to int32_t or %d->%"PRIi32" to fix codacy.

cristiantpop:
Changelog V4:

- Fix astyle

cristiantpop:
Changelog V5:

- Fix astyle

cristiantpop:
The driver works now (a clean sine wave can be observed after a capture).

drivers/axi_core/spi_engine/spi_engine.h
adrimbarean: We can't guarantee this because it would depend on HDL notation. If there is no other way we should leave it like this, I think.

drivers/axi_core/spi_engine/spi_engine.c
```diff
@@ -140,6 +140,23 @@ static int32_t spi_engine_set_transfer_width(struct spi_desc *desc,
...
+ */
+void spi_engine_set_speed(struct spi_desc *desc,
+			  uint32_t speed_hz)
+{
+	struct spi_engine_desc	*desc_extra;
+
+	desc_extra = desc->extra;
+
+	desc_extra->clk_div =  desc_extra->ref_clk_hz /
+			       (2 * speed_hz) - 1;
```
adrimbarean, line 157: I see this method of calculating the divide is used in at least two places. Maybe it could be a define or inline function to avoid changing one and not the other in the future.

mirceacaprioru: Was this change ignored or missed ? The argument was valid in my opinion

PR #715, opened 2020-07-30
drivers:platform:aducm3029: add power driver

adrimbarean:
This module contains a piece of code used by all ADICUP3029 projects to
initialize the main and peripheral clock of the chip.

### DIFF
--- a/drivers/platform/aducm3029/platform_init.c
+++ b/drivers/platform/aducm3029/platform_init.c
@@ -1,0 +1,59 @@
+/***************************************************************************//**
+ *   @file   platform_init.c
+ *   @brief  ADuCM3029 platform initialization source.
+ *   @author Andrei Drimbarean (Andrei.Drimbarean@analog.com)
+********************************************************************************
+ * Copyright 2020(c) Analog Devices, Inc.
+ *
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *  - Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ *  - Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in
+ *    the documentation and/or other materials provided with the
+ *    distribution.
+ *  - Neither the name of Analog Devices, Inc. nor the names of its
+ *    contributors may be used to endorse or promote products derived
+ *    from this software without specific prior written permission.
+ *  - The use of this software may or may not infringe the patent rights
+ *    of one or more patent holders.  This license does not release you
+ *    from the requirement that you obtain separate licenses from these
+ *    patent holders to use this software.
+ *  - Use of the software either in source or binary form, must be run
+ *    on or directly connected to an Analog Devices Inc. component.
+ *
+ * THIS SOFTWARE IS PROVIDED BY ANALOG DEVICES "AS IS" AND ANY EXPRESS OR
+ * IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, NON-INFRINGEMENT,
+ * MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED.
+ * IN NO EVENT SHALL ANALOG DEVICES BE LIABLE FOR ANY DIRECT, INDIRECT,
+ * INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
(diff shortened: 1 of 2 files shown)


drivers/platform/aducm3029/power.c
```diff
@@ -0,0 +1,62 @@
+/***************************************************************************//**
+ *   @file   power.c
+ *   @brief  ADuCM3029 application power controller.
+ *   @author Andrei Drimbarean (Andrei.Drimbarean@analog.com)
+********************************************************************************
+ * Copyright 2019(c) Analog Devices, Inc.
```
amiclaus, line 6: 2020

drivers/platform/aducm3029/power.c
```diff
@@ -0,0 +1,62 @@
...
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+*******************************************************************************/
+
+#include <drivers/pwr/adi_pwr.h>
+#include "error.h"
+
+/**
+ * Initialize the power controller and set the core and peripherals clock
```
amiclaus, line 44: @brief

drivers/platform/aducm3029/power.c
```diff
@@ -0,0 +1,62 @@
...
+ * @params void
+ *
+ * @return 0 in case of success or negative error code otherwise.
+ */
+int32_t pwr_setup()
+{
+	int32_t ret;
+
+	ret = adi_pwr_Init();
+	if(ret != SUCCESS)
```
amiclaus, line 56: Can the error code be different from `FAILURE`? Otherwise return `FAILURE` directly.

drivers/platform/aducm3029/power.h
```diff
@@ -0,0 +1,47 @@
+/***************************************************************************//**
+ *   @file   power.h
+ *   @brief  ADuCM3029 application power controller header.
+ *   @author Andrei Drimbarean (Andrei.Drimbarean@analog.com)
+********************************************************************************
+ * Copyright 2019(c) Analog Devices, Inc.
```
amiclaus, line 6: 2020

drivers/platform/aducm3029/power.c
```diff
@@ -0,0 +1,62 @@
...
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+*******************************************************************************/
+
+#include <drivers/pwr/adi_pwr.h>
+#include "error.h"
+
+/**
+ * Initialize the power controller and set the core and peripherals clock
+ * divider.
+ *
```
amiclaus, line 46: no spaces

drivers/platform/aducm3029/power.c
```diff
@@ -0,0 +1,62 @@
...
+ * @params void
+ *
+ * @return 0 in case of success or negative error code otherwise.
```
amiclaus, line 49: return 0 -> return SUCCESS

amiclaus:
A few minor comments. Rest looks good. Also tested the code and it works fine.

mchindri:
- My opinion is that we should call the funciton/files somthing like platform_initialize
- Also I think we can also call the adi_initComponents(); function

drivers/platform/aducm3029/power.c
adrimbarean: The there are several error codes that the functions in this method return, but I think a return FAILURE will suffice for now, until we implement multiple error codes ourselves.

adrimbarean:
> 
> 
>     * My opinion is that we should call the funciton/files somthing like platform_initialize
> 
>     * Also I think we can also call the adi_initComponents(); function

adrimbarean:
@mchindri on your second point, the main function is usually generated calling the `adi_initComponents();` so I feel like is not necessary to add it in the platform_init module. I will rename it, because it might end up doing more in the future.

adrimbarean:
V1 changes:
 - Update license notice to 2020 from 2019;
 - Rename files `platform_init`;
 - Return `FAILURE`;
 - Update DOxygen documentation.

dbogdan:
I agree with @mchindri, I would use a different name too.

adrimbarean:
V2 changes:
 - update header description and header guard to include the name change;
 - use specific void parameters for `pwr_setup()` function.

adrimbarean:
V3 changes:
 - remove `@param` from `pwr_setup()` function since parameters are `void`.

dbogdan (approved):

PR #718, opened 2020-08-04
ad9081: add zcu102 support

buha:
HDL design for zcu102 exists, this commit performs necessary
modifications to make no-OS/projects/ad9081 work on zcu102.

Tested, works well on vcu118 + ad9081-fmca-ebz and zcu102 + ad9081-fmca-ebz.

### DIFF
--- a/drivers/adc/ad9081/ad9081.c
+++ b/drivers/adc/ad9081/ad9081.c
@@ -483,6 +483,7 @@ static int32_t ad9081_setup(struct ad9081_phy *phy)
 	uint64_t status64;
 	uint64_t sample_rate;
 	uint64_t rx_lane_rate_kbps;
+	uint32_t timeout;
 
 	clk_recalc_rate(phy->dev_clk, &dev_frequency_hz);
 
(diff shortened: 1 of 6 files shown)


drivers/adc/ad9081/ad9081.c
```diff
@@ -707,18 +707,37 @@ static int32_t ad9081_setup(struct ad9081_phy *phy)
 	mdelay(10);
 
 	if (phy->jesd_rx_clk) {
-		mdelay(10);
-		ret = clk_enable(phy->jesd_rx_clk);
-		if (ret < 0) {
+		int timeout = 2000;
```
mchindri, line 710: you can declare timeout at the beginning of the function and only initialize it here.
And also for consistency you should use int32_t instead of int.

drivers/adc/ad9081/ad9081.c
```diff
@@ -707,18 +707,37 @@ static int32_t ad9081_setup(struct ad9081_phy *phy)
...
+		while(timeout) {
+			ret = clk_enable(phy->jesd_rx_clk);
+			if (ret) {
```
mchindri, line 713: Can the changes in this file be in a separate commit?
"Add timeout for clk_enable " or something like that

projects/ad9081/src/app_parameters.h
```diff
@@ -44,28 +44,48 @@
...
+#define PHY_CS			0
 
 #ifdef QUAD_MXFE
-#define ADF4371_CS	0
-#define HMC7043_CS	4
+#define ADF4371_CS		0
+#define HMC7043_CS		4
 
-#define PHY_RESET		41
+#define PHY_RESET		GPIO_OFFSET + 41
```
mchindri, line 59:
```suggestion
#define PHY_RESET		(GPIO_OFFSET + 41)
```

mchindri: same below

drivers/adc/ad9081/ad9081.c
buha: I prefer variables in the context where they are used but I think we have a rule about this so I'll move it at the top

drivers/adc/ad9081/ad9081.c
buha: Yeah good point

projects/ad9081/src/app_parameters.h
buha: yeah, not really needed here but I'll do it anyway, it looks better

buha:
V2:
- split the ad9081.c driver change into different commit, also moved timeout at the top of the function
- #define parantheses

buha:
V3: 
- rebase on master (might get rid of ad77681 related astyle error)

PR #723, opened 2020-08-13
drivers: adc: add AD7798/AD7799 driver

amiclaus:
Add driver implementation for the AD7798/AD7799 ADC.

More information:
https://www.analog.com/media/en/technical-documentation/data-sheets/AD7798_7799.pdf

### DIFF
--- a/drivers/adc/ad7799/ad7799.c
+++ b/drivers/adc/ad7799/ad7799.c
@@ -1,0 +1,444 @@
+/***************************************************************************//**
+ *   @file   ad7799.c
+ *   @brief  Implementation of AD7798/AD7799 Driver.
+ *   @author Antoniu Miclaus (antoniu.miclaus@analog.com)
+********************************************************************************
+ * Copyright 2020(c) Analog Devices, Inc.
+ *
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *  - Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ *  - Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in
+ *    the documentation and/or other materials provided with the
+ *    distribution.
+ *  - Neither the name of Analog Devices, Inc. nor the names of its
+ *    contributors may be used to endorse or promote products derived
+ *    from this software without specific prior written permission.
+ *  - The use of this software may or may not infringe the patent rights
+ *    of one or more patent holders.  This license does not release you
+ *    from the requirement that you obtain separate licenses from these
+ *    patent holders to use this software.
+ *  - Use of the software either in source or binary form, must be run
+ *    on or directly connected to an Analog Devices Inc. component.
+ *
+ * THIS SOFTWARE IS PROVIDED BY ANALOG DEVICES "AS IS" AND ANY EXPRESS OR
+ * IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, NON-INFRINGEMENT,
+ * MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED.
+ * IN NO EVENT SHALL ANALOG DEVICES BE LIABLE FOR ANY DIRECT, INDIRECT,
+ * INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, INTELLECTUAL PROPERTY RIGHTS, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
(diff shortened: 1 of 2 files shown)


amiclaus:
V2:
- remove extra line

drivers/adc/ad7799/ad7799.h
```diff
@@ -0,0 +1,231 @@
...
+/* AD7799 Reset Sequence */
+#define AD7799_RESET_DATA	 0xFF
+
+/**
+ * @enum ad7799_type
+ * @brief Device type (AD7798/AD7799)
+ */
+enum ad7799_type {
+	/** AD7798 device*/
+	ID_AD7798,
```
buha, line 155: You may use this enum to drop the ID definitions above:
```
enum ad7799_type {
	/** AD7798 device*/
	ID_AD7798 = 0x8,
	/** AD7799 device */
	ID_AD7799 = 0x9
};
```

drivers/adc/ad7799/ad7799.c
```diff
@@ -0,0 +1,444 @@
...
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+*******************************************************************************/
+
+/******************************************************************************/
+/***************************** Include Files **********************************/
+/******************************************************************************/
+#include "ad7799.h"
```
buha, line 43: I think consensus is <> first and then "" includes.

drivers/adc/ad7799/ad7799.c
```diff
@@ -0,0 +1,444 @@
...
+	uint8_t buff_size;
+
+	*reg_data = 0;
+
+	buff_size = device->reg_size[reg_addr];
+
+	buff[0] = AD7799_COMM_READ | AD7799_COMM_ADDR(reg_addr);
+
+	/* Fill buffer with zeros for the read operation */
+	for (i = 1; i < buff_size + 1 ; i++)
```
buha, line 99: memset zeros

drivers/adc/ad7799/ad7799.c
```diff
@@ -0,0 +1,444 @@
...
+	return ad7799_write(device, AD7799_REG_CONF, reg_data);
+}
+
+/**
+ * @brief Get the ADC gain
+ * @param device - The device structure.
+ * @param  gain - the gain value from the register.
+ * @return SUCCESS in case of success, negative error code otherwise.
+ */
+int32_t ad7799_get_gain(struct ad7799_dev *device, uint8_t *gain)
```
buha, line 265: You could drop this function in favour of a `device->gain` member that works as a cached setting.

drivers/adc/ad7799/ad7799.h
```diff
@@ -0,0 +1,231 @@
...
+#define AD7799_REG_IO	    	5 /* IO Register (RO, 8-bit) */
+#define AD7799_REG_OFFSET   	6 /* Offset Register (RW, 24-bit) */
+#define AD7799_REG_FULLSCALE	7 /* Full-Scale Register (RW, 24-bit) */
+
+/* AD7799 Polarity */
+#define AD7799_BIPOLAR 		0 /* Bipolar bit */
+#define AD7799_UNIPOLAR		1 /* Unipolar bit */
+
+/* Communications Register Bit Designations (AD7799_REG_COMM) */
+#define AD7799_COMM_WEN		(1 << 7) /* Write Enable */
```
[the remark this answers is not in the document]
mirceacaprioru: well these are just commands so you always just append them depending on what you are doing. I usually use MSK only for register read writes. In my opinion it is ok as is.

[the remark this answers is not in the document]
mirceacaprioru: I am not a big fan of decimal values with register. Hex is the norm.

drivers/adc/ad7799/ad7799.h
```diff
@@ -0,0 +1,231 @@
...
+ * @struct ad7799_dev
+ * @brief AD7798/AD7799 Device description
+ */
+struct ad7799_dev {
+	/** SPI Descriptor */
+	struct spi_desc *spi_desc;
+	/** Chip type (AD7798/AD7799) */
+	uint8_t chip_type;
+	/** Register size */
+	const uint32_t *reg_size;
```
mirceacaprioru, line 170: uint8_t here. Let's try to use as less memory as we can from the start.

drivers/adc/ad7799/ad7799.c
amiclaus: This function was created only for test purposes (when testing the driver in conjunction with IIO). 
I thought I can leave it since it was already implemented.

Otherwise, if required, I can remove it.

In my opinion, adding `gain` member to the structure implies adding also members for other driver specific parameters (polarity, channel selection, ref detect, etc.)

amiclaus:
V3:

- remove ID defines and update `enum ad7799_type`
- use `unint8_t` for `reg_size`
- use `memset` when performing read operation
- use hex values for defines
- rearrange includes

drivers/adc/ad7799/ad7799.c
buha: I agree, adding `gain` would imply adding other structure members too (that's what we do in other drivers).
If you don't wanna readback these values I guess deleting this function would be fine too.

amiclaus:
V4:
- resolve newly appeared Travis build issues

PR #725, opened 2020-08-17
tools:scripts: fix bug in the windows.mk makefile

adrimbarean:
The windows makefile has to call the "xsdk.bat" batch script instead of the
"xsdk" on which is a shell script.


adrimbarean:
V1:
 - removed new line at the end of the file.

PR #726, opened 2020-08-17
ad9361 sync

dbogdan:
Sync no-OS and Linux drivers.


adrimbarean (approved):
Looks good to me.

nitpick: looks like AStyle is still complaining.

dbogdan:
@amiclaus , you are right - my fault. @adrimbarean , I've fixed the real astyle complaints.

PR #729, opened 2020-08-20
GPIO tristate API

adrimbarean:
Suggestion on how to implement a method to disable output and input buffers for GPIOs after they have been enabled.

This approach makes a new API, which seems logical to me considering there are two different APIs to enable input and output direction.

Any thoughts or suggestions?

### DIFF
--- a/drivers/platform/aducm3029/gpio.c
+++ b/drivers/platform/aducm3029/gpio.c
@@ -208,7 +208,9 @@ int32_t gpio_get_direction(struct gpio_desc *desc, uint8_t *direction)
 /**
  * @brief Set the value of the specified GPIO.
  * @param desc - The GPIO descriptor.
- * @param value - The value: GPIO_HIGH or GPIO_LOW
+ * @param value - The value: GPIO_HIGH, GPIO_LOW or GPIO_HIGH_Z
+ *                Choosing GPIO_HIGH_Z will deactivate the input and output
+ *                buffers.
  * @return \ref SUCCESS in case of success, \ref FAILURE otherwise.
  */
 int32_t gpio_set_value(struct gpio_desc *desc, uint8_t value)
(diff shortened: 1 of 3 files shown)


adrimbarean:
V1 changes:
 - solve AStyle.

adrimbarean:
V2 changes:
 - fix function return check mistake.

drivers/platform/aducm3029/gpio.c
```diff
@@ -205,6 +205,28 @@ int32_t gpio_get_direction(struct gpio_desc *desc, uint8_t *direction)
...
+ * @param desc - The GPIO descriptor.
+ * @return \ref SUCCESS in case of success, \ref FAILURE otherwise.
+ */
+int32_t gpio_tristate(struct gpio_desc *desc)
+{
+	int32_t ret;
+
+	ret = adi_gpio_OutputEnable(PORT(desc->number), PIN(desc->number),
+				    false);
+	if ((ret != ADI_GPIO_SUCCESS) && (ret != ADI_GPIO_NOT_INITIALIZED))
```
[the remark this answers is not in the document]
adrimbarean: You are right, we should not continue in this case. At the moment I believed that the `ADI_GPIO_NOT_INITIALIZED` was referring to the hardware driver not being initialized, but it's about the driver.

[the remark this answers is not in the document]
adrimbarean: Fixed.

drivers/platform/aducm3029/gpio.c
```diff
@@ -205,6 +205,28 @@ int32_t gpio_get_direction(struct gpio_desc *desc, uint8_t *direction)
 	return SUCCESS;
 }
 
+/**
+ * @brief Place GPIO in tristate mode.
+ * @param desc - The GPIO descriptor.
+ * @return \ref SUCCESS in case of success, \ref FAILURE otherwise.
+ */
+int32_t gpio_tristate(struct gpio_desc *desc)
```
buha, line 213: I would add another output gpio state:
GPIO_HIGH_Z 
And use gpio_set_value(..., GPIO_HIGH_Z).

This approach would be backwards compatible and you could get rid of this function.

adrimbarean: The thing to watch out here is to not change functionality for any of the current projects. As it currently stands if the second parameter of `gpio_set_value` is anything but zero, the GPIO will be set to high state. I will be introduceing a new case, where if the second parameter is 2 the GPIO will be high impedance. I will leave the rest of the cases to bring it to high state.

adrimbarean:
V3 change log:
 - change implementation to avoid adding a new API and update the existing one to get desired functionality.

include/gpio.h
```diff
@@ -110,7 +120,7 @@ int32_t gpio_get_direction(struct gpio_desc *desc,
 
 /* Set the value of the specified GPIO. */
 int32_t gpio_set_value(struct gpio_desc *desc,
-		       uint8_t value);
+		       enum gpio_values value);
```
mchindri, line 123: You need to change this header in all platform implementations

adrimbarean:
V4 change log:
 - make `value` parameter `uint8_t` to maintain backwards compatibility.

mchindri (approved):
LGTM

amiclaus (approved):
LGTM.

mirceacaprioru (approved):

PR #733, opened 2020-08-24
axi_adc_core: add delay_set and delay_calibration

amiclaus:
These functions are required for porting the ad9467_fmc project.

Reference: https://github.com/analogdevicesinc/no-OS/commit/570893fb1b333d5abee930430d429a8087460141

### DIFF
--- a/drivers/axi_core/axi_adc_core/axi_adc_core.c
+++ b/drivers/axi_core/axi_adc_core/axi_adc_core.c
@@ -151,6 +151,118 @@ void axi_adc_idelay_set(struct axi_adc *adc,
 }
 
 /***************************************************************************//**
+ * @brief axi_adc_delay_set
+*******************************************************************************/
+int32_t axi_adc_delay_set(struct axi_adc *adc,
+			  uint32_t no_of_lanes,
+			  uint32_t delay)
+{
+	uint32_t i;
+	uint32_t rdata;
+	uint32_t pcore_version;
+
+	axi_adc_read(adc, 0x0, &pcore_version);
+	pcore_version >>= 16;
+	if (pcore_version < 9) {
+		printf(" pcore_version is : %d\n\r", pcore_version);
+		printf(" DRIVER DOES NOT SUPPORT PCORE VERSIONS OLDER THAN 10 !");
+		return FAILURE;
+	} else {
+		for (i = 0; i < no_of_lanes; i++) {
+			axi_adc_idelay_set(adc, i, delay);
+			axi_adc_read(adc, AXI_ADC_REG_DELAY(i), &rdata);
+			if (rdata != delay) {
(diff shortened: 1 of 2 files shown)


drivers/axi_core/axi_adc_core/axi_adc_core.c
```diff
@@ -150,6 +150,118 @@ void axi_adc_idelay_set(struct axi_adc *adc,
 
 }
 
+/***************************************************************************//**
+ * @brief axi_adc_delay_set
```
adrimbarean, line 154: Since this is an API I think we will benefit from more descriptive comments.

drivers/axi_core/axi_adc_core/axi_adc_core.c
```diff
@@ -150,6 +150,118 @@ void axi_adc_idelay_set(struct axi_adc *adc,
...
+
+	axi_adc_read(adc, 0x0, &pcore_version);
+	pcore_version >>= 16;
+	if (pcore_version < 9) {
+		printf(" pcore_version is : %d\n\r", pcore_version);
+		printf(" DRIVER DOES NOT SUPPORT PCORE VERSIONS OLDER THAN 10 !");
+		return FAILURE;
+	} else {
+		for (i = 0; i < no_of_lanes; i++) {
+			axi_adc_write(adc,((0x200 + i)*4), delay);
```
adrimbarean, line 172: I think you can use `AXI_ADC_REG_DELAY` define here:
```suggestion
			axi_adc_write(adc,AXI_ADC_REG_DELAY(i), delay);
```
Same below.

drivers/axi_core/axi_adc_core/axi_adc_core.c
amiclaus: This approach was used here to keep the consistency with the rest of the file (all functions in `axi_adc_core` have short descriptions like this one). I believe adding proper header comments to the entire file is subject for another commit.

drivers/axi_core/axi_adc_core/axi_adc_core.c
amiclaus: The `AXI_ADC_REG_DELAY` has different definition, therefore I think it won't work properly if i use it.

The macro is used for `axi_adc_idelay_set` function which is outside of the commit's scope.

amiclaus: My bad, you are right here, `AXI_ADC_REG_DELAY` can be used. Will do the change.

amiclaus:
V2:
- use macro definition `AXI_ADC_REG_DELAY` and `axi_adc_idelay_set` function where possible

dbogdan (approved):

PR #741, opened 2020-08-31
pwm.h: Add header interface for PWM generator.

cristiantpop:
Interface for PWM generator.

### DIFF
--- a/include/pwm.h
+++ b/include/pwm.h
@@ -1,0 +1,134 @@
+/***************************************************************************//**
+ *   @file   pwm.h
+ *   @brief  Header file of PWM Interface
+ *   @author Cristian Pop (cristian.pop@analog.com)
+********************************************************************************
+ * Copyright 2020(c) Analog Devices, Inc.
+ *
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *  - Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ *  - Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in
+ *    the documentation and/or other materials provided with the
+ *    distribution.
+ *  - Neither the name of Analog Devices, Inc. nor the names of its
+ *    contributors may be used to endorse or promote products derived
+ *    from this software without specific prior written permission.
+ *  - The use of this software may or may not infringe the patent rights
+ *    of one or more patent holders.  This license does not release you
+ *    from the requirement that you obtain separate licenses from these
+ *    patent holders to use this software.
+ *  - Use of the software either in source or binary form, must be run
+ *    on or directly connected to an Analog Devices Inc. component.
+ *
+ * THIS SOFTWARE IS PROVIDED BY ANALOG DEVICES "AS IS" AND ANY EXPRESS OR
+ * IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, NON-INFRINGEMENT,
+ * MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED.
+ * IN NO EVENT SHALL ANALOG DEVICES BE LIABLE FOR ANY DIRECT, INDIRECT,
+ * INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, INTELLECTUAL PROPERTY RIGHTS, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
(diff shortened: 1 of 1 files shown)


drivers/axi_core/axi_pwmgen/axi_pwmgen.h
```diff
@@ -0,0 +1,134 @@
...
+/******************************************************************************/
+#include <stdint.h>
+#include <stdbool.h>
+
+/******************************************************************************/
+/*************************** Types Declarations *******************************/
+/******************************************************************************/
+/**
+ * @enum axi_pwmgen_polarity
+ * @brief Possible events for uart interrupt
```
scuciurean, line 53: UART interrupt?

drivers/axi_core/axi_pwmgen/axi_pwmgen.h
```diff
@@ -0,0 +1,134 @@
...
+ */
+enum axi_pwmgen_polarity {
```
scuciurean, line 55: Also, I don't think that the pwmgen supports inverted polarity

drivers/axi_core/axi_pwmgen/axi_pwmgen.h
```diff
@@ -0,0 +1,134 @@
...
+/* Set duty cycle of PWM generator device */
+int32_t axi_pwmgen_set_duty_cycle(struct axi_pwmgen_desc *desc,
+				  uint32_t duty_cycle_ns);
+
+/* Get period of PWM generator device */
+int32_t axi_pwmgen_get_duty_cycle(struct axi_pwmgen_desc *desc,
+				  uint32_t *duty_cycle_ns);
+
+/* Set polarity of PWM generator device */
+int32_t axi_pwmgen_set_polarity(struct axi_pwmgen_desc *desc,
```
dbogdan, line 131: If we have "get" for all the other functions, I would add it for "polarity" too.

dbogdan:
By the way, why are we adding only the header? I would do this only if it would be a generic include file.

cristiantpop:
V2:

- Add `pwm_set_polarity` function interface.

- Move file and rename it from `drivers/axi_core/axi_pwmgen/axi_pwmge.h` to `includes/pwm.h`

scuciurean (approved):

include/pwm.h
```diff
@@ -0,0 +1,138 @@
...
+int32_t pwm_remove(struct pwm_desc *desc);
+
+/* Enable PWM generator device */
+int32_t pwm_enable(struct pwm_desc *desc);
+
+/* Disable PWM generator device */
+int32_t pwm_disable(struct pwm_desc *desc);
+
+/* Set period of PWM generator device */
+int32_t pwm_set_period(struct pwm_desc *desc,
```
amiclaus, line 115: Just a curiosity on my side: why did you prefer `period` over `frequency`?

I believe the frequency is a more common term when using PWMs.

cristiantpop: The unit preferred in linux is period. Since duty_cycle is in ns, it makes sense to have PWM period, probably.

include/pwm.h
```diff
@@ -0,0 +1,138 @@
...
+	PWM_POLARITY_INVERSED,
+};
+
+/**
+ * @struct pwm_init_param
+ * @brief  Structure containing the init parameters needed by the PWM generator
+ */
+struct pwm_init_param {
+	/** Base address where the PWM generator core is situated */
+	uint32_t base_addr;
```
mchindri, line 66: This should be in extra

include/pwm.h
```diff
@@ -0,0 +1,138 @@
...
+	/** PWM generator reference clock rate , in Hz */
+	uint32_t clk_rate_hz;
+	/** PWM generator period */
+	uint32_t period_ns;
+	/** PWM generator duty cycle */
+	uint32_t duty_cycle_ns;
+	/** PWM generator polarity */
+	enum pwm_polarity polarity;
+	/** PWM generator enabled */
+	bool enabled;
```
mchindri, line 76: Why enabled is needed as init param?

include/pwm.h
```diff
@@ -0,0 +1,138 @@
...
+ * @struct pwm_init_param
+ * @brief  Structure containing the init parameters needed by the PWM generator
+ */
+struct pwm_init_param {
+	/** Base address where the PWM generator core is situated */
+	uint32_t base_addr;
+	/** PWM generator reference clock rate , in Hz */
+	uint32_t clk_rate_hz;
+	/** PWM generator period */
+	uint32_t period_ns;
```
mchindri, line 70:
```suggestion
	/** PWM frequency in Hz */
	uint32_t frequency;
```
What clk_rate_hz is used for? Can be calculated from the specified frequency or is a platform specific field?

include/pwm.h
mchindri: I'am more familiar with frequency too and with duty expressed in percents, But our direction is to follow how it is done in linux is ok to me

include/pwm.h
```diff
@@ -0,0 +1,138 @@
...
+
+/******************************************************************************/
+/*************************** Types Declarations *******************************/
+/******************************************************************************/
+/**
+ * @enum pwm_polarity
+ * @brief Possible polarities of the PWM signal
+ */
+enum pwm_polarity {
+	PWM_POLARITY_NORMAL,
```
adrimbarean, line 56: Can you specify with comments which one is idle low and which is idle high? It might make it easier to understand.

include/pwm.h
cristiantpop: True

include/pwm.h
cristiantpop: Not necessary, it can be removed and the default state should be disabled

include/pwm.h
cristiantpop: Is the reference clock to generate the pwm

cristiantpop:
V3:

- Add comments for `pwm_polarity`;

- Add `void *extra`, device specific parameters.

amiclaus (approved):

PR #745, opened 2020-09-03
axi_pwm: Add support for AXI PWM generator.

cristiantpop:
This patch adds support for the AXI PWM Generator. This device is an
FPGA-implemented peripheral used as PWM signal generator and can be
interfaced with AXI4. The register map of this peripheral makes possible
the configuration of the output signal, in terms of period and duty
cycle.

### DIFF
--- a/drivers/axi_core/axi_pwmgen/axi_pwm.c
+++ b/drivers/axi_core/axi_pwmgen/axi_pwm.c
@@ -1,0 +1,285 @@
+/*******************************************************************************
+ *   @file   axi_pwm.c
+ *   @brief  Driver for the Analog Devices AXI PWM generator.
+ *   @author Cristian Pop (cristian.pop@analog.com)
+********************************************************************************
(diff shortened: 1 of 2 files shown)


cristiantpop:
V2:

- Fix some white spaces

drivers/axi_core/axi_pwmgen/axi_pwm.c
```diff
@@ -0,0 +1,285 @@
...
+
+/******************************************************************************/
+/************************** Functions Implementation **************************/
+/******************************************************************************/
+
+/**
+ * @brief Enable PWM generator device.
+ *
+ * @param [in] desc - Decriptor containing PWM generator parameters.
+ * @return \ref SUCCESS in case of success, \ref FAILURE otherwise.
```
buha, line 75: (If I remember correctly) axi_io_write/read only returns SUCCESS

buha (approved):

PR #746, opened 2020-09-03
Ad469x

cristiantpop:
Add ad469x driver

### DIFF
--- a/drivers/adc/ad469x/ad469x.c
+++ b/drivers/adc/ad469x/ad469x.c
@@ -1,0 +1,565 @@
+/***************************************************************************//**
+ *   @file   ad469x.c
+ *   @brief  Implementation of ad69x Driver.
+ *   @author Cristian Pop (cristian.pop@analog.com)
+********************************************************************************
+ * Copyright 2020(c) Analog Devices, Inc.
+ *
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *  - Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ *  - Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in
+ *    the documentation and/or other materials provided with the
+ *    distribution.
+ *  - Neither the name of Analog Devices, Inc. nor the names of its
+ *    contributors may be used to endorse or promote products derived
+ *    from this software without specific prior written permission.
+ *  - The use of this software may or may not infringe the patent rights
+ *    of one or more patent holders.  This license does not release you
+ *    from the requirement that you obtain separate licenses from these
+ *    patent holders to use this software.
+ *  - Use of the software either in source or binary form, must be run
+ *    on or directly connected to an Analog Devices Inc. component.
+ *
+ * THIS SOFTWARE IS PROVIDED BY ANALOG DEVICES "AS IS" AND ANY EXPRESS OR
+ * IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, NON-INFRINGEMENT,
+ * MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED.
+ * IN NO EVENT SHALL ANALOG DEVICES BE LIABLE FOR ANY DIRECT, INDIRECT,
+ * INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, INTELLECTUAL PROPERTY RIGHTS, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+*******************************************************************************/
+
+/******************************************************************************/
+/***************************** Include Files **********************************/
+/******************************************************************************/
+#include "stdio.h"
+#include "stdlib.h"
+#include "ad469x.h"
+#include "spi_engine.h"
+#include "delay.h"
+#include "error.h"
+
+/******************************************************************************/
+/********************** Macros and Constants Definitions **********************/
+/******************************************************************************/
+#define AD469x_TEST_DATA 0xEA
+
+/******************************************************************************/
+/************************** Functions Implementation **************************/
+/******************************************************************************/
+
+/**
+ * Read from device.
+ * @param dev - The device structure.
+ * @param reg_addr - The register address.
+ * @param reg_data - The register data.
+ * @return SUCCESS in case of success, negative error code otherwise.
+ */
+int32_t ad469x_spi_reg_read(struct ad469x_dev *dev,
+			    uint16_t reg_addr,
+			    uint8_t *reg_data)
+{
+	int32_t ret;
+	uint8_t buf[3];
+
+	ret = spi_engine_set_transfer_width(dev->spi_desc, dev->reg_data_width);
+	if (ret != SUCCESS)
+		return ret;
+
+	spi_engine_set_speed(dev->spi_desc, dev->reg_access_speed);
+
+	buf[0] = (1 << 7) | ((reg_addr >> 8) & 0x7F);
+	buf[1] = 0xFF & reg_addr;
+	buf[2] = 0xFF;
+
+	ret = spi_write_and_read(dev->spi_desc, buf, 3);
+	*reg_data = buf[1];
+
+	ret = spi_engine_set_transfer_width(dev->spi_desc,
+					    dev->capture_data_width);
+	if (ret != SUCCESS)
+		return ret;
+
+	spi_engine_set_speed(dev->spi_desc, dev->spi_desc->max_speed_hz);
+
+	return ret;
+}
+
+/**
+ * Write to device.
+ * @param dev - The device structure.
+ * @param reg_addr - The register address.
+ * @param reg_data - The register data.
+ * @@eturn SUCCESS in case of success, negative error code otherwise.
+ */
+int32_t ad469x_spi_reg_write(struct ad469x_dev *dev,
+			     uint16_t reg_addr,
+			     uint8_t reg_data)
+{
+	int32_t ret;
+	uint8_t buf[3];
+
+	ret = spi_engine_set_transfer_width(dev->spi_desc, dev->reg_data_width);
+	if (ret != SUCCESS)
+		return ret;
+
+	spi_engine_set_speed(dev->spi_desc, dev->reg_access_speed);
+
+	buf[0] = ((reg_addr >> 8) & 0x7F);
+	buf[1] = 0xFF & reg_addr;
+	buf[2] = reg_data;
+
+	ret = spi_write_and_read(dev->spi_desc, buf, 3);
+
+	ret = spi_engine_set_transfer_width(dev->spi_desc,
+					    dev->capture_data_width);
+	if (ret != SUCCESS)
+		return ret;
+
+	spi_engine_set_speed(dev->spi_desc, dev->spi_desc->max_speed_hz);
+
+	return ret;
+}
+
+/**
+ * SPI read from device using a mask.
+ * @param dev - The device structure.
+ * @param reg_addr - The register address.
+ * @param mask - The mask.
+ * @param data - The register data.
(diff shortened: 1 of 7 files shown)


drivers/adc/ad469x/ad469x.c
```diff
@@ -0,0 +1,431 @@
...
+ * @param [in] dev - ad469x_dev device handler.
+ * @return \ref SUCCESS in case of success, \ref FAILURE otherwise.
+ */
+int32_t ad469x_exit_conversion_mode(struct ad469x_dev *dev)
+{
+	uint8_t buf[1];
+
+	buf[0] = AD469x_CMD_REG_CONFIG_MODE;
+
+	return spi_write_and_read(dev->spi_desc, buf, 1);
```
mchindri, line 208:
```suggestion
	uint8_t buf = AD469x_CMD_REG_CONFIG_MODE;

	return spi_write_and_read(dev->spi_desc, &buf, 1);
```

drivers/adc/ad469x/ad469x.c
```diff
@@ -0,0 +1,431 @@
...
+	ad469x_enter_conversion_mode(dev);
+	if (ret != SUCCESS)
+		goto error_spi;
+
+	ret = spi_engine_set_transfer_width(dev->spi_desc,
+					    spi_eng_init_param->data_width);
+	if (ret != SUCCESS)
+		goto error_spi;
+
+	spi_engine_set_speed(dev->spi_desc, dev->spi_desc->max_speed_hz);
```
mchindri, line 378: I see that to write to register the spi_engine_speed is changed.
If in the app there is a call to ad469x_spi_reg_read, spi_speed must be changed too?

cristiantpop:
V2:

- Remove macros: `AD469x_REG_READ, AD469x_REG_WRITE`

- Change speed to every register access

drivers/adc/ad469x/ad469x.c
cristiantpop: Yes, solved by changing speed to every register access

cristiantpop:
V3:

- add V2 the changes to the appropriate commits

cristiantpop:
V4:

- fix doxygen

cristiantpop:
It seems that we have a Travis build here:
`./adc/ad469x/ad469x.h:47:28: fatal error: clk_axi_clkgen.h: No such file or directory`
It seems that this clock is used in driver not in the project. Do you think that we should move this include from:

https://github.com/analogdevicesinc/no-OS/blob/ada32b5deed651a3ad7a0fff5e004cbd7006c38f/drivers/adc/ad469x/ad469x.h#L47

into:

https://github.com/analogdevicesinc/no-OS/blob/94391eee2c37f272899ea4f8bd27fe28039ca8b4/projects/ad469x_fmcz/src/ad469x_fmcz.c#L56

drivers/adc/ad469x/ad469x.c
```diff
@@ -0,0 +1,448 @@
...
+	int32_t ret;
+	uint8_t buf[3];
+
+	ret = spi_engine_set_transfer_width(dev->spi_desc, dev->reg_data_width);
+	if (ret != SUCCESS)
+		return ret;
+
+	spi_engine_set_speed(dev->spi_desc, dev->reg_access_speed);
+
+	buf[0] = ( (~(1 << 7)) & ((reg_addr >> 8) & 0x7F) );
```
mchindri, line 117:
```suggestion
	buf[0] = ((reg_addr >> 8) & 0x7F);
```
 (~(1 << 7)) is the same as 0x7F, you don't need to apply the mask twice

projects/ad469x_fmcz/src/ad469x_fmcz.c
```diff
@@ -0,0 +1,190 @@
...
+	};
+
+	Xil_ICacheEnable();
+	Xil_DCacheEnable();
+
+	ret = ad469x_init(&dev, &ad469x_init_param);
+	if (ret < 0)
+		return ret;
+
+	ret = spi_engine_offload_init(dev->spi_desc,
```
mchindri, line 154: Should we access spi_desc from dev in main? 
Isn't it better to have a ad469x_read and ad469x_write functions?

drivers/adc/ad469x/ad469x.h
```diff
@@ -0,0 +1,236 @@
...
+	/** Busy on gp3 */
+	AD469x_busy_gp3 = 1,
+};
+
+/**
+  * @enum ad469x_supported_dev_ids
+  * @brief Supported devices
+  */
+enum ad469x_supported_dev_ids {
+	ID_AD4695,
```
amiclaus, line 151: question: are these parts yet released?

drivers/adc/ad469x/ad469x.h
```diff
@@ -0,0 +1,236 @@
...
+/******************************************************************************/
+/*************************** Types Declarations *******************************/
+/******************************************************************************/
+/**
+ * @enum ad469x_interface_mode
+ * @brief Possible device communication modes
+ */
+enum ad469x_interface_mode {
+	/** Read/Write registers */
+	AD469x_IF_REGISTER_MODE = 0,
```
amiclaus, line 115: just a matter of preference:
`AD469x_IF_REGISTER_MODE` -> `AD469x_REGISTER_MODE`
`AD469x_IF_CONVERSION_MODE` -> `AD469x_CONVERSION_MODE`

drivers/adc/ad469x/ad469x.c
```diff
@@ -0,0 +1,448 @@
...
+	if (ret != SUCCESS)
+		return ret;
+
+	/** Reset to configure pins */
+	if (init_param->gpio_resetn) {
+		ret = gpio_direction_output(dev->gpio_resetn, false);
+		if (ret != SUCCESS)
+			return ret;
+
+		mdelay(100);
```
amiclaus, line 201: is the delay required?

amiclaus (commented):
> 
> 
> It seems that we have a Travis build here:
> `./adc/ad469x/ad469x.h:47:28: fatal error: clk_axi_clkgen.h: No such file or directory`
> It seems that this clock is used in driver not in the project. Do you think that we should move this include from:
> 
> https://github.com/analogdevicesinc/no-OS/blob/ada32b5deed651a3ad7a0fff5e004cbd7006c38f/drivers/adc/ad469x/ad469x.h#L47
> 
> into:
> 
> https://github.com/analogdevicesinc/no-OS/blob/94391eee2c37f272899ea4f8bd27fe28039ca8b4/projects/ad469x_fmcz/src/ad469x_fmcz.c#L56
> 
> ?

I don't think you should move the include in the main app, since the driver implementation itself is dependent on axi_clkgen functions.

I added here a PR that should fix the problem: #749

drivers/adc/ad469x/ad469x.c
cristiantpop: Yes, is necessary for reset

drivers/adc/ad469x/ad469x.h
cristiantpop: I don't think so

amiclaus (commented):
A link to the ADC datasheet in the commit description would be nice.

cristiantpop:
V5:

- remove access `spi_desc` in `main`?

-Implement  `ad469x_read_data` function

drivers/adc/ad469x/ad469x.c
```diff
@@ -0,0 +1,524 @@
...
+#include "stdbool.h"
+#include "ad469x.h"
+#include "spi_engine.h"
+#include "delay.h"
+#include "error.h"
+
+/******************************************************************************/
+/********************** Macros and Constants Definitions **********************/
+/******************************************************************************/
+#define AD469x_TEST_DATA 0xEA
```
amiclaus, line 55: Is this value contained in the driver specifications (datasheet value)? If so, I think it can be moved to `ad469x.h`

drivers/adc/ad469x/ad469x.c
```diff
@@ -0,0 +1,524 @@
...
+ * @return \ref SUCCESS in case of success, \ref FAILURE otherwise.
+ */
+int32_t ad469x_set_channel_sequence(struct ad469x_dev *dev,
+				    enum ad469x_channel_sequencing seq)
+{
+	int32_t ret;
+
+	switch (seq) {
+	case AD469x_single_cycle:
+		ret = ad469x_spi_write_mask(dev,
```
mirceacaprioru, line 226: check the ret after each operation do not use |=. This will not tell you where the operation failed.

drivers/adc/ad469x/ad469x.c
```diff
@@ -0,0 +1,524 @@
...
+	msg.commands = spi_eng_msg_cmds;
+	msg.no_commands = ARRAY_SIZE(spi_eng_msg_cmds);
+	msg.rx_addr = (uint32_t)buf;
+	msg.commands_data = commands_data;
+
+	ret = spi_engine_offload_transfer(dev->spi_desc, msg, samples * 2);
+	if (ret != SUCCESS)
+		return ret;
+
+	Xil_DCacheInvalidateRange(msg.rx_addr, samples * 4);
```
mirceacaprioru, line 401: why is there a call to a specific xilinx function ? The driver should be platform agnostic.

drivers/adc/ad469x/ad469x.c
```diff
@@ -0,0 +1,524 @@
...
+
+	if (data != AD469x_TEST_DATA)
+		goto error_spi;
+
+	ret = ad469x_set_busy(dev, AD469x_busy_gp0);
+	if (ret != SUCCESS)
+		goto error_spi;
+
+	ret = ad469x_set_channel_sequence(dev, AD469x_single_cycle);
+	ad469x_enter_conversion_mode(dev);
```
mirceacaprioru, line 469: check ret before entering conversion mode. Maybe the channels sequence failed so there is not need to enter this mode before checking.

mirceacaprioru (changes requested):
changes required

drivers/adc/ad469x/ad469x.c
```diff
@@ -0,0 +1,524 @@
...
+			     uint16_t reg_addr,
+			     uint8_t mask,
+			     uint8_t *data)
+{
+	uint8_t reg_data[3];
+	int32_t ret;
+
+	ret = ad469x_spi_reg_read(dev, reg_addr, reg_data);
+	*data = (reg_data[1] & mask);
+
```
adrimbarean, line 152: You could also use `find_first_set_bit` from `util` here to return the shifted value of the field.

drivers/adc/ad469x/ad469x.c
```diff
@@ -0,0 +1,524 @@
...
+ */
+int32_t ad469x_spi_write_mask(struct ad469x_dev *dev,
+			      uint16_t reg_addr,
+			      uint8_t mask,
+			      uint8_t data)
+{
+	uint8_t reg_data;
+	int32_t ret;
+
+	ret = ad469x_spi_reg_read(dev, reg_addr, &reg_data);
```
adrimbarean, line 172: It would probably be good to check `ret` after this instruction and not do the rest if it didn't work.

drivers/adc/ad469x/ad469x.c
```diff
@@ -0,0 +1,524 @@
...
+	reg_data &= ~mask;
+	reg_data |= data;
+	ret |= ad469x_spi_reg_write(dev, reg_addr, reg_data);
+
+	return ret;
```
adrimbarean, line 177: Can directly `return ad469x_spi_reg_write(dev, reg_addr, reg_data);` if ret was checked before.

drivers/adc/ad469x/ad469x.c
```diff
@@ -0,0 +1,524 @@
...
+{
+	int32_t ret;
+
+	ret = gpio_get_optional(&dev->gpio_resetn, init_param->gpio_resetn);
+	if (ret != SUCCESS)
+		return ret;
+
+	/** Reset to configure pins */
+	if (init_param->gpio_resetn) {
+		ret = gpio_direction_output(dev->gpio_resetn, false);
```
adrimbarean, line 198: If by some reason GPIO_LOW is different on a platform than 0, it will probably be safer to let the platform driver handle it. I suggest using GPIO_LOW instead of false.

drivers/adc/ad469x/ad469x.c
cristiantpop: No, it isn't, is just a random peak

drivers/adc/ad469x/ad469x.c
cristiantpop: I think is not a good idea, since mask can have more than one set bit

cristiantpop:
V6:

- Add `parameters.h`.

- Check the ret after each operation, remove `|=`. This will not tell you where the operation failed.

- Check return values, where chacks are missing.

- Remove xilinx platform dependency. Add `dcache_invalidate_range` function pointer.

- Use `GPIO_LOW` and `GPIO_HIGH` for seting GPIOs.

cristiantpop:
V7:

- Rename `ad469x_interface_mode` elements.

- Replace:
	`buf[0] = ( (~(1 << 7)) & ((reg_addr >> 8) & 0x7F) );`
	to:
	`buf[0] = ((reg_addr >> 8) & 0x7F);`

cristiantpop:
V8: 

- trigger  build

mchindri (approved):
From my side is all ok

mirceacaprioru (approved):

PR #750, opened 2020-09-09
drivers: platform: linux: gpio: Keep direction/value files opened

dbogdan:
... until gpio_remove() is called.

### DIFF
--- a/drivers/platform/linux/linux_gpio.c
+++ b/drivers/platform/linux/linux_gpio.c
@@ -51,6 +51,21 @@
 #include <sys/ioctl.h>
 
 /******************************************************************************/
+/*************************** Types Declarations *******************************/
+/******************************************************************************/
+
+/**
+ * @struct linux_gpio_desc
+ * @brief Linux platform specific GPIO descriptor
+ */
+struct linux_gpio_desc {
+	/** /sys/class/gpio/gpio"number"/direction file descriptor */
+	int direction_fd;
+	/** /sys/class/gpio/gpio"number"/value file descriptor */
+	int value_fd;
+};
+
+/******************************************************************************/
 /************************ Functions Definitions *******************************/
 /******************************************************************************/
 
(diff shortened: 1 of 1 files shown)


drivers/platform/linux/linux_gpio.c
```diff
@@ -50,6 +50,21 @@
 #include <unistd.h>
 #include <sys/ioctl.h>
 
+/******************************************************************************/
```
mchindri, line 53: commit title should be I thing 75 characters long

dbogdan: What do you mean? It has only 65 characters. "... until gpio_remove() is called." is the description :).

drivers/platform/linux/linux_gpio.c
```diff
@@ -73,34 +89,59 @@ int32_t gpio_get(struct gpio_desc **desc,
 	if (!descriptor)
 		return FAILURE;
 
+	linux_desc = (struct linux_gpio_desc*) malloc(sizeof(struct linux_gpio_desc));
```
mchindri, line 92:
```suggestion
	linux_desc = (struct linux_gpio_desc*)calloc(1, sizeof(*linux_desc));
```
To use *linux_desc is how linux coding style recommands it
Calloc is prefered because it initializes the memory with 0 and some errors in code can be found faster

drivers/platform/linux/linux_gpio.c
```diff
@@ -73,34 +89,59 @@ int32_t gpio_get(struct gpio_desc **desc,
...
+
+	descriptor->extra = linux_desc;
 	descriptor->number = param->number;
 
 	fd = open("/sys/class/gpio/export", O_WRONLY);
 	if (fd < 0) {
 		printf("%s: Can't open device\n\r", __func__);
-		free(descriptor);
-		return FAILURE;
+		goto free;
```
mchindri, line 102: nitpick: maybe an other label name can be used to not be the same as the function name.

dbogdan:
v2:
 - use calloc instead of malloc;
 - rename goto labels.

mchindri (approved):

PR #757, opened 2020-09-17
Ad400x fmcz

cristiantpop:
Small fix


amiclaus (approved):

amiclaus (commented):
Actually I have one comment:
Is the gpio used in this project?
Otherwise it shouldn't be included in the makefile

PR #764, opened 2020-09-22
tools:scripts:src_model: fix IGNORED_FILES variable

adrimbarean:
The generic.mk file actually looks for the variable IGNORED_FILES instead
of IGNORE_FILES. https://github.com/analogdevicesinc/no-OS/blob/master/tools/scripts/generic.mk#L47
Change example accordingly.


amiclaus (approved):
LGTM

mchindri (approved):

PR #768, opened 2020-09-23
iio_demo: Fix xilinx sdk project build

mchindri:
Some compiler flags and libraries where not added to the project.
No it can build.

### DIFF
--- a/tools/scripts/linux.mk
+++ b/tools/scripts/linux.mk
@@ -134,7 +134,7 @@ LDFLAGS = -T $(LSCRIPT)							\
 ifeq (xilinx,$(strip $(PLATFORM)))
 
 # Define the platform compiler switch
-CFLAGS += -D XILINX_PLATFORM						\
+CFLAGS += -DXILINX_PLATFORM						\
 	 -fdata-sections						\
 	 -ffunction-sections 						\
 	 -O2								\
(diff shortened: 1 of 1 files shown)


tools/scripts/generic.mk
```diff
@@ -8,6 +8,10 @@ rwildcard = $(wildcard $1$2) $(foreach d,$(wildcard $1*),$(call rwildcard,$d/,$2
 #                          COMMON INITIALIZATION
 #------------------------------------------------------------------------------
 
+#This compiler flags will be added to project if they are found in CFLAGS
+ADDITONAL_FLAGS = XILINX_PLATFORM ADUCM_PLATFORM ALTERA_PLATFORM \
```
buha, line 12: I'm not sure if this is the right approach. If I want to add a completely new flag, I need to edit this line too which is not ideal.

We should propagate everything that starts with -D from CFLAGS to the SDK.

mchindri: done

mchindri:
v2:
 - Do not use ADDITONAL_FLAGS. Add all -D flags.
 - add commit to move commands inside if

adrimbarean:
I think the first commit message needs to be updated to reflect the new approach.

mchindri:
v3:
 - Update first commit message

adrimbarean (approved):
Looks good to me.

amiclaus (approved):

buha:
These defines were added automatically by the scripts modified in this PR.
Screenshot from 2020-09-30 12-43-40

PR #773, opened 2020-09-28
drivers: gpio_get_optional fix

amiclaus:
The check of the initialization parameters should be
performed properly before accessing the platform specific function.

The check is kept also inside each platform specific function to make
sure that the drivers still work if used independently on the generic
gpio interface.


adrimbarean (approved):
Looks good to me.

PR #775, opened 2020-09-30
Ad469x

cristiantpop:
Add oversampling support and other small fixes

### DIFF
--- a/drivers/adc/ad469x/ad469x.c
+++ b/drivers/adc/ad469x/ad469x.c
@@ -261,6 +261,7 @@ int32_t ad469x_set_channel_sequence(struct ad469x_dev *dev,
 					    AD469x_SETUP_CYC_CTRL_SINGLE(0));
 		if (ret != SUCCESS)
 			return ret;
+		dev->num_slots = 0;
 
 		break;
 
@@ -285,6 +286,7 @@ int32_t ad469x_set_channel_sequence(struct ad469x_dev *dev,
 					    AD469x_SETUP_CYC_CTRL_SINGLE(1));
 		if (ret != SUCCESS)
 			return ret;
+		dev->num_slots = 0;
 
 		break;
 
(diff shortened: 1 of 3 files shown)


drivers/adc/ad469x/ad469x.c
```diff
@@ -309,13 +309,6 @@ int32_t ad469x_set_channel_sequence(struct ad469x_dev *dev,
 		if (ret != SUCCESS)
 			return ret;
 
-		ret = ad469x_spi_write_mask(dev,
```
adrimbarean, line 309: This could be merged with the commit that added the configuration to `ad469x_adv_sequence_set_num_slots()`.

drivers/adc/ad469x/ad469x.c
```diff
@@ -628,6 +630,7 @@ int32_t ad469x_init(struct ad469x_dev **device,
 	dev->capture_data_width = init_param->capture_data_width;
 	dev->dev_id = init_param->dev_id;
 	dev->dcache_invalidate_range = init_param->dcache_invalidate_range;
+	dev->ch_sequence = AD469x_standard_seq;
```
adrimbarean, line 786: nitpick: naming of `AD469x_standard_seq` is a bit odd as it mixes upper and lower case letters.

projects/ad469x_fmcz/src/ad469x_fmcz.c
```diff
@@ -164,9 +164,9 @@ int main()
 		if (ret != SUCCESS)
 			return ret;
 
-		for (i = 0; i < AD469x_EVB_SAMPLE_NO; i++) {
-			printf("ADC %"PRIu32": %"PRIu32" \n", i, buf[i]);
-		}
+		for (i = 0; i < AD469x_EVB_SAMPLE_NO; i += 3)
+			printf("ADC sample: %"PRIu32", ch0: %"PRIu32", ch1:%"PRIu32", temp: %"PRIu32"\n",
```
buha, line 179: I'm curious, why do we use PRIu32 instead of %lu ?

buha (approved):

PR #778, opened 2020-10-05
iio_axi_adc: get true sampling frequency

buha:
It is sometimes not possible to get the exact sampling frequency from
HDL registers. Application may know more about the number of lanes and
other parameters that affect the true sampling rate of an ADC.
    
By adding this function pointer to iio_axi_adc, the user may provide a
custom function that knows how to get the exact sampling rate if an axi_adc.

This PR also updates the Navassa project to make use of the new function
pointer in iio_axi_adc.

Depends on #771 , only last 2 commits are relevant for review here.


buha:
V2:
- rebase on latest #771

dbogdan (approved):

dbogdan:
This can be rebased on top of the current master branch.

buha:
Rebased on master due to #771 being merged to master already.

PR #781, opened 2020-10-06
include: print_log: add messages helpers

amiclaus:
Add debug/error/warning messages helpers used when running
projects/drivers.

Example of usage inside main application source files:

`#define LOG_LEVEL 4`
`#include "print_log.h"`

### DIFF
--- a/include/print_log.h
+++ b/include/print_log.h
@@ -1,0 +1,110 @@
+/***************************************************************************//**
+ *   @file   print_log.h
+ *   @brief  Print messages helpers.
+ *   @author Antoniu Miclaus (antoniu.miclaus@analog.com)
+********************************************************************************
+ * Copyright 2020(c) Analog Devices, Inc.
+ *
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *  - Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ *  - Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in
+ *    the documentation and/or other materials provided with the
+ *    distribution.
+ *  - Neither the name of Analog Devices, Inc. nor the names of its
+ *    contributors may be used to endorse or promote products derived
+ *    from this software without specific prior written permission.
+ *  - The use of this software may or may not infringe the patent rights
+ *    of one or more patent holders.  This license does not release you
+ *    from the requirement that you obtain separate licenses from these
+ *    patent holders to use this software.
+ *  - Use of the software either in source or binary form, must be run
+ *    on or directly connected to an Analog Devices Inc. component.
+ *
+ * THIS SOFTWARE IS PROVIDED BY ANALOG DEVICES "AS IS" AND ANY EXPRESS OR
+ * IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, NON-INFRINGEMENT,
+ * MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED.
+ * IN NO EVENT SHALL ANALOG DEVICES BE LIABLE FOR ANY DIRECT, INDIRECT,
+ * INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, INTELLECTUAL PROPERTY RIGHTS, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+*******************************************************************************/
+
+#ifndef PRINT_LOG_H_
+#define PRINT_LOG_H_
+
+#include <stdio.h>
+
+#define LOG_EMERG	0x0
+#define LOG_ALERT	0x1
+#define LOG_CRIT	0x2
+#define LOG_ERR		0x3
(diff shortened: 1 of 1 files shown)


dbogdan:
We need to have something similar to Linux's console_loglevel - look here: https://elinux.org/Debugging_by_printing - "kernel compares the log level of the message to the console_loglevel (a kernel variable) and if the priority is higher (i.e. a lower value) than the console_loglevel the message will be printed to the current console".

dbogdan:
Regarding the naming, I would use the "pr_" prefix (e.g., pr_err(), pr_info(), ...). Since we don't have a real kernel, printk.h doesn't sound good for me either.

amiclaus:
V2:

- add print log levels
- use different namings

amiclaus:
V3:

- minor fix

amiclaus:
V4:
- remove `print` string from each log level

include/print_log.h
```diff
@@ -0,0 +1,110 @@
+/***************************************************************************//**
+ *   @file   print_log.h
```
mchindri, line 2: why not debug.h ?

include/print_log.h
```diff
@@ -0,0 +1,110 @@
...
+#define LOG_EMERG	0x0
+#define LOG_ALERT	0x1
+#define LOG_CRIT	0x2
+#define LOG_ERR		0x3
+#define LOG_WARNING	0x4
+#define LOG_NOTICE	0x5
+#define LOG_INFO	0x6
+#define LOG_DEBUG	0x7
+
+#if defined(LOG_LEVEL) && LOG_LEVEL >= LOG_EMERG && LOG_LEVEL <= LOG_DEBUG
```
mchindri, line 54: I think the last condition is redundant: (LOG_LEVEL <= LOG_DEBUG)

include/print_log.h
```diff
@@ -0,0 +1,110 @@
...
+#define pr_emerg(fmt, args...) printf("Emergency: %s:%s:%d(): " \
+fmt, __FILE__, __func__, __LINE__, ##args)
+#else
+#define pr_emerg(fmt, args...)
+#endif
+
+#if defined(LOG_LEVEL) && LOG_LEVEL >= LOG_ALERT && LOG_LEVEL <= LOG_DEBUG
+#define pr_alert(fmt, args...) printf("Alert: %s:%s:%d(): " \
```
mchindri, line 62: I think is better if we indent the first text: 
Ex
```
Alert   :
Error   : 
Critical:
```

include/print_log.h
```diff
@@ -0,0 +1,110 @@
...
+#include <stdio.h>
+
+#define LOG_EMERG	0x0
+#define LOG_ALERT	0x1
+#define LOG_CRIT	0x2
+#define LOG_ERR		0x3
+#define LOG_WARNING	0x4
+#define LOG_NOTICE	0x5
+#define LOG_INFO	0x6
+#define LOG_DEBUG	0x7
```
mchindri, line 52: Why not use just 4 Levels:
Error, Warning, Info and Debug?
And for linux compatiblity use: 
`#define pr_alert(fmt, args...) pr_err(fmt, args...)`

include/print_log.h
amiclaus: The file contains multiple log levels. Debug is only one of them.

include/print_log.h
amiclaus: I think we should not let the user define LOG_LEVELS outside the boundaries.

include/print_log.h
amiclaus: Agree.

include/print_log.h
amiclaus: I think this is a matter of preference. I tried to align as much as possible with the linux log levels:
https://elinux.org/Debugging_by_printing

amiclaus:
V5:

- indent log level string

include/print_log.h
```diff
@@ -0,0 +1,110 @@
...
+
+#if defined(LOG_LEVEL) && LOG_LEVEL >= LOG_NOTICE && LOG_LEVEL <= LOG_DEBUG
+#define pr_notice(fmt, args...) printf("Notice   : %s:%s:%d(): " \
+fmt, __FILE__, __func__, __LINE__, ##args)
+#else
+#define pr_notice(fmt, args...)
+#endif
+
+#if defined(LOG_LEVEL) && LOG_LEVEL >= LOG_INFO && LOG_LEVEL <= LOG_DEBUG
+#define pr_info(fmt, args...) printf("Info     : %s:%s:%d(): " \
```
dbogdan, line 97: My preference is to not add a prefix to the pr_info() string. If we would agree to do this, the additional spaces for the rest of the cases won't be necessary anymore.

include/print_log.h
```diff
@@ -0,0 +1,110 @@
...
+
+#if defined(LOG_LEVEL) && LOG_LEVEL >= LOG_CRIT && LOG_LEVEL <= LOG_DEBUG
+#define pr_crit(fmt, args...) printf("Critical : %s:%s:%d(): " \
+fmt, __FILE__, __func__, __LINE__, ##args)
+#else
+#define pr_crit(fmt, args...)
+#endif
+
+#if defined(LOG_LEVEL) && LOG_LEVEL >= LOG_ERR && LOG_LEVEL <= LOG_DEBUG
+#define pr_err(fmt, args...) printf("Error    : %s:%s:%d(): " \
```
dbogdan, line 76: Parsing the log would be easier if we will have "ERR:" instead of "Error:". Other messages might contain "Error" without being printed by pr_err() (e.g., "No Errors. Good to Go!"). "ERR" will be less likely present somewhere else. Same for the rest of the prints: EMERG, ALERT, CRIT, ERR, WARNING, NOTICE, DEBUG are better from my point of view.

include/print_log.h
amiclaus: Ok, I will remove the prefix for the pr_info(). But the rest of the cases still have prefixes with different lengths. Should I keep there the spaces or no?

include/print_log.h
amiclaus: I agree. Should we keep the spaces for the alignment of the new prefixes?

amiclaus:
V6:
- use uppercase letters for prefixes
- remove prefix for `pr_info`

amiclaus:
Waiting for feedback if to add or not extra spaces for indenting the content after the prefixes.

include/print_log.h
dbogdan: I wouldn't keep them.

include/print_log.h
dbogdan: From my point of view, no.

dbogdan (approved):

mchindri (approved):

PR #782, opened 2020-10-06
Iio fix

cristiantpop:
Iio fix

### DIFF
--- a/libraries/iio/iio_types.h
+++ b/libraries/iio/iio_types.h
@@ -58,7 +58,9 @@
 enum iio_chan_type {
 	IIO_VOLTAGE,
 	IIO_CURRENT,
-	IIO_ALTVOLTAGE
+	IIO_ALTVOLTAGE,
+	/* All new types must be added before this field */
+	IIO_LAST_TYPE
 };
 
 /**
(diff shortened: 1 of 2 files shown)


dbogdan (approved):

libraries/iio/iio_types.h
```diff
@@ -121,6 +121,12 @@ struct iio_channel {
 	struct iio_attribute **attributes;
 	/** if true, the channel is an output channel */
 	bool ch_out;
+	/** channel offset
+	 * Ex:
+	 * offset 0, ch[0, 1]: voltage0, voltage1
+	 * offset 2, ch[2, 5]: altvoltage0, altvoltage1, altvoltage2, altvoltage3
+	 * */
+	int32_t offset;
```
mchindri, line 129: I prefer a change like this:
https://github.com/analogdevicesinc/no-OS/commit/b757cbd3f57cd08914eeadd1131789ee5c03aa2d
This way the user don't need to set the offset value. Also it is a configuration that must be done from iio not from the app side.

cristiantpop: That is a good idea, thanks!

cristiantpop:
V2:

- Change offset calculation to Mihail proposal: b757cbd

mchindri (approved):

PR #783, opened 2020-10-07
drivers: jesd204: axi_adxcvr: Fix adxcvr_init return

scuciurean:
In a case where the adxcvr_clk_set_rate() will fail to find a satyisfying
lane rate, the function will return an error code, without saving the xcvr
descriptor to the given address.
This change first assigns the descriptor and returns the error code after.

Signed-off-by: Sergiu Cuciurean <sergiu.cuciurean@analog.com>

### DIFF
--- a/drivers/axi_core/jesd204/axi_adxcvr.c
+++ b/drivers/axi_core/jesd204/axi_adxcvr.c
@@ -411,7 +411,7 @@ int32_t adxcvr_init(struct adxcvr **ad_xcvr,
 	if (xcvr->lane_rate_khz && xcvr->ref_rate_khz) {
 		ret = adxcvr_clk_set_rate(xcvr, xcvr->lane_rate_khz, xcvr->ref_rate_khz);
 		if (ret)
-			return ret;
+			goto err;
 	}
 
 	*ad_xcvr = xcvr;


amiclaus (approved):
LGTM.

Nitpick in commit message: `satyisfying` -> `satisfying`

drivers/axi_core/jesd204/axi_adxcvr.c
```diff
@@ -408,15 +408,12 @@ int32_t adxcvr_init(struct adxcvr **ad_xcvr,
 		}
 	}
 
-	if (xcvr->lane_rate_khz && xcvr->ref_rate_khz) {
+	if (xcvr->lane_rate_khz && xcvr->ref_rate_khz)
 		ret = adxcvr_clk_set_rate(xcvr, xcvr->lane_rate_khz, xcvr->ref_rate_khz);
```
mchindri, line 412: Shouldn't cleanup be done if an error occurs?

dbogdan:
I'm not convinced :). I would keep the old approach but free(xcvr) in case of failure.
If the specified lane_rate_khz can't be set, the user has two options: to choose another value or to not specify one at all.

scuciurean:
Changelog -> V2:
- Free the memory if the transceiver can't find a lane rate with a vco frequency between the required ranges.

dbogdan (approved):

dbogdan:
I'm happy now. :)

mchindri (approved):

PR #784, opened 2020-10-08
Makefile fix

mchindri:
This changes were tested with iio_demo and daq2 to build both in terminal and in sdk


amiclaus (commented):
Nitpick:
for `tools:scripts: Fix build errors` the commit message should comply the 50/72 format rule.

tools/scripts/linux.mk
```diff
@@ -109,8 +109,6 @@ LIB_TINYIIOD_PATH = $(LIBRARIES)/libtinyiiod/build
 LIB_TINYIIOD = tinyiiod
 TINYIIOD_STD_TYPES = _USE_STD_INT_TYPES
 CFLAGS += -D $(TINYIIOD_STD_TYPES)
-EXTRA_LIBS += $(LIB_TINYIIOD)
-EXTRA_LIBS_PATHS += $(LIB_TINYIIOD_PATH)
```
scuciurean, line 113: If those are removed, what's happening with 
 `
ADD_LIBRARIES = $(foreach lib, $(EXTRA_LIBS), $(call add_library_cmd,$(lib)))
ADD_LIBRARIES_PATH = $(foreach path, $(EXTRA_LIBS_PATHS), $(call lib_serach_path_cmd,$(path)))
`
At the moment, the projects don't use the generic.mk file

mchindri: They are added by create_project.tcl . That's the error was to add the here too, causing the duplication error.

mchindri:
v2:
 - Fix commit message to fit 50/72 rule

adrimbarean (approved):
Looks good to me.

PR #788, opened 2020-10-14
.gitignore: Add more files to be excluded

mchindri:
These files are binary file and there is no need to be tracked by git.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,4 +11,10 @@ build
 dfp_drivers
 aducm_project
 aducm_build
+.vscode
 
+*.a
+*.o
+*.d
+*.hex
+*.hdf


.gitignore
```diff
@@ -12,3 +12,8 @@ dfp_drivers
 aducm_project
 aducm_build
 
+*.a
+*.o
+*.d
+*.hex
+*.hdf
```
buha, line 19: I like seeing my .hdf's with a git status :D

cristiantpop:
I didn't see your pull request :)

.gitignore
cristiantpop: Also I would like to see here: *.a, .vscode
I am for *.hdf in .gitignore

mchindri:
v2: 
 - Added .vscode . (.a was already added)

.gitignore
mchindri: As per slack vote I will keep .hdf

scuciurean:
I think we can remove _.a_ from the gitignore file. Maybe in the future we will add device firmwares in this format.

mchindri:
> I think we can remove _.a_ from the gitignore file. Maybe in the future we will add device firmwares in this format.

They can be added manually with "git add -f somefile.a" and then they will be tracked normally.

PR #790, opened 2020-10-15
libraries:iio: Add modifier to channel type.

kister-jimenez:
Added channel properties to make it as close as possible to linux iio:
- .channel
- .channel2
- .modified
- .indexed

The modified is a property used in linux iio on the channel spec that 
when set to true, the channel2 property is used as the selector for the
modifier. For example, if the type is IIO_ANGL_VEL this will use 
"anglvel" as the channel type. If modified is true and channel2 is set 
to IIO_MOD_X, the type becomes "anglvel_x".

This also affects the indexing which uses the channel and indexed 
property were added. if channel indexed is true, the value set in the 
channel property is attached to the channel. For example, "altvoltage1"
is channel type altvoltage, channel=1 and indexed=true.

Signed-off-by: Kister Jimenez <kister.jimenez@analog.com>

### DIFF
--- a/libraries/iio/iio.c
+++ b/libraries/iio/iio.c
@@ -130,7 +130,6 @@ struct iio_desc {
 	uint32_t		xml_size;
 	uint32_t		xml_size_to_last_dev;
 	uint32_t		dev_count;
-	int32_t			id_offsets[IIO_LAST_TYPE];
 	struct uart_desc	*uart_desc;
 #ifdef ENABLE_IIO_NETWORK
 	/* FIFO for socket descriptors */
@@ -314,9 +313,13 @@ static char *get_channel_id(enum iio_chan_type type)
 	return "";
 }
 
-static inline void _print_ch_id(char *buff, enum iio_chan_type type, int32_t id)
+static inline void _print_ch_id(char *buff, struct iio_channel *ch)
 {
-	sprintf(buff, "%s%d", get_channel_id(type), (int)id);
+	if(ch->indexed) {
+		sprintf(buff, "%s%d", get_channel_id(ch->ch_type), (int)ch->channel);
+	} else {
+		sprintf(buff, "%s", get_channel_id(ch->ch_type));
+	}
 }
 
 /**
@@ -333,8 +336,7 @@ static inline struct iio_channel *iio_get_channel(const char *channel,
 	char	ch_id[20];
 
 	while (channels[i]) {
-		_print_ch_id(ch_id, channels[i]->ch_type,
-			     channels[i]->reserved);
+		_print_ch_id(ch_id, channels[i]);
 		if (!strcmp(channel, ch_id) && (channels[i]->ch_out == ch_out))
 			return channels[i];
 		i++;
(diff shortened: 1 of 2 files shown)


libraries/iio/iio_types.h
```diff
@@ -95,15 +107,15 @@ struct iio_attribute {
...
+	uint8_t realbits;
 	/** Realbits + padding */
-	uint8_t			storagebits;
+	uint8_t storagebits;
 	/** Shift right by this before masking out realbits. */
-	uint8_t			shift;
+	uint8_t shift;
 	/** True if big endian, false if little endian */
-	bool			is_big_endian;
+	bool is_big_endian;
```
mchindri, line 118: It is easier to see the variable names if they are indented

libraries/iio/iio_types.h
```diff
@@ -1,4 +1,4 @@
-/***************************************************************************//**
+/***************************************************************************/ /**
```
mchindri, line 1: remove added space

mchindri (commented):
I would split this in 3 commits.
- Get ch_num from scan_index
- Add indexed field
- Add modifiers
-

kister-jimenez:
> I would split this in 3 commits.
> 
> * Get ch_num from scan_index
> * Add indexed field
> * Add modifiers

These 3 are needed by the ch type modifier to work. The way ch_num was previously accessed was dependent on the number after the channel type in which, if we implement indexed and modified fields, will not work because these two may cause the channel type to not have a number after it

I added the modifiers_name and iio_modifier enums, for the modifier to work also. So maybe what I can split is the addition of iio_chan_type entries to suppor IIO_ANGL_VEL, and IIO_TEMP. The next commit will be the addition of fields. Then, last lastly the usage of the modifier in iio along with the iio_modifier enum and iio_modifier_names array.

Is that ok?

mchindri:
> > I would split this in 3 commits.
> > 
> > * Get ch_num from scan_index
> > * Add indexed field
> > * Add modifiers
> 
> These 3 are needed by the ch type modifier to work. The way ch_num was previously accessed was dependent on the number after the channel type in which, if we implement indexed and modified fields, will not work because these two may cause the channel type to not have a number after it
> 
> I added the modifiers_name and iio_modifier enums, for the modifier to work also. So maybe what I can split is the addition of iio_chan_type entries to suppor IIO_ANGL_VEL, and IIO_TEMP. The next commit will be the addition of fields. Then, last lastly the usage of the modifier in iio along with the iio_modifier enum and iio_modifier_names array.
> 
> Is that ok?

It should be ok.

kister-jimenez:
V2. 
-Split in multiple commits.
- Reverted the spacing of some affected fields.
- Removed the space accidentally added in iio_types.h

libraries/iio/iio_types.h
```diff
@@ -127,6 +133,12 @@ struct iio_channel {
 	bool ch_out;
 	/** Reserved. Id offset */
 	int32_t reserved;
```
cristiantpop, line 146: Is this variable used still somewhere?

cristiantpop (approved):

libraries/iio/iio.c
```diff
@@ -597,15 +596,12 @@ static ssize_t iio_ch_write_attr(const char *device_id, const char *channel,
 	if (!dev)
 		return -ENOENT;
 
-	ch_info.ch_out = ch_out;
-	ch_info.ch_num = iio_get_channel_number(channel);
```
mchindri, line 597: You can also remove iio_get_channel_number function because it is not used.

libraries/iio/iio_types.h
```diff
@@ -117,6 +117,12 @@ struct iio_channel {
 	char			*name;
 	/** Chanel type */
 	enum iio_chan_type	ch_type;
+	/** Channel number when the same channel type */
+	int 			channel;
```
mchindri, line 132: Nitpick: I think it makes more sense if you add this channel in the 3-rd commit and  channel2 in the 4-th commit.

mchindri (approved):

PR #793, opened 2020-10-16
spi_engine: update write/read function

amiclaus:
Do not skip the first byte. This implies shifting the data byte in each
driver that uses SPI Engine when performing a read operation.


adrimbarean (approved):
Looks good to me.

amiclaus:
V2:
- fix astyle

dbogdan (approved):

PR #801, opened 2020-10-27
Buffer iio

cristiantpop:


cristiantpop:
V2:

- fix srk.mk: remove new line separator before `endif`

PR #809, opened 2020-11-03
Miscellaneous GPIO fixes

dbogdan:


adrimbarean (approved):
Looks good to me.

drivers/platform/linux/linux_gpio.h
```diff
@@ -0,0 +1,47 @@
+/*******************************************************************************
+ *   @file   linux/linux_gpio.h
```
mchindri, line 2: For consistency it should be gpio_extra.h

dbogdan: We shouldn't do it anymore since now, multiple types of GPIO controllers can be instantiated in the same project.

amiclaus (approved):

PR #810, opened 2020-11-03
Remove legacy ad9361 and extend current version

dbogdan:


buha (approved):
really well split into commits in a logical way !

drivers/rf-transceiver/ad9361/common.h
```diff
@@ -43,18 +43,7 @@
...
-/******************************************************************************/
-#define EIO			5	/* I/O error */
-#define EAGAIN		11	/* Try again */
-#define ENOMEM		12	/* Out of memory */
-#define EFAULT		14	/* Bad address */
-#define ENODEV		19	/* No such device */
-#define EINVAL		22	/* Invalid argument */
-#define EOPNOTSUPP	45	/* Operation not supported on transport endpoint */
-#define ETIMEDOUT	110	/* Connection timed out */
+#include "error.h"
```
mchindri, line 46: why not remove this file and include this headers where needed?

mchindri (approved):

drivers/rf-transceiver/ad9361/common.h
dbogdan: It might be an option too. In this commit, I tried to make as few changes as I could.

PR #824, opened 2020-11-13
Remove unused function and add ch name checking in xml.

kister-jimenez:
Some iio library changes:
- removed unused function: int32_t iio_get_channel_number(const char *ch)
- added channel name checking before printing name field in xml. This makes the name field optional so tools like iio-osc and iio_info will use the channel id as names if not provided.


mchindri (approved):
LGTM

dbogdan (approved):

amiclaus (approved):

PR #798, opened 2020-10-20
Xilinx  makefile improvements [WIP]

mchindri:
See last commit

PR #830, opened 2020-11-20
Xilinx makefile improvements [UPDATED]

mchindri:
This is the same as https://github.com/analogdevicesinc/no-OS/pull/798 (Closed by mistake and can't reopen)

V2:
 - Split changes in multiple commits
 - Fix https://github.com/analogdevicesinc/no-OS/pull/798#issuecomment-725267643

### DIFF
--- a/tools/scripts/aducm.mk
+++ b/tools/scripts/aducm.mk
@@ -120,14 +120,10 @@ $(call escape_spaces,$(CCES_HOME))/ARM/packs/ARM/CMSIS/*/CMSIS/Core)
 ifneq ($(words $(CCES_HOME)), 1)
 
 ifeq ($(wildcard $(PLATFORM_TOOLS)/dfp_drivers),)
-$(warning ERROR:$(ENDL)\
-CCES_HOME dir have spaces. To avoid this you can install CCES in a path without\
-spaces.$(ENDL)$(ENDL) Or you can copy the dfp into noos running:$(ENDL)$(ENDL)\
- make install_dfp$(ENDL))
-else
-DFP_DRIVERS = $(PLATFORM_TOOLS)/dfp_drivers
-
+#Create link to dfp to noos if it contains spaces (default on windows)
+DUMMY_VAR := $(shell $(call make_dir_link,$(CCES_HOME),$(PLATFORM_TOOLS)/dfp_drivers))
 endif
+DFP_DRIVERS = $(PLATFORM_TOOLS)/dfp_drivers
 
 else
 
--- a/tools/scripts/aducm.mk
+++ b/tools/scripts/aducm.mk
@@ -71,9 +71,6 @@ endif
 #                             DFP DEPENDENCIES                          
 #------------------------------------------------------------------------------
 
-#Get all src files that are not in SRC_DRIS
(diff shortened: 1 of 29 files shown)


mchindri:
v3:
 - Fix run rule for xilinx.
 - Fix CFLAGS in libraries

tools/scripts/aducm.mk
```diff
@@ -120,14 +120,10 @@ $(call escape_spaces,$(CCES_HOME))/ARM/packs/ARM/CMSIS/*/CMSIS/Core)
...
 
 ifeq ($(wildcard $(PLATFORM_TOOLS)/dfp_drivers),)
-$(warning ERROR:$(ENDL)\
-CCES_HOME dir have spaces. To avoid this you can install CCES in a path without\
-spaces.$(ENDL)$(ENDL) Or you can copy the dfp into noos running:$(ENDL)$(ENDL)\
- make install_dfp$(ENDL))
-else
-DFP_DRIVERS = $(PLATFORM_TOOLS)/dfp_drivers
-
+#Create link to dfp to noos if it contains spaces (default on windows)
```
adrimbarean, line 59: I think I prefer copying the DFP sources instead of linking. If for whatever reason the user decides to try and change some of the DFP sources during a development process, this shouldn't affect other projects.

mchindri: If it modifies the dfp for one project (the copy) an other project will use the same dfp

mchindri:
v4:
 - Add LINK_SRCS feature
 - Remove some code used for debugging

mchindri:
V5:
 - Use hard links instead of soft links

scuciurean:
If you want to move the content from a file to another, make sure you don't lose the history of the old one. If not, the new file will have only one commit with a description linked to some commits that were related to the old file.

mchindri:
v6:
 - move setting of PLATFORM to generic_variables.mk to be available in src.mk
 - remove USE_OLD_MAKEFILE = n used for debug.
   If we merge this branch, by default the old make aproach will be used in order to not brake project.
   In order to use the new one use 'make USE_OLD_MAKEFILE = n'

mchindri:
v7:
 - Update commit description for xilinx.mk to reflect from where it came from. (Link to linux.mk and windows.mk at a specific commit)

buha:
`make USE_OLD_MAKEFILE = n`
This is awkward for the user. I think the new approach should be used by default, there shouldn't be any old build system, only the new one. If you think transition could take long, do something like `make LEGACY_BUILD=y` but I would avoid it altogether.

mchindri:
> make LEGACY_BUILD=y

I need

mchindri:
> `make USE_OLD_MAKEFILE = n`
> This is awkward for the user. I think the new approach should be used by default, there shouldn't be any old build system, only the new one. If you think transition could take long, do something like `make LEGACY_BUILD=y` but I would avoid it altogether.

I need to add also the altera build for a complete transition. That's why I chose this approach

mchindri:
v8:
- By default use new build and use LEGACY_BUILD=y for altera and old build. Will document this on wiki

buha (approved):

mchindri:
If no more comments will merge :)

PR #833, opened 2020-11-23
spi_engine: driver updates

amiclaus:
- rearrange SPI commands order 
- start SPI transfer before the DMA 

Changes required for ad7616-sdz project update.

**NOTE:** changes tested on https://github.com/analogdevicesinc/no-OS/tree/master/projects/ad738x_fmcz

Signed-off-by: Antoniu Miclaus <antoniu.miclaus@analog.com>

PR #832, opened 2020-11-23
spi_engine: expose dma flags attribute

amiclaus:
Allow the user to configure the dma flags.

The default value is DMA_CYCLIC.

Signed-off-by: Antoniu Miclaus <antoniu.miclaus@analog.com>

PR #834, opened 2020-11-23
AD7616 update driver and project

amiclaus:
Update AD7616 driver and AD7616-sdz project to the latest platform drivers.

**NOTE:** Pull Request is dependent on: #833 and #832 


drivers/adc/ad7616/ad7616.c
```diff
@@ -191,16 +195,16 @@ int32_t ad7616_spi_write(ad7616_dev *dev,
  * @param reg_data - The register data.
  * @return 0 in case of success, negative error code otherwise.
  */
-int32_t ad7616_par_read(ad7616_dev *dev,
-						uint8_t reg_addr,
-						uint16_t *reg_data)
+int32_t ad7616_par_read(struct ad7616_dev *dev,
```
adrimbarean, line 198: Does this change mean that the driver will support the specific HDL core for this device? Do we want the device driver to do this?

drivers/adc/ad7616/ad7616.c
```diff
@@ -356,83 +358,218 @@ int32_t ad7616_set_mode(ad7616_dev *dev,
...
-						GPIO_HIGH);
+	ad7616_core_setup(dev);
+	if (dev->interface == AD7616_SERIAL) {
+		ret = spi_init(&dev->spi_desc, init_param->spi_param);
+		spi_engine_set_speed(dev->spi_desc, dev->reg_access_speed);
+	}
+	if (ret != SUCCESS)
+		return ret;
+
+	ret |= gpio_get_optional(&dev->gpio_hw_rngsel0,
```
cristiantpop, line 524: If you check the `ret`  value right after the function call is not necessary to `|` it

drivers/adc/ad7616/ad7616.c
amiclaus: The driver is dependent on the HDL core. (see also the old driver implementation: https://github.com/analogdevicesinc/no-OS/blob/d0bbfe561c64e539e5ab64dd8473f93847faa1db/ad7616-sdz/ad7616.c#L194-L226 )

This approach is meant to remove the ad7616_core layer, since the serial capture is already covered by the new spi_engine.
And for the parallel capture, the ad7616_core contains mainly AXI_IO read/write wrappers.

drivers/adc/ad7616/ad7616.c
amiclaus: you are right. will change.

amiclaus:
V2:
- fix doxygen
- fix drivers build (add new commit for the drivers makefile)
- remove unnecessary logical operation on the `ret` variable.

drivers/adc/ad7616/ad7616.c
```diff
@@ -356,83 +358,218 @@ int32_t ad7616_set_mode(ad7616_dev *dev,
...
-		ret |= gpio_set_value(&dev->gpio_dev,
-						dev->gpio_reset,
-						GPIO_HIGH);
+	ad7616_core_setup(dev);
+	if (dev->interface == AD7616_SERIAL) {
+		ret = spi_init(&dev->spi_desc, init_param->spi_param);
+		spi_engine_set_speed(dev->spi_desc, dev->reg_access_speed);
+	}
+	if (ret != SUCCESS)
+		return ret;
```
mchindri, line 522: dev should be freed in case of error.

amiclaus:
V3:

- fix astyle
- free `dev` if `spi_init` fails.

adrimbarean (approved):
Looks good to me.

PR #837, opened 2020-11-25
Add UART Linux Support

dbogdan:


include/uart.h
```diff
@@ -49,6 +49,49 @@
 /*************************** Types Declarations *******************************/
 /******************************************************************************/
 
+/**
+ * @enum uart_size
+ * @brief UART character size (number of data bits) options.
+ */
+typedef enum uart_size {
```
cristiantpop, line 56: Linux coding style tends to avoid using typedefs

drivers/platform/linux/linux_uart.c
```diff
@@ -0,0 +1,286 @@
...
+ * @param param - The structure that contains the UART parameters.
+ * @return SUCCESS in case of success, FAILURE otherwise.
+ */
+int32_t uart_init(struct uart_desc **desc, struct uart_init_param *param)
+{
+	struct linux_uart_init_param *linux_init;
+	struct linux_uart_desc *linux_desc;
+	struct uart_desc *descriptor;
+	speed_t speed;
+	char path[64];
```
cristiantpop, line 84: Use `uint8_t` for consistency.

drivers/platform/linux/linux_uart.c
```diff
@@ -0,0 +1,286 @@
...
+		goto free_desc;
+
+	linux_desc->terminal = (struct termios*) malloc(sizeof(struct termios));
+	if (!linux_desc->terminal)
+		goto free_linux_desc;
+
+	descriptor->extra = linux_desc;
+	linux_init = param->extra;
+
+	snprintf(path, sizeof(path), "/dev/%s", linux_init->device_id);
```
cristiantpop, line 101: You should check if the string fitted in path

drivers/platform/linux/linux_uart.c
```diff
@@ -0,0 +1,286 @@
...
+			linux_desc->terminal->c_cflag |= CS7;
+			break;
+		case UART_CS_8:
+			linux_desc->terminal->c_cflag |= CS8;
+			break;
+	}
+
+	linux_desc->terminal->c_cflag &= ~PARENB;
+	linux_desc->terminal->c_cflag &= ~PARODD;
+	switch(param->parity) {
```
cristiantpop, line 183: You should add a default, since there are more than two possible parities and we are handling only two for the momment.

drivers/platform/linux/linux_uart.c
```diff
@@ -0,0 +1,286 @@
...
+	int32_t ret;
+
+	linux_desc = desc->extra;
+
+	ret = close(linux_desc->fd);
+	if (ret < 0) {
+		printf("%s: Can't close device\n\r", __func__);
+		return FAILURE;
+	}
+
```
cristiantpop, line 236: Free this:
	free(linux_desc->terminal);
	free(linux_desc);

drivers/platform/linux/linux_uart.h
```diff
@@ -0,0 +1,52 @@
+/*******************************************************************************
+ *   @file   linux/linux_i2c.h
```
cristiantpop, line 2: Change to linux_uart

drivers/platform/linux/linux_uart.h
```diff
@@ -0,0 +1,52 @@
...
+/**
+ * @struct linux_uart_init_param
+ * @brief Structure holding the initialization parameters for Linux platform
+ * specific UART parameters.
+ */
+struct linux_uart_init_param {
+	/** UART device ID (/dev/"device_id") */
+	const char *device_id;
+};
+
```
adrimbarean, line 51: Do we need here declarations for the source functions?

include/uart.h
dbogdan: Done.

drivers/platform/linux/linux_uart.c
dbogdan: I'm not convinced that we should do this for the C standard libraries - it might trigger some cast warnings, isn't it?

drivers/platform/linux/linux_uart.c
dbogdan: Done.

drivers/platform/linux/linux_uart.c
dbogdan: Done.

drivers/platform/linux/linux_uart.c
dbogdan: Done.

drivers/platform/linux/linux_uart.h
dbogdan: Done.

drivers/platform/linux/linux_uart.h
dbogdan: They are present in the "uart.h".

dbogdan:
v2:
 - remove typedefs;
 - check snprintf() result;
 - add default case for switch(param->size) and switch(param->parity);
 - fix uart_remove();
 - update comments;
 - update uart_write() for O_NONBLOCK.

drivers/platform/linux/linux_uart.c
buha: i also think there might be warnings, char is ok for strings

drivers/platform/linux/linux_uart.c
```diff
@@ -0,0 +1,298 @@
...
+/******************************************************************************/
+
+/**
+ * @struct linux_uart_desc
+ * @brief Linux platform specific UART descriptor
+ */
+struct linux_uart_desc {
+	/** /dev/"device_id" file descriptor */
+	int fd;
+	struct termios *terminal;
```
buha, line 65: doxygen will likely complain about undocumented terminal

drivers/platform/linux/linux_uart.c
```diff
@@ -0,0 +1,298 @@
...
+			speed = B9600;
+			break;
+		case 19200:
+			speed = B19200;
+			break;
+		case 38400:
+			speed = B38400;
+			break;
+		default:
+			goto free;
```
buha, line 163: maybe this should be:
```
ret = -EINVAL;
goto free;
```

And free: shoud end with:
```
return ret;
```

For more meaningful error handling.

Also memory allocations above could set ret = -ENOMEM etc.

drivers/platform/linux/linux_uart.c
```diff
@@ -0,0 +1,298 @@
...
+{
+	struct linux_uart_desc *linux_desc;
+	ssize_t count = 0;
+	int32_t ret;
+
+	linux_desc = desc->extra;
+
+	while (count < bytes_number) {
+		ret = write(linux_desc->fd, data, bytes_number);
+		if (ret != -1)
```
adrimbarean, line 268: Would it be safer to check for <= -1?

adrimbarean (approved):
Looks good to me.

drivers/platform/linux/linux_uart.c
dbogdan: I guess it might be safer even if at this point, write() returns -1 on error.

dbogdan:
v3:
 - update write()/read() returned value check;
 - use -ENOMEM, -EINVAL;
 - comment terminal member of struct linux_uart_desc.

dbogdan:
v4:
 - Fix astyle reported issues

dbogdan:
v5:
 - use -ENOENT.

PR #838, opened 2020-11-26
Update axi adc to new buffer functions

mchindri:
It was tested on ad9361

### DIFF
--- a/projects/ad9081/src/app_iio.c
+++ b/projects/ad9081/src/app_iio.c
@@ -56,6 +56,11 @@
 /************************** Functions Implementation **************************/
 /******************************************************************************/
 
+struct iio_data_buffer g_read_buff = {
+	.buff = (void *)ADC_DDR_BASEADDR,
+	.size = 0xFFFFFFFF,
+};
+
 /**
  * @brief Application IIO setup.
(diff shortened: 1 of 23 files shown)


projects/ad9081/src/app_iio.c
```diff
@@ -56,6 +56,11 @@
 /************************** Functions Implementation **************************/
 /******************************************************************************/
 
+struct iio_data_buffer g_read_buff = {
+		.buff = (void *)ADC_DDR_BASEADDR,
+		.size = 0xFFFFFFFF,
```
amiclaus, line 61: the `size` value seems to be propagated throughout all projects that support IIO. is it possible to add a macro for this?

mchindri:
v2:
 - Astyle

cristiantpop (approved):

mchindri:
v3:
 - More astyle

projects/ad9081/src/app_iio.c
mchindri: I think I can change this if needed in a other PR for both ADC and DAC. Will make more sense together.

amiclaus (approved):

PR #855, opened 2020-12-08
projects: ad9739a: add IIO Support

amiclaus:
Allow the user to read/write via UART using IIO Osc.

### DIFF
--- a/projects/ad9739a-fmc-ebz/src/app_iio.c
+++ b/projects/ad9739a-fmc-ebz/src/app_iio.c
@@ -1,6 +1,6 @@
 /***************************************************************************//**
- *   @file   parameters.h
- *   @brief  Platform dependent parameters.
+ *   @file   app_iio.c
+ *   @brief  Application IIO setup.
  *   @author Antoniu Miclaus (antoniu.miclaus@analog.com)
 ********************************************************************************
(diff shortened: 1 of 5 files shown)


amiclaus (commented):
V2:
- fix header comment

adrimbarean (approved):
Looks good to me.

projects/ad9739a-fmc-ebz/src/app_iio.c
```diff
@@ -0,0 +1,124 @@
...
+#include "irq_extra.h"
+
+/******************************************************************************/
+/************************ Variables Definitions *******************************/
+/******************************************************************************/
+static struct irq_ctrl_desc *irq_desc;
+
+struct iio_data_buffer g_read_buff = {
+	.buff = (void *)DAC_DDR_BASEADDR,
+	.size = 0xFFFFFFFF,
```
adrimbarean, line 60: nitpick: should we use a define here to indicate why this size is chosen? Something like ULONG_MAX from limits.h maybe.

amiclaus: This topic was already discussed here:
https://github.com/analogdevicesinc/no-OS/pull/838

I think this should be done in all projects in a separate commit/PR.

PR #865, opened 2020-12-10
IIO attributes update

mchindri:
- priv field was added to be able to set the same function for multiple attributes
- attributes are now set as array of structures not array of pointers. This will avoid to create structures just to pass a reference
- minimal attributes test on ad9361 was done using iio-osciloscope


mchindri:
v2: 
 - Fix doxygen

mchindri:
v3:
 - Rebase onto master
 - Compare iio_info with the version in master: The structure is the same, just the value of fastlock_save was changed which I assume changed by testing the board

cristiantpop (approved):

PR #866, opened 2020-12-10
Add info about submodules in readme

mchindri:
Preview: (without the green lines)

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,6 +18,29 @@ Code documentation is automatically generated using Doxygen tool, available at [
 
 ## Build
 
+Clone this repository:
+```
+git clone https://github.com/analogdevicesinc/no-OS
+```
+This repo is using different submodules, so depending of your needs you should initialize some of them:
+
+- **Tinyiiod**: To use projects which support iio
+```
+  git submodule update --init libraries/iio/libtinyiiod
+```
+- **STM32CubeF4**: To use projects for stm32 platforms
+```
+  git submodule update --init libraries/stm32/STM32CubeF4
+```
+- **Mqtt client**: To use projects that are using the mqtt client to connect to an mqtt borker
+```
+  git submodule update --init libraries/mqtt/paho.mqtt.embedded-c
+```
+- **MbedTLS**: To use the network layer with TLS integrated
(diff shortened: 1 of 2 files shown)


buha:
I'm skeptical about this.
If we really want to automate this we should use git hooks and not make to do it.

mchindri:
v2:
 - Revers PR. Remove initialization of submodules

README.md
```diff
@@ -18,6 +18,9 @@ Code documentation is automatically generated using Doxygen tool, available at [
 
 ## Build
 
+It is recomanded to clone noos with submodules:
```
dbogdan, line 21: There is a typo ("recomanded" -> "recommended"). Or maybe, we can say:
"When cloning the no-OS repository, please pass the --recurse-submodules to the git clone command so the included submodules will be also cloned."

buha: Maybe we should think a bit more about what to recommend. I have a PR that adds /libraries/stm32/STM32CubeF4 as a submodule from ST github. If we have projects that will use stm32f0 or f1, we will have to add those as well which makes for lots of submodules to add at each no-OS clone. Maybe we should clone no-OS by recursively cloning submodules based on depth somehow so that not all the submodules get cloned the first time, only the critical ones.

mchindri:
v3:
 - Update text: See preview in description
 - Add stm32 submodule in description too

README.md
```diff
@@ -18,6 +18,29 @@ Code documentation is automatically generated using Doxygen tool, available at [
 
 ## Build
 
+Clone this repository:
```
amiclaus, line 21: Shouldn't we add this information in the build guides, rather than main README of the no-OS repository?

buha: Yeah we should add a section in the build guides with this info

buha:
We decided to keep in no-OS only dependencies that are widely used in all projects.

Therefore, platform specific dependencies (such as stm32cube chip family packages) will be downloaded separately by the user if needed.

This way, we can have a simple command for cloning the no-OS:

git clone --recursive https://...no-OS.git

Please see #885 for the background of this.

PR #867, opened 2020-12-10
libraries:iio: Make channels as array

mchindri:
This approach is similar to linux and channels can be added using
defines

It was tested on ad9361 and it generates the same output of iio_info as the current master


mchindri:
v2:
 -Rebase onto master and fix conflict

dbogdan (approved):

PR #875, opened 2020-12-16
stm32 (stm32f4) support in makefiles

buha:
This adds the possibility to build a project for the STM32
microcontrollers using:

make PLATFORM=stm32

The family is detected automatically from file names.

The build/ directory is used only to store the .o, .elf
and other build artifacts and, unlike for xilinx and intel
projects, it is not used to store an SDK project. The reason
behind this is that STM32CubeIDE does not support project
creation from command line. Therefore the build is entirely
maintained by us using our build system.

A new rule was added specifically for this platform, which is:

make PLATORM=stm32 debug

This rule has as prerequisite that the shell PATH contains

OPENOCD_BIN
OPENOCD_SCRIPTS

environment variables. If they are properly configured, the
rule creates an .openocd and a .gdb config script, starts
an openocd server and starts a gdb debug session with the .elf.

Additionaly, like for other platforms, one can flash the chip
and run execution by

make PLATFORM=stm32 run

Screenshot from 2020-12-16 22-16-26

### DIFF
--- a/tools/scripts/generic.mk
+++ b/tools/scripts/generic.mk
@@ -35,18 +35,16 @@ INCS     := $(filter-out $(ALL_IGNORED_FILES),$(INCS))
 #------------------------------------------------------------------------------
 
 ifeq (aducm3029,$(strip $(PLATFORM)))
-#Aducm3029 makefile
-include $(NO-OS)/tools/scripts/aducm.mk
-
+	include $(NO-OS)/tools/scripts/aducm.mk
+else ifeq (stm32,$(strip $(PLATFORM)))
+	include $(NO-OS)/tools/scripts/stm32.mk
 else
-#Xilnx and altera makefiles
-ifeq ($(OS), Windows_NT)
-include $(NO-OS)/tools/scripts/windows.mk
-else
-include $(NO-OS)/tools/scripts/linux.mk
-endif #($(OS), Windows_NT)
-
-endif #(aducm3029,$(strip $(PLATFORM)))
+	ifeq ($(OS), Windows_NT)
+		include $(NO-OS)/tools/scripts/windows.mk
+	else
+		include $(NO-OS)/tools/scripts/linux.mk
+	endif
+endif
 
 list:
 ifeq ($(OS), Windows_NT)
@@ -153,11 +151,12 @@ INCLUDE			?= $(NO-OS)/include
 PLATFORM_DRIVERS	?= $(NO-OS)/drivers/platform/$(PLATFORM)
 
 #USED IN MAKEFILE
-PROJECT_NAME		= $(notdir $(PROJECT))
-BUILD_DIR		?= $(PROJECT)/build
+PROJECT_NAME	= $(notdir $(PROJECT))
+BUILD_DIR_NAME	?= build
+BUILD_DIR		?= $(PROJECT)/$(BUILD_DIR_NAME)
 OBJECTS_DIR		= $(BUILD_DIR)/objs
 WORKSPACE		?= $(BUILD_DIR)
-PLATFORM_TOOLS		= $(NO-OS)/tools/scripts/platform/$(PLATFORM)
+PLATFORM_TOOLS	= $(NO-OS)/tools/scripts/platform/$(PLATFORM)
 BINARY			= $(BUILD_DIR)/$(PROJECT_NAME).elf
 
 ifneq ($(words $(NO-OS)), 1)
@@ -193,6 +192,10 @@ ifeq '' '$(HARDWARE)'
 $(error 'No HARDWARE for aducm3029 found. Add pinmux_config.c file')
 endif
 include $(NO-OS)/tools/scripts/aducm.mk
+endif
+
+ifeq 'stm32' '$(PLATFORM)'
+include $(NO-OS)/tools/scripts/stm32.mk
 endif
 
 #------------------------------------------------------------------------------
@@ -250,7 +253,14 @@ REL_SRCS = $(addprefix $(OBJECTS_DIR)/,$(call get_relative_path,$(SRCS)))
 OBJS = $(REL_SRCS:.c=.o)
 
 REL_ASM_SRCS = $(addprefix $(OBJECTS_DIR)/,$(call get_relative_path,$(ASM_SRCS)))
-ASM_OBJS = $(REL_ASM_SRCS:.S=.o)
+ASM_OBJS_s = $(REL_ASM_SRCS:.s=.o)
+ifneq ($(REL_ASM_SRCS),$(ASM_OBJS_s))
+	ASM_OBJS += $(ASM_OBJS_s)
+endif
+ASM_OBJS_S = $(REL_ASM_SRCS:.S=.o)
+ifneq ($(REL_ASM_SRCS),$(ASM_OBJS_S))
+	ASM_OBJS += $(ASM_OBJS_S)
+endif
 
 #Add to include all directories containing a .h file
 EXTRA_INC_PATHS += $(sort $(foreach dir, $(INCS),$(dir $(dir))))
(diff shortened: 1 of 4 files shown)


buha:
V2: .s and .S files are both handled

tools/scripts/generic_variables.mk
```diff
@@ -3,11 +3,14 @@
...
 # It assumes that the makefile is invoked from No-OS/project/some_project
 
-# User must set PLATFORM variable between:
+# For these platforms, the user may optionally pass the PLATFORM variable:
 # PLATFORM = xilinx
 # PLATFORM = altera
 # PLATFORM = aducm3029
 
+# For these platforms, the user must set the PLATFORM variable:
+# PLATFORM = stm32f4
```
buha, line 12: should be stm32, will fix

buha:
V3: styling and comments changes

dbogdan (approved):

tools/scripts/stm32.mk
```diff
@@ -0,0 +1,95 @@
...
+	-Wl,--start-group \
+	-lc \
+	-lm \
+	-Wl,--end-group
+
+CC = arm-none-eabi-gcc
+AS = arm-none-eabi-gcc
+AR = arm-none-eabi-ar
+
+rwildcard=$(foreach d,$(wildcard $(1:=/*)),$(call rwildcard,$d,$2) $(filter $(subst *,%,$2),$d))
```
mchindri, line 31: You can use rwildcard from generic.mk or is this different?

tools/scripts/stm32.mk
```diff
@@ -0,0 +1,95 @@
...
+TARGETCFG = target/stm32f4x.cfg
+else
+$(error Couldn't detect stm32 family.$(ENDL))
+endif
+
+openocd_paths:
+ifndef OPENOCD_SCRIPTS
+	$(error OPENOCD_SCRIPTS not found in shell environment.$(ENDL))
+endif
+
```
mchindri, line 47: If you have this extra line, $(error OPENOCD_BIN not found in shell environment.$(ENDL))
will show when running the makefile. not for the targer openocd_paths

tools/scripts/stm32.mk
```diff
@@ -0,0 +1,95 @@
...
+ifndef OPENOCD_BIN
+	$(error OPENOCD_BIN not found in shell environment.$(ENDL))
+endif
+
+.PHONY: $(BINARY).openocd
```
mchindri, line 52: Use PHONEY += $(BINARY).openocd
At the end of generic.mk  .PHONEY = $(PHONEY)

mchindri: Anyway is there a reason to set this rules as PHONY ?
I think every time you call run will be overwritten with the same value.
If you don't set PHONY  then they will be written once.

tools/scripts/stm32.mk
buha: didn't see it, will try to use it

tools/scripts/stm32.mk
buha: I mean, I don't get why maintain a PHONY list when it doesn't add any value ? You still have to add one line before each rule so you're not even saving typing.

I agree for the other comment, I need to see if they really need to be rebuilt all the time.

tools/scripts/stm32.mk
buha: Not sure if I understand the comment. It only tests it when PLATFORM is stm32 and rule is debug.

tools/scripts/generic.mk
```diff
@@ -250,7 +253,14 @@ REL_SRCS = $(addprefix $(OBJECTS_DIR)/,$(call get_relative_path,$(SRCS)))
...
 REL_ASM_SRCS = $(addprefix $(OBJECTS_DIR)/,$(call get_relative_path,$(ASM_SRCS)))
-ASM_OBJS = $(REL_ASM_SRCS:.S=.o)
+ASM_OBJS_s = $(REL_ASM_SRCS:.s=.o)
+ifneq ($(REL_ASM_SRCS),$(ASM_OBJS_s))
+	ASM_OBJS += $(ASM_OBJS_s)
+endif
+ASM_OBJS_S = $(REL_ASM_SRCS:.S=.o)
+ifneq ($(REL_ASM_SRCS),$(ASM_OBJS_S))
+	ASM_OBJS += $(ASM_OBJS_S)
+endif
```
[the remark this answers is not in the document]
buha: Will test this and include your change if it works, it's shorter ;)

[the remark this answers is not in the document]
buha: Although what do you do when you have a mix of .asm .s and .S suffixes ?

tools/scripts/generic.mk
```diff
@@ -154,7 +152,8 @@ PLATFORM_DRIVERS	?= $(NO-OS)/drivers/platform/$(PLATFORM)
 
 #USED IN MAKEFILE
 PROJECT_NAME		= $(notdir $(PROJECT))
-BUILD_DIR		= $(PROJECT)/build
+BUILD_DIR_NAME		= build
+BUILD_DIR		= $(PROJECT)/$(BUILD_DIR_NAME)
```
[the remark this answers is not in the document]
buha: Yeah sure, but I saw you have it in your PR as well so let's just leave it as-is and whoever needs to fix conflicts will pick the ?= version

tools/scripts/stm32.mk
buha: Did a test regarding usefulness of those rules being PHONY: For example, the user may wanna edit the gdb script to uncomment some of the commented lines in stm32.mk or add new persistent gdb commands at startup. If the rule is not marked as .PHONY, it will not update the file build/project.elf.gdb.

tools/scripts/generic.mk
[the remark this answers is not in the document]
mchindri: It can be done using wildcard somehow but for aducm3029 there is just on .S file and I assume for stm32 there are only .s files so I think there is no need to treat that case for the moment

tools/scripts/stm32.mk
buha: oh ok, the empty line should not be there, got it ! will fix

tools/scripts/stm32.mk
mchindri: I took the PHONY approach from the makefile in linux OS, I don't know why :)
anyway you should use .PHONEY += something otherwise  with := will be overwritten.

tools/scripts/stm32.mk
buha: will remove the empty line !

tools/scripts/generic.mk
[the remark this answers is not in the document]
mchindri:
```suggestion
ASM_SUFFIX = $(suffix $(REL_ASM_SRCS))
ASM_OBJS  = $(REL_ASM_SRCS:.s=.o)
ASM_OBJS  := $(ASM_OBJS:.S=.o)
```

tools/scripts/stm32.mk
buha: so conclusion here is I'm removing .PHONY: dependence on .openocd and .gdb rules

buha:
V4:
- rebased on master due to conflicts with recent changes

tools/scripts/stm32.mk
mchindri: https://github.com/analogdevicesinc/no-OS/blob/7728d2e1d9653112fcbaefa86e736b65292e900b/tools/scripts/generic.mk#L9

mchindri: Didn't it work?

amiclaus (approved):

PR #876, opened 2020-12-17
platform/stm32: gpio and delay 

buha:
With this modification it's possible to use functions from gpio.h and delay.h on stm32 microcontrollers.

### DIFF
--- a/drivers/platform/stm32/stm32_delay.c
+++ b/drivers/platform/stm32/stm32_delay.c
@@ -1,0 +1,61 @@
+/***************************************************************************//**
+ *   @file   stm32/delay.c
+ *   @brief  Implementation of stm32 delay functions.
+ *   @author Darius Berghe (darius.berghe@analog.com)
+********************************************************************************
+ * Copyright 2020(c) Analog Devices, Inc.
+ *
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *  - Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ *  - Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in
+ *    the documentation and/or other materials provided with the
+ *    distribution.
+ *  - Neither the name of Analog Devices, Inc. nor the names of its
+ *    contributors may be used to endorse or promote products derived
+ *    from this software without specific prior written permission.
+ *  - The use of this software may or may not infringe the patent rights
+ *    of one or more patent holders.  This license does not release you
+ *    from the requirement that you obtain separate licenses from these
+ *    patent holders to use this software.
+ *  - Use of the software either in source or binary form, must be run
+ *    on or directly connected to an Analog Devices Inc. component.
+ *
+ * THIS SOFTWARE IS PROVIDED BY ANALOG DEVICES "AS IS" AND ANY EXPRESS OR
+ * IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, NON-INFRINGEMENT,
+ * MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED.
+ * IN NO EVENT SHALL ANALOG DEVICES BE LIABLE FOR ANY DIRECT, INDIRECT,
+ * INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, INTELLECTUAL PROPERTY RIGHTS, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+*******************************************************************************/
+#include "delay.h"
+#include "stm32_hal.h"
+
+/**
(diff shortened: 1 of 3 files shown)


dbogdan:
It looks good. I would only add the stm32_ prefix in the files' names (we have started doing this for other platforms too: https://github.com/analogdevicesinc/no-OS/tree/master/drivers/platform/linux, https://github.com/analogdevicesinc/no-OS/tree/master/drivers/platform/xilinx).

dbogdan (approved):

drivers/platform/stm32/gpio.c
```diff
@@ -0,0 +1,320 @@
...
+/************************ Functions Definitions *******************************/
+/******************************************************************************/
+
+/**
+ * @brief Prepare the GPIO decriptor.
+ * @param desc - The GPIO descriptor.
+ * @param param - The structure that contains the GPIO parameters.
+ * @return SUCCESS in case of success, FAILURE otherwise.
+ */
+int32_t _gpio_init(struct gpio_desc *desc,
```
mchindri, line 69: static int32_t ...

drivers/platform/stm32/gpio.c
```diff
@@ -0,0 +1,320 @@
...
+		       const struct gpio_init_param *param)
+{
+	struct gpio_desc *descriptor;
+	struct stm32_gpio_desc *extra;
+	int32_t ret;
+
+	descriptor = (struct gpio_desc *)malloc(sizeof(*descriptor));
+	if (!descriptor) {
+		ret = -ENOMEM;
+		goto error;
```
mchindri, line 147: no need to go to error. Will free a NULL pointer.
Same for extra

drivers/platform/stm32/gpio.c
```diff
@@ -0,0 +1,320 @@
...
+	return ret;
+}
+
+/**
+ * @brief Obtain the GPIO decriptor.
+ * @param desc - The GPIO descriptor.
+ * @param param - GPIO initialization parameters
+ * @return SUCCESS in case of success, FAILURE otherwise.
+ */
+int32_t stm32_gpio_get(struct gpio_desc **desc,
```
mchindri, line 137: neet to check if desc or param is NULL

drivers/platform/stm32/gpio.c
```diff
@@ -0,0 +1,320 @@
...
+	int32_t ret = 0;
+	struct stm32_gpio_desc *extra = desc->extra;
+	struct stm32_gpio_desc *pextra = param->extra;
+	GPIO_InitTypeDef gis;
+
+	if (!param)
+		return -EINVAL;
+
+	/* enable gpio port in RCC */
+	if (pextra->port == GPIOA)
```
mchindri, line 81: nitpick: why not using a switch?

drivers/platform/stm32/gpio.c
```diff
@@ -0,0 +1,320 @@
...
+ */
+int32_t stm32_gpio_get_optional(struct gpio_desc **desc,
+				const struct gpio_init_param *param)
+{
+	if(param == NULL) {
+		*desc = NULL;
+		return SUCCESS;
+	}
+
+	return stm32_gpio_get(desc, param);
```
mchindri, line 184: Why not implementing this in gpio.c (the generic one)

drivers/platform/stm32/gpio.c
```diff
@@ -0,0 +1,320 @@
...
+
+/**
+ * @brief Free the resources allocated by gpio_get().
+ * @param desc - The GPIO descriptor.
+ * @return SUCCESS in case of success, FAILURE otherwise.
+ */
+int32_t stm32_gpio_remove(struct gpio_desc *desc)
+{
+	if (desc != NULL) {
+		free(desc->extra);
```
mchindri, line 195: it will be useful to check extra too

drivers/platform/stm32/gpio.c
```diff
@@ -0,0 +1,320 @@
...
+
+	return 0;
+}
+
+/**
+ * @brief Enable the input direction of the specified GPIO.
+ * @param desc - The GPIO descriptor.
+ * @return SUCCESS in case of success, FAILURE otherwise.
+ */
+int32_t stm32_gpio_direction_input(struct gpio_desc *desc)
```
mchindri, line 207: I think it will be good to check desc and extra on all function. Maybe for performance we can surround them in #ifdef DEBUG or something ?

drivers/platform/stm32/gpio.c
buha: switch doesnt work for pointers unless you cast, it looks ugly with casts all over
GPIOA is a pointer

drivers/platform/stm32/gpio.c
buha: if pointer is null, free does nothing, no need to check

drivers/platform/stm32/gpio.c
buha: true

drivers/platform/stm32/gpio.c
buha: freeing null pointer is ok

drivers/platform/stm32/gpio.c
buha: true

drivers/platform/stm32/gpio.c
buha: You mean why do we implement platform specific gpio_get_optional functions ? good question, historical reasons i guess

drivers/platform/stm32/gpio.c
buha: yeah good point, will check all of them

buha: an no, we shouldn't do #ifdef DEBUG because some of these functions are exposed to the user directly and the user may pass NULL pointers and the system would crash into a hard fault handler, we wanna avoid that so i will test all of them in this file

drivers/platform/stm32/gpio.c
buha: https://stackoverflow.com/questions/6084218/is-it-good-practice-to-free-a-null-pointer-in-c

buha:
V2:
- renamed gpio.c to stm32_gpio.c in platform/stm32/
- implemented review feedback so far

buha:
V3:
-renamed all files to have stm32_ prefix

mchindri (approved):

amiclaus (approved):
LGTM.

PR #879, opened 2020-12-17
projects: ad9739a: add DMA support

amiclaus:
Add DAC DMA example for the ad9739a project.

### DIFF
--- a/projects/ad9739a-fmc-ebz/src/ad9739a_fmc_ebz.c
+++ b/projects/ad9739a-fmc-ebz/src/ad9739a_fmc_ebz.c
@@ -137,11 +137,6 @@ int main(void)
 
 	/* AXI DAC Core */
 	struct axi_dac_channel ad9739a_channels[1];
-	ad9739a_channels[0].dds_dual_tone = 0;
-	ad9739a_channels[0].dds_frequency_0 = 33*1000*1000;
-	ad9739a_channels[0].dds_phase_0 = 0;
-	ad9739a_channels[0].dds_scale_0 = 250000;
-	ad9739a_channels[0].sel = AXI_DAC_DATA_SEL_DDS;
 
 	struct axi_dac_init ad9739a_core_param = {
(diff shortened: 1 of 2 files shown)


buha (approved):

projects/ad9739a-fmc-ebz/src/ad9739a_fmc_ebz.c
```diff
@@ -169,6 +164,33 @@ int main(void)
...
+	extern const uint16_t sine_lut[128];
+	struct axi_dmac_init ad9739a_dmac_init_param = {
+		.name = "ad9739a_dmac",
+		.base = TX_DMA_BASEADDR,
+		.direction = DMA_MEM_TO_DEV,
+		.flags = DMA_CYCLIC
+	};
+	struct axi_dmac *ad9739a_dmac_desc;
+
+	ad9739a_channels[0].sel = AXI_DAC_DATA_SEL_DMA;
```
mchindri, line 177: I think we should use functions like _axi_dac_set_datasel_ to modify this parameters no directly the structures.

mchindri (changes requested):

projects/ad9739a-fmc-ebz/src/ad9739a_fmc_ebz.c
amiclaus: I wouldn't let the channel attributes completely uninitialized.

Also, if you call directly axi_dac_set_datasel, and you want to change all channels, you will have to add raw negative value as parameter. 
`axi_dac_set_datasel(ad9739a_core, -1, AXI_DAC_DATA_SEL_DMA)`
For one channel:
`axi_dac_set_datasel(ad9739a_core, 0, AXI_DAC_DATA_SEL_DMA)`

I'm not a fan of this approach, since these values are not mapped accordingly yet.

For now `axi_dac_data_setup` wraps properly these kind of calls.

mchindri (approved):

PR #887, opened 2020-12-24
platform/stm32: spi

buha:
Implements the link between our spi layer to the stm32 HAL_SPI layer.

The stm32 project in the projects/ folder must have a stm32_hal.h file
that includes specific headers for the chip family in-use.

### DIFF
--- a/drivers/platform/stm32/stm32_spi.c
+++ b/drivers/platform/stm32/stm32_spi.c
@@ -1,0 +1,234 @@
+/***************************************************************************//**
+ *   @file   stm32/stm32_spi.c
+ *   @brief  Implementation of stm32 spi driver.
+ *   @author Darius Berghe (darius.berghe@analog.com)
+********************************************************************************
+ * Copyright 2020(c) Analog Devices, Inc.
+ *
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *  - Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ *  - Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in
+ *    the documentation and/or other materials provided with the
+ *    distribution.
+ *  - Neither the name of Analog Devices, Inc. nor the names of its
+ *    contributors may be used to endorse or promote products derived
+ *    from this software without specific prior written permission.
+ *  - The use of this software may or may not infringe the patent rights
+ *    of one or more patent holders.  This license does not release you
+ *    from the requirement that you obtain separate licenses from these
+ *    patent holders to use this software.
+ *  - Use of the software either in source or binary form, must be run
+ *    on or directly connected to an Analog Devices Inc. component.
+ *
+ * THIS SOFTWARE IS PROVIDED BY ANALOG DEVICES "AS IS" AND ANY EXPRESS OR
+ * IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, NON-INFRINGEMENT,
+ * MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED.
+ * IN NO EVENT SHALL ANALOG DEVICES BE LIABLE FOR ANY DIRECT, INDIRECT,
+ * INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, INTELLECTUAL PROPERTY RIGHTS, PROCUREMENT OF SUBSTITUTE GOODS OR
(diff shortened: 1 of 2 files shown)


drivers/platform/stm32/stm32_spi.c
```diff
@@ -0,0 +1,232 @@
...
+	uint32_t input_clock;
+	const uint32_t prescaler_default = SPI_BAUDRATEPRESCALER_64;
+	const uint32_t prescaler_min = SPI_BAUDRATEPRESCALER_2;
+	const uint32_t prescaler_max = SPI_BAUDRATEPRESCALER_256;
+	uint32_t prescaler_reg = 0u;
+
+	if (!desc || !param)
+		return -EINVAL;
+
+	*desc = calloc(1,sizeof(**desc));
```
amiclaus, line 74: nitpick: `calloc(1, sizeof(**desc));` (add space between function parameters)

drivers/platform/stm32/stm32_spi.c
```diff
@@ -0,0 +1,232 @@
...
+ * @brief Initialize the SPI communication peripheral.
+ * @param desc - The SPI descriptor.
+ * @param param - The structure that contains the SPI parameters.
+ * @return SUCCESS in case of success, FAILURE otherwise.
+ */
+int32_t stm32_spi_init(struct spi_desc **desc,
+		       const struct spi_init_param *param)
+{
+	int32_t ret;
+	uint32_t input_clock;
```
amiclaus, line 65: for consistency, I would create a local spi descriptor that is passed at the end of the function to the input parameter `desc`.

example:
https://github.com/analogdevicesinc/no-OS/blob/e1347d0c670f0f4f26e2e942968f8f587f6f1373/drivers/platform/aducm3029/spi.c#L127

https://github.com/analogdevicesinc/no-OS/blob/e1347d0c670f0f4f26e2e942968f8f587f6f1373/drivers/platform/aducm3029/spi.c#L178

drivers/platform/stm32/stm32_spi.c
```diff
@@ -0,0 +1,232 @@
...
+	if (!desc || !desc->extra)
+		return -EINVAL;
+
+	sdesc = desc->extra;
+	HAL_SPI_DeInit(&sdesc->hspi);
+	gpio_remove(sdesc->chip_select);
+	free(desc->extra);
+	free(desc);
+	return 0;
+}
```
amiclaus, line 193: maybe `return SUCCESS`

drivers/platform/stm32/stm32_spi.c
```diff
@@ -0,0 +1,232 @@
...
+				 uint16_t bytes_number)
+{
+	int32_t	ret;
+	struct stm32_spi_desc *sdesc;
+
+	if (!desc || !desc->extra || !data)
+		return -EINVAL;
+
+	if (!bytes_number)
+		return 0;
```
amiclaus, line 213: maybe `return SUCCESS`

drivers/platform/stm32/stm32_spi.c
```diff
@@ -0,0 +1,232 @@
...
+
+	sdesc = desc->extra;
+
+	ret = gpio_set_value(sdesc->chip_select, GPIO_LOW);
+	if (ret < 0)
```
amiclaus, line 218: maybe `ret != SUCCESS`

drivers/platform/stm32/stm32_spi.c
buha: yeah, true !

drivers/platform/stm32/stm32_spi.c
buha: yeah, that could do
i don't see any advantage other than consistency but i will change it

drivers/platform/stm32/stm32_spi.c
buha: Unfortunately there is a name clash on stm32 platform with SUCCESS being already defined in STM32 HAL.
The only elegant way of overcoming this was use <errno.h> and bypass our "error.h" and use 0 instead of SUCCESS.

buha:
V2: 
- implemented remarks above

drivers/platform/stm32/stm32_spi.c
```diff
@@ -0,0 +1,234 @@
...
+{
+	int32_t ret;
+	uint32_t input_clock;
+	const uint32_t prescaler_default = SPI_BAUDRATEPRESCALER_64;
+	const uint32_t prescaler_min = SPI_BAUDRATEPRESCALER_2;
+	const uint32_t prescaler_max = SPI_BAUDRATEPRESCALER_256;
+	uint32_t prescaler_reg = 0u;
+	struct spi_desc	*spi_desc;
+
+	if (!spi_desc || !param)
```
amiclaus, line 72: Why are you checking `if (!spi_desc)` before the memory allocation?

Shouldn't it be only `if (!param)`?

buha: Good catch, it should be desc.... because desc is dereferenced lower with *desc = spi_desc.
This was a bad refactoring :D

buha:
V3:
- fixed bad refactoring

amiclaus (approved):

PR #894, opened 2021-01-06
stm32 uart

buha:
Add the UART layer and stdio redirection for stm32 platform.

Example usage:
```
	struct uart_desc *uart;

        // stm32 specific stuff
	struct stm32_uart_init_param uip_extra = {
		.Mode = UART_MODE_TX_RX,
		.HwFlowCtl = UART_HWCONTROL_NONE,
		.OverSampling = UART_OVERSAMPLING_16,
	};

        // generic stuff
	struct uart_init_param uip = {
		.device_id = 2,
		.baud_rate = 115200,
		.size = UART_CS_8,
		.parity = UART_PAR_NO,
		.stop = UART_STOP_1,
		.extra = &uip_extra,
	};

	ret = uart_init(&uart, &uip);
	if (ret < 0)
		return ret;

        // bind stdout and stdin to the uart instance and have printf/scanf use uart as their operating medium
	stm32_uart_stdio(uart); 
```

### DIFF
--- a/drivers/platform/stm32/stm32_uart.c
+++ b/drivers/platform/stm32/stm32_uart.c
@@ -1,0 +1,244 @@
+/***************************************************************************//**
+ *   @file   stm32/stm32_uart.c
+ *   @brief  Source file of UART driver for STM32
+ *   @author Darius Berghe (darius.berghe@analog.com)
+********************************************************************************
+ * Copyright 2020(c) Analog Devices, Inc.
+ *
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *  - Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ *  - Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in
+ *    the documentation and/or other materials provided with the
+ *    distribution.
+ *  - Neither the name of Analog Devices, Inc. nor the names of its
+ *    contributors may be used to endorse or promote products derived
+ *    from this software without specific prior written permission.
+ *  - The use of this software may or may not infringe the patent rights
+ *    of one or more patent holders.  This license does not release you
+ *    from the requirement that you obtain separate licenses from these
+ *    patent holders to use this software.
+ *  - Use of the software either in source or binary form, must be run
+ *    on or directly connected to an Analog Devices Inc. component.
+ *
+ * THIS SOFTWARE IS PROVIDED BY ANALOG DEVICES "AS IS" AND ANY EXPRESS OR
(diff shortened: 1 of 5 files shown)


drivers/platform/stm32/stm32_uart.c
```diff
@@ -0,0 +1,244 @@
...
+{
+	struct stm32_uart_desc *sud;
+
+	if (!desc)
+		return -EINVAL;
+
+	sud = desc->extra;
+	HAL_UART_DeInit(&sud->huart);
+	free(desc->extra);
+	free(desc);
```
mchindri, line 184: Nitpic (styling):Add empty line before return.

drivers/platform/stm32/stm32_uart.c
```diff
@@ -0,0 +1,244 @@
...
+	int32_t ret;
+
+	if (!desc || !desc->extra || !data)
+		return -EINVAL;
+
+	if (!bytes_number)
+		return 0;
+
+	sud = desc->extra;
+
```
mchindri, line 208: nitpick: extra line

drivers/platform/stm32/stm32_uart.h
```diff
@@ -0,0 +1,69 @@
...
+#include "uart.h"
+#include "stm32_hal.h"
+
+/**
+ * @struct stm32_uart_init_param
+ * @brief Specific initialization parameters for stm32 UART.
+ */
+struct stm32_uart_init_param {
+	/** Specifies the Receive or Transmit mode. */
+	uint32_t Mode;
```
mchindri, line 53: use lower case and _
https://www.kernel.org/doc/html/v4.10/process/coding-style.html#naming

buha: I am going to pass on this one because they are named like that in STM32 HAL and I want consistency with that, otherwise it gets confusing.

mchindri:
Can platform/stm32: uart stdio redirection be implemented using generic uart? will be useful for aducm3029 too.

buha:
Good question, but the redirection is pretty platform specific.
For example, aducm would certainly need to implement a function similar to stm32_uart_stdio().
Also _write and _read make calls to stm32 HAL here and perhaps the aducm HAL is different if it is not CMSIS HAL.
So... I think it's not worth to have a generic one, too many quirks to overcome.

drivers/platform/stm32/stm32_uart.c
buha: done

drivers/platform/stm32/stm32_uart.c
buha: done

drivers/platform/stm32/stm32_uart.h
buha: eventually did the change

buha:
V2:
- implemented remarks

dbogdan (approved):

mchindri (approved):

PR #880, opened 2020-12-18
Create aducm3029_adc and iio project

mchindri:
A minimal driver for the built-in adc in the aducm uc was developed and in order to don't die locally a sample project was created too.  Maybe the project can be added to the adicup repository but that is just the last commit.
In order to capture some data on the adc a pwm driver was also created.
The project support also iio and a simple gpio controller was added to the iio device to be able to control gpios from iio clients.
The firsts 3 commits are small fixes that can be moved to a separate commit if needed.
Some screen shots:
osc:
image

uart:
image
image

PR #900, opened 2021-01-07
iio_demo stm32 support

buha:
This PR makes it possible to run projects/iio_demo on stm32f446


projects/iio_demo/src/app/main.c
```diff
@@ -45,7 +45,7 @@
 #include "iio_demo_dev.h"
 #include "app_config.h"
 #include "parameters.h"
-#include "error.h"
+#include <errno.h>
```
mchindri, line 48: errno is included in error.h

projects/iio_demo/src/app/main.c
```diff
@@ -71,9 +77,58 @@ static uint8_t out_buff[MAX_SIZE_BASE_ADDR];
...
+#ifdef STM32_PLATFORM
+void Error_Handler(void)
+{
+	/* User can add his own implementation to report the HAL error return state */
+	__disable_irq();
+	while (1) {
+	}
+}
+
+void SystemClock_Config(void)
```
mchindri, line 89: Will this be the same for all project using stm32?
Can this be added to drivers/platform/stm32/platform_init.c like https://github.com/analogdevicesinc/no-OS/pull/880/commits/31145ea6d4d531a57df53f1270d1f78935838458 ?

projects/iio_demo/src/app/main.c
buha: there are troubles with SUCCESS being defined multiple times and wanted to avoid them

projects/iio_demo/src/app/main.c
buha: kind of, yes
HAL_Init and SystemClock_Config will have to be called in each project but they are both autogenerated (by STM32CubeIDE) and custom for each project, the user needs to copy them anyway to their project and I think it's useless to create another function (say platform_init) that calls these two functions. Also, some users may initialize more stuff than these two functions do, so better keep things simple I guess

projects/iio_demo/src.mk
```diff
@@ -1,9 +1,23 @@
-#See No-OS/tool/scripts/src_model.mk for variable description
-
+# See No-OS/tool/scripts/src_model.mk for variable description
 SRC_DIRS += $(PROJECT)/src/app
 SRC_DIRS += $(NO-OS)/iio/iio_demo
 SRC_DIRS += $(NO-OS)/iio/iio_app
 
+LIBRARIES += iio
```
adrimbarean, line 6: I think this line is now redundant since the makefiles now only use the TINYIIOD define to determine if compiling the iio library is necessary. @mchindri can you confirm this?

projects/iio_demo/src/hal/main.h
```diff
@@ -0,0 +1,83 @@
+/* USER CODE BEGIN Header */
```
adrimbarean, line 1: nitpick: should main.h be in the "hardware abstraction layer" folder? I'm thinking it should go into the app folder with it's .c counterpart.

projects/iio_demo/src/hal/stm32f4xx_hal_conf.h
```diff
@@ -0,0 +1,486 @@
+/**
```
adrimbarean, line 1: Is this a generated source, or source from somewhere else? I am thinking that maybe there is a way to avoid having it whole here if we don't need all of it.

projects/iio_demo/src/hal/main.h
buha: Yes, it looks weird, but unfortunately that's what STM32CubeIDE creates, and I think it's easier for the user to simply move the files to the hal/ directory without renaming.

projects/iio_demo/src/hal/stm32f4xx_hal_conf.h
buha: Everything in hal/ is generated, including this file, and this file is very important in activating/deactivating stuff in the STM32 HAL. The only change I applied on the hal/* files is astyle.

projects/iio_demo/src.mk
mchindri: yes, LIBRARIES += iio is not needed anymore. TINYIIOD=y is enough

projects/iio_demo/src/app/main.c
mchindri, line 48: You need to replace FAILURE with -1 in platform_init

projects/iio_demo/src.mk
buha: removed it in latest version.

buha:
V2:
- removed LIBRARIES += iio
- removed some unnecessary HAL files from build

dbogdan:
@buha, doxygen generation fails.

projects/iio_demo/src/app/main.c
buha: fixed in latest version

buha:
V3:
- fix aducm build
- fix doxy build

dbogdan (approved):

PR #901, opened 2021-01-08
tools:script:xilinx.mk: Fix heap replacement

mchindri:
It seems that only the first branch of the if is executed when
running make. This seems to be beacause the if is evaluated before
ARCH is available.
Anyway, both replacements can be executed because only one will take
have effect, the one that matches the HEAP value.


amiclaus (commented):
Isn't there any other workaround to avoid calling a function that **hopefully** does nothing each time?

mchindri:
> Isn't there any other workaround to avoid calling a function that **hopefully** does nothing each time?

I don't have other idea to solve this. And yes, one instruction will do nothing.

adrimbarean:
Should the first replace_heap call just be removed completely?

mchindri:
> Should the first replace_heap call just be removed completely?

No, ps7 will have the heap of 0x800 and it needs to match this value in order to replace it.

dbogdan (approved):

PR #914, opened 2021-01-14
projects:iio_ad7124: example project

adrimbarean:
Add example project with iio_ad7124.

### DIFF
--- a/projects/ad7124/Makefile
+++ b/projects/ad7124/Makefile
@@ -1,0 +1,10 @@
+# Uncomment to use the desired platform
+# PLATFORM = xilinx
+# PLATFORM = altera
+PLATFORM = aducm3029
+
+include ../../tools/scripts/generic_variables.mk
+
+include src.mk
+
+include ../../tools/scripts/generic.mk
--- a/projects/ad7124/pinmux_config.c
+++ b/projects/ad7124/pinmux_config.c
@@ -1,0 +1,48 @@
+/*
+ **
+ ** Source file generated on November 18, 2019 at 09:16:12.
+ **
+ ** Copyright (C) 2011-2019 Analog Devices Inc., All Rights Reserved.
+ **
+ ** This file is generated automatically based upon the options selected in
+ ** the Pin Multiplexing configuration editor. Changes to the Pin Multiplexing
+ ** configuration should be made by changing the appropriate options rather
+ ** than editing this file.
+ **
+ ** Selected Peripherals
+ ** --------------------
+ ** SPI1 (SCLK, MISO, MOSI, CS_0)
+ ** UART0 (Tx, Rx)
+ **
+ ** GPIO (unavailable)
+ ** ------------------
+ ** P0_10, P0_11, P1_06, P1_07, P1_08, P1_09
+ */
+
+#include <sys/platform.h>
+#include <stdint.h>
+
+#define SPI1_SCLK_PORTP1_MUX  ((uint16_t) ((uint16_t) 1<<12))
+#define SPI1_MISO_PORTP1_MUX  ((uint16_t) ((uint16_t) 1<<14))
+#define SPI1_MOSI_PORTP1_MUX  ((uint32_t) ((uint32_t) 1<<16))
+#define SPI1_CS_0_PORTP1_MUX  ((uint32_t) ((uint32_t) 1<<18))
+#define UART0_TX_PORTP0_MUX  ((uint32_t) ((uint32_t) 1<<20))
+#define UART0_RX_PORTP0_MUX  ((uint32_t) ((uint32_t) 1<<22))
+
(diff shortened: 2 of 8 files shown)


adrimbarean:
V1:
 - fix astyle and doxygen

projects/iio_ad7124/src/main.c
```diff
@@ -0,0 +1,134 @@
+/***************************************************************************//**
+ *   @file   iio_demo/src/main.c
+ *   @brief  Implementation of Main Function.
+ *   @author Cristian Pop (cristian.pop@analog.com)
```
cristiantpop, line 4: Hey It is me!!! :)

cristiantpop: Check all headers

projects/iio_ad7124/src/ad7124_regs.c
```diff
@@ -0,0 +1,110 @@
...
+* LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+* ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+* (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+* SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+*******************************************************************************/
+
+
+#include "ad7124_regs.h"
+
+struct ad7124_st_reg ad7124_regs[AD7124_REG_NO] = {
```
buha, line 44: my only comment would be, aren't these registers supposed to be in the driver ?

projects/iio_ad7124/src/ad7124_regs.c
```diff
@@ -0,0 +1,110 @@
+/***************************************************************************//**
+*   @file    ad7124_regs.c
+*   @brief   AD7124 register file.
+*   @devices AD7124-4, AD7124-8
```
amiclaus, line 4: This is not a standard doxygen command.

projects/iio_ad7124/src/ad7124_regs.h
```diff
@@ -0,0 +1,51 @@
+/***************************************************************************//**
+*   @file    ad7124_regs.h
+*   @brief   AD7124 register file.
+*   @devices AD7124-4, AD7124-8
```
amiclaus, line 4: This is not a standard doxygen command.

projects/iio_ad7124/src/ad7124_regs.c
adrimbarean: This is an initialization register array for the part. The values of registers here will be applied to the device in the initialization function. This will be different for every application.

adrimbarean:
V2:
 - mentioned right names in headers;
 - fixed wrong doxygen commend.

projects/iio_ad7124/src/ad7124_regs.c
mchindri: Then maybe is better to rename the variable to something like: ad7124_init_regs_default.

mchindri: maybe is better to move them to the .h file. This way one file will be needed.

projects/iio_ad7124/src/main.c
```diff
@@ -0,0 +1,134 @@
...
+#include "irq.h"
+#include "irq_extra.h"
+#include "uart.h"
+#include "uart_extra.h"
+#include "iio_ad7124.h"
+#include "ad7124_regs.h"
+#include "spi_extra.h"
+#include "iio_app.h"
+
+#ifdef USE_TCP_SOCKET
```
mchindri, line 57: I think this is needed only in iio_app.h

adrimbarean:
V3:
 - remove unneeded code in main.c;
 - rename default register array to ''ad7124_init_regs_default".

projects/iio_ad7124/src/main.c
adrimbarean: I did not move the array in the header file because the original driver is in this format and want to keep it to better serve as example. Should we change the driver itself? https://github.com/analogdevicesinc/no-OS/tree/master/drivers/adc/ad7124

buha (approved):

mchindri (approved):

mchindri:
image

Maybe you can rename the project to ad7124 some other projects have iio_support but they don't have the iio_ suffix

dbogdan:
> Maybe you can rename the project to ad7124 some other projects have iio_support but they don't have the iio_ suffix

I agree.

adrimbarean:
V4:
 - rename project to ad7124.

dbogdan (approved):

PR #919, opened 2021-01-20
Ad5933 feature update

pmallick-adi:
Added settling time function

Changed return-type for ad5933_calculate_impedance to a structure

Function now returns magnitude and phase with the intent to calculate the impedance in the application code

Added M_PI definition for calculation of radToDeg180ovPi


pmallick-adi:
Will do the suggested changes

PR #923, opened 2021-01-26
Misc makefile fixes

mchindri:


amiclaus (approved):
LGTM.
one question: were the makefile changes tested at least on one no-OS project (besides iio_demo)?

projects/iio_demo/src.mk
```diff
@@ -27,12 +27,19 @@ SRC_DIRS += $(NO-OS)/network
 SRCS	 += $(NO-OS)/util/circular_buffer.c
 SRCS	 += $(PLATFORM_DRIVERS)/delay.c
 SRCS	 += $(PLATFORM_DRIVERS)/timer.c
+INCS	 += $(INCLUDE)/delay.h			\
```
buha, line 30: why were these files missing ? was this a hidden problem by the fact that the source files were not linked but all copied ?
if this is the case, then other projects may hide such problems too

adrimbarean (approved):
Looks good to me.

buha:
Also, I think we need an easy way to generate a standalone project (with copied files). I think the current way would be 
`make LINK_SRCS=n` which is not easy to remember for users.

A rule like this one could make things easier:
`make standalone`

mchindri:
> Also, I think we need an easy way to generate a standalone project (with copied files). I think the current way would be
> `make LINK_SRCS=n` which is not easy to remember for users.
> 
> A rule like this one could make things easier:
> `make standalone`

I'am working on that, but I have some issues with it. That is how I found the problem in iio_demo.
I will come with a PR just for that.

projects/iio_demo/src.mk
mchindri: It was working  because coping src isn't well implemented. I will solve that in a pr for make standalone.

mchindri:
> LGTM.
> one question: were the makefile changes tested at least on one no-OS project (besides iio_demo)?

yes

dbogdan (approved):

dbogdan:
If another PR will implement "make standalone", we should be good to go with this one.

PR #925, opened 2021-01-26
Make standalone

mchindri:
I spitted the changes as much as I could.
I tested them only together and that's way there are all in one PR. 

This is how the build directory will look like:
It will be easy to find sources for both standalone and linked files.
image

### DIFF
--- a/projects/iio_demo/src.mk
+++ b/projects/iio_demo/src.mk
@@ -63,6 +63,8 @@ SRCS += $(PLATFORM_DRIVERS)/stm32_delay.c \
 ASM_SRCS = $(PROJECT)/src/startup/startup_stm32f446retx.s
 
 INCS += $(PROJECT)/src/hal/stm32/main.h \
+	$(INCLUDE)/delay.h \
+	$(PLATFORM_DRIVERS)/stm32_uart_stdio.h \
 	$(PROJECT)/src/hal/stm32/stm32f4xx_hal_conf.h \
 	$(PROJECT)/src/hal/stm32/stm32f4xx_it.h \
 	$(PROJECT)/src/hal/stm32/stm32_hal.h \
--- a/tools/scripts/aducm.mk
+++ b/tools/scripts/aducm.mk
@@ -74,7 +74,7 @@ endif
 #                             DFP DEPENDENCIES                          
 #------------------------------------------------------------------------------
 
-PROJECT_BUILD = $(BUILD_DIR)/$(PROJECT_NAME)
+CCES_PROJ_NAME = $(notdir $(PROJECT_BUILD))
 
 #DFP Files
 DFP_FILES = $(call rwildcard,$(DFP_DRIVERS),*.c)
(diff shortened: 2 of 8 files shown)


dbogdan (approved):

tools/scripts/generic.mk
```diff
@@ -207,7 +207,7 @@ endif
 
 #Add to include all directories containing a .h file
 EXTRA_INC_PATHS += $(sort $(foreach dir, $(INCS),$(dir $(dir))))
-CFLAGS += $(addprefix -I,$(EXTRA_INC_PATHS))
+CFLAGS += $(addprefix -I,$(EXTRA_INC_PATHS) $(PLATFORM_INCS))
```
buha, line 223: I see you defined PLATFORM_INCS in aducm.mk but how about xilinx/altera/stm32 ?

tools/scripts/aducm.mk
```diff
@@ -81,20 +81,18 @@ DFP_FILES = $(call rwildcard,$(DFP_DRIVERS),*.c)
 #Not for aducm3029
 DFP_IGNORED_FILES += $(DFP_DRIVERS)/flash/adi_flash_data.c $(DFP_DRIVERS)/rtc/adi_rtc_data.c
 
-ADUCM_SRCS += $(filter-out $(DFP_IGNORED_FILES), $(DFP_FILES))
+PLATFORM_SRCS += $(filter-out $(DFP_IGNORED_FILES), $(DFP_FILES))
```
buha, line 84: the changes in this commit should be sqashed with the previous commit, they both modify aducm.mk regarding PLATFORM_SRCS

buha (changes requested):
this whole PR breaks stm32 builds

mchindri:
v2:
 - Fix stm32 make by running make update_srcs on stm32 too. This way standalone stm32 projects can be created and also will reveal hidden missing files from src.mk like delay.h in the next commit.
 - Add delay.h to src.mk needed by stm32

projects/iio_demo/src.mk
```diff
@@ -62,6 +62,7 @@ SRCS += $(PLATFORM_DRIVERS)/stm32_delay.c \
 
 ASM_SRCS = $(PROJECT)/src/startup/startup_stm32f446retx.s
 
+INCS += $(INCLUDE)/delay.h
```
buha, line 65: Looks like there were indeed some hidden problems like this missing delay.h, I would completely move it out of the #ifdef area in src.mk and have it included on all platforms, but it's as you wish.

However, when I tried to build the iio_demo with latest change, I still got an error about `stm32_uart_stdio.h` not being found. So could you please add to INCS `$(PLATFORM_DRIVERS)/stm32_uart_stdio.h` in this commit ? This is also a hidden problem that I introduced and it makes sense to fix it in this commit.

tools/scripts/generic.mk
mchindri: They are added directly to CFLAGS and it is ok. In aducm is separate to don't added these directories to the projects because they are already there.

tools/scripts/aducm.mk
mchindri: The previous one is about INCS and reordering them. There was a bug with the order.

mchindri:
v3:
 - Add also $(PLATFORM_DRIVERS)/stm32_uart_stdio.h to fix stm32 build
 - Update description

projects/iio_demo/src.mk
mchindri: Done

mchindri:
v4:
 - Fix also src.mk in adxrs290 project

buha (approved):
Great, now the PR looks correct.
However, we must merge with care because there might be hidden issues in various projects' src.mk

tools/scripts/stm32.mk
```diff
@@ -99,4 +100,5 @@ debug: all openocd_paths $(BINARY).openocd $(BINARY).gdb
 		-c "init" &);
 	arm-none-eabi-gdb --command=$(BINARY).gdb
 
-stm32_project: ;
\ No newline at end of file
+stm32_project:
+	$(MUTE) $(MAKE) --no-print-directory update_srcs
```
amiclaus, line 104: nitpick: no new line

amiclaus (approved):
One minor comment.
The rest lgtm as long as the changes were properly tested.

PR #930, opened 2021-02-03
TDM generic + stm32 TDM using SAI

buha:
This PR implements a generic TDM interface that may be used to send/receive data using serial audio peripherals of several microcontrollers. TDM slave implementation on STM32 is also given in this PR.

### DIFF
--- a/drivers/platform/stm32/stm32_tdm.c
+++ b/drivers/platform/stm32/stm32_tdm.c
@@ -1,0 +1,221 @@
+/***************************************************************************//**
+ *   @file   stm32/stm32_tdm.c
+ *   @brief  Implementation of TDM driver using STM32's SAI peripheral.
+ *   @author Darius Berghe (darius.berghe@analog.com)
+********************************************************************************
+ * Copyright 2020(c) Analog Devices, Inc.
+ *
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *  - Redistributions of source code must retain the above copyright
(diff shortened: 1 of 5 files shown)


mchindri:
Why TDM interface was added with ops?
Spi and i2c was created this why in order to support controllers or different implementation in the same platform.
Will this be needed in this case?

buha:
Good question. I just went with the flow. I could see different implementations on stm32 platform - for example SPI has an I2S mode and SAI also can be configured in I2S mode. But to be honest, they will probably never get mixed so ops would not be needed.

cristiantpop (approved):

drivers/platform/stm32/stm32_tdm.c
```diff
@@ -0,0 +1,219 @@
...
+	tdesc->hsai.SlotInit.SlotNumber = param->slots_per_frame;
+	tdesc->hsai.SlotInit.SlotActive = (1u << param->slots_per_frame) - 1u;
+	if (HAL_SAI_Init(&tdesc->hsai) != HAL_OK) {
+		ret = -EIO;
+		goto error;
+	}
+
+	*desc = tdm_desc;
+
+	return SUCCESS;
```
amiclaus, line 162: Is `SUCCESS` defined in `errno.h` ? 
Otherwise I would include `error.h` in this file.

Then you can replace all `return 0;` with `return SUCCESS` (See line 185 and 206)

amiclaus (approved):
Just a question on my side.

Rest looks good.

buha:
V2:
- included "error.h" and now using return SUCCESS instead of return 0

dbogdan (approved):

PR #931, opened 2021-02-04
CN0531 IIO development

adrimbarean:
Add first draft of IIO driver for ad5791 and example IIO project for cn0531, the PMOD using ad5791.

### DIFF
--- a/drivers/dac/ad5791/ad5791.c
+++ b/drivers/dac/ad5791/ad5791.c
@@ -160,7 +160,7 @@ int32_t ad5791_set_register_value(struct ad5791_dev *dev,
 	status = spi_write_and_read(dev->spi_desc,
 				    write_command,
 				    3);
-	if(status != 3) {
+	if(status != 0) {
 		return -1;
 	}
 
@@ -177,21 +177,22 @@ int32_t ad5791_set_register_value(struct ad5791_dev *dev,
  *                          AD5791_REG_CTRL         - Control register
  *                          AD5791_REG_CLR_CODE     - Clearcode register
  *                          AD5791_CMD_WR_SOFT_CTRL - Software control register
+ * @param value - Pointer to the register data.
  *
  * @return dataRead        - The register's value or negative error code.
 *******************************************************************************/
 int32_t ad5791_get_register_value(struct ad5791_dev *dev,
-				  uint8_t register_address)
+				  uint8_t register_address,
+				  uint32_t *value)
 {
 	uint8_t register_word[3] = {0, 0, 0};
-	uint32_t data_read = 0x0;
 	int8_t status = 0;
 
 	register_word[0] = (AD5791_READ | AD5791_ADDR_REG(register_address)) >> 16;
 	status = spi_write_and_read(dev->spi_desc,
 				    register_word,
 				    3);
-	if(status != 3) {
+	if(status != 0) {
 		return -1;
 	}
 	register_word[0] = 0x00;
@@ -200,14 +201,14 @@ int32_t ad5791_get_register_value(struct ad5791_dev *dev,
 	status = spi_write_and_read(dev->spi_desc,
 				    register_word,
 				    3);
-	if(status != 3) {
+	if(status != 0) {
 		return -1;
 	}
-	data_read = ((int32_t)register_word[0] << 16) |
-		    ((int32_t)register_word[1] <<  8) |
-		    ((int32_t)register_word[2] <<  0);
+	*value = ((int32_t)register_word[0] << 16) |
+		 ((int32_t)register_word[1] <<  8) |
+		 ((int32_t)register_word[2] <<  0);
 
-	return data_read;
+	return status;
 }
 
 /***************************************************************************//**
@@ -227,14 +228,14 @@ int32_t ad5791_dac_ouput_state(struct ad5791_dev *dev,
 {
 	uint32_t old_ctrl = 0;
 	uint32_t new_ctrl = 0;
+	uint32_t val;
 	int32_t status = 0;
 
-	status = ad5791_get_register_value(dev,
-					   AD5791_REG_CTRL);
+	status = ad5791_get_register_value(dev, AD5791_REG_CTRL, &val);
 	if(status < 0) {
 		return status;
 	}
-	old_ctrl = status;
+	old_ctrl = val;
 	/* Clear DACTRI and OPGND bits. */
 	old_ctrl = old_ctrl & ~(AD5791_CTRL_DACTRI | AD5791_CTRL_OPGND);
 	/* Sets the new state provided by the user. */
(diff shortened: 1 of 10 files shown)


projects/cn0531/src/parameters.h
```diff
@@ -0,0 +1,108 @@
+/***************************************************************************//**
+ *   @file   ad9361/src/parameters.h
```
cristiantpop, line 2: Fix header

projects/cn0531/pinmux_config.c
```diff
@@ -0,0 +1,48 @@
...
+ **
+ ** GPIO (unavailable)
+ ** ------------------
+ ** P0_10, P0_11, P1_06, P1_07, P1_08, P1_09
+ */
+
+#include <sys/platform.h>
+#include <stdint.h>
+
+#define SPI1_SCLK_PORTP1_MUX  ((uint16_t) ((uint16_t) 1<<12))
```
cristiantpop, line 25: Maybe use BIT() macro

projects/cn0531/src.mk
```diff
@@ -0,0 +1,19 @@
+#See No-OS/tool/scripts/src_model.mk for variable description
+
+SRC_DIRS += $(PROJECT)/src
+SRC_DIRS += $(NO-OS)/iio/iio_app
+
+# Add to SRCS source files to be build in the project
+SRCS += $(NO-OS)/drivers/dac/ad5791/ad5791.c \
```
mchindri, line 7: SRC_DIRS += $(NO-OS)/drivers/dac/ad5791

projects/cn0531/src/main.c
```diff
@@ -0,0 +1,128 @@
...
+#include "irq.h"
+#include "irq_extra.h"
+#include "uart.h"
+#include "uart_extra.h"
+#include "ad5791.h"
+#include "iio_ad5791.h"
+#include "spi_extra.h"
+#include "iio_app.h"
+
+#include <sys/platform.h>
```
mchindri, line 56: non of this headers are needed, there are in platform_init

projects/cn0531/src/main.c
```diff
@@ -0,0 +1,128 @@
...
+#include "adi_initialize.h"
+#include <drivers/pwr/adi_pwr.h>
+
+#define MAX_SIZE_BASE_ADDR		1024
+
+static uint8_t out_buff[MAX_SIZE_BASE_ADDR];
+
+#define DAC_DDR_BASEADDR	((uint32_t)out_buff)
```
mchindri, line 64: You should move these lines to parameters.h inside the ADUCM3029 define. Otherwishe DAC_DDR_BASEADDR will be redefined

projects/cn0531/pinmux_config.c
mchindri: This is a autogenerated file for aducm3029

projects/cn0531/src/main.c
mchindri: or remove xilinx stuff from parameters

projects/cn0531/src/parameters.h
```diff
@@ -0,0 +1,108 @@
...
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+*******************************************************************************/
+#ifndef __PARAMETERS_H__
+#define __PARAMETERS_H__
+
+/******************************************************************************/
+/***************************** Include Files **********************************/
+/******************************************************************************/
+
+#ifdef XILINX_PLATFORM
```
amiclaus, line 46: how do you check the defined platform, if the file containing the defines is not included here?

mchindri: If is defined at compile time it will work: -DXILINX_PLAFORM , it is done inside the makefile. Anyway maybe app_config.h can be included.

adrimbarean:
V1:
 - fix comment header;
 - use SRC_DIRS;
 - remove unnecessary includes and defines;
 - rebase on master to pass azure tests.

amiclaus (approved):

drivers/dac/ad5791/ad5791.c
```diff
@@ -210,6 +211,26 @@ int32_t ad5791_get_register_value(struct ad5791_dev *dev,
...
 }
 
+/**
+ * @brief Get register value standard function.
+ * @param [in] dev - Pointer to the driver handler.
+ * @param [in] addr - Address of the register.
+ * @param [out] value - Pointer to the value of the register.
+ * @return 0 in case of success, error code otherwise.
+ */
+int32_t ad5791_get_register_value2(struct ad5791_dev *dev, uint8_t addr,
```
dbogdan, line 221: In this case, I would only update the original ad5791_get_register_value() function instead of creating a wrapper.

adrimbarean:
V2:
 - update ad5791_get_register_value() instead of creating a wrapper

adrimbarean:
V3:
 - fix doxygen

mchindri (changes requested):

drivers/dac/ad5791/ad5791.c
```diff
@@ -227,10 +228,10 @@ int32_t ad5791_dac_ouput_state(struct ad5791_dev *dev,
 {
 	uint32_t old_ctrl = 0;
 	uint32_t new_ctrl = 0;
+	uint32_t val;
 	int32_t status = 0;
 
-	status = ad5791_get_register_value(dev,
-					   AD5791_REG_CTRL);
+	status = ad5791_get_register_value(dev, AD5791_REG_CTRL, &val);
```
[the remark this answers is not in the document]
adrimbarean: Solved. Good catch, thank you!

adrimbarean:
V4:
 - Update api to correctly use ad5791_get_register_value function;

mchindri (approved):
I would put iio structures in .h

drivers/dac/ad5791/iio_ad5791.h
```diff
@@ -0,0 +1,105 @@
...
+/******************************************************************************/
+ssize_t ad5791_iio_get_raw(void *device, char *buf, size_t len,
+			   const struct iio_ch_info *channel, intptr_t priv);
+ssize_t ad5791_iio_set_raw(void *device, char *buf, size_t len,
+			   const struct iio_ch_info *channel, intptr_t priv);
+
+/*****************************************************************************/
+/***************************** Constant definition ***************************/
+/*****************************************************************************/
+static struct scan_type const ad5791_iio_scan_type = {
```
mirceacaprioru, line 60: what is the benefit of defining variables inside the header ? And we just use the static to make them similar to global variable why not just use global variable insidethe .c file.

drivers/dac/ad5791/iio_ad5791.h
```diff
@@ -0,0 +1,105 @@
...
+#include "iio.h"
+#include "util.h"
+#include "ad5791.h"
+
+#define AD5791_CH_NO 1
+
+/******************************************************************************/
+/************************ Functions Declarations ******************************/
+/******************************************************************************/
+ssize_t ad5791_iio_get_raw(void *device, char *buf, size_t len,
```
mirceacaprioru, line 52: these functions are here only to be used in the static variable. They are just internal function to this driver so should not be called outside of it.

adrimbarean:
V5:
 - define iio structures in .h;

Personally I would still advise against it for a few reasons:
 1. It's usually done in the source file (e.g. [linux:ad5791](https://github.com/analogdevicesinc/linux/blob/master/drivers/iio/dac/ad5791.c#L281), [linux:ad7124](https://github.com/analogdevicesinc/linux/blob/master/drivers/iio/adc/ad7124.c#L162)).
 2. It's usually considered to be best practice. ([stackoverflow opinion](https://stackoverflow.com/questions/24261345/declaring-global-variables-in-the-header-file-or-the-c-source-file))
 3. For them to work we have to extend the scope of the attribute functions beyond their translation unit (i.e. [making them non-static](https://stackoverflow.com/questions/24261345/declaring-global-variables-in-the-header-file-or-the-c-source-file)). This means that there will be access to internals of the driver from where there should be none and we have avoided this in the past too (e.g [uart driver](https://github.com/analogdevicesinc/no-OS/blob/master/drivers/platform/aducm3029/uart.c#L211))

adrimbarean:
V6:
 - change iio name from "cn0531" to "ad5791" ([here](https://github.com/analogdevicesinc/no-OS/blob/cb29e6e50c726ae2bf242795d0a3c8d89f2e3a44/projects/cn0531/src/main.c#L118))

dbogdan (approved):

PR #936, opened 2021-02-12
adf5902: add driver and project

amiclaus:
Add driver and application for ADF5902.


amiclaus:
V2:
- remove extra empty lines

drivers/frequency/adf5902/adf5902.c
```diff
@@ -0,0 +1,1046 @@
...
+/************************** Functions Implementation **************************/
+/******************************************************************************/
+
+/**
+ * @brief Writes 4 bytes of data to ADF5902.
+ * @param dev - The device structure.
+ * @param data - Data value to write.
+ * @return Returns 0 in case of success or negative error code otherwise.
+ */
+int32_t adf5902_write(struct adf5902_dev *dev, uint32_t data)
```
adrimbarean, line 59: Is it possible to have register read ad write functions that have the standard signature?
Standard register read function: xxxxxx_reg_read(device *ptr, uint32_t addr, uint32_t *val)
Standard register write function: xxxxxx_reg_read(device *ptr, uint32_t addr, uint32_t val)

amiclaus:
v3:
- remove more extra empty lines

drivers/frequency/adf5902/adf5902.c
```diff
@@ -0,0 +1,1046 @@
...
+	return SUCCESS;
+}
+
+/**
+ * @brief Initializes the ADF5902.
+ * @param device - The device structure.
+ * @param init_param - The structure containing the device initial parameters.
+ * @return Returns 0 in case of success or negative error code.
+ */
+int32_t adf5902_init(struct adf5902_dev **device,
```
adrimbarean, line 151: The init function seems a bit large and complicated. Can it be separated in multiple blocks based on functionality? This can help a reader understand the code. The recalibrate and compute_frequency functions will also benefit from this.

drivers/frequency/adf5902/adf5902.h
```diff
@@ -0,0 +1,585 @@
...
+#define ADF5902_BUFF_SIZE_BYTES		4
+
+/******************************************************************************/
+/*************************** Types Declarations *******************************/
+/******************************************************************************/
+
+struct slope {
+	/* Deviation Word */
+	int16_t dev_word;
+
```
adrimbarean, line 424: nitpick: seems like extra line here.

drivers/frequency/adf5902/adf5902.c
amiclaus: Yes, I can simplify a bit the initialization flow.

drivers/frequency/adf5902/adf5902.c
amiclaus: The register addresses are 5-bit long, so it will require more data shifting and arrangement within the read/write functions to properly transmit/receive the bytes.

Moreover, all the other ADF* parts from this section are using this format:
https://github.com/analogdevicesinc/no-OS/blob/b7312f038fa915933b77142d86101dd024658a0c/drivers/frequency/adf4156/adf4156.h#L239
https://github.com/analogdevicesinc/no-OS/blob/b7312f038fa915933b77142d86101dd024658a0c/drivers/frequency/adf4157/adf4157.h#L233
https://github.com/analogdevicesinc/no-OS/blob/b7312f038fa915933b77142d86101dd024658a0c/drivers/frequency/adf4350/adf4350.h#L206

drivers/frequency/adf5902/adf5902.h
amiclaus: I added extra line between each parameter for a better readability. But I can remove them all.

drivers/frequency/adf5902/adf5902.c
```diff
@@ -0,0 +1,1046 @@
...
+	struct adf5902_dev *dev;
+
+	dev = (struct adf5902_dev *)calloc(1, sizeof(*dev));
+	if (!dev)
+		return FAILURE;
+
+	/* GPIO Chip Enable */
+	ret = gpio_get(&dev->gpio_ce, init_param->gpio_ce_param);
+	if (ret != SUCCESS)
+		return ret;
```
mchindri, line 165: need to free dev on failure

drivers/frequency/adf5902/adf5902.c
```diff
@@ -0,0 +1,1046 @@
...
+
+	dev->ramp_status = init_param->ramp_status;
+
+	if(init_param->ramp_mode != ADF5902_CONT_SAWTOOTH &&
+	    init_param->ramp_mode != ADF5902_SAWTOOTH_BURST &&
+	    init_param->ramp_mode != ADF5902_CONTINUOUS_TRIANGULAR &&
+	    init_param->ramp_mode != ADF5902_SINGLE_RAMP_BURST)
+		return FAILURE;
+
+	dev->ramp_mode = init_param->ramp_mode;
```
mchindri, line 311: I would add a function just for checking init param.
I would call it at the beginning of this function before allocating dev.

drivers/frequency/adf5902/adf5902.c
```diff
@@ -0,0 +1,1046 @@
...
+	    init_param->ramp_status != ADF5902_TEMP_SENS_TO_ADC)
+		return FAILURE;
+
+	dev->ramp_status = init_param->ramp_status;
+
+	if(init_param->ramp_mode != ADF5902_CONT_SAWTOOTH &&
+	    init_param->ramp_mode != ADF5902_SAWTOOTH_BURST &&
+	    init_param->ramp_mode != ADF5902_CONTINUOUS_TRIANGULAR &&
+	    init_param->ramp_mode != ADF5902_SINGLE_RAMP_BURST)
+		return FAILURE;
```
mchindri, line 309: Nitpick: Instead of returning just FAILURE, I would return a different value for each if
Like -0x100, -0x200, .. (leaving the first byte empty in order to not be confused with codes from ernno.h)
This will be useful for debuging.
Or at least -EINVAL for all :)

drivers/frequency/adf5902/adf5902.c
mchindri: Allways 4 bytes will be written? 
I agree with Andrei that the standard register function are a better approach.
Anyway I think it will require a lot of changes so I don't know what to say for this one.
But for the future I would use the standard ones

drivers/frequency/adf5902/adf5902.c
```diff
@@ -0,0 +1,1046 @@
...
+
+	/* Set ADC clock divider value for 1MHz ADC clock */
+	dev->adc_clk_div = dev->f_pfd / ADF5902_ADC_CLK_DIV_1MHZ;
+
+	if (dev->adc_clk_div > ADF5902_ADC_MAX_CLK_DIVIDER)
+		return FAILURE;
+
+	/* Master Reset */
+	ret = adf5902_write(dev, ADF5902_REG7 | ADF5902_REG7_RESERVED |
```
mchindri, line 345: What about adding something similar to: https://github.com/analogdevicesinc/linux/blob/master/drivers/iio/regmap/iio-regmap.c
In order to initialize the register from an array of commands?
(Not in this PR)

projects/adf5902_sdz/src.mk
```diff
@@ -0,0 +1,33 @@
...
+#     Shared variables:							       #
+#	- PROJECT							       #
+#	- DRIVERS							       #
+#	- INCLUDE							       #
+#	- PLATFORM_DRIVERS						       #
+#	- NO-OS								       #
+#									       #
+################################################################################
+
+# Uncomment to select the profile
```
mchindri, line 12: remove this line

mchindri: or add profiles to select :)

drivers/frequency/adf5902/adf5902.c
amiclaus: Splitted up into multiple functions the device initialization function.

drivers/frequency/adf5902/adf5902.h
amiclaus: Removed.

drivers/frequency/adf5902/adf5902.c
amiclaus: added unconditional jumps throughout the entire init function.

projects/adf5902_sdz/src.mk
amiclaus: removed.

drivers/frequency/adf5902/adf5902.c
amiclaus: added `return -EINVAL` for all init_param checks.

drivers/frequency/adf5902/adf5902.c
amiclaus: added separate function for init_param check.

amiclaus:
V4:
- split initialization function into multiple sub-functions for better readability
- add separate function for init_param checks + return -EINVAL error code
- add unconditional jumps inside main init function
- remove more extra empty lines

drivers/frequency/adf5902/adf5902.c
```diff
@@ -0,0 +1,940 @@
...
+		     struct adf5902_init_param *init_param)
+{
+	int32_t ret;
+	uint32_t i;
+	struct adf5902_dev *dev;
+
+	/* Check ADF5902 Initialization Paramteres */
+	ret = adf5902_check_init_param(init_param);
+	if(ret != SUCCESS)
+		goto error;
```
mchindri, line 495: You can just return ret

drivers/frequency/adf5902/adf5902.c
```diff
@@ -0,0 +1,940 @@
...
+
+	dev = (struct adf5902_dev *)calloc(1, sizeof(*dev));
+	if (!dev)
+		goto error;
```
mchindri, line 499: return -ENOMEM

drivers/frequency/adf5902/adf5902.c
amiclaus: if the unconditional jump is already implemented in this function, I would use it as much as I can and avoid mixing it with direct returns.

drivers/frequency/adf5902/adf5902.c
amiclaus: Done

drivers/frequency/adf5902/adf5902.c
amiclaus: Actually, I will remove the `error:` completely I `return ret;` :)

amiclaus:
V5:
- remove `error:` and `return ret;` instead
- `return -ENOMEM`

amiclaus:
V6:
- change write/readback function headers for adf5902 driver
- add 2 extra commits for IIO support

drivers/frequency/adf5902/adf5902.c
```diff
@@ -0,0 +1,944 @@
...
+ * @param data - Data value to write.
+ * @return Returns 0 in case of success or negative error code otherwise.
+ */
+int32_t adf5902_write(struct adf5902_dev *dev, uint8_t reg_addr, uint32_t data)
+{
+	uint8_t buff[ADF5902_BUFF_SIZE_BYTES];
+
+	data = data | reg_addr;
+
+	buff[0] = data >> 24;
```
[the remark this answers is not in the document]
mchindri: Same for read

drivers/frequency/adf5902/iio_adf5902.h
```diff
@@ -0,0 +1,52 @@
...
+#include "iio_types.h"
+#include "adf5902.h"
+
+/** IIO Descriptor */
+const struct iio_device adf5902_iio_descriptor = {
+	.debug_reg_read = (int32_t (*)())adf5902_readback,
+	.debug_reg_write = (int32_t (*)())adf5902_write,
+};
+
+#endif //IIO_ADF5902_H
```
mchindri, line 52: Nitpick: new line needed

projects/adf5902_sdz/src/app_iio.c
```diff
@@ -0,0 +1,122 @@
+/***************************************************************************//**
```
mchindri, line 1: You can use iio/iio_app (see iio_demo).
Otherwise we will duplicate the same iio code in every project

drivers/frequency/adf5902/adf5902.c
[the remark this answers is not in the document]
amiclaus: Is there any no-OS driver that uses this approach? Otherwise I would change all no-OS drivers with the suggested improvement in a separate commit.

projects/adf5902_sdz/src/app_iio.c
amiclaus: Is there any actual no-OS project that uses this approach (not demo projects)? Otherwise I would change (and test) all no-OS projects with the suggested improvement in a separate PR/commit .

amiclaus:
V6:
- fix doxygen complains
- add new line at end of file

projects/adf5902_sdz/src/app_iio.c
mchindri: https://github.com/analogdevicesinc/no-OS/blob/master/projects/ad7124-8pmdz
When iio_app was added only iio_demo was modified as example for new projects.
The others projects where not modified.

amiclaus:
V7:
- remove extra files added to commits

amiclaus:
V8:
- align function headers between `.c` and `.h` files for the adf5902 driver.

drivers/frequency/adf5902/adf5902.c
```diff
@@ -0,0 +1,946 @@
...
+
+	if (dev->f_pfd > ADF5902_MAX_FREQ_PFD)
+		return FAILURE;
+
+	dev->int_div = (uint16_t)(dev->rf_out / (dev->f_pfd * 2));
+
+	if (dev->int_div > ADF5902_MAX_INT_MSB_WORD)
+		return FAILURE;
+
+	dev->frac_msb = (uint16_t)((((float)(dev->rf_out) / (dev->f_pfd * 2)) -
```
mhennerich, line 136: Why is this computed twice? It's one 25-bit FRACT value! 
Avoid floating point math!
Also your grater than checks are superfluous and wrong.

drivers/frequency/adf5902/adf5902.c
```diff
@@ -0,0 +1,946 @@
...
+		dev->freq_cal_div += 1;
+
+	if(dev->freq_cal_div > ADF5902_MAX_FREQ_CAL_DIV)
+		return FAILURE;
+
+	/* Set clock divider value */
+	dev->clk1_div = dev->f_pfd / ADF5902_CLK1_DIV_25KHZ;
+
+	/* f_PFD/25kHz <= clk1_div */
+	if (dev->f_pfd % ADF5902_CLK1_DIV_25KHZ)
```
mhennerich, line 163: DIV_ROUND_UP() ???

drivers/frequency/adf5902/adf5902.c
```diff
@@ -0,0 +1,946 @@
...
+ * @param dev - The device structure.
+ * @return Returns 0 in case of success or negative error code.
+ */
+static int32_t adf5902_vco_freq_param(struct adf5902_dev *dev)
+{
+
+	dev->f_pfd = (dev->ref_in * (1 + dev->ref_doubler_en) /
+		      (dev->ref_div_factor *(1 + dev->ref_div2_en)));
+
+	if (dev->f_pfd > ADF5902_MAX_FREQ_PFD)
```
mhennerich, line 128: Can't we compute that all internally? Try to maximize PFD

drivers/frequency/adf5902/adf5902.c
amiclaus: This is the approach used in the datasheet. Should I go for a different approach, or keep as much as I can a correlation between code and documentation?
https://www.analog.com/media/en/technical-documentation/data-sheets/ADF5902.pdf
(page 33 of 39)
image

The checks are present to make sure that the computed values do not exceed the size of the registers, since the data size/type of each parameter does not constrain this (sometimes).

drivers/frequency/adf5902/adf5902.c
amiclaus: Will do.

drivers/frequency/adf5902/adf5902.c
amiclaus: This is how documentation suggests the PFD computation based on the input parameters:
https://www.analog.com/media/en/technical-documentation/data-sheets/ADF5902.pdf
(Page 11 of 39)
image

What do you mean by "Try to maximize PFD"?

dbogdan: For example, in your formula, the f_pfd depends on ref_div_factor. You should calculate the ref_div_factor (instead of providing it in the initialization structure) so the f_pfd value will be the maximum allowed one (still lower than ADF5902_MAX_FREQ_PFD).

amiclaus: I understand, thanks! I was using as reference the ADF5902 Evaluations Software from analog.com, where all these parameters were configured manually by the user (including ref_div_factor). I will compute them inside the driver.

amiclaus:
V9:
- use `DIV_ROUND_UP` macro provided by `util.h`
- remove `ref_div_factor` from initialization parameters of the device, since it is now computed inside the driver.

amiclaus:
V10:
- remove floating point math

drivers/frequency/adf5902/adf5902.c
mhennerich: I don't know who wrote the ds. For me this looks stupid. Do we need to repeat stupid things if we can do it better?

drivers/frequency/adf5902/adf5902.c
```diff
@@ -0,0 +1,937 @@
...
+		dev->f_pfd = (dev->ref_in * (1 + dev->ref_doubler_en) /
+			      (dev->ref_div_factor *(1 + dev->ref_div2_en)));
+	} while (dev->f_pfd > ADF5902_MAX_FREQ_PFD);
+
+	dev->int_div = (uint16_t)(dev->rf_out / (dev->f_pfd * 2));
+
+	if (dev->int_div > ADF5902_MAX_INT_MSB_WORD)
+		return FAILURE;
+
+	dev->frac_msb = (uint16_t)(((dev->rf_out) * pow(2, 12) / (dev->f_pfd * 2)) -
```
dbogdan, line 139: pow(2, x) is less efficient than shifting (1 << x), so I wouldn't use it in these cases.

amiclaus: done.

drivers/frequency/adf5902/adf5902.c
amiclaus: Changed the approach: fractional part is computed once, then the MSB and LSB are extracted. 
Let me know if this looks fine for you.

amiclaus:
V11:
- improve the way of computing fractional part.
- remove pow() and use raw bit shifting instead.

adrimbarean (approved):

amiclaus:
V12:
- use iio_app
- modify IIO driver after latest changes on master branch

projects/adf5902_sdz/src/adf5902_sdz.c
```diff
@@ -215,6 +222,14 @@ int main(void)
 
 	pr_info("ADF5902 Temperature value: %.2f degC \n", temperature);
 
+#ifdef IIO_SUPPORT
+	struct iio_app_device devices[] = {
+		IIO_APP_DEVICE("adf5902_dev", dev, &adf5902_iio_descriptor,
+			       NULL, NULL),
+	};
+	return iio_app_run(devices, NUMBER_OF_DEVICES);
```
mchindri, line 230: nitpick: ARRAY_SIZE(devices)

mchindri (approved):

amiclaus:
V13:
- use array size

PR #939, opened 2021-02-16
tools:scripts:Build_projects.py: Script for building all projects

mchindri:
Idea to build all projects from one scripts. This will be useful for testing and for CI
In iio_demo:builds.json is an example of an input file for the script for a project with all possible configurations.
Will be good to discuses the format of that file.


mchindri:
https://github.com/analogdevicesinc/no-OS/pull/942 was created

PR #943, opened 2021-02-22
CI: Build all aducm projects

mchindri:
With this pr and a some modification in azure all aducm project binaries will be added here:
https://github.com/analogdevicesinc/no-OS/releases
This will support all other platforms to, but for the moment just aducm has been tested.


mirceacaprioru:
Why are these projects in the no-os repo ? why not use the ADICUP3029 ?

tools/scripts/build_projects.py
```diff
@@ -0,0 +1,76 @@
...
+							(to_blue(build_name),
+							to_blue(project),
+							to_blue(cur_platform)))
+						build_dir_name = 'build_%s_%s' % (cur_platform, build_name)
+						binary = os.path.join(build_dir_name, "%s_%s.elf" % (project, build_name))
+						export_file = os.path.join(projet_dir, binary)
+						cmd = 'make -C %s %s BUILD_DIR_NAME=%s\
+ LOCAL_BUILD=n LINK_SRCS=n BINARY=%s VERBOSE=y -j%d ' % (projet_dir, flags, build_dir_name, binary, multiprocessing.cpu_count() - 1)
+						run_cmd(cmd + 'ra')
+						if (cur_platform == 'aducm3029'):
```
buha: I don't think this kind of stuff should belong here. If you need to do aducm specific actions, you should add an optional postbuild: or postprocessing: rule to build system, and each platform may or may not implement it.

This script should contain the logic to build any project - period, it shouldn't know anything specific.

tools/scripts/aducm.mk
```diff
@@ -133,6 +133,9 @@ HEX = $(basename $(BINARY)).hex
 #                                 RULES                              
 #------------------------------------------------------------------------------
 
+PHONY += hex
+hex: $(HEX)
```
buha, line 137: hex: is only available for aducm and this creates a nightmare when documenting, we need to keep track of all platform-specific make rules and document each separately. Instead, we should converge on a set of universal make rules for all platforms. This could be part of a postprocessing: or postbuild: step that every platform .mk may implement or not (optional)

tools/scripts/build_projects.py
```diff
@@ -22,11 +22,11 @@ def parse_input():
 			    formatter_class=argparse.RawTextHelpFormatter)
```
buha, line 22: these changes to build_projects.py should belong to previous commit.
otherwise, the jsons are really nice and clean, that's gonna be an improvement over what we had so far ;)

mchindri:
> Why are these projects in the no-os repo ? why not use the ADICUP3029 ?

CI support will added for adicup repo too. I think they where added here for easier support.

mchindri:
v2: 
 - Move changes to good commit
 -  Create post_build rule.

tools/scripts/aducm.mk
mchindri: done

tools/scripts/build_projects.py
mchindri: done

tools/scripts/build_projects.py
mchindri: just the replacement of .elf to .hex is keept.

dbogdan (approved):

PR #944, opened 2021-02-22
Set up CI for build projects

mchindri:
This pipeline will use a localpool to build noos projects

### DIFF
--- a/build_projects.yml
+++ b/build_projects.yml
@@ -1,0 +1,26 @@
+trigger:
+- master
+- dev/*
+
+pr:
+- master
+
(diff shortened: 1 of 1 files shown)


buha (commented):
Could you please add more details to the commit message ? I think it should at least be clear from the commit message what CI is being used to configure with this file.

How we do CI is not gonna be documented elsewhere (i guess) and we better have something in the git log for those who read it...

mchindri:
v2:
 - Update commit description

dbogdan (approved):

PR #946, opened 2021-02-22
include: util: add bit swap macro

amiclaus:
Macro that swaps bits inside a byte.

Useful for drivers that support different data orientation types: LSB
first / MSB first.

Example of usage:
image


include/util.h
```diff
@@ -96,6 +96,16 @@
 	((((x) & 0xff000000) >> 24) | (((x) & 0x00ff0000) >>  8) | \
 	 (((x) & 0x0000ff00) <<  8) | (((x) & 0x000000ff) << 24))
 
+#define bit_swap_constant_8(x) \
```
mirceacaprioru, line 99: is this not already done here https://github.com/analogdevicesinc/no-OS/blob/42412dc58eeffb75f6fbfc28d074619d8f6e8bd6/include/util.h#L80 ?

amiclaus: not really. `swap(x, y)` interchanges the values between 2 variables. this function rearranges all bits from a byte, where LSB becomes MSB and vice-versa.

mchindri (approved):

PR #948, opened 2021-02-25
Print Log update

amiclaus:
- Remove FILE, LINE and FUNCTION information from the low priority log
levels.
- Print code line before function name.
- Set default log level to 6 inside `print_log.h`.
- Remove the log level defines from projects where the default one is
used.

### DIFF
--- a/include/print_log.h
+++ b/include/print_log.h
@@ -80,29 +80,25 @@ fmt, __FILE__, __func__, __LINE__, ##args)
 #endif
 
 #if defined(LOG_LEVEL) && LOG_LEVEL >= LOG_WARNING && LOG_LEVEL <= LOG_DEBUG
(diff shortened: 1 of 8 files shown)


cristiantpop (approved):

include/print_log.h
```diff
@@ -42,6 +42,10 @@
 
 #include <stdio.h>
 
+#ifndef LOG_LEVEL
+#define LOG_LEVEL 0x6
```
buha, line 46: I'd just move this below these defines and 
#ifndef LOG_LEVEL
#define LOG_LEVEL LOG_INFO
#endif

amiclaus: Sure.

amiclaus:
V2:
- move `ifndef` below log level defines.

buha (approved):

PR #958, opened 2021-03-08
[4] Add spi_transfer and new spi PL driver

mchindri:
New xilinx pl spi drivers using resgisters will be needed in order to do multiple transfers with CS low.
Also 2 commit with spi modifications where added in order to get feedback.

### DIFF
--- a/drivers/platform/xilinx/spi_extra.h
+++ b/drivers/platform/xilinx/spi_extra.h
@@ -73,7 +73,7 @@ enum xil_spi_type {
 /**
  * @struct xil_spi_init_param
  * @brief Structure holding the initialization parameters for Xilinx platform
- * specific SPI parameters.
+ * specific SPI parameters when using xil_platform_ops.
  */
 typedef struct xil_spi_init_param {
 	/** Xilinx architecture */
@@ -106,6 +106,11 @@ extern const struct spi_platform_ops spi_eng_platform_ops;
  * @brief Xilinx specific SPI platform ops structure
  */
 extern const struct spi_platform_ops xil_platform_ops;
+
+/**
+ * @brief Xilinx SPI PS register functions. No extra param needed
+ */
+extern const struct spi_platform_ops xil_spi_reg_ops_pl;
 
 /******************************************************************************/
 /************************ Functions Declarations ******************************/
--- a/drivers/platform/xilinx/xilinx_spi_pl.c
+++ b/drivers/platform/xilinx/xilinx_spi_pl.c
@@ -1,0 +1,348 @@
+/*******************************************************************************
+ *   @file   xilinx/xilinx_spi_pl.c
+ *   @brief  Driver for spi pl device using registers
+ *   @author Mihail Chindris (mihail.chindris@analog.com)
+********************************************************************************
+ * Copyright 2021(c) Analog Devices, Inc.
+ *
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *  - Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ *  - Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in
+ *    the documentation and/or other materials provided with the
+ *    distribution.
+ *  - Neither the name of Analog Devices, Inc. nor the names of its
+ *    contributors may be used to endorse or promote products derived
+ *    from this software without specific prior written permission.
+ *  - The use of this software may or may not infringe the patent rights
+ *    of one or more patent holders.  This license does not release you
+ *    from the requirement that you obtain separate licenses from these
+ *    patent holders to use this software.
+ *  - Use of the software either in source or binary form, must be run
+ *    on or directly connected to an Analog Devices Inc. component.
+ *
+ * THIS SOFTWARE IS PROVIDED BY ANALOG DEVICES "AS IS" AND ANY EXPRESS OR
+ * IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, NON-INFRINGEMENT,
+ * MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED.
+ * IN NO EVENT SHALL ANALOG DEVICES BE LIABLE FOR ANY DIRECT, INDIRECT,
+ * INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, INTELLECTUAL PROPERTY RIGHTS, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+*******************************************************************************/
+
+#include <xparameters.h>
+
+#ifdef XPAR_XSPI_NUM_INSTANCES
+
+/******************************************************************************/
+/***************************** Include Files **********************************/
+/******************************************************************************/
+
+#include <xspi.h>
+#include <stdlib.h>
+#include <stdbool.h>
+#include "spi.h"
+#include "error.h"
+#include "delay.h"
+#include "util.h"
+
+/******************************************************************************/
+/********************** Macros and Constants Definitions **********************/
+/******************************************************************************/
+
+#define XSP_DEFAULT_SR_VALUE	0x25
+#define XSP_DUMMY_DATA		0xFF
+
+/*
+ * Used to store current spi device configuration in order to not update it for
+ * each spi transaction if not needed.
+ */
+static struct spi_desc g_spi_desc[XPAR_XSPI_NUM_INSTANCES];
+
+/******************************************************************************/
+/*************************** Types Declarations *******************************/
+/******************************************************************************/
+
+/* Xilinx PL spi specific data */
+struct xspi_desc {
+	uint32_t base_addr;
+	uint32_t disable_slaves_mask;
+	uint32_t fifo_depth;
+	uint8_t  cs_asserted;
+};
+
+/******************************************************************************/
+/************************ Functions Definitions *******************************/
+/******************************************************************************/
+
+/* Read device register */
+static inline uint32_t _read(struct xspi_desc *xdesc, uint32_t reg)
+{
+	return Xil_In32(xdesc->base_addr + reg);
+}
+
+/* Write device register */
+static inline void _write(struct xspi_desc *xdesc, uint32_t reg, uint32_t val)
+{
+	Xil_Out32(xdesc->base_addr + reg, val);
+}
+
+/* Update device register. Only bits set to 1 in mask will be modified */
+static inline void _wr_mask(struct xspi_desc *xdesc, uint32_t reg,
+			    uint32_t val, uint32_t mask)
+{
+	uint32_t tmp;
+
+	tmp = _read(xdesc, reg);
+	tmp &= ~mask;
+	tmp |= val & mask;
+	_write(xdesc, reg, tmp);
+}
+
+/* Update spi device configuration if needed */
+static void _update_mode(struct spi_desc *desc)
+{
+	uint32_t		val;
+	uint32_t		mask;
+
+	if (desc->mode == g_spi_desc[desc->device_id].mode &&
+	    desc->bit_order == g_spi_desc[desc->device_id].bit_order)
+		return ;
+
+	/* Update mode */
+	val = desc->mode | SPI_CPOL ? XSP_CR_CLK_POLARITY_MASK : 0;
+	val |= desc->mode | SPI_CPHA ? XSP_CR_CLK_PHASE_MASK : 0;
+	mask = XSP_CR_CLK_POLARITY_MASK | XSP_CR_CLK_PHASE_MASK;
+
+	val |= desc->bit_order == SPI_BIT_ORDER_LSB_FIRST ?
+	       XSP_CR_LSB_MSB_FIRST_MASK : 0;
+	mask |= XSP_CR_LSB_MSB_FIRST_MASK;
+
+	_wr_mask(desc->extra, XSP_CR_OFFSET, val, mask);
+	g_spi_desc[desc->device_id] = *desc;
+}
+
+/* Configure SPI PL device */
+static int32_t _xil_spi_init_dev(struct xspi_desc *xdesc)
+{
+	uint32_t val;
+
+	/* Software reset */
+	_write(xdesc, XSP_SRR_OFFSET, XSP_SRR_RESET_MASK);
+	mdelay(1);
+
+	/* Check if success reset */
+	val = _read(xdesc, XSP_SR_OFFSET);
+	if (val != XSP_DEFAULT_SR_VALUE)
+		return -ENODEV;
+
+	/* Disable interrupts. Drivers only works with pooled mode */
+	_write(xdesc, XSP_IIER_OFFSET, 0);
+
+	/* Deasert chip selects */
+	_write(xdesc, XSP_SSR_OFFSET, 0xFFFFFFFF);
+
+	/* Get fifo depth */
+	while (!(_read(xdesc, XSP_SR_OFFSET) & XSP_SR_TX_FULL_MASK))
+		_write(xdesc, XSP_DTR_OFFSET, XSP_DUMMY_DATA);
+	xdesc->fifo_depth = _read(xdesc, XSP_TFO_OFFSET) + 1;
+
+	/* Configure spi device */
+	val = XSP_CR_ENABLE_MASK |
+	      XSP_CR_MANUAL_SS_MASK |
+	      XSP_CR_TXFIFO_RESET_MASK |
+	      XSP_CR_RXFIFO_RESET_MASK |
+	      XSP_CR_MASTER_MODE_MASK |
+	      XSP_CR_TRANS_INHIBIT_MASK;
+
+	_write(xdesc, XSP_CR_OFFSET, val);
+
+	return SUCCESS;
+}
+
+/* Initialize spi_desc structure and device */
+static int32_t xil_spi_init_pl(struct spi_desc **desc,
+			       const struct spi_init_param *param)
+{
+	struct spi_desc		*ldesc;
+	struct xspi_desc	*xdesc;
+	int32_t			err;
+	XSpi_Config		*cfg;
+
+	if (!desc || !param || param->device_id >= XPAR_XSPI_NUM_INSTANCES)
+		return -EINVAL;
+
+	cfg = XSpi_LookupConfig(param->device_id);
+	if (!cfg) {
+		err = -EINVAL;
+		goto error;
+	}
+
+	if (param->chip_select > cfg->NumSlaveBits)
+		return -EINVAL;
+
+	ldesc = (struct spi_desc *)calloc(1, sizeof(*ldesc));
+	if (!ldesc) {
+		err = -ENOMEM;
+		goto error;
+	}
+
+	xdesc = (struct xspi_desc *)calloc(1, sizeof(*xdesc));
+	if (!xdesc) {
+		err = -ENOMEM;
+		goto error;
+	}
+
+	ldesc->bit_order = param->bit_order;
+	ldesc->chip_select = param->chip_select;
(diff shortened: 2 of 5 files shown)


drivers/platform/xilinx/xilinx_spi_pl.c
```diff
@@ -0,0 +1,301 @@
...
+{
+	struct spi_desc		*ldesc;
+	struct xspi_desc	*xdesc;
+	int32_t			err;
+	XSpi_Config		*cfg;
+
+	if (!desc || !param)
+		return -EINVAL;
+
+	err = -EINVAL;
```
buha, line 162: This is not a good way to do it. 
`err` should be set only when you encounter an error like below
```
if (!cfg) {
    err = ...
    goto error;
}

I agree it does the job, but It's awkard to read as it is.

File: drivers/platform/xilinx/xilinx_spi_pl.c [hunk 2]
```diff
@@ -0,0 +1,301 @@
...
+#include <stdlib.h>
+#include <stdbool.h>
+#include "spi.h"
+#include "error.h"
+#include "delay.h"
+#include "util.h"
+
+#define XSP_DEFAULT_SR_VALUE	0x25
+
+#if XPAR_SPI_0_FIFO_EXIST==1
```
buha, line 54: I see SPI_0 here.
Don't SPI_1 & others need to be treated ? I think we define SPI0 SPI1 and SPI2 in spi.h so this implementation should follow.

drivers/platform/xilinx/xilinx_spi_pl.c
```diff
@@ -0,0 +1,301 @@
...
+
+/* Update device register. Only bits set to 1 in mask will be modified */
+static inline void _update(struct xspi_desc *xdesc, uint32_t reg,
+				uint32_t val, uint32_t mask)
+{
+	uint32_t old;
+
+	old = _read(xdesc, reg);
+	old &= ~mask;
+	old |= val & mask;
```
buha, line 91: maybe I'm too pendantic here but at this line `old` is actually `new` so the naming is confusing...
I'd use something like `val` or `tmp` that has neutral meaning

drivers/platform/xilinx/xilinx_spi_pl.c
```diff
@@ -0,0 +1,301 @@
...
+{
+	struct xspi_desc	*xdesc;
+
+	if (!desc || !desc->extra)
+		return -EINVAL;
+
+	xdesc = desc->extra;
+	if (assert)
+		_write(xdesc, XSP_SSR_OFFSET, ~(1 << desc->chip_select));
+
```
buha, line 250: remove the empty line before else

drivers/platform/xilinx/xilinx_spi_pl.c
```diff
@@ -0,0 +1,301 @@
...
+static int32_t xil_spi_cs_assert(struct spi_desc *desc, uint8_t assert)
+{
+	struct xspi_desc	*xdesc;
+
+	if (!desc || !desc->extra)
+		return -EINVAL;
+
+	xdesc = desc->extra;
+	if (assert)
+		_write(xdesc, XSP_SSR_OFFSET, ~(1 << desc->chip_select));
```
buha, line 249: I don't know much about the hardware here but i know that for xilinx PS, the chip select can be used as a mask to control 8 chip selects using 3 bits (CS Decode I think it's called)

The way the implementation is done here, it would assert / deassert all of them which I think it is not what you intend.
I'm not sure if the PL implementation has this CS Decode mode but it'd be worth investigating.

drivers/platform/xilinx/xilinx_spi_pl.c
```diff
@@ -0,0 +1,301 @@
...
+
+	else
+		_write(xdesc, XSP_SSR_OFFSET, xdesc->disable_slaves_mask);
+
+	return SUCCESS;
+}
+
+/* If start is 0 stop stransfer, else start it */
```
buha, line 257: typo in comment

drivers/platform/aducm3029/spi.c
```diff
@@ -134,7 +134,7 @@ int32_t spi_init(struct spi_desc **desc,
 		return FAILURE;
 
 	config = param->extra;
-	if (config->spi_channel >= NB_SPI_DEVICES ||
+	if (param->device_id >= NB_SPI_DEVICES ||
```
buha, line 137: typo in commit message

drivers/platform/xilinx/spi_extra.h
```diff
@@ -107,6 +107,8 @@ extern const struct spi_platform_ops spi_eng_platform_ops;
  */
 extern const struct spi_platform_ops xil_platform_ops;
```
cristiantpop, line 108: "..multiple transfers will **be** added..."

drivers/platform/xilinx/xilinx_spi_pl.c
```diff
@@ -0,0 +1,301 @@
...
+
+	xil_spi_cs_assert(desc, 1);
+	_xil_spi_start_transfer(xdesc, 1);
+
+	rx_remaining = bytes_number;
+	tx_remaining = bytes_number;
+	while (tx_remaining || rx_remaining) {
+		_wr_fifo(xdesc, data + (bytes_number - tx_remaining),
+				&tx_remaining);
+		_rd_fifo(xdesc, data + (bytes_number - rx_remaining),
```
cristiantpop, line 285: Are you reading data before a `_xil_spi_start_transfer` call?

mchindri: _xil_spi_start_transfer(xdesc, 1) will start the transfer
_xil_spi_start_transfer(xdesc, 0) will stop the transfer
This code is between them.

cristiantpop: True

drivers/platform/xilinx/xilinx_spi_pl.c
mchindri: From the datasheet: "At most, one bit can be asserted Low."
So my understanding is that CS Decode won't work for this device

drivers/platform/aducm3029/spi_extra.h
```diff
@@ -51,19 +51,6 @@
 /*************************** Types Declarations *******************************/
 /******************************************************************************/
 
-/**
- * @enum spi_channel
- * @brief Available SPI channels on the ADuCM3029
- */
-enum spi_channel {
```
mirceacaprioru, line 58: this change is not related to the commit description

mchindri: this is the same as device_id

drivers/platform/xilinx/xilinx_spi_pl.c
mchindri: Lazy me wanted to have less code lines. Changed anyway.

mchindri:
v2:
 - Resolve types and comment observations.
 - Add 2 commits extending spi interface for multiple messages sending and implementing it on the PL driver

drivers/platform/xilinx/xilinx_spi_pl.c
```diff
@@ -326,10 +335,81 @@ static int32_t xil_spi_write_and_read_pl(struct spi_desc *desc, uint8_t *data,
...
+	rx.buff = head->rx_buff;
+	rx.i = 0;
+	rx.is_write = 0;
+
+	tx.msg = head;
+	tx.buff = head->tx_buff;
+	tx.i = 0;
+	tx.is_write = 1;
+
+	xil_spi_cs_assert(desc, 1);
```
buha, line 393: This allows multiple transfers with CS asserted but in my opinion is not flexible enough. Like on linux, and like we discussed in the meeting, there should be a possibility to assert/deassert CS at the beggining / end of each transfer since we are at it.

drivers/platform/xilinx/xilinx_spi_pl.c
```diff
@@ -326,10 +335,81 @@ static int32_t xil_spi_write_and_read_pl(struct spi_desc *desc, uint8_t *data,
 	return SUCCESS;
 }
 
+/* Read/write FIFOs to/from buffers and update it structure accordingly */
+static void inline _transfer_and_update(struct xspi_desc *xdesc,
```
buha, line 339: Please provide an argument why chosing this FIFO approach is better compared to a simple ARRAY approach

drivers/spi/spi.c
```diff
@@ -85,3 +85,34 @@ int32_t spi_write_and_read(struct spi_desc *desc,
...
 	return desc->platform_ops->write_and_read(desc, data, bytes_number);
 }
+
+/**
+ * @brief  Iterate over head list and send all spi messages
+ * @param desc - The SPI descriptor.
+ * @param head - First link of the list with messages to be sent
+ * @return SUCCESS in case of success, negativ error code otherwise.
+ */
+int32_t spi_write_and_read_list(struct spi_desc *desc,
```
buha, line 95: I think `spi_write_and_read` is named like this for historical reasons.
In SPI there is always write and read at the same time on MOSI and MISO respectively.
A better name would have been `spi_transfer` (2 words vs 4 words).
In this case, since it's a new function and we can name it whatever, we could use `spi_multi_transfer` (3 words vs 5 words)

mchindri:
v3:
 - Rebase onto #998 
 - Add cs_change field
 - Rename spi_write_and_read_list to spi_transfer

drivers/platform/aducm3029/spi_extra.h
mchindri: Change now realted to #998

drivers/platform/xilinx/xilinx_spi_pl.c
mchindri: It is easy to create the list (no need to handle array size and allocation).

drivers/spi/spi.c
mchindri: renamed to spi_transfer

mchindri:
v4:
 - Rebase onto master

drivers/platform/xilinx/xilinx_spi_pl.c
```diff
@@ -0,0 +1,417 @@
...
+/********************** Macros and Constants Definitions **********************/
+/******************************************************************************/
+
+#define XSP_DEFAULT_SR_VALUE	0x25
+
+/*
+ * Used to store current spi device configuration in order to not update it for
+ * each spi transaction if not needed.
+ */
+static struct spi_desc g_spi_desc;
```
scuciurean, line 66: What happends if there are two spi controllers?

mchindri: in this implementations fails :)
Will create an array

include/spi.h
```diff
@@ -160,4 +191,8 @@ int32_t spi_write_and_read(struct spi_desc *desc,
 			   uint8_t *data,
 			   uint16_t bytes_number);
 
+/* Iterate over head list and send all spi messages. */
+int32_t spi_transfer(struct spi_desc *desc, struct spi_msg_list *head);
```
buha, line 195: First off, i think the user shouldn't need to care how you implemented this and whether they should pass the head or the tail or whatever as parameter. The naming of `head` then is rather unfortunate, better naming would be `msg_list` or `xfer_list`.

include/spi.h
```diff
@@ -83,6 +83,35 @@ typedef enum spi_bit_order {
 	SPI_BIT_ORDER_LSB_FIRST = 1,
 } spi_bit_order;
 
+#define SPI_MSG(_tx_buff, _rx_buff, _bytes_number, _cs_change, _next) \
```
buha, line 86: Secondly, I'd like to see an example of 3-4 messages initialized one after another, to me it seems it's going to be a pain in the ass for the user to remember to initialize .next....

mchindri: I used it like: 
image

drivers/platform/xilinx/xilinx_spi_pl.c
```diff
@@ -0,0 +1,332 @@
...
+	/* Software reset */
+	_write(xdesc, XSP_SRR_OFFSET, XSP_SRR_RESET_MASK);
+	mdelay(1);
+
+	/* Check if success reset */
+	val = _read(xdesc, XSP_SR_OFFSET);
+	if (val != XSP_DEFAULT_SR_VALUE)
+		return -ENODEV;
+
+	/* Disable interrupts. Drivers only works with pooled mode */
```
mirceacaprioru, line 155: nit: polling mode

drivers/platform/xilinx/xilinx_spi_pl.c
```diff
@@ -0,0 +1,332 @@
...
+static inline uint32_t _wr_fifo(struct xspi_desc *xdesc, uint8_t *data,
+				uint32_t len)
+{
+	uint32_t i;
+	uint32_t to_write;
+	uint32_t tx_available;
+
+	tx_available = xdesc->fifo_depth - _read(xdesc, XSP_TFO_OFFSET);;
+	to_write = min(tx_available, len);
+	for (i = 0; i < to_write; ++i)
```
mirceacaprioru, line 257: this for will render your inline useless. Do not use inline the compiler is smart enough to determine if a static function should be inlined or not;

https://www.kernel.org/doc/html/v4.10/process/coding-style.html#the-inline-disease

drivers/spi/spi.c
```diff
@@ -85,3 +85,34 @@ int32_t spi_write_and_read(struct spi_desc *desc,
...
+	int32_t ret;
+
+	if (!desc || !desc->platform_ops)
+		return -EINVAL;
+
+	if (desc->platform_ops->transfer)
+		return desc->platform_ops->transfer(desc, head);
+
+	while (head) {
+		if (head->rx_buff != head->tx_buff)
```
mirceacaprioru, line 107: if these two pointers have to be the same why do you need two for the same thing ?

drivers/platform/xilinx/xilinx_spi_pl.c
```diff
@@ -323,10 +326,92 @@ static int32_t xil_spi_write_and_read_pl(struct spi_desc *desc, uint8_t *data,
 	return SUCCESS;
 }
 
+/* Read/write FIFOs to/from buffers and update it structure accordingly.
+ * Return */
+static inline uint8_t _transfer_and_update(struct xspi_desc *xdesc,
```
mirceacaprioru, line 331: inline useless for such a large function.

drivers/platform/xilinx/xilinx_spi_pl.c
```diff
@@ -0,0 +1,417 @@
...
+	uint32_t		len;
+	uint8_t			cs_change;
+
+	total = 0;
+	cs_change = 0;
+	do {
+		len = it->msg->bytes_number - it->i;
+		if (it->is_write)
+			i = _wr_fifo(xdesc, it->buff + it->i, len);
+		else if (it->buff)
```
mirceacaprioru, line 345: why this check? isn't it the se with _rd_fifo(xdesc, NULL, len); if it->buff == NULL ?

drivers/platform/xilinx/xilinx_spi_pl.c
```diff
@@ -0,0 +1,417 @@
...
+	struct msg_iterator	rx;
+	struct msg_iterator	tx;
+
+	if (!desc || !desc->extra)
+		return -EINVAL;
+
+	xdesc = desc->extra;
+	_update_mode(desc);
+
+	rx.msg = head;
```
mirceacaprioru, line 386: this iterator just complicates things. Just use msg_list you already have there the tx and rx buffer an you actually don't use the index outside of _transfer_and_update so in can be a tmp variable there.

drivers/platform/xilinx/xilinx_spi_pl.c
```diff
@@ -0,0 +1,332 @@
...
+}
+
+/* Write device register */
+static inline void _write(struct xspi_desc *xdesc, uint32_t reg, uint32_t val)
+{
+	Xil_Out32(xdesc->base_addr + reg, val);
+}
+
+/* Update device register. Only bits set to 1 in mask will be modified */
+static inline void _update(struct xspi_desc *xdesc, uint32_t reg,
```
adrimbarean, line 106: nitpick: i saw the name `write_mask` being used before. [example](https://github.com/analogdevicesinc/no-OS/blob/cbd8405892eeadc4383a6d8ddc3644fb9a6f4743/drivers/adc/ad7091r5/ad7091r5.c#L141)
Should we do it like this?

drivers/spi/spi.c
```diff
@@ -85,3 +85,34 @@ int32_t spi_write_and_read(struct spi_desc *desc,
...
 	return desc->platform_ops->write_and_read(desc, data, bytes_number);
 }
+
+/**
+ * @brief  Iterate over head list and send all spi messages
+ * @param desc - The SPI descriptor.
+ * @param head - First link of the list with messages to be sent
+ * @return SUCCESS in case of success, negativ error code otherwise.
+ */
+int32_t spi_transfer(struct spi_desc *desc,
```
adrimbarean, line 95: I think the name and/or the description should better emphasize that this is a continuous transfer of multiple messages.

drivers/platform/xilinx/xilinx_spi_pl.c
mirceacaprioru: nevermind this just saw that it->buff gets incremented with it->i

drivers/platform/xilinx/xilinx_spi_pl.c
```diff
@@ -0,0 +1,332 @@
...
+	uint32_t tmp;
+	uint32_t i;
+
+	rx_available = _read(xdesc, XSP_RFO_OFFSET) + 1;
+	to_read = min(rx_available, len);
+
+	for (i = 0; i < to_read; ++i) {
+		tmp = _read(xdesc, XSP_DRR_OFFSET);
+		if (data)
+			data[i] = tmp;
```
adrimbarean, line 278: To confirm: Is this check here instead of at the beginning of the function to be able to call the function to empty the receive FIFO?
If yes, I am not sure the function should do that. I think the FIFO should just be allowed to overwrite itself or throw an interrupt that handles buffer overflow.

drivers/platform/xilinx/xilinx_spi_pl.c
mchindri: will rename it

drivers/spi/spi.c
mchindri: This is in spi.c, if a platform don't defines a spi_transfer function, this will be used.
The check is because spi_write_and_read will be used and the buffer to be passed will be the same for write and read.

drivers/platform/xilinx/xilinx_spi_pl.c
mchindri: I will think of something to make the function cleaner

drivers/platform/xilinx/xilinx_spi_pl.c
adrimbarean: Then why not put the check at the beginning to be done only once? Outside of the `for` loop, I mean.

mchindri: can you give me an example, I don't really understand what you mean.

adrimbarean: `	uint32_t i;

	if (!data)
		return 0;

	rx_available = _read(xdesc, XSP_RFO_OFFSET) + 1;`

mchindri: For each byte yo write a new read byte will be added to the fifo, therefor it needs to be discarded otherwise on the next transfer with a read buff, old data will be written.

drivers/spi/spi.c
adrimbarean: I understand due to the other discussion that this is not necessarily the case. Resolved.

mchindri:
TODO: 
if cs change set to 0 on last transfer, keep low until new transfer or writing to another device.
https://github.com/analogdevicesinc/linux/blob/83ce6e3b1894d512c1c41ae0391bad8f46c10343/include/linux/spi/spi.h#L815

include/spi.h
buha: So if I understand this correctly you'd need to declare:
msg1
msg2
msg3

Then do:
SPI_MSG(msg1 parameters, &msg2)
SPI_MSG(msg2 parameters, &msg3)
SPI_MSG(msg3 parameters, NULL)

All this verbosity could dissapear with a simple array like
transfers[] = {
 { msg1 parameters },
 {msg2 parameters },
 {msg3 parameters },
 };
 
I really don't like that the user needs to take care about next, that's internal implementation specifics that should be hidden. If you really want to do it with list, do it with a list, but construct your list from a user-friendly user interface such as an array.

mchindri:
v5:
 - use array of g_spi_desc
 - use array of messages instead of list
 - remove inline from some functions
 - fix typo
 - rename _update to _wr_mask
 - refactor spi_transfer to improve its functionality:
 Ex: If just write buffer: don't read rx
     If just read buffer: send 0xFF
     If multiple messages, send as close as posible one from an other.
 - If CS was not deasserted, and an other CS is needed, assert the new one.
 - rebase onto master

mchindri:
v6:
 - Fix doxygen

scuciurean (approved):

mchindri:
v7:  
 - Initielize to_inc_rx = 0; It wasn't initialized

PR #962, opened 2021-03-10
tools:scripts:build_projects.py:Improve log

mchindri:
Print error messages with red.
Preppend each log line with a arrow (->).
Red if is related to an error message, green otherwise.

### DIFF
--- a/tools/scripts/build_projects.py
+++ b/tools/scripts/build_projects.py
@@ -38,17 +38,28 @@ def parse_input():
 		args.platform, args.build_name)
 
 ERR = 0
+LOG_START = " -> "
+
+def log(msg):
+	print(TGREEN + LOG_START + TWHITE + msg)
+
+def log_err(msg):
+	print(TRED + LOG_START + msg + TWHITE)
+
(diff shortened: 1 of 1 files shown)


mchindri:
Will look like this:
image

And for errors will be red arrows.

tools/scripts/build_projects.py
```diff
@@ -38,17 +38,28 @@ def parse_input():
 		args.platform, args.build_name)
 
 ERR = 0
+LOG_START = " -> "
```
mirceacaprioru, line 41: do we need this ? If there is a an error will they also get the " -> " sign ?

Don't have a strong opinion on it. We can go with it as is.

mirceacaprioru (approved):

tools/scripts/build_projects.py
mchindri: yes, -> will be also on error. 
This will help indentation following the change in color of -> will make easier to find the error.

amiclaus (approved):

PR #964, opened 2021-03-11
Changes done to no-OS to use it as it is. 

pmallick-adi:
So our team has been using drivers from no-OS repository for our application firmware development.
We would like to use the no-OS repository as it is i.e. using it directly without making any changes in our local cloned repository. In order to achieve that we would need the below suggested changes to be implemented in the no-OS repository.
Care has been taken that these changes wouldn’t affect the other projects that are dependent or part of the no-OS repository. 

Below mentioned are the issues that were blocking us from using no-OS directly:  
1)	The ERROR value in the irq_uart_event_e ( in the “irq.h” file inside “no-OS/include” directory ) conflicts with a lot of ST HAL libraries that Mbed is using. 
There ERROR value seems to be generic and hence conflicts with other pre-defined macros that come with the compiler. 

2)	The <sys/types.h> file seems to be GCC specific , hence not supported by IAR and ARM compilers and is used only in “ iio_types.h” and the main reason is to get ‘ssize_t’ typedef.
Thereby creating a dependency on the compiler for the files using it, making it less portable and compiler dependent.

Proposed Solution:
1)	Prefix the irq event names with "IRQ_" as this would be more consistent and match the naming convention used for 'irq_trig_level' enum.

2)	Adding compat.h file to the “no-OS/include” directory which has compiler specific inclusion of the “ssize_t”  and further #including this file to “iio_types.h” file, will makes it accessible to all the files dependent on it.


buha:
This looks good to me, but you'd need to make sure that the checks pass, because it seems the modifications about the ssize_t breaks something.
Click on any red X symbol and `View more on Azure Pipelines` to see the error log, CTRL+F for `error`.

mchindri:
Instead of the second commit it would be better to change the return value to int32_t instead of creating a compat.h file.
The return values came from the tinyiiod submodule which have its own compat.h, I think a cast in iio.c should be enought to solve the issue.
First commit looks good.

pmallick-adi:
> Instead of the second commit it would be better to change the return value to int32_t instead of creating a compat.h file.
> The return values came from the tinyiiod submodule which have its own compat.h, I think a cast in iio.c should be enought to solve the issue.
> First commit looks good.

Are you suggesting to change to return value to ssize_t to int32_t , where it is used ?

mchindri:
> > Instead of the second commit it would be better to change the return value to int32_t instead of creating a compat.h file.
> > The return values came from the tinyiiod submodule which have its own compat.h, I think a cast in iio.c should be enought to solve the issue.
> > First commit looks good.
> 
> Are you suggesting to change to return value to ssize_t to int32_t , where it is used ?

Yes, just in iio.c and iio.h

pmallick-adi:
> > > Instead of the second commit it would be better to change the return value to int32_t instead of creating a compat.h file.
> > > The return values came from the tinyiiod submodule which have its own compat.h, I think a cast in iio.c should be enought to solve the issue.
> > > First commit looks good.
> > 
> > 
> > Are you suggesting to change to return value to ssize_t to int32_t , where it is used ?
> 
> Yes, just in iio.c and iio.h

From the failed logs what I could infer, was there is some conflict between "ssize_t", "size_t" declared in compat.h file and the one present in "sys/types.h" file. 
Also there are other files which happen to use the ssize_t typedef ( example xml.c ) without directly #including "sys/types.h" . "ssize_t" can also be accessed through "stdio.h" as this file #includes <sys/types.h> file, and similarly "size_t" can be accessed through stdio.h" as this file #include <stddef.h> which has this typedef defined at line 209. 

Having said this we were aware of these errors, so to avoid this we did add a macro check for the GCC compiler as mentioned in the image below, which seemed to work on our windows systems with GCC compiler. From the error logs it seems that the macro check in the compat.h file has failed to identify the GCC compiler. 

image

buha:
@mchindri ,
What you propose is a workaround. Even if ssize_t currently it's only used in one place, there is a possibility that it will get used in the future and the workaround won't hold.

@pmallick-adi ,
I think we can move forward with the compat.h as you suggested.
We had a compat.h at some point, lost it because it was kind of useless, but since you use other compilers too (we only use gcc AFAIK) it's good to resurrect compat.h back.
Just please look at my previous comment regarding the failing tests and try to get the tests pass by editing the commits and force pushing back - tests will get triggered again and you will get new results.

pmallick-adi:
> @mchindri ,
> What you propose is a workaround. Even if ssize_t currently it's only used in one place, there is a possibility that it will get used in the future and the workaround won't hold.
> 
> @pmallick-adi ,
> I think we can move forward with the compat.h as you suggested.
> We had a compat.h at some point, lost it because it was kind of useless, but since you use other compilers too (we only use gcc AFAIK) it's good to resurrect compat.h back.
> Just please look at my previous comment regarding the failing tests and try to get the tests pass by editing the commits and force pushing back - tests will get triggered again and you will get new results.

Hello @buha, 
In order to resolve the build tests, I would need help from you to understand the points mentioned below: 

1) Do you include "**libtinyiiod**" repository as a part of the no-OS repository while performing the automated test ? It has been included as a submodule in the no-OS repository, but not sure the if you are including it during the build process. 

2) Are you defining "**_USE_STD_INT_TYPES**" macro anywhere during the build process ?

mchindri:
> > @mchindri ,
> > What you propose is a workaround. Even if ssize_t currently it's only used in one place, there is a possibility that it will get used in the future and the workaround won't hold.
> > @pmallick-adi ,
> > I think we can move forward with the compat.h as you suggested.
> > We had a compat.h at some point, lost it because it was kind of useless, but since you use other compilers too (we only use gcc AFAIK) it's good to resurrect compat.h back.
> > Just please look at my previous comment regarding the failing tests and try to get the tests pass by editing the commits and force pushing back - tests will get triggered again and you will get new results.
> 
> Hello @buha,
> In order to resolve the build tests, I would need help from you to understand the points mentioned below:
> 
> 1. Do you include "**libtinyiiod**" repository as a part of the no-OS repository while performing the automated test ? It has been included as a submodule in the no-OS repository, but not sure the if you are including it during the build process.
> 2. Are you defining "**_USE_STD_INT_TYPES**" macro anywhere during the build process ?

I think this will be a better fix for the issue. https://github.com/analogdevicesinc/no-OS/pull/970/commits/bdd3781f714021884835388e99218be7ed24640d
(Last commit of #970 in case some force push is done)
Can you test it and give as feedback if it solves your use case?
If still some compatibility issues exist, I think it would be good to fix them here: https://github.com/analogdevicesinc/libtinyiiod/blob/master/compat.h

Regarding your questions:
 - We use tinyiiod to build a good part of our noos projects from: https://github.com/analogdevicesinc/no-OS/tree/master/projects
 - Yes, in the build process we always use _USE_STD_INT_TYPES.

pmallick-adi:
> > > @mchindri ,
> > > What you propose is a workaround. Even if ssize_t currently it's only used in one place, there is a possibility that it will get used in the future and the workaround won't hold.
> > > @pmallick-adi ,
> > > I think we can move forward with the compat.h as you suggested.
> > > We had a compat.h at some point, lost it because it was kind of useless, but since you use other compilers too (we only use gcc AFAIK) it's good to resurrect compat.h back.
> > > Just please look at my previous comment regarding the failing tests and try to get the tests pass by editing the commits and force pushing back - tests will get triggered again and you will get new results.
> > 
> > 
> > Hello @buha,
> > In order to resolve the build tests, I would need help from you to understand the points mentioned below:
> > 
> > 1. Do you include "**libtinyiiod**" repository as a part of the no-OS repository while performing the automated test ? It has been included as a submodule in the no-OS repository, but not sure the if you are including it during the build process.
> > 2. Are you defining "**_USE_STD_INT_TYPES**" macro anywhere during the build process ?
> 
> I think this will be a better fix for the issue: Last commit of #970
> Can you test it and give as feedback if it solves your use case?
> If still some compatibility issues exist, I think it would be good to fix them here: https://github.com/analogdevicesinc/libtinyiiod/blob/master/compat.h
> 
> Regarding your questions:
> 
> * We use tinyiiod to build a good part of our noos projects from: https://github.com/analogdevicesinc/no-OS/tree/master/projects
> * Yes, in the build process we always use _USE_STD_INT_TYPES.

@mchindri  The Last commit of [#970](https://github.com/analogdevicesinc/no-OS/pull/970) definitely helps. 
The only change we had to do the compat.h file to include arm compiler definition for ssize_t. 

I would like to know if you are passing the "_USE_STD_INT_TYPES" macro to the automatic build tests ? ( which gets triggered whenever a pull request is made)

buha:
@pmallick-adi 
It looks like _USE_STD_INT_TYPES is **only** defined when building no-OS projects with IIO support.
To me this sounds like we need to move that compat.h from libtinyiiod to no-OS.
Any thoughts on this @dbogdan @cristiantpop ?

pmallick-adi:
> @pmallick-adi
> It looks like _USE_STD_INT_TYPES is **only** defined when building no-OS projects with IIO support.
> To me this sounds like we need to move that compat.h from libtinyiiod to no-OS.
> Any thoughts on this @dbogdan @cristiantpop ?

@buha As mentioned earlier the [build fails](https://dev.azure.com/AnalogDevices/no-OS/_build/results?buildId=2024&view=logs&j=60d1ba4b-5db0-5461-607f-0c2a12524714&t=59f7d250-5a88-5157-974b-6c5f973eca2c&l=2018) as it couldn't find the compat.h file from the libtinyiiod repository.

mchindri:
> > @pmallick-adi
> > It looks like _USE_STD_INT_TYPES is **only** defined when building no-OS projects with IIO support.
> > To me this sounds like we need to move that compat.h from libtinyiiod to no-OS.
> > Any thoughts on this @dbogdan @cristiantpop ?
> 
> @buha I have added compat.h file in order to support uses ssize_t in GCC as well ARM compiler. Have done some changes to the compat.h file for the same. [analogdevicesinc/libtinyiiod#39](https://github.com/analogdevicesinc/libtinyiiod/pull/39)
> 
> As mentioned earlier the [build fails](https://dev.azure.com/AnalogDevices/no-OS/_build/results?buildId=2024&view=logs&j=60d1ba4b-5db0-5461-607f-0c2a12524714&t=59f7d250-5a88-5157-974b-6c5f973eca2c&l=2018) as it couldn't find the compat.h file .

Xml.h and xml.c seem not to be used anymore. I think they can be removed. @analogdevicesinc/noos

mchindri (approved):

buha:
It seems xml.c and xml.h are not used anymore.
However, this is an error on our side, we should've removed xml.c and xml.h files from all src.mk when we did this change 87802a90cc3bc5e5744a5a7004b64ff629998679

@pmallick-adi,
Would you be kind to add a commit to this PR that removes all occurences ?
```
cd no-OS/projects
find . -name "src.mk" -exec sed -i '/xml\.c/d' {} \;
find . -name "src.mk" -exec sed -i '/xml\.h/d' {} \;
git add -u
git commit # perhaps with --ammend
```

pmallick-adi:
> It seems xml.c and xml.h are not used anymore.
> However, this is an error on our side, we should've removed xml.c and xml.h files from all src.mk when we did this change [87802a9](https://github.com/analogdevicesinc/no-OS/commit/87802a90cc3bc5e5744a5a7004b64ff629998679)
> 
> @pmallick-adi,
> Would you be kind to add a commit to this PR that removes all occurences ? With this there will likely be no more errors on automatic tests.
> 
> ```
> cd no-OS/projects
> find . -name "src.mk" -exec sed -i '/xml\.c/d' {} \;
> find . -name "src.mk" -exec sed -i '/xml\.h/d' {} \;
> git add -u
> git commit # perhaps with --ammend
> ```

Sure @buha . I'll do these changes.

buha (approved):
last change looks good, will merge

buha:
Thanks for the contribution @pmallick-adi

PR #980, opened 2021-03-22
stm32: spi implementation with direct reg access

buha:
stm32 SPI implementation used the HAL functions for performing
the SPI transfer and CS toggling (separate, with GPIO, due to
stm32 SPI issue).

This was painfully slow and had to be done by direct register
access.

The new implementation is much faster, probably about 3x from
my measurements.

### DIFF
--- a/drivers/platform/stm32/stm32_spi.c
+++ b/drivers/platform/stm32/stm32_spi.c
@@ -38,6 +38,7 @@
 *******************************************************************************/
 #include <stdlib.h>
 #include <errno.h>
+#include "util.h"
 #include "gpio.h"
 #include "stm32_gpio.h"
 #include "spi.h"
@@ -205,9 +206,11 @@ int32_t stm32_spi_write_and_read(struct spi_desc *desc,
 				 uint8_t *data,
 				 uint16_t bytes_number)
 {
-	int32_t	ret;
+	uint8_t *tx = data;
+	uint8_t *rx = data;
 	struct stm32_spi_desc *sdesc;
-
+	struct stm32_gpio_desc *gdesc;
+	SPI_TypeDef * SPIx;
 	if (!desc || !desc->extra || !data)
 		return -EINVAL;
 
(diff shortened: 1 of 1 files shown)


drivers/platform/stm32/stm32_spi.c
```diff
@@ -205,30 +206,33 @@ int32_t stm32_spi_write_and_read(struct spi_desc *desc,
 				 uint8_t *data,
 				 uint16_t bytes_number)
 {
-	int32_t	ret;
+	uint8_t *tx = data;
+	uint8_t *rx = data;
 	struct stm32_spi_desc *sdesc;
-
+	struct stm32_gpio_desc *gdesc;
```
mchindri, line 212: it seems that gpio is not used anymore so it can be removed from other places too:
For example https://github.com/analogdevicesinc/no-OS/blob/62e4fa56b595be7a6c415f14a48a9202310547a2/drivers/platform/stm32/stm32_spi.h#L68
And set https://github.com/analogdevicesinc/no-OS/blob/62e4fa56b595be7a6c415f14a48a9202310547a2/drivers/platform/stm32/stm32_spi.h#L55
as uint32_t

drivers/platform/stm32/stm32_spi.c
```diff
@@ -205,30 +206,33 @@ int32_t stm32_spi_write_and_read(struct spi_desc *desc,
...
-		else
-			ret = -EIO;
+	gdesc = sdesc->chip_select->extra;
+	SPIx = sdesc->hspi.Instance;
+
+	gdesc->port->BSRR = BIT(sdesc->chip_select->number) << 16;
+	__HAL_SPI_ENABLE(&sdesc->hspi);
+	while(bytes_number--) {
+		while(!(SPIx->SR & SPI_SR_TXE))
+			;
```
mirceacaprioru, line 228: nitpick: why the extra line.

drivers/platform/stm32/stm32_spi.c
```diff
@@ -205,30 +206,33 @@ int32_t stm32_spi_write_and_read(struct spi_desc *desc,
...
+		*(volatile uint8_t *)&SPIx->DR = *tx++;
+		while(!(SPIx->SR & SPI_SR_RXNE))
+			;
```
mirceacaprioru, line 231: same nitpick here

drivers/platform/stm32/stm32_spi.c
buha: Actually it is still used.
I use the STM32 HAL to initialize the GPIO but then use the 'desc->extra->chip_select' to access the registers of the gpio.

drivers/platform/stm32/stm32_spi.c
buha: for clarity,
```
while(!(SPIx->SR & SPI_SR_TXE));
// next line
```
when reading the above code, one could think // next line is under while (easy to visually miss the semicolon)

mchindri (approved):

PR #983, opened 2021-03-23
Ad7689 update

buha:
- fix issue visible when sequencer was enabled
- update openocd and gdb scripts (unrelated to driver but used them to test the chip with fast CPU and APB clocks)


mirceacaprioru (approved):

mirceacaprioru:
i don't think the fail is related to this PR but it is worth checking and making sure.

buha:
Yeah, looks unrelated, I rebased branch on updated master, errors should disappear.

amiclaus (approved):

PR #987, opened 2021-03-26
drivers:frequency:adf5355: initial commit

amiclaus:
Add no-OS support for ADF5355, ADF4355, ADF4355-2, ADF4355-3, ADF5356.

More info: https://www.analog.com/media/en/technical-documentation/data-sheets/ADF5355.pdf

Note: `adf5355_clk_recalc_rate`, `adf5355_clk_set_rate` and `adf5355_clk_round_rate` function were added to align the driver with the linux version and ease the integration with certain projects.

### DIFF
--- a/drivers/frequency/adf5355/adf5355.c
+++ b/drivers/frequency/adf5355/adf5355.c
@@ -1,0 +1,529 @@
+/***************************************************************************//**
+ *   @file   adf5355.c
+ *   @brief  Implementation of adf5355 Driver.
+ *   @author Antoniu Miclaus (antoniu.miclaus@analog.com)
+********************************************************************************
+ * Copyright 2021(c) Analog Devices, Inc.
+ *
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *  - Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ *  - Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in
+ *    the documentation and/or other materials provided with the
+ *    distribution.
+ *  - Neither the name of Analog Devices, Inc. nor the names of its
+ *    contributors may be used to endorse or promote products derived
+ *    from this software without specific prior written permission.
+ *  - The use of this software may or may not infringe the patent rights
+ *    of one or more patent holders.  This license does not release you
+ *    from the requirement that you obtain separate licenses from these
+ *    patent holders to use this software.
+ *  - Use of the software either in source or binary form, must be run
+ *    on or directly connected to an Analog Devices Inc. component.
+ *
+ * THIS SOFTWARE IS PROVIDED BY ANALOG DEVICES "AS IS" AND ANY EXPRESS OR
+ * IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, NON-INFRINGEMENT,
+ * MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED.
+ * IN NO EVENT SHALL ANALOG DEVICES BE LIABLE FOR ANY DIRECT, INDIRECT,
+ * INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, INTELLECTUAL PROPERTY RIGHTS, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+*******************************************************************************/
+
+/******************************************************************************/
+/***************************** Include Files **********************************/
+/******************************************************************************/
+#include <inttypes.h>
+#include <stdio.h>
+#include "error.h"
+#include <malloc.h>
+#include "delay.h"
+#include "util.h"
+#include "adf5355.h"
+
+/******************************************************************************/
+/************************** Functions Implementation **************************/
+/******************************************************************************/
+
+/**
+ * SPI register write to device.
+ * @param dev - The device structure.
+ * @param reg - The register address.
+ * @param data - The register data.
+ *              NOTE: if raw data is writtend to the device and not using the
+ *              predefined register bits masks, a 4 bit shif to the right
+ *              is required, since the 4 bits of each register correspond
+ *              to the register address.
+ * @return SUCCESS in case of success, negative error code otherwise.
+ */
+static int32_t adf5355_write(struct adf5355_dev *dev,
+			     uint8_t reg_addr,
+			     uint32_t data)
+{
+	uint8_t buf[ADF5355_SPI_NO_BYTES];
+	data = data | reg_addr;
+
+	buf[0] = data >> 24;
+	buf[1] = data >> 16;
+	buf[2] = data >> 8;
+	buf[3] = data;
+
+	return spi_write_and_read(dev->spi_desc, buf, ARRAY_SIZE(buf));
+}
+
+/**
+ * Compute PLL parameters.
+ * @param vco - The VCO frequency.
+ * @param pfd - The PFD frequency.
+ * @param integer - The integer division factor.
+ * @param fract1 - The fractionality.
+ * @param fract2 - The auxiliary fractionality.
+ * @param mod2 - The auxiliary modulus.
+ * @return None.
+ */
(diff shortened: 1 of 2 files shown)


amiclaus:
V2:
- add REG10 write in the frequency update sequence for recommended V_tune calibration.

drivers/frequency/adf5355/adf5355.c
```diff
@@ -0,0 +1,518 @@
...
+
+/******************************************************************************/
+/************************** Functions Implementation **************************/
+/******************************************************************************/
+
+/**
+ * SPI register write to device.
+ * @param dev - The device structure.
+ * @param reg - The register address.
+ * @param val - The register data.
```
adrimbarean, line 59: Doxygen might flag if names are not the same in the comments and code.

drivers/frequency/adf5355/adf5355.c
```diff
@@ -0,0 +1,518 @@
...
+ * @return SUCCESS in case of success, negative error code otherwise.
+ */
+static int32_t adf5355_write(struct adf5355_dev *dev,
+			     uint8_t reg_addr,
+			     uint32_t data)
+{
+	uint8_t buf[ADF5355_SPI_NO_BYTES];
+	data = data | reg_addr;
```
adrimbarean, line 67: This seems a bit odd. Is `data` supposed to be in a specific format that (i think shifted by 8 bits in this case)? If yes, it would be good to document it, but it might not be necessary if you just set do:
`buff[0] = data >> 16;
 buff[1] = data >> 8;
 buff[2] = data;
 buff[3] = reg_addr;`

drivers/frequency/adf5355/adf5355.c
```diff
@@ -0,0 +1,518 @@
...
+	}
+
+	gcd_div = greatest_common_divisor(*fract2, *mod2);
+	*mod2 /= gcd_div;
+	*fract2 /= gcd_div;
+}
+
+/**
+ * ADF5355 Register configuration
+ * @param dev - The device structure.
```
adrimbarean, line 119: nitpick: sync_all not documented. Not too bad, but might help with maintenance in the future.

drivers/frequency/adf5355/adf5355.c
amiclaus: the reg_addr is only 4 bits, the rest of 4 are data bytes.
I know this is a strange approach, but I used it because there were remarks on other similar drivers that the read/write function signature should comply to the IIO support.
Same thing happens here:
https://github.com/analogdevicesinc/no-OS/blob/ca092f47e8ed60e7fb94b265fdcf27da12b6c3d9/drivers/frequency/adf5902/adf5902.c#L64

drivers/frequency/adf5355/adf5355.c
```diff
@@ -0,0 +1,518 @@
...
+ * @param rounded_rate - The closest possible rate of desired rate.
+ * @return SUCCESS in case of success, negative error code otherwise.
+ */
+int32_t adf5355_clk_round_rate(struct adf5355_dev *dev, uint64_t rate,
+			       uint64_t *rounded_rate)
+{
+	*rounded_rate = rate;
+
+	return SUCCESS;
+}
```
adrimbarean, line 340: Is this necessary?

amiclaus: I mentioned in the note of the PR:

> Note: `adf5355_clk_recalc_rate`, `adf5355_clk_set_rate` and `adf5355_clk_round_rate` function were added to align the driver with the linux version and ease the integration with certain projects.

See another example: 
https://github.com/analogdevicesinc/no-OS/blob/ca092f47e8ed60e7fb94b265fdcf27da12b6c3d9/drivers/frequency/adf4371/adf4371.c#L709-L721

drivers/frequency/adf5355/adf5355.c
amiclaus: It's strange that doxygen did not catch this issue. Will solve.

drivers/frequency/adf5355/adf5355.h
```diff
@@ -0,0 +1,295 @@
...
+ * INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, INTELLECTUAL PROPERTY RIGHTS, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+*******************************************************************************/
+
+/* Register Definition */
+#define ADF5355_REG(x)                          (x)
```
adrimbarean, line 41: Is this necessary?

drivers/frequency/adf5355/adf5355.c
amiclaus: You are right. Will fix.

drivers/frequency/adf5355/adf5355.h
amiclaus: I oscillated between this approach and defining each register address:
```
#define ADF5355_REG0 0x0
#define ADF5355_REG1 0x1
#define ADF5355_REG2 0x2
........
```
I preferred reducing the lines of code written, but keeping the code readability

amiclaus:
V3:
- fix doxygen issues mentioned above in comments.

drivers/frequency/adf5355/adf5355.c
adrimbarean: Understood. So is there any specific documentation about the format of the parameters, for example data or register address needing to be shifted prior to passing them to the function? Assuming the user does not use the defines in the header.

amiclaus: The defines (register bits masks) are handling this data shifting. So if the customer will use our driver as it is, there shouldn't be any problems.
But I agree on documenting the data/register address handling. But not sure where it should be done. Maybe in the function header comment?

amiclaus:
V4:
- add additional information in the function header comment of `adf5355_write` about the required composition of the input parameter `data`.

drivers/frequency/adf5355/adf5355.c
amiclaus: Added a note in the function header comment.

drivers/frequency/adf5355/adf5355.c
```diff
@@ -0,0 +1,523 @@
...
+ * @param rate - Channel rate.
+ * @return SUCCESS in case of success, negative error code otherwise.
+ */
+int32_t adf5355_clk_set_rate(struct adf5355_dev *dev, uint32_t chan,
+			     uint64_t rate)
+{
+	if (chan >= dev->num_channels)
+		return FAILURE;
+
+	return adf5355_set_freq(dev, rate, chan);
```
cristiantpop, line 329: Probably the logic inside `adf5355_set_freq` should be moved inside `adf5355_clk_round_rate`, and only some registers settings should take place in `adf5355_set_freq` function, otherwise the output `rounded_rate` of function `adf5355_clk_round_rate` and the actual value set with `adf5355_clk_set_rate` might be different.
This functions should be used together like in this case:
  https://github.com/analogdevicesinc/no-OS/blob/1365adbebfef34edb32a5ce89dce603f1bb93451/projects/adrv9009/src/app/app_clocking.c#L459
`clk_round_rate` called first to determine valid rate
https://github.com/analogdevicesinc/no-OS/blob/1365adbebfef34edb32a5ce89dce603f1bb93451/projects/adrv9009/src/app/app_clocking.c#L487
`clk_set_rate` second to set the determined valid rate

drivers/frequency/adf5355/adf5355.c
```diff
@@ -0,0 +1,523 @@
...
+ * @param dev - The device structure.
+ * @param reg - The register address.
+ * @param data - The register data.
+ *              NOTE: if raw data is writtend to the device and not using the
+ *              predefined register bits masks, a 4 bit shif to the right
+ *              is required, since the 4 bits of each register correspond
+ *              to the register address.
+ * @return SUCCESS in case of success, negative error code otherwise.
+ */
+static int32_t adf5355_write(struct adf5355_dev *dev,
```
PLG-AMC, line 66: I think you forgot to add something very important, the option to enable or disable the chip select before sending the command. By the way, shouldn't there be a mask for each case ? I mean, buff[0] = (data & 0x00FFFFFF) and so on?

Thanks

drivers/frequency/adf5355/adf5355.c
amiclaus: This looks like a valid approach to me.
The only isssue I see is the alignment with other clock chip drivers on no-OS/linux which use more or less the current approach:
https://github.com/analogdevicesinc/no-OS/tree/master/drivers/frequency/adf4371
https://github.com/analogdevicesinc/linux/blob/master/drivers/iio/frequency/adf5355.c
Even for hmc7044, you showed an example of usage of the driver, but if we have a look at the driver implementation in the no-OS repository, the frequency parameters computation is done mainly in the `setup` function

adrimbarean (approved):
Looks good to me.

amiclaus:
V5:
- make sure that the channel used to generate output frequency (via `set_freq()`) is enabled.

amiclaus:
V6:
- fix typo introduced with latest modifications

PR #988, opened 2021-03-26
[2] Small improvements in CI

mchindri:
Will be rebase after #985 is merged (ignore first 3 commits)

### DIFF
--- a/tools/scripts/build_projects.py
+++ b/tools/scripts/build_projects.py
@@ -14,14 +14,14 @@ TWHITE = '\033[39m' #Withe text
 description_help='''Build noos projects
 Examples:\n
 	Build all noos projects
-    	>python build_projects.py ..\.. export_dir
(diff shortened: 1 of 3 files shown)


tools/scripts/build_projects.py
```diff
@@ -54,7 +54,11 @@ def log_success(msg):
 log_file = DEFAULT_LOG_FILE
 
 def run_cmd(cmd):
-	log(cmd)
+	tmp = cmd.split(' ')
```
buha, line 57: This commit "tools:scripts:build_projects.py: Improve log" could be merged with the previous commit that adds logging to have logging "improved" from the first shot :)

mchindri: They are to separated things. One moves logs to files and this one just make pretty the normal log

mchindri:
V2: 
 - Remove last commit. A server link won't be used anymore.

PR #991, opened 2021-03-29
Fix cces build

mchindri:


buha (approved):
xsdk/cces could be launched as separate process but let's investigate that later

tested this and works fine !

PR #993, opened 2021-03-30
[1] Add builds.json files for some projects

mchindri:

PR #995, opened 2021-03-30
[2] Add remaing builds.json files and fix projects

mchindri:
Will rebase onto #993 when merged


cristiantpop (approved):

mchindri:
v2:
 - Fix astyle

mchindri:
v3:
 - Add remaining projects.
 - Edit some commits titles

mchindri:
All builds seems to run in almost 8 minutes :)

projects/adrv9009/builds.json
```diff
@@ -0,0 +1,25 @@
...
+                  "hardware": [
+                        "adrv9009_zc706",
+                        "adrv9009_zcu102"
+                  ]
+            },
+            "dma_example": {
+                  "flags" : "NEW_CFLAGS=-DDAC_DMA_EXAMPLE",
+                  "hardware": [
+                        "adrv9009_zc706",
+                        "adrv9009_zcu102"
```
buha, line 14: this project can run on more combinations of hardware but it's ok as-is, we can extend later on

projects/adrv9001/builds.json
```diff
@@ -0,0 +1,40 @@
+{
+      "xilinx": {
+            "demo": {
+                  "flags": "",
+                  "hardware": [
+                        "adrv9001_zc706",
+                        "adrv9001_zcu102",
+                        "adrv9001_zcu102_lvds",
+                        "adrv9001_zed"
```
buha, line 9: We do have HDL that has support for ZED + ADRV9001 but the no-OS project was never updated to support that.
zc706 is also with question mark.

mchindri: Will be update or should be removed from the build?

mchindri: removed zed and zc706 for the moment

PR #996, opened 2021-03-30
drivers: rename amplifiers section

amiclaus:
`instr-amplif` -> `amplifiers`

Move all drivers to the new section.

Use same naming as on linux.


mirceacaprioru:
-> make update_srcs
 -> make all
 -> ERROR
 -> See log logs/ada4250_ardz_aducm3029_simple.txt -- Use cat (linux) or type (windows) to see colored output

Why does the build script no longer print the error logs ?

amiclaus:
V2:
- adapt makefile to new driver path

mirceacaprioru (approved):

PR #1000, opened 2021-04-02
Remove old transfer_dev_to_mem and update ad7134 to use iio_app

mchindri:
Remove old transfer_dev_to_mem function and use new one. Then remove it from iio.
Add support for register access.
Use iio_app

### DIFF
--- a/iio/iio_ad9361/iio_ad9361.c
+++ b/iio/iio_ad9361/iio_ad9361.c
@@ -2296,10 +2296,6 @@ int32_t iio_ad9361_init(struct iio_ad9361_desc **desc,
 	iio_ad9361_inst->dev_descriptor.attributes = global_attributes;
 	iio_ad9361_inst->dev_descriptor.debug_attributes = NULL;
 	iio_ad9361_inst->dev_descriptor.buffer_attributes = NULL;
-	iio_ad9361_inst->dev_descriptor.transfer_dev_to_mem = NULL;
-	iio_ad9361_inst->dev_descriptor.transfer_mem_to_dev = NULL;
-	iio_ad9361_inst->dev_descriptor.read_data = NULL;
-	iio_ad9361_inst->dev_descriptor.write_data = NULL;
 	*desc = iio_ad9361_inst;
 
 	return SUCCESS;
--- a/iio/iio_app/iio_app.c
+++ b/iio/iio_app/iio_app.c
@@ -36,6 +36,9 @@
  * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
  * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 *******************************************************************************/
+
+#ifdef IIO_SUPPORT
+
 #include <stdlib.h>
 #include "iio.h"
 #include "iio_app.h"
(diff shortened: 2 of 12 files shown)


mchindri:
v2:
 - Fix CI problem

dbogdan (approved):

dbogdan:
By the way, was this tested on the real hardware?

mchindri:
> By the way, was this tested on the real hardware?

Yes, but I have some issue 

> By the way, was this tested on the real hardware?

Yes, but there is still a issue here:  https://github.com/analogdevicesinc/no-OS/blob/a26ec68e988a748973d30fd2f6850b4e9913d80d/drivers/adc/ad713x/iio_dual_ad713x.c#L123
It has different behavior than https://github.com/analogdevicesinc/no-OS/blob/a26ec68e988a748973d30fd2f6850b4e9913d80d/projects/ad713x_fmcz/src/ad713x_fmc.c#L277
I asked @cristiantpop to take a look at the PR maybe I am missing something.

iio/iio_ad9361/iio_ad9361.c
```diff
@@ -2296,10 +2296,6 @@ int32_t iio_ad9361_init(struct iio_ad9361_desc **desc,
 	iio_ad9361_inst->dev_descriptor.attributes = global_attributes;
 	iio_ad9361_inst->dev_descriptor.debug_attributes = NULL;
 	iio_ad9361_inst->dev_descriptor.buffer_attributes = NULL;
-	iio_ad9361_inst->dev_descriptor.transfer_dev_to_mem = NULL;
```
amiclaus, line 2299: It looks that you are doing here changes also to other projects that are unrelated to ad7134.

libraries/iio/iio.c
```diff
@@ -819,12 +819,6 @@ static uint32_t bytes_to_samples(struct iio_interface *intf, uint32_t bytes)
 static ssize_t iio_transfer_dev_to_mem(const char *device, size_t bytes_count)
 {
 	struct iio_interface *iio_interface = iio_get_interface(device);
-
-	if (iio_interface->dev_descriptor->transfer_dev_to_mem)
```
amiclaus, line 823: Shouldn't library changes be subject for another PR? Since these changes might influence also other projects than ad7134.

iio/iio_app/iio_app.c
```diff
@@ -36,6 +36,9 @@
  * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
  * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 *******************************************************************************/
+
+#ifdef IIO_SUPPORT
```
amiclaus, line 40: another change that has to do with more than ad7134.

mchindri:
v3: 
 - Fix _iio_ad713x_read_dev. No it works fine. (tested on hardware)

libraries/iio/iio.c
mchindri: Only this projects was still using this functions. The scope was to remove them so the project has to be updated. 
Anyway, the changes are in separated commits.

iio/iio_app/iio_app.c
mchindri: it is a commit unrelated to ad7134.

amiclaus: I was a bit confused about the PR title.

mchindri: My mistake, I changed it :)

amiclaus (approved):

PR #1001, opened 2021-04-02
drivers: amplifiers: ada4230 support

amiclaus:
Pin strap version of ADA4250.

Gain and Reference buffer are controlled via GPIOs.

Calibration features not available.

### DIFF
--- a/drivers/amplifiers/ada4250/ada4250.c
+++ b/drivers/amplifiers/ada4250/ada4250.c
@@ -60,6 +60,9 @@ int32_t ada4250_write(struct ada4250_dev *dev, uint8_t reg_addr,
 {
 	uint8_t buff[ADA4250_BUFF_SIZE_BYTES];
 
+	if(dev->device_id != ADA4250)
+		return -ENODEV;
+
 	buff[0] = ADA4250_SPI_WRITE_CMD | reg_addr;
 	buff[1] = data;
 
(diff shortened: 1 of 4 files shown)


drivers/amplifiers/ada4250/ada4250.c
```diff
@@ -272,29 +314,62 @@ int32_t ada4250_init(struct ada4250_dev **device,
...
+
+		ret = gpio_get_optional(&dev->gpio_g1, init_param->gpio_g1_param);
+		if (ret != SUCCESS)
+			goto error_g2;
+
+		ret = gpio_direction_output(dev->gpio_g1, GPIO_LOW);
+		if (ret != SUCCESS)
+			goto error_g1;
+
+		ret = gpio_get_optional(&dev->gpio_g0, init_param->gpio_g0_param);
```
mchindri, line 358: If you use gpio_get_optional shouldn't later check every time if gpio_g0 is not NULL?

amiclaus: Actually, `gpio_get` will fit better here, since the gpios are mandatory for ada4230.

amiclaus (commented):
V2:
- use `gpio_get` instead of `gpio_get_optional`

mchindri (approved):

dbogdan (approved):

PR #1005, opened 2021-04-06
aducm i2c bug fix: bitrate/address may not be init

buha:
Bitrate/Address may not be initialized when last_address (global)
or last_bitrate (global) are equal to the new address/bitrate.

This change uses impossible values for address or bitrate to "mark"
them as uninitialized on driver removal.

---------------
For a bit more context, address 0xff is not possible on classical I2C (0-127 addresses only) and a bitrate of 0 doesn't make sense, hence those values.

### DIFF
--- a/drivers/platform/aducm3029/i2c.c
+++ b/drivers/platform/aducm3029/i2c.c
@@ -47,6 +47,9 @@
 #include <drivers/i2c/adi_i2c.h>
 #include <drivers/gpio/adi_gpio.h>
 
+#define UNINITIALIZED_BITRATE 0
+#define UNINITIALIZED_ADDRESS 0xFF
+
 /******************************************************************************/
 /*************************** Types Declarations *******************************/
 /******************************************************************************/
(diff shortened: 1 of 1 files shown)


drivers/platform/aducm3029/i2c.c
```diff
@@ -164,6 +166,8 @@ int32_t i2c_remove(struct i2c_desc *desc)
 		i2c_handler = NULL;
 	}
 	free(desc);
+	last_address = UNINITIALIZED_ADDRESS;
```
mchindri, line 169: This makes sense to be inside: if (nb_created_desc == 0)

drivers/platform/aducm3029/i2c.c
```diff
@@ -65,10 +65,12 @@ static uint8_t adi_i2c_buffer[ADI_I2C_MEMORY_SIZE]
 __attribute__((aligned (4)));
 
 /** Save the current state of the bitrate to not change it each time */
-static uint32_t last_bitrate = 0;
+#define UNINITIALIZED_BITRATE 0
```
mchindri, line 68: styling: I would put defines together upper in the file, not mixed with variables

drivers/platform/aducm3029/i2c.c
buha: yeah true, fixed it, funny how i knew it's supposed to go there and placed it outside :)))

drivers/platform/aducm3029/i2c.c
buha: i moved them somewhere up

buha:
V2:
- implement the two suggestions above

mchindri (approved):

dbogdan (approved):

PR #1006, opened 2021-04-07
drivers:platform:aducm3029: i2c read fix

adrimbarean:
Setting bRepeatedStart=true for a transaction does not only ommit the stop
bit of the current transaction, but actually sends the repeated start as
well as a prologue data sequence.
This means that calling a read without a
stop bit and thn a write for aducm3029 actually sends the reapeated start,
then the I2C address and then a Stop bit directly, meaning the write
operation is not atomic anymore.
At best this is a benign bug and might still work, with hardware errors and
reading wrong registers at worst.

Capture of the write operation with bRepeatedStart=true
image
Image credit: @buha 

Capture of a full register read after the fix.
image

Bug discovered by @buha.

### DIFF
--- a/drivers/platform/aducm3029/i2c.c
+++ b/drivers/platform/aducm3029/i2c.c
@@ -46,6 +46,7 @@
 #include <stdlib.h>
 #include <drivers/i2c/adi_i2c.h>
 #include <drivers/gpio/adi_gpio.h>
+#include <string.h>
 
 #define UNINITIALIZED_BITRATE 0
 #define UNINITIALIZED_ADDRESS 0xFF
@@ -72,6 +73,17 @@ static uint32_t last_bitrate = UNINITIALIZED_BITRATE;
 
 /** Save the current slave_address to not change it each time */
 static uint8_t last_address = UNINITIALIZED_ADDRESS;
+
+/**
+ * @struct i2c_extra
+ * @brief I2C platform specific parameters.
+ */
+struct aducm_i2c_extra {
+	/** Pointer to the prologue data. */
+	uint8_t *prologue_data;
+	/** Prologue size. */
+	uint8_t prologue_size;
+};
 
 /******************************************************************************/
 /************************ Functions Definitions *******************************/
@@ -117,6 +129,8 @@ static uint32_t set_transmission_configuration(struct i2c_desc *desc)
 int32_t i2c_init(struct i2c_desc **desc,
 		 const struct i2c_init_param *param)
 {
+	struct aducm_i2c_extra *aducm_i2c;
+
 	if (!desc || !param)
 		return -EINVAL;
 
@@ -144,9 +158,10 @@ int32_t i2c_init(struct i2c_desc **desc,
 		}
 	}
 
+	aducm_i2c = calloc(1, sizeof(*aducm_i2c));
 	(*desc)->max_speed_hz = param->max_speed_hz;
 	(*desc)->slave_address = param->slave_address;
-	(*desc)->extra = NULL;
+	(*desc)->extra = aducm_i2c;
 	nb_created_desc++;
 
 	return SUCCESS;
@@ -159,6 +174,8 @@ int32_t i2c_init(struct i2c_desc **desc,
  */
 int32_t i2c_remove(struct i2c_desc *desc)
 {
+	struct aducm_i2c_extra *aducm_i2c = desc->extra;
+
 	if (!desc)
 		return -EINVAL;
 	nb_created_desc--;
@@ -168,6 +185,9 @@ int32_t i2c_remove(struct i2c_desc *desc)
 		last_address = UNINITIALIZED_ADDRESS;
 		last_bitrate = UNINITIALIZED_BITRATE;
 	}
+	if (aducm_i2c->prologue_data)
+		free(aducm_i2c->prologue_data);
+	free(aducm_i2c);
 	free(desc);
 
 	return SUCCESS;
@@ -188,12 +208,15 @@ int32_t i2c_write(struct i2c_desc *desc,
 		  uint8_t bytes_number,
 		  uint8_t stop_bit)
 {
+	struct aducm_i2c_extra *aducm_i2c = desc->extra;
+
 	if (!desc)
 		return -EINVAL;
 
 	ADI_I2C_TRANSACTION trans[1];
 	uint32_t errors;
 	int32_t ret;
+	uint8_t *temp_ptr = NULL;
 
 	ret = set_transmission_configuration(desc);
 	if (ret < 0)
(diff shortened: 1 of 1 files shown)


drivers/platform/aducm3029/i2c.c
```diff
@@ -206,7 +211,17 @@ int32_t i2c_write(struct i2c_desc *desc,
 		return SUCCESS;
 	}
 
-	trans->bRepeatStart = (stop_bit == 1) ? 0 : 1;
+	if (stop_bit == 0) {
+		prologue_size = bytes_number;
+		prologue_data = malloc(bytes_number);
```
buha, line 216: So, the change works on the assumption that the `write(stop_bit=0)` followed by a `read(stop_bit=1)`.
I think we should somehow enforce this workflow, nu sure how exactly. But if we don't enforce it, and the user performs a `write(stop_bit=0)` followed by a `write(stop_bit=0)`, you'll end up with a memory leak (that global pointer allocated again) and possibly other side effects.

I otherwhise agree with the change, also tried it in ideal conditions and behaves as expected.

drivers/platform/aducm3029/i2c.c
```diff
@@ -73,6 +74,10 @@ static uint32_t last_bitrate = UNINITIALIZED_BITRATE;
 /** Save the current slave_address to not change it each time */
 static uint8_t last_address = UNINITIALIZED_ADDRESS;
 
+/* Data to send in prologue */
+static uint8_t *prologue_data = NULL;
```
mchindri, line 78: I would add this inside desc.

drivers/platform/aducm3029/i2c.c
mchindri: I have 2 proposals:
 - create a static function for write/read in order to support stop bit also for read  and not duplicate code.
    Regarding the  memory leak, if multiple calls with stop_bit = 0 are executed then, maybe realloc can be used or just overwrite the previous prolog freeing the pointer. Memory can be deallocated in i2c_close if no free was called
 - Create a i2c_readwrite function with prolog which can be used to read a register and return -EINVAL because aducm is not directly supporting stop bit.

drivers/platform/aducm3029/i2c.c
adrimbarean: i2c does not have an extra header. Do we want to add it just for this?

adrimbarean:
V1: 
 - realloc if write with stop_bit=0 is called more than once consecutively;
 - mark the stop_bit option in i2c_read as irrelevant in documentation;
 - free prologue memory in i2c_remove if the transaction was never fulfilled.

drivers/platform/aducm3029/i2c.c
```diff
@@ -167,6 +172,8 @@ int32_t i2c_remove(struct i2c_desc *desc)
 		i2c_handler = NULL;
 		last_address = UNINITIALIZED_ADDRESS;
 		last_bitrate = UNINITIALIZED_BITRATE;
+		if (prologue_data)
+			free(prologue_data);
```
buha, line 176: Given it's a global I think it should be also set to NULL after freeing.

drivers/platform/aducm3029/i2c.c
mchindri: you can declare here aducm_i2c_desc and add to i2c_desc->extra.
You don't need an extra file

drivers/platform/aducm3029/i2c.c
```diff
@@ -206,7 +214,22 @@ int32_t i2c_write(struct i2c_desc *desc,
 		return SUCCESS;
 	}
 
-	trans->bRepeatStart = (stop_bit == 1) ? 0 : 1;
+	if (stop_bit == 0) {
+		prologue_size = bytes_number;
+		if (prologue_data) {
+			temp_ptr = realloc(prologue_data, bytes_number);
```
mchindri, line 220: need to check realloc return

drivers/platform/aducm3029/i2c.c
```diff
@@ -244,14 +267,26 @@ int32_t i2c_read(struct i2c_desc *desc,
...
+
 	trans->pData = data;
 	trans->nDataSize = bytes_number;
 	trans->bReadNotWrite = 1;
-	if (ADI_I2C_SUCCESS != adi_i2c_ReadWrite(i2c_handler, trans, &errors))
-		return -EIO;
+	ret = adi_i2c_ReadWrite(i2c_handler, trans, &errors);
 
-	return SUCCESS;
+	if (prologue_size != 0) {
```
mchindri, line 285: add this cleanup on i2c_remove too

drivers/platform/aducm3029/i2c.c
mchindri: Can a realloc(prologue_data, bytes_number + old_prologue_size) and then
memcpy(prologue_data + old_proloque_size, data, bytes_number) be a better approach?
This will enable write(stop_bit = 0) write(stop_bit = 0) and then read(stop_bit = 1).

mchindri (commented):
This approach works only for:
write(stop_bit = 0) and then read(stop_bit = 1)
Is for example read(stop_bit = 0), write(stop_bit = 1)  a possible case ? shouldn't it be treated too?

adrimbarean:
> This approach works only for:
> write(stop_bit = 0) and then read(stop_bit = 1)
> Is for example read(stop_bit = 0), write(stop_bit = 1) a possible case ? shouldn't it be treated too?

The reason of the repeated start is to write the address of the register the user wants to read and then read it without releasing the bus so that another master on the line does not change the register again and cause bugs (as Mark [posted in Slack](https://analogdevicessdg.slack.com/archives/CRW7W4J7J/p1618409690129900)). Moreover some old, very orthodox devices might clear the address register when the bus is open. But there is no case (that I know of) I which you want to read with a repeated start. IMO we don't need to cover that case, just make sure it does not cause a glitch.

drivers/platform/aducm3029/i2c.c
adrimbarean: Should be done [here](https://github.com/analogdevicesinc/no-OS/pull/1006/files#diff-8c18182bade514b4db3b40669c1670e0da5e78446b38f209375c20083431a4b6R175).

drivers/platform/aducm3029/i2c.c
adrimbarean: Will do.

drivers/platform/aducm3029/i2c.c
adrimbarean: We can do that, but do we want to? What would be the reason to do this instead of just overwriting?

buha (approved):
The globals could be moved to desc->extra but this not a big issue since aducm3029 has only 1 i2c anyway

I'm approving this because the other method (generate a stop condition in between write and read) is not supported by current chips such as ad7746, not just older ones.

adrimbarean:
V2:
 - add global prologue variables to an extra structure;
 - check return value of realloc.

adrimbarean:
V3:
 - fix CI errors

PR #1008, opened 2021-04-08
ADA4250 updates

amiclaus:
- remove offset range current values
- remove range define
- rework range/bias defines


drivers/amplifiers/ada4250/ada4250.c
```diff
@@ -203,7 +203,7 @@ static int32_t ada4250_set_range(struct ada4250_dev *dev,
 		return -ENODEV;
 
 	return ada4250_update(dev, ADA4250_REG_SNSR_CAL_CNFG, ADA4250_RANGE_SET_MSK,
-			      ADA4250_RANGE_SET(ADA4250_RANGE(range)));
+			      ADA4250_RANGE_SET(range));
```
mchindri, line 206: Nitpick: I think this commit should be squashed with previous. Otherwise incorrect values can be set between them.

mchindri (approved):

buha (approved):

PR #1012, opened 2021-04-09
drivers:amplifiers:ada4250: fix polarity meaning

amiclaus:
According to the datasheet, 0 corresponds to negative offset values and
1 corresponds to positive offset values. 

image

### DIFF
--- a/drivers/amplifiers/ada4250/ada4250.c
+++ b/drivers/amplifiers/ada4250/ada4250.c
@@ -283,10 +283,10 @@ int32_t ada4250_set_offset(struct ada4250_dev *dev, int32_t offset)
 
 	if (offset < 0) {
 		dev->offset_uv = (-1) * offset_raw * vlsb;
-		return ada4250_write(dev, ADA4250_REG_SNSR_CAL_VAL, (1 << 8 | abs(offset_raw)));
+		return ada4250_write(dev, ADA4250_REG_SNSR_CAL_VAL, offset_raw);
 	} else {
 		dev->offset_uv = offset_raw * vlsb;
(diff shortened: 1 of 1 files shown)


drivers/amplifiers/ada4250/ada4250.c
```diff
@@ -283,10 +283,10 @@ int32_t ada4250_set_offset(struct ada4250_dev *dev, int32_t offset)
 
 	if (offset < 0) {
 		dev->offset_uv = (-1) * offset_raw * vlsb;
-		return ada4250_write(dev, ADA4250_REG_SNSR_CAL_VAL, (1 << 8 | abs(offset_raw)));
+		return ada4250_write(dev, ADA4250_REG_SNSR_CAL_VAL, offset_raw);
```
mchindri, line 286: I would sugest in order to make possible negative offset too
```suggestion
                val = offset > 0 ? 1 << 8 : 0;
                val |= abs(offset_raw);
		return ada4250_write(dev, ADA4250_REG_SNSR_CAL_VAL, val);
```
you can reuse val for the second call.

amiclaus: sure, i can do it like that

amiclaus: but I need to store also the computed value for the offset in uV:
`dev->offset_uv = (-1) * offset_raw * vlsb;`
or
dev->offset_uv = * offset_raw * vlsb;

and there is no `second call`
it's either offset < 0 or offset >= 0

mchindri (approved):

dbogdan (approved):

PR #1017, opened 2021-04-14
drivers:platform:aducm3029: i2c read fix

adrimbarean:
Setting bRepeatedStart=true for a transaction does not only ommit the stop
bit of the current transaction, but actually sends the repeated start as
well as a prologue data sequence.
This means that calling a read without a
stop bit and thn a write for aducm3029 actually sends the reapeated start,
then the I2C address and then a Stop bit directly, meaning the write
operation is not atomic anymore.
At best this is a benign bug and might still work, with hardware errors and
reading wrong registers at worst.

This is a variant solution to the #1006 

### DIFF
--- a/drivers/platform/aducm3029/i2c.c
+++ b/drivers/platform/aducm3029/i2c.c
@@ -255,3 +255,43 @@ int32_t i2c_read(struct i2c_desc *desc,
 
 	return SUCCESS;
 }
+
+/**
+ * @brief Read data after writing with repeated start (useful for
+ *        reading registers).
+ * @param desc - Descriptor of the I2C device
+ * @param write_data - Buffer that stores the data to be written.
+ * @param wr_bytes_number - Number of bytes to write.
+ * @param read_data - Buffer that stores the data that will be read.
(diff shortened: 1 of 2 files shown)


adrimbarean:
V1:
 - solve astyle and doxygen problems

drivers/platform/aducm3029/i2c.c
```diff
@@ -255,3 +255,43 @@ int32_t i2c_read(struct i2c_desc *desc,
...
+int32_t i2c_writeread(struct i2c_desc *desc,
+		      uint8_t *write_data,
+		      uint8_t wr_bytes_number,
+		      uint8_t *read_data,
+		      uint8_t rd_bytes_number)
+{
+	if (!desc)
+		return -EINVAL;
+
+	ADI_I2C_TRANSACTION trans[1];
```
scuciurean, line 278: Why initialize as array if it has one element?

buha:
After some discussions, this approach is not good because it impacts all platforms.
I have approved the other method #1006 and closing this one.

PR #1018, opened 2021-04-15
projects: adrv9008-1/adrv9008-2 support

amiclaus:
Add support for ADRV9008-1 and ADRV9008-2.

### DIFF
--- a/projects/adrv9009/builds.json
+++ b/projects/adrv9009/builds.json
@@ -7,6 +7,20 @@
                         "adrv9009_zcu102"
                   ]
             },
+	    "adrv9008-1": {
+		"flags": "NEW_CFLAGS=-DADRV9008_1",
+		"hardware": [
+		      "adrv9009_zc706",
+		      "adrv9009_zcu102"
+		]
+	    },
+	    "adrv9008-2": {
+		"flags": "NEW_CFLAGS=-DADRV9008_2",
+		"hardware": [
+		      "adrv9009_zc706",
+		      "adrv9009_zcu102"
+		]
+	    },
             "dma_example": {
                   "flags" : "NEW_CFLAGS=-DDAC_DMA_EXAMPLE",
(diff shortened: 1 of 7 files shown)


scuciurean:
Isn't this forcing you to use either tx or rx?
I would use `#ifdef ADRV9008_1`/`#ifdef ADRV9008_1` and for adrv9009 define both as active.

amiclaus (commented):
> 
> 
> Isn't this forcing you to use either tx or rx?
> I would use `#ifdef ADRV9008_1`/`#ifdef ADRV9008_2` and for adrv9009 define both as active.

I used #ifndefs so that the defafult configuration remains the same. 

As stated also here:
https://github.com/analogdevicesinc/no-OS/blob/b7c03e86ed8dbf1cb7c276eb933cfe74ac4b746b/projects/adrv9009/src/app/app_config.h#L53-L55

Not changing anything with respect to the defines, will run the application as it is. (for adrv9009)

scuciurean:
> > Isn't this forcing you to use either tx or rx?
> > I would use `#ifdef ADRV9008_1`/`#ifdef ADRV9008_2` and for adrv9009 define both as active.
> 
> I used #ifndefs so that the defafult configuration remains the same.
> 
> As stated also here:
> 
> https://github.com/analogdevicesinc/no-OS/blob/b7c03e86ed8dbf1cb7c276eb933cfe74ac4b746b/projects/adrv9009/src/app/app_config.h#L53-L55
> 
> Not changing anything with respect to the defines, will run the application as it is. (for adrv9009)

Oh, you are right

scuciurean (approved):

dbogdan:
initCalMask needs to be updated too.

amiclaus:
V2:
- update `initCalMask`

dbogdan (approved):

PR #1020, opened 2021-04-19
Linux userspace makefile and configuration for debugging with VSCode

mchindri:
Beside building and debugging: 
Problems are configured to be shown in the problems tab in order to go to them in one click
image

### DIFF
--- a/tools/scripts/linux.mk
+++ b/tools/scripts/linux.mk
@@ -1,0 +1,13 @@
+CC = gcc
+
+BINARY		= $(BUILD_DIR)/$(PROJECT_NAME).out
+
+PLATFORM_RELATIVE_PATH = $1
+PLATFORM_FULL_PATH = $1
+
+CFLAGS +=  -g3 \
+		-DLINUX_PLATFORM \
+
+$(PROJECT_TARGET):
+	$(MUTE) $(call mk_dir, $(BUILD_DIR)) $(HIDE)
+	$(MUTE) $(call set_one_time_rule,$@)
--- a/tools/scripts/generic.mk
+++ b/tools/scripts/generic.mk
@@ -308,9 +308,13 @@ $(OBJECTS_DIR)/%.o: $$(call get_full_path, %).S | $$(@D)/.
 	@$(call print,[AS] $(notdir $<))
 	$(MUTE) $(AS) -c $(ASFLAGS) $< -o $@
 
+ifneq ($(strip $(LSCRIPT)),)
+LSCRIPT_FLAG = -T$(LSCRIPT) $(LDFLAGS)
+endif
+
 $(BINARY): $(LIB_TARGETS) $(OBJS) $(ASM_OBJS) $(LSCRIPT)
 	@$(call print,[LD] $(notdir $(OBJS)))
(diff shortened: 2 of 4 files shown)


tools/scripts/platform/linux/.vscode/c_cpp_properties.json
```diff
@@ -0,0 +1,19 @@
+{
+    "configurations": [
```
buha, line 2: As much as I like the idea of being able to debug graphically, I think we shouldn't maintain ourselves config files for third-party IDE's. We do it for XSDK because we use Xilinx SDK. We do it for CrossCore because we use ADUCM etc. but VSCode is just an arbitrary choice that no-OS users might not agree with. On top of this, VSCode is going to evolve and somebody will need to maintain the files in this commit, do we want that ?

mchindri: I think is better than nothing. If some don't like to debug with VSCode, make is still available without it.
It seems that I forgot to add a make run :)

mchindri: Regarding the maintainance, maybe we can write some where it was tested with version x.x.x.

buha: It's not better than nothing - we already support 3 IDE's, why should we have a 4th if there is no need ?
There is no standard IDE for the linux platform. In fact, I think most people on that platform prefer vim + plugins.

tools/scripts/platform/linux/.vscode/launch.json
```diff
@@ -0,0 +1,16 @@
+{ 
+    "version": "0.2.0", 
+    "configurations": [
+        {
+            "name": "(gdb) Launch", 
+            "type": "cppdbg", 
+            "request": "launch",
+            "miDebuggerPath": "/usr/bin/gdb", 
+            "externalConsole": false, 
+            "targetArchitecture": "arm",
```
scuciurean, line 10: This won't work for microblaze

tools/scripts/linux.mk
```diff
@@ -0,0 +1,16 @@
...
+PLATFORM_FULL_PATH = $1
+
+CFLAGS +=  -ggdb \
+		-DLINUX_PLATFORM \
+
+$(PROJECT_TARGET):
+	$(MUTE) $(call mk_dir, $(BUILD_DIR)) $(HIDE)
+	$(MUTE) $(call mk_dir, $(PROJECT_BUILD)) $(HIDE)
+	$(MUTE) $(call copy_folder, $(PLATFORM_TOOLS)/.vscode, $(PROJECT_BUILD)/.vscode) $(HIDE)
+	$(MUTE) python $(PLATFORM_TOOLS)/config_build.py $(NO-OS) $(PROJECT) $(BINARY) $(HIDE)
```
scuciurean, line 15: What happends if the host has a very dated python version? For example 2.7

mchindri:
V2:
-Remove debug commit from this PR

dbogdan (approved):

PR #1022, opened 2021-04-21
drivers: pwm: axi_pwmgen: Add offset and multiple channel support

scuciurean:


scuciurean:
Changelog -> V2:
- Fix axi_io name typo

include/pwm.h
```diff
@@ -127,6 +131,14 @@ int32_t pwm_set_duty_cycle(struct pwm_desc *desc,
 int32_t pwm_get_duty_cycle(struct pwm_desc *desc,
 			   uint32_t *duty_cycle_ns);
 
+/* Set duty cycle of PWM generator device */
```
mchindri, line 134: Fix comment. Next function too

include/pwm.h
```diff
@@ -70,6 +70,8 @@ struct pwm_init_param {
 	uint32_t period_ns;
 	/** PWM generator duty cycle */
 	uint32_t duty_cycle_ns;
+	/** PWM generator offset */
```
mchindri, line 73: Maybe you can add more in the comment about what the offset is, I'am just understood reading the commit description.

drivers/axi_core/axi_pwmgen/axi_pwm.c
```diff
@@ -68,6 +68,28 @@
...
+/**
+ * @brief Write the pwmgen axi register using a bitmask
+ *
+ * @param [in] base - Device baseaddress
+ * @param [in] offset - Device register offset
+ * @param [in] mask - Bitmask used to write
+ * @param [in] data - Data to write in register
+ * @return \ref SUCCESS in case of success, \ref FAILURE otherwise.
+ */
+int32_t axi_pwmgen_write_mask(uint32_t base, uint32_t offset, uint32_t mask,
```
mchindri, line 87: Shouldn't this be static?

include/pwm.h
mchindri: Anyway. Isn't this a feature that should be in extra?
For example, an MCU can generate a PMW using timers, I don't know if there would be a reliable way to synchronize all the timers to have a desired offset between different PWMs.

drivers/axi_core/axi_pwmgen/axi_pwm.c
```diff
@@ -228,6 +247,15 @@ int32_t pwm_init(struct pwm_desc **desc,
 
 	axi_init = param->extra;
 
+	ret = axi_io_read(axi_desc->base_addr,
+			  AXI_PWMGEN_REG_NPWM,
+			  &data);
```
mchindri, line 293: it seems data is not used.

drivers/axi_core/axi_pwmgen/axi_pwm.c
```diff
@@ -228,6 +247,15 @@ int32_t pwm_init(struct pwm_desc **desc,
...
+	if (ret != SUCCESS)
+		return FAILURE;
+
+	if (axi_init->channel > (ret - 1))
```
mchindri, line 297: shouldn't data be instead of ret?

include/pwm.h
scuciurean: If the controller supports polarity, you can use for example inverted polarity and the duty cycle width will represent a time offset relative to the moment when the pwm was enabled

scuciurean: Or for example you are using two timers to generate an interrupt based pwm. If they both start counting from 0, the rising edge will be alligned, but if you set a different value inside the first timer's counter, you will get a phase shift between the signals

scuciurean:
Changelog -> V3:
- Fix NPWM register read in the init function

drivers/axi_core/axi_pwmgen/axi_pwm.c
```diff
@@ -68,6 +68,29 @@
...
+				     uint32_t data)
+{
+	int32_t ret;
+	uint32_t temp;
+
+	ret = axi_io_read(base, offset, &temp);
+	if (ret != 0)
+		return ret;
+
+	return axi_io_write(base, offset, (temp & ~mask) | data);
```
buha, line 98: I think data should also be masked, otherwise you could be setting bits outside of the mask
```
(temp & ~mask) | (data & mask)

File: include/pwm.h [hunk 7]
```diff
@@ -131,11 +131,11 @@ int32_t pwm_set_duty_cycle(struct pwm_desc *desc,
 int32_t pwm_get_duty_cycle(struct pwm_desc *desc,
 			   uint32_t *duty_cycle_ns);
 
-/* Set duty cycle of PWM generator device */
+/* Set the offset of PWM generator device */
```
buha, line 134: make sure to merge this commit with one of the previous commits that changes the same files

include/pwm.h
buha: I'm also a bit confused, is this a phase offset ? If yes, you could write it like that, then it's much more clear

scuciurean: I agree, the 'offset' naming came out of the hdl registers. Will change

scuciurean:
Changelog-> V4:
- Rename 'offset' to 'phase'

drivers/axi_core/axi_pwmgen/axi_pwm.c
```diff
@@ -68,6 +68,29 @@
...
+				     uint32_t data)
+{
+	int32_t ret;
+	uint32_t temp;
+
+	ret = axi_io_read(base, offset, &temp);
+	if (ret != 0)
+		return ret;
+
+	return axi_io_write(base, offset, (temp & ~mask) | (data & mask));
```
adrimbarean, line 98: Is it worth it to use `find_first_set_bit` from `util` to shift data before applying it to the register? This saves the user from having to shift it.

adrimbarean (approved):
Looks good to me. One comment, but can approve like this too.

buha (approved):

buha:
I think you can remove the draft label

scuciurean:
> I think you can remove the draft label

Still waiting for the HDL to be merged first

drivers/axi_core/axi_pwmgen/axi_pwm.c
cristiantpop: Other implementations are similar to this, with shift from the user. It will be confusing to have two implementations

PR #1026, opened 2021-04-21
drivers:Makefile: Fix make clean

mchindri:
With find, ignored directories where also added to SRCS and they
can't and shoulden't be removed.

### DIFF
--- a/drivers/Makefile
+++ b/drivers/Makefile
@@ -41,4 +41,4 @@ OBJS = $(SRCS:.c=.o)
 all: $(OBJS)
 
 clean:
-	rm -f $(OBJS)
+	rm -f *.o


amiclaus (changes requested):
Still the object files created with `make` are not removed with `make clean`.

mchindri:
v2:
 - really fix make clean :)

amiclaus:
> 
> 
> v2:
> 
>     * really fix make clean :)

Not really :) It looks like the build is now failing.
Was `CC = arm-none-eabi-gcc` removed on purpose?

mchindri:
v3:
 - Fix again

amiclaus (approved):
Works now 🎉

PR #1028, opened 2021-04-22
Some fixes

mchindri:
Is drivers/platform/linux/platform_drivers still used?


tools/scripts/linux.mk
```diff
@@ -11,3 +11,6 @@ CFLAGS +=  -g3 \
 $(PROJECT_TARGET):
 	$(MUTE) $(call mk_dir, $(BUILD_DIR)) $(HIDE)
 	$(MUTE) $(call set_one_time_rule,$@)
+
+linux_run: $(BINARY)
+	$(BINARY)
```
amiclaus, line 16: nitpick:
newline

drivers/platform/linux/linux_gpio.c
```diff
@@ -117,14 +120,26 @@ int32_t linux_gpio_get(struct gpio_desc **desc,
...
 	}
 
 	sprintf(path, "/sys/class/gpio/gpio%d/value", descriptor->number);
-	linux_desc->value_fd = open(path, O_WRONLY);
+	timeout = GPIO_TIMEOUT_MS;
+	while (--timeout) {
+		linux_desc->value_fd = open(path, O_WRONLY);
+		if (linux_desc->value_fd >= 0)
+			break;
+		mdelay(1);
```
amiclaus, line 141: is the delay required?

amiclaus (commented):
+ build issues

drivers/platform/linux/linux_gpio.c
mchindri: I would say yes.
If I set a timeout without delay it will depend on the CPU speed.
I tested without delay I think 1000 times and it sometimes don't work.

mchindri:
v2:
 - Add new line
 - Fix astyle

drivers/platform/linux/linux_gpio.c
amiclaus: Then I believe a comment would be helpful pointing out why delay was added.
This approach I believe is used in other drivers also, if not specified in the datasheet.

mchindri:
v3:
 - Edit commit of gpio_get fix

drivers/platform/linux/linux_gpio.c
mchindri: I edited the commit message

amiclaus (approved):

drivers/platform/linux/linux_spi.c
```diff
@@ -189,11 +189,44 @@ int32_t linux_spi_remove(struct spi_desc *desc)
...
+
+{
+	struct spi_ioc_transfer *tr;
+	struct linux_spi_desc	*linux_desc;
+	int			ret;
+	uint32_t		i;
+
+	linux_desc = desc->extra;
+
+	tr = (struct spi_ioc_transfer *)calloc(3, sizeof(*tr));
```
adrimbarean, line 204: Is this freed anywhere?

tools/scripts/linux.mk
```diff
@@ -11,3 +11,6 @@ CFLAGS +=  -g3 \
 $(PROJECT_TARGET):
 	$(MUTE) $(call mk_dir, $(BUILD_DIR)) $(HIDE)
 	$(MUTE) $(call set_one_time_rule,$@)
+
+linux_run: $(BINARY)
```
adrimbarean, line 15: Why not just run?

adrimbarean:
Could you change the name to something more suggestive?

tools/scripts/linux.mk
mchindri: when running `make run` this rule will be called

mchindri:
> Could we change the name to something more suggestive?

Which name?

tools/scripts/linux.mk
buha: not related to this commit but yeah I'm also wondering why this was implemented with [platform]_run: in generic,mk when you could simply have a run: rule in the [platform].mk

drivers/platform/linux/linux_uart.c
```diff
@@ -105,7 +105,7 @@ int32_t uart_init(struct uart_desc **desc, struct uart_init_param *param)
 	linux_init = param->extra;
 
 	ret = snprintf(path, sizeof(path), "/dev/%s", linux_init->device_id);
-	if (ret < 0 || ret >= sizeof(path)) {
+	if (ret < 0 || ret >= (int)sizeof(path)) {
```
buha, line 108: let's try to use explicit size for types everywhere (if possible)
here, it's possible, you know what type ret is

mchindri: I don't really understand what do you want me to do :)

tools/scripts/linux.mk
mchindri: I though was nice to search for the rules in one file (generic.mk)

mchindri:
v4:
 - free allocated memory

drivers/platform/linux/linux_spi.c
mchindri: done

PR #1029, opened 2021-04-22
Aducm make debug

buha:
As a side effect, this PR enables debugger flags for all aducm builds, this should be fine, we do it anyway on all platforms and we don't have debug/release separation.

### DIFF
--- a/tools/scripts/aducm.mk
+++ b/tools/scripts/aducm.mk
@@ -106,18 +106,17 @@ GENERIC_FLAGS += -DCORE0 -D_RTE_ -D__ADUCM3029__ -D__SILICON_REVISION__=0xffff -
 GENERIC_DEBUG_FLAGS = -g -gdwarf-2 -D_DEBUG
 
 GENERIC_RELEASE_FLAGS = -DNDEBUG 
-C_RELEASE_FLAGS = -O2
 
 CFLAGS += -DADUCM_PLATFORM
 
 ASFLAGS	+= $(GENERIC_FLAGS) $(PLATFORM_INCS) -x assembler-with-cpp 
 CFLAGS	+= $(GENERIC_FLAGS) -Wall -ffunction-sections -fdata-sections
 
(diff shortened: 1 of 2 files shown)


tools/scripts/aducm.mk
```diff
@@ -162,6 +162,24 @@ aducm3029_run: all
...
+	@echo tb main >> $(BINARY).gdb
+	@echo c >> $(BINARY).gdb
+	@echo refresh >> $(BINARY).gdb
+
+.PHONY: debug
+debug: all $(BINARY).gdb
+	(openocd -s "$(OPENOCD_SCRIPTS)" -f interface/cmsis-dap.cfg \
+		-s "$(ADUCM_DFP)/openocd/scripts" -f target/aducm3029.cfg \
+		-c "init" &);
+	/usr/bin/arm-none-eabi-gdb -tui --command=$(BINARY).gdb || \
```
mchindri, line 180: this doesn't seems to be a path that work on windows :)
arm-none-eabi-gdb should be enough

buha: look at the PR  comment, it's explained there

buha: the problem is that the one that comes with the toolchain doesn't provide -tui, so I put this in-place until they wil ship one with the -tui option
it won't fail on windows

buha: So actually i realised it will fail on windows but not for that reason.
& (launch as separate process) and || (or operator) are actually bash operators, this is what will not work on windows after all :) I am changing the commit message accordingly.

I think we should add this stuff gradually - first have make debug on linux, then extend to windows.
Is that ok with you ?

buha:
V2:
- changed commit message
- remove optimizations - they make the debugger jump weirdly in C code - perhaps we could think to add debug/release separation at some point in no-OS

buha:
V3:
- use gdb's `tui enable` command instead of cmdline switch for stm32

adrimbarean (approved):
Looks good to me.

PR #1031, opened 2021-04-23
[cn0552 p1] ad7746 driver updates

buha:
AD7746 driver updates

Requires #1006 to be merged to work.

### DIFF
--- a/drivers/adc/ad7746/ad7746.c
+++ b/drivers/adc/ad7746/ad7746.c
@@ -378,7 +378,7 @@ static inline int32_t _ad7746_write_2byte(struct ad7746_dev *dev, uint8_t reg,
 	if (!dev)
 		return -EINVAL;
 
-	buf[0] = (uint8_t)(val >> 16);
+	buf[0] = (uint8_t)(val >> 8);
 	buf[1] = (uint8_t)val;
 
 	return ad7746_reg_write(dev, reg, buf, 2);
--- a/drivers/adc/ad7746/ad7746.c
+++ b/drivers/adc/ad7746/ad7746.c
@@ -465,6 +465,9 @@ int32_t ad7746_get_vt_data(struct ad7746_dev *dev, uint32_t *vt_data)
 		   ((uint32_t)dev->buf[1] << 8) |
 		   dev->buf[0];
 
+	if (dev->setup.config.md == AD7746_MODE_SINGLE)
+		dev->setup.config.md = AD7746_MODE_IDLE;
+
 	return SUCCESS;
 }
 
@@ -503,6 +506,9 @@ int32_t ad7746_get_cap_data(struct ad7746_dev *dev, uint32_t *cap_data)
 	*cap_data = ((uint32_t)dev->buf[0] << 16) |
 		    ((uint32_t)dev->buf[1] << 8) |
 		    dev->buf[0];
+
+	if (dev->setup.config.md == AD7746_MODE_SINGLE)
+		dev->setup.config.md = AD7746_MODE_IDLE;
 
 	return SUCCESS;
 }
--- a/drivers/adc/ad7746/ad7746.c
+++ b/drivers/adc/ad7746/ad7746.c
@@ -69,6 +69,8 @@ int32_t ad7746_init(struct ad7746_dev **device,
 	if (ret < 0)
 		goto error_1;
 
+	dev->id = init_param->id;
+
 	ret = ad7746_reset(dev);
 	if (ret < 0)
 		goto error_2;
--- a/drivers/adc/ad7746/ad7746.c
+++ b/drivers/adc/ad7746/ad7746.c
@@ -512,3 +512,47 @@ int32_t ad7746_get_cap_data(struct ad7746_dev *dev, uint32_t *cap_data)
 
 	return SUCCESS;
 }
+
+/***************************************************************************//**
+ * @brief Perform offset/gain calibration
+ *
+ * @param dev - Device descriptor pointer.
+ * @param md - AD7746 calibration mode specifier.
+ *
+ * @return return code.
+ *         Example: -EINVAL - Wrong input values.
+ *                  -EIO - I2C Communication error.
+ *                  SUCCESS - No errors encountered.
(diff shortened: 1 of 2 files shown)


projects/cn0552/src/app/headless.c
```diff
@@ -0,0 +1,100 @@
...
+
+#include <sys/platform.h>
+#include "adi_initialize.h"
+#include <drivers/pwr/adi_pwr.h>
+
+#include "iio.h"
+#include "iio_ad7746.h"
+#include "iio_app.h"
+#include "ad7746.h"
+#include "irq.h"
```
amiclaus, line 51: I believe you don't need all these includes like irq, irq_extra, uart, uart_extra if you are using iio_app.

amiclaus (commented):
General comment:
In my opinion there are too many commits for one Pull Request.
I would do separate Pull Requests and add them progressively.
1. AD7746 fixes
2. IIO updates
3. AD7746 IIO driver + CN0552 project.

projects/cn0552/src/app/headless.c
buha: yeah i agree about too many commits, there many changes
will see about those includes, maybe i can cut some out

drivers/adc/ad7746/ad7746.h
```diff
@@ -182,6 +182,8 @@ struct ad7746_init_param {
 struct ad7746_dev {
 	i2c_desc *i2c_dev;
 	uint8_t buf[AD7746_NUM_REGISTERS + 1u];
+	uint8_t capdac[2][2];
```
mchindri, line 185: Can 2 and 2 be a define?

include/util.h
```diff
@@ -111,6 +111,8 @@
 
 #define UNUSED_PARAM(x) ((void)x)
 
+#define shift_right(x, s) ((x) < 0 ? -(-(x) >> (s)) : (x) >> (s))
```
mchindri, line 114: If the variable is signed, isn't this done by default?

libraries/iio/iio.c
```diff
@@ -284,27 +293,6 @@ static ssize_t iio_phy_write(const char *buf, size_t len)
 	return -EINVAL;
 }
 
-/* Get string for channel id from channel type */
-static char *get_channel_id(enum iio_chan_type type)
```
mchindri, line 288: Function using get_channel_id should be updated in this commit

libraries/iio/iio_types.h
```diff
@@ -51,6 +51,18 @@
 /*************************** Types Declarations *******************************/
 /******************************************************************************/
 
+
+enum iio_val {
```
mchindri, line 55: Where is this used?
It seems not related to this commit

libraries/iio/iio.c
```diff
@@ -296,19 +296,19 @@ static ssize_t iio_phy_write(const char *buf, size_t len)
...
+				sprintf(buff, "%s%d-%s%d", iio_chan_type_string[ch->ch_type],
+					(int)ch->channel, iio_chan_type_string[ch->ch_type],
 					(int)ch->channel2);
 			else
-				sprintf(buff, "%s%d", get_channel_id(ch->ch_type),
+				sprintf(buff, "%s%d", iio_chan_type_string[ch->ch_type],
 					(int)ch->channel);
 		} else {
-			sprintf(buff, "%s", get_channel_id(ch->ch_type));
+			sprintf(buff, "%s", iio_chan_type_string[ch->ch_type]);
```
mchindri, line 311: Below changes are related with a previous commit

libraries/iio/iio_types.h
```diff
@@ -161,6 +167,8 @@ struct iio_channel {
 	 *  for angular rate when applied to channel2 will make make the
 	 *  IIO_ANGL_VEL have anglvel_x which corresponds to the x-axis. */
 	int 			channel2;
+	/** Driver specific identifier. */
+	unsigned long		address;
```
mchindri, line 171: Is the linux naming?
Doesn't it make more sense to be void *context or something like that?

include/util.h
buha: no, nothing guarantees you that

libraries/iio/iio_types.h
buha: i don't think changing names will improve anything - it will actually make it worse, every time you wanna port an iio driver implementation between linux <-> no-OS you need to think about which corresponds to which

libraries/iio/iio.c
buha: of course, i did it in a different commit, will try to move changes around

buha:
V2:
- implemented feedback above (moved commits around, remove useless includes)

buha:
V3:
- split into multiple PRs

include/util.h
mchindri: It seems to be compiler dependent (https://en.wikipedia.org/wiki/Bitwise_operations_in_C#Right_shift_%3E%3E) 
Good to know :)

drivers/adc/ad7746/ad7746.c
```diff
@@ -504,5 +509,49 @@ int32_t ad7746_get_cap_data(struct ad7746_dev *dev, uint32_t *cap_data)
...
+	    md != AD7746_MODE_GAIN_CALIB)
+		return -EINVAL;
+
+	c->md = md;
+	ret = ad7746_set_config(dev, *c);
+	if (ret < 0)
+		return ret;
+
+	do {
+		mdelay(20);
```
amiclaus, line 545: is this a datasheet related value? or trial/error value? might fit a code comment here :)

buha: yeah it needs a comment !

adrimbarean (approved):
Technically the last commit can go in the PR that will use it, but approve.

buha:
V4: 
- add chip id
- comment that delay
- remove capdac stuff from here -> got moved to iio part of the driver

mchindri (approved):

dbogdan (approved):

PR #1034, opened 2021-04-26
[cn0552 p3] ad7746 iio driver + cn0552 project

buha:
CN0552 board uses the ad7746 cdc chip.
Create a project for it and the iio part of the driver.


drivers/adc/ad7746/iio_ad7746.c
```diff
@@ -0,0 +1,714 @@
...
+
+static const uint8_t ad7746_cap_filter_rate_table[][2] = {
+	{91, 11 + 1}, {84, 12 + 1}, {50, 20 + 1}, {26, 38 + 1},
+	{16, 62 + 1}, {13, 77 + 1}, {11, 92 + 1}, {9, 110 + 1},
+};
+
+static int ad7746_store_cap_filter_rate_setup(struct ad7746_dev *chip,
+		int val)
+{
+	unsigned int i;
```
amiclaus: maybe use `stdint` typedefs here.
Check entire code.

buha: good point, completely forgot about this

buha:
V2:
- iio part of the driver has now an init/remove function
- perform some additional stuff at init (capdac, chip id)

drivers/adc/ad7746/iio_ad7746.c
buha: so i've done some modifications in this regard but couldn't do all of them because in someplaces i used int because existing iio_types.h uses types like that. I think in those cases we should chean iio up. Until then I fixed only the independent int / unsigned int occurences.

buha:
V3:
- IIO_CAPACITANCE moved completely to part3 (here)

mchindri (approved):

drivers/adc/ad7746/iio_ad7746.h
```diff
@@ -0,0 +1,62 @@
...
+
+#ifndef IIO_AD7746_H
+#define IIO_AD7746_H
+
+#include "iio.h"
+
+struct ad7746_iio_dev {
+	struct ad7746_dev *ad7746_dev;
+	struct iio_device *iio_dev;
+	// capdac[0] - first channel, capdac[1] - second channel
```
amiclaus: are these lines still required?

buha:
V4:
- rebase on master

libraries/iio/iio_types.h
```diff
@@ -73,8 +73,7 @@ enum iio_chan_type {
 	IIO_ALTVOLTAGE,
 	IIO_ANGL_VEL,
 	IIO_TEMP,
-	/* All new types must be added before this field */
-	IIO_LAST_TYPE
+	IIO_CAPACITANCE
```
scuciurean, line 76: Wasn't IIO_LAST_TYPE there for a reason? Also, i would move this change to a separate commit

buha: no reason that I could see, completely unused.

drivers/adc/ad7746/iio_ad7746.h
buha: you mean the comments ? i would say yes, they help

buha:
V4:
- fixed typo caliscale -> calibscale

amiclaus (approved):

PR #1041, opened 2021-05-05
iio: moved iio folders and iio.c

ranechita:
Moved iio device files and folders in the coresponding driver
And moved iio.c and iio.h from libraries back to no-Os iio.

Signed-off-by: Ramona Alexandra Nechita <ramona.nechita@analog.com>


mchindri (commented):
Move iio_types.h to and remove libraries/iio/Makefile which is not used anymore

amiclaus (commented):
Since this draft involves modifications on multiple projects, I'd suggest either renaming the branch using the `dev/` prefix (example: https://github.com/analogdevicesinc/no-OS/tree/dev/adpd410x_iio) or converting the draft to an actual PR. In this manner, the build jobs (build `Projects`) will be triggered and give you a quick insight on the correctness of the changes.

ranechita:
v2

- rebased master
- moved iio_types and deleted libraries/iio/Makefile
- updated iio comments

iio/iio.c
```diff
@@ -1247,6 +1247,8 @@ static uint32_t iio_generate_device_xml(struct iio_device *device, char *name,
  * @param dev_descriptor - Device descriptor
  * @param name - Name to identify the registered device
  * @param dev_instance - Opened instance of the device
+ * @param read_buff - read buffer
```
amiclaus, line 1250: The changes in this file should be subject for another commit.

amiclaus (commented):
it looks like "iio_types.h" was completely removed alongside the Makefile

iio/iio.c
ranechita: I was thinking it fits in the same commit because they were just small changes so the build doesn't fail, but I can make a separate commit just with fixes in general

amiclaus (approved):
looks good now.

PR #1044, opened 2021-05-05
redefine SUCCESS and FAILURE if necessary

buha:
It makes sense for SUCCESS/FAILURE to be redefined using our own
(no-OS) convention rather than mix things up.

One currently known scenario is the aducm3029 platform which
defines FAILURE as 1, contrary to our -1 expectation. This leads
to bad behaviour.

### DIFF
--- a/include/error.h
+++ b/include/error.h
@@ -46,12 +46,14 @@
 /********************** Macros and Constants Definitions **********************/
 /******************************************************************************/
 
-#ifndef SUCCESS
+#ifdef SUCCESS
+#undef SUCCESS
+#endif
 #define SUCCESS		0
+#ifdef FAILURE
+#undef FAILURE
 #endif
-#ifndef FAILURE
 #define FAILURE		-1
-#endif
 
 #ifndef __ELASTERROR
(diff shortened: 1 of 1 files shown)


amiclaus (commented):
Will these new definitions be used also in the BSP drivers that were developed around their own standardized defines? I'm just thinking if this approach might cause problems outside of our scope.

buha:
I'm also wondering about the impact of this.
From what I can tell so far, yes - the BSP will use our new definition of SUCCESS/FAILURE. The BSPs that define SUCCESS/FAILURE are stm32 and aducm3029. For stm32, the definitions are identical - no issue. In the case of aducm3029 the definition of SUCCESS is identical, but FAILURE=1 instead of our -1. In my trials on aducm3029 this change improves things instead of breaking.

amiclaus (approved):
Since the drivers are platform agnostic they might be used with other platforms than xilinx, altera, aducm3029, where issues might appear. Approving PR for now, we will see if complains will appear or not in the future 💣  .

adrimbarean (approved):

PR #1055, opened 2021-07-09
Add parameterized JTAG filtering to PS init

tfcollins:
This change allows more advanced filtering of the JTAG targets using the
jtag_cable_name filter. This is required when you have multiple JTAG
cables connected to the same machine. Not providing the argument
JTAG_CABLE_ID will fall back to the standard behavior.

This is required for the hardware testing hardness.

Example usage:
```
make run JTAG_CABLE_ID=210299A567FE
```

This has been tested with and without this new argument with ADRV9361-Z7035 boards and the Xilinx SDK 2019.1.

Signed-off-by: Travis F. Collins <travis.collins@analog.com>

### DIFF
--- a/tools/scripts/generic.mk
+++ b/tools/scripts/generic.mk
@@ -10,6 +10,7 @@ export AR
 export EXTRA_INC_PATHS
 export FLAGS_WITHOUT_D
 export PROJECT_BUILD
+export JTAG_CABLE_ID
 #------------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


buha (approved):
please sign the CLA, it's a formality all no-OS contributors have signed

this is a very welcome addition, i thought about the need of this some time ago :)

amiclaus (approved):

PR #1056, opened 2021-07-14
Add adpd410x support for reading a large number of bytes from the FIFO at once

NSagan271:
Add function to adpd410x.c to read a specified number of bytes from the
FIFO using the minimum number of register reads. Also add support for
reading a specified number of bytes (up to 255) from a register with one
i2c or spi read.

When reading large amounts of data from the FIFO (on the order of 100 - 512 bytes),
this greatly improves efficiency, as opposed to reading one packet from the FIFO
at a time.

### DIFF
--- a/drivers/photo-electronic/adpd410x/adpd410x.c
+++ b/drivers/photo-electronic/adpd410x/adpd410x.c
@@ -62,37 +62,74 @@ int32_t adpd410x_reg_read(struct adpd410x_dev *dev, uint16_t address,
 			  uint16_t *data)
 {
 	int32_t ret;
-	uint8_t buff[] = {0, 0, 0, 0};
+	uint8_t buff[] = {0, 0};
+
+	ret = adpd410x_reg_read_bytes(dev, address, (uint8_t *) buff, 2);
+	if (ret != SUCCESS) {
+		return ret;
+	}
+
+	*data = ((uint16_t)buff[0] << 8) & 0xff00;
+	*data |= buff[1] & 0xff;
+
+	return SUCCESS;
+}
+
+/**
+ * @brief Read a specified number of bytes from device register.
+ * @param dev - Device handler.
+ * @param address - Register address.
+ * @param data - Pointer to the register value container.
+ * @param num_bytes - number of bytes to read. Max 255 for ADPD4101 (I2C).
+ * @return SUCCESS in case of success, FAILURE otherwise.
+ */
+int32_t adpd410x_reg_read_bytes(struct adpd410x_dev *dev, uint16_t address,
+				uint8_t *data, uint16_t num_bytes)
+{
+	int32_t ret;
+	uint8_t i;
+	uint8_t *buff;
 
 	switch (dev->dev_type) {
 	case ADPD4100:
+		buff = (uint8_t *) calloc(num_bytes + 2, sizeof(*buff));
 		buff[0] = field_get(ADPD410X_UPPDER_BYTE_SPI_MASK, address);
 		buff[1] = (address << 1) & ADPD410X_LOWER_BYTE_SPI_MASK;
 
-		ret = spi_write_and_read(dev->dev_ops.spi_phy_dev, buff, 4);
-		if(ret != SUCCESS)
+		ret = spi_write_and_read(dev->dev_ops.spi_phy_dev, buff, num_bytes + 2);
+		if(ret != SUCCESS) {
+			free(buff);
 			return ret;
+		}
+		for (i = 0; i < num_bytes; i++) {
+			data[i] = buff[i + 2];
+		}
 		break;
 	case ADPD4101:
+		// Number of bytes for an I2C read is an 8-bit number, or at most 255
+		if (num_bytes > 255)
+			return FAILURE;
+		buff = (uint8_t *) calloc(2, sizeof(*buff));
 		buff[0] = field_get(ADPD410X_UPPDER_BYTE_I2C_MASK, address);
 		buff[0] |= 0x80;
 		buff[1] = address & ADPD410X_LOWER_BYTE_I2C_MASK;
 
 		/* No stop bit */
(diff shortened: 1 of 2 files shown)


adrimbarean:
Try to limit the commit title to 75 characters.

drivers/photo-electronic/adpd410x/adpd410x.c
```diff
@@ -62,37 +62,74 @@ int32_t adpd410x_reg_read(struct adpd410x_dev *dev, uint16_t address,
...
+	int32_t ret;
+	uint8_t i;
+	uint8_t *buff;
+
+	if (num_bytes > 255)
+		return FAILURE;
 
 	switch (dev->dev_type) {
 	case ADPD4100:
+		buff = (uint8_t *) calloc(num_bytes + 2, 1);
```
adrimbarean, line 98: Would be better to have `buff = (uint8_t *) calloc(num_bytes + 2, sizeof(uint8_t));`
It is clearer this way.

drivers/photo-electronic/adpd410x/adpd410x.c
```diff
@@ -62,37 +62,74 @@ int32_t adpd410x_reg_read(struct adpd410x_dev *dev, uint16_t address,
...
+ * @return SUCCESS in case of success, FAILURE otherwise.
+ */
+int32_t adpd410x_reg_read_bytes(struct adpd410x_dev *dev, uint16_t address,
+				uint8_t *data, uint16_t num_bytes)
+{
+	int32_t ret;
+	uint8_t i;
+	uint8_t *buff;
+
+	if (num_bytes > 255)
```
adrimbarean, line 93: Is there a reason for this cap or is it arbitrary?

drivers/photo-electronic/adpd410x/adpd410x.c
```diff
@@ -551,22 +542,88 @@ static int32_t adpd410x_get_data_packet(struct adpd410x_dev *dev,
...
+		ret = adpd410x_reg_read_bytes(dev, ADPD410X_REG_FIFO_DATA,
+					      data_byte_buff + bytes_read,
+					      next_packet_size);
+		if(ret != SUCCESS)
+			goto fifo_free_return;
+
+		bytes_read += next_packet_size;
+	}
+
+	if (data == NULL)
```
adrimbarean, line 593: Would it make more sense to check this at the beginning, before doing any reads?

adrimbarean (commented):
Looks good. A couple of items and I think we should be set.

drivers/photo-electronic/adpd410x/adpd410x.c
NSagan271: For I2C, reads are at most 255 bytes. SPI doesn't seem to have this limit, so I'll move the cap to the ADPD4101 section.

drivers/photo-electronic/adpd410x/adpd410x.c
```diff
@@ -62,37 +62,74 @@ int32_t adpd410x_reg_read(struct adpd410x_dev *dev, uint16_t address,
...
+ * @param data - Pointer to the register value container.
+ * @param num_bytes - number of bytes to read, max 255
+ * @return SUCCESS in case of success, FAILURE otherwise.
+ */
+int32_t adpd410x_reg_read_bytes(struct adpd410x_dev *dev, uint16_t address,
+				uint8_t *data, uint16_t num_bytes)
+{
+	int32_t ret;
+	uint8_t i;
+	uint8_t *buff;
```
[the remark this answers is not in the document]
NSagan271: I found out that, though I2C reads are at most 255 bytes, SPI reads can be longer (the bytes_number parameter for an SPI read is a 16-bit integer). I think in that case it might be better to dynamically allocate memory?

NSagan271:
Updated PR, addressing all comments except [https://github.com/analogdevicesinc/no-OS/pull/1056#discussion_r671274019](https://github.com/analogdevicesinc/no-OS/pull/1056#discussion_r671274019).

adrimbarean (approved):
Looks good to me.

buha (approved):
looks good to me, any other opinions ?

drivers/photo-electronic/adpd410x/adpd410x.c
```diff
@@ -62,37 +62,74 @@ int32_t adpd410x_reg_read(struct adpd410x_dev *dev, uint16_t address,
...
 			return ret;
+		}
+		for (i = 0; i < num_bytes; i++) {
+			data[i] = buff[i + 2];
+		}
 		break;
 	case ADPD4101:
+		// Number of bytes for an I2C read is an 8-bit number, or at most 255
+		num_bytes = min(num_bytes, 255);
+		return FAILURE;
```
mchindri, line 111: this is not ok here, or should be inside a if.

drivers/photo-electronic/adpd410x/adpd410x.c
```diff
@@ -62,37 +62,74 @@ int32_t adpd410x_reg_read(struct adpd410x_dev *dev, uint16_t address,
...
+			free(buff);
 			return ret;
+		}
+		for (i = 0; i < num_bytes; i++) {
+			data[i] = buff[i + 2];
+		}
 		break;
 	case ADPD4101:
+		// Number of bytes for an I2C read is an 8-bit number, or at most 255
+		num_bytes = min(num_bytes, 255);
```
mchindri, line 110: I won't do this. Return failure if num_bytes > 255 or read all bytes inside a for.
A user will expect all values to be read.

mchindri (changes requested):

NSagan271:
Updated PR, addressing @mchindri's requested changes. I am returning FAILURE if num_bytes > 255 and the device type is ADPD4101.

mchindri (approved):

PR #1059, opened 2021-07-23
tools: scripts: libraries.mk remove git command

mchindri:
In order to use git on windows it needs to be installed making harder
the build process. If the version is hardcoded, posible issues related
with the git command are avoided.


buha:
I think whoever downloads no-OS does it via git clone --recursive.
Since obtaining no-OS is done via git, git is there during build.

There is the scenario where somebody downloads a zip with no-OS.
Question is, do we want to bother with that scenario ? My answer would be no, but let's see what others say.

adrimbarean:
Will close for now.

PR #1061, opened 2021-07-28
fmcjesdadc1: Update project because of HDL update

hotoleanudan:
Updated the project as a consequence of the HDL project update (the ADC data path was modified from having two 64 bit paths with 2 channels each, to one 128 bit path with 4 channels).  The modification is that there is now a single
instance of the JESD TPL, the cpack and the ADC data dma.

Signed-off-by: Dan Hotoleanu <dan.hotoleanu@analog.com>


amiclaus (commented):
Please replace `ad9250_0_core`, `ad9250_1_core`,  `ad9250_0_dmac`,  `ad9250_1_dmac` also in the IIO support section from the `fmcjesdadc1.c`, otherwise the build will fail. (as stated also in the Azure Pipelines build job).

I will handle futher IIO updates in a separate PR after this one is merged.

projects/fmcjesdadc1/src/app/fmcjesdadc1.c
```diff
@@ -562,27 +536,24 @@ int main(void)
 	ad9250_transfer(ad9250_1_device);
 
 	// capture data with DMA
-	axi_dmac_transfer(ad9250_0_dmac, ADC_0_DDR_BASEADDR, 16384 * 2);
-	axi_dmac_transfer(ad9250_1_dmac, ADC_1_DDR_BASEADDR, 16384 * 2);
-
-	Xil_DCacheFlush();
```
buha, line 568: Probably this Xil_DCacheFlush(); should be kept

adrimbarean (approved):
Looks good to me. We should re-run the automatic checks when the new HDL passed. I think the build is failing because the HDL master hasn't changed yet.

amiclaus:
> 
> 
> Looks good to me. We should re-run the automatic checks when the new HDL passed. I think the build is failing because the HDL master hasn't changed yet.

As far as I know, the hdf/xsa files used for CI build on no-OS are taken from: https://github.com/analogdevicesinc/no-OS/blob/eca1dd972c078a4e24565592a9575ae7ae3cedbd/tools/scripts/build_projects.py#L97

I requested from @hotoleanudan the hdl output files so we can update them for fmcjesdadc1.

amiclaus (commented):
> 
> 
> Looks good to me. We should re-run the automatic checks when the new HDL passed. I think the build is failing because the HDL master hasn't changed yet.

As far as I know, the hdf/xsa files used for CI build on no-OS are taken from: https://github.com/analogdevicesinc/no-OS/blob/eca1dd972c078a4e24565592a9575ae7ae3cedbd/tools/scripts/build_projects.py#L97

I requested from @hotoleanudan the hdl output files so we can update them for fmcjesdadc1.

buha (approved):

mchindri:
v2:
 - Rebase onto master to build with fixed link

projects/fmcjesdadc1/src/devices/adi_hal/parameters.h
```diff
@@ -64,15 +64,12 @@
 #define GPIO_OFFSET				54
 #endif
 
-#define ADC_0_DDR_BASEADDR			(XPAR_DDR_MEM_BASEADDR + 0x800000)
-#define ADC_1_DDR_BASEADDR			(XPAR_DDR_MEM_BASEADDR + 0x900000)
+#define ADC_DDR_BASEADDR			(XPAR_DDR_MEM_BASEADDR + 0x800000)
```
mchindri, line 67: This needs to be modified on line 53 too.
image

mchindri: Also a change in app_iio.c is needed too:
image

amiclaus (commented):
V3:
- add  back `Xil_DCacheFlush();`
- fix `ADC_DDR_BASEADDR` macro related issues
(changes made by me after talking with @hotoleanudan)

amiclaus (commented):
V4:
- fix IIO buffer size

mchindri (approved):

PR #1062, opened 2021-08-02
drivers:adpd410x: fix SPI reg write

amiclaus:
Make sure the Write Bit is set.


adrimbarean (approved):
Thanks for the patch.

dbogdan (approved):

PR #1067, opened 2021-08-16
projects:fmcjesdadc1: update IIO app

amiclaus:
Update IIO support source files for the fmcjesdadc1 project after the
latest HDL changes.

Related to: #1061 

### DIFF
--- a/projects/fmcjesdadc1/src/app/app_iio.c
+++ b/projects/fmcjesdadc1/src/app/app_iio.c
@@ -59,12 +59,7 @@
 static struct irq_ctrl_desc *irq_desc;
 #endif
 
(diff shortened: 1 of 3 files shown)


mchindri (commented):
Changes seems to be ok.
My comment will be to change the project to use the new iio_app: https://github.com/analogdevicesinc/no-OS/tree/master/iio/iio_app because you are doing modification in this direction and it will be easy to test.
The scope of the new iio_app was to update the project gradually when doing modification on the project using similar iio_app files and doing a test is easy.

mchindri (approved):
Last comment can be implemented in a new PR :)

PR #1068, opened 2021-08-18
drivers: ad7799: update ADC channel read

amiclaus:
drivers: ad7799: update ADC channel read

Implementation of the read channel function is no more constrained to
mV.

The data read from the device is now converted based on the precision
attribute in either mV or uV.

### DIFF
--- a/drivers/adc/ad7799/ad7799.c
+++ b/drivers/adc/ad7799/ad7799.c
@@ -235,17 +235,22 @@ int32_t ad7799_get_channel(struct ad7799_dev *device, uint8_t ch,
 }
 
 /**
- * @brief Read data in mV from specific ADC channel.
+ * @brief Read data from specific ADC channel with specified precision.
  * @param device - The device structure.
  * @param ch - The ADC channel.
- * @param data_mv - The content of the data converted in mV.
+ * @param data_scaled - The content of the data in mV/uV.
  * @return SUCCESS in case of success, negative error code otherwise.
  */
-int32_t ad7799_read_channel_mv(struct ad7799_dev *device, uint8_t ch,
-			       int32_t *data_mv)
+int32_t ad7799_read_channel(struct ad7799_dev *device, uint8_t ch,
+			    int32_t *data_scaled)
 {
 	int32_t ret;
 	uint32_t data, temp;
+
+	uint32_t vref_scaled = device->vref_mv;
+
+	if (device->precision)
+		vref_scaled *= 1000;
 
 	ret = ad7799_get_channel(device, ch, &data);
 	if(ret)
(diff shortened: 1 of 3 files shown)


drivers/adc/ad7799/ad7799.c
```diff
@@ -235,17 +235,18 @@ int32_t ad7799_get_channel(struct ad7799_dev *device, uint8_t ch,
...
  * @return SUCCESS in case of success, negative error code otherwise.
  */
-int32_t ad7799_read_channel_mv(struct ad7799_dev *device, uint8_t ch,
-			       int32_t *data_mv)
+int32_t ad7799_read_channel_scaled(struct ad7799_dev *device, uint8_t ch,
+				   int32_t *data_scaled)
 {
 	int32_t ret;
 	uint32_t data, temp;
+	float vref_scaled = device->vref * device->scale;
```
mchindri, line 249: I'am not sure if we should use float. Some MCU can't do floating point operations

drivers/adc/ad7799/ad7799.c
```diff
@@ -235,17 +235,18 @@ int32_t ad7799_get_channel(struct ad7799_dev *device, uint8_t ch,
...
+ * @brief Read scaled data from specific ADC channel.
  * @param device - The device structure.
  * @param ch - The ADC channel.
- * @param data_mv - The content of the data converted in mV.
+ * @param data_scaled - The content of the scaled data
  * @return SUCCESS in case of success, negative error code otherwise.
  */
-int32_t ad7799_read_channel_mv(struct ad7799_dev *device, uint8_t ch,
-			       int32_t *data_mv)
+int32_t ad7799_read_channel_scaled(struct ad7799_dev *device, uint8_t ch,
```
mchindri, line 244: nitpick: I'am not sure if scaled shoud be added to the name. I would go for just removing mv and just specify in the comment the format of the expected result.

drivers/adc/ad7799/ad7799.h
```diff
@@ -188,8 +190,10 @@ struct ad7799_init_param {
 	uint8_t gain;
 	/** Unipolar/Bipolar Coding */
 	bool polarity;
-	/** Reference Voltage in mV */
-	uint32_t vref_mv;
+	/** Reference Voltage in V */
+	float vref;
+	/** Voltage Scale **/
+	uint32_t scale;
```
mchindri, line 196: nitpick: Isn't better an
```
enum ad7799_precision { 
AD7799_PRECISION_UV,
AD7799_PRECISION_MV
}
```
?

mchindri: Or at least a better documentation for the parameter

amiclaus: Since the `float` comment is valid, I will use this approach and constrain the precision only to uV and mV.

The `scale` idea came up from the linux IIO framework: `IIO_CHAN_INFO_SCALE`

amiclaus (commented):
V2:
- remove `float` type
- constrain ADC channel precision to only mV and uV.

mchindri (approved):

buha (approved):

PR #1071, opened 2021-08-19
Implemented Support to ADT7xxx Family

jansunil:
1.) Added support to devices from the ADT7xxx family.
2.)The code flow has accommodated selection of devices based on SPI or I2C interfacing, corresponding to the target device selected-  Example: ADT7420 /ADT7320.To achieve this, macros corresponding to the register map information for ADT7320 has also been added.
3.) The individual functions would thus include a communication interface check, depending on which, the  register definitions are passed as arguments to the function calls serving the purpose.

Signed-off-by: Janani Sunil <janani.sunil@analog.com>

### DIFF
--- a/drivers/temperature/adt7420/adt7420.c
+++ b/drivers/temperature/adt7420/adt7420.c
@@ -3,7 +3,7 @@
  *   @brief  Implementation of ADT7420 Driver.
  *   @author DBogdan (dragos.bogdan@analog.com)
 ********************************************************************************
- * Copyright 2012(c) Analog Devices, Inc.
+ * Copyright 2012, 2019, 2021(c) Analog Devices, Inc.
  *
  * All rights reserved.
  *
@@ -40,8 +40,35 @@
 /******************************************************************************/
 /***************************** Include Files **********************************/
 /******************************************************************************/
+#include <stdint.h>
 #include <stdlib.h>
+#include "spi.h"
+#include "i2c.h"
+#include "error.h"
 #include "adt7420.h"
+
+const struct adt7420_chip_info chip_info[] = {
+	[ID_ADT7410] = {
+		.resolution = 16,
+		.communication = I2C,
+	},
+	[ID_ADT7420] = {
+		.resolution = 16,
+		.communication = I2C,
+	},
+	[ID_ADT7422] = {
+		.resolution = 16,
+		.communication = I2C,
+	},
+	[ID_ADT7310] = {
+		.resolution = 16,
+		.communication = SPI,
+	},
+	[ID_ADT7320] = {
+		.resolution = 16,
+		.communication = SPI,
+	}
+};
 
 /***************************************************************************//**
  * @brief Reads the value of a register.
@@ -51,44 +78,57 @@
  *
  * @return register_value  - Value of the register.
 *******************************************************************************/
-uint8_t adt7420_get_register_value(struct adt7420_dev *dev,
-				   uint8_t register_address)
+uint16_t adt7420_get_register_value(struct adt7420_dev *dev,
+				    uint8_t register_address)
 {
-	uint8_t register_value = 0;
-
-	i2c_write(dev->i2c_desc,
-		  &register_address,
-		  1,
-		  0);
-	i2c_read(dev->i2c_desc,
-		 &register_value,
-		 1,
-		 1);
-
-	return register_value;
+	if (adt7420_is_spi(dev))
+		return adt7420_spi_get_register_value(dev, register_address);
+	else
+		return adt7420_i2c_get_register_value(dev, register_address);
 }
 
 /***************************************************************************//**
- * @brief Sets the value of a register.
+ * @brief Sets the value of a register SPI/I2C.
  *
  * @param dev              - The device structure.
  * @param register_address - Address of the register.
- * @param register_value   - Value of the register.
+ * @param num_data_bytes   - Number of bytes.
+ * @param data             - Data to be written in input register.
+ * @param mask		   - Bit Mask of the bit to be written
+ * @param value		   - Value of the bit
  *
- * @return None.
+ * @return SUCCESS in case of success, FAILURE otherwise.
 *******************************************************************************/
-void adt7420_set_register_value(struct adt7420_dev *dev,
-				uint8_t register_address,
-				uint8_t register_value)
+int32_t set_register_value(struct adt7420_dev *dev,
+			   uint8_t register_address,
+			   uint8_t num_data_bytes,
+			   uint8_t *data, uint8_t mask, uint8_t value)
 {
-	uint8_t data_buffer[2] = {0, 0};
 
-	data_buffer[0] = register_address;
-	data_buffer[1] = register_value;
-	i2c_write(dev->i2c_desc,
-		  data_buffer,
-		  2,
-		  1);
+	uint8_t data_buffer[3] = {0, 0, 0};
+
+	data[0] &= ~mask;
+	data[0] |= value;
+	data_buffer[1] = data[0];
+	data_buffer[2] = data[1];
+
+	if (adt7420_is_spi(dev)) {
+		/*Form the SPI register read command byte. Bits [5:3] of command byte
+		 holds the register address.*/
+		data_buffer[0] = (register_address << 3) & ADT7320_WRITE_MASK_CMD;
+		if (spi_write_and_read(dev->spi_desc, data_buffer, num_data_bytes) != SUCCESS)
+			return FAILURE;
+	} else {
+		data_buffer[0] = register_address;
+
+		if (i2c_write(dev->i2c_desc,
+			      data_buffer,
+			      num_data_bytes,
+			      1) != SUCCESS)
+			//no repeat start
+			return FAILURE;
+	}
+	return SUCCESS;
 }
 
 /***************************************************************************//**
@@ -110,24 +150,51 @@ int32_t adt7420_init(struct adt7420_dev **device,
 {
 	struct adt7420_dev *dev;
 	int32_t status;
-	uint8_t test = 0;
+	uint8_t device_connected_check = 0;
 
 	dev = (struct adt7420_dev *)malloc(sizeof(*dev));
 	if (!dev)
 		return -1;
 
-	/* I2C */
-	status = i2c_init(&dev->i2c_desc, &init_param.i2c_init);
+	dev->active_device = init_param.active_device;
 
-	/* Device Settings */
-	dev->resolution_setting = init_param.resolution_setting;
+	if (adt7420_is_spi(dev))
+		status = spi_init(&dev->spi_desc, &init_param.interface_init.spi_init);
+	else
+		status = i2c_init(&dev->i2c_desc, &init_param.interface_init.i2c_init);
 
-	test   = adt7420_get_register_value(dev, ADT7420_REG_ID);
-	if(test != ADT7420_DEFAULT_ID)
-		status = -1;
+	if (status != FAILURE) {
+		/* Device Settings */
+		dev->resolution_setting = init_param.resolution_setting;
 
-	*device = dev;
+		/* Reset device to default values to ensure all internal circuitry is properly initialised*/
+		status = adt7420_reset(dev);
+		if (status < 0)
+			goto adt7420_init_error;
 
+		/*Register read to ensure that next read will be valid - acts as 200us delay while
+		  device resets*/
+		if (adt7420_is_spi(dev))
+			device_connected_check = adt7420_get_register_value(dev, ADT7320_REG_ID);
+		else
+			device_connected_check = adt7420_get_register_value(dev, ADT7420_REG_ID);
+
+		device_connected_check >>= 4; // Manufacturer ID
+
+		if(device_connected_check != ADT7xxx_ID) // AD7xxx ID Check
+			status = FAILURE;
+		else
+			status = SUCCESS;
+
+		*device = dev;
+
+		return status;
+	}
+
+	return status;
+
+adt7420_init_error:
+	adt7420_remove(dev);
 	return status;
 }
 
@@ -142,35 +209,17 @@ int32_t adt7420_remove(struct adt7420_dev *dev)
 {
 	int32_t ret;
 
-	ret = i2c_remove(dev->i2c_desc);
-
+	if (!(adt7420_is_spi(dev)))
+		ret = i2c_remove(dev->i2c_desc);
+	else
+		ret = spi_remove(dev->spi_desc);
 	free(dev);
 
 	return ret;
 }
 
 /***************************************************************************//**
- * @brief Resets the ADT7420.
- *        The ADT7420 does not respond to I2C bus commands while the default
- *        values upload (approximately 200 us).
- *
- * @param dev - The device structure.
- *
- * @return None.
-*******************************************************************************/
-void adt7420_reset(struct adt7420_dev *dev)
-{
-	uint8_t register_address = ADT7420_REG_RESET;
-
-	i2c_write(dev->i2c_desc,
-		  &register_address,
-		  1,
-		  1);
-	dev->resolution_setting = 0;
-}
-
-/***************************************************************************//**
- * @brief Sets the operational mode for ADT7420.
+ * @brief Sets the operational mode for ADT7420/ADT7320.
  *
  * @param dev  - The device structure.
  * @param mode - Operation mode.
@@ -184,16 +233,23 @@ void adt7420_reset(struct adt7420_dev *dev)
 void adt7420_set_operation_mode(struct adt7420_dev *dev,
 				uint8_t mode)
 {
-	uint8_t register_value = 0;
+	uint8_t register_value [2] = { 0, 0 };
+	uint8_t device_reg_address;
 
-	register_value  = adt7420_get_register_value(dev, ADT7420_REG_CONFIG);
-	register_value &= ~ADT7420_CONFIG_OP_MODE(ADT7420_OP_MODE_SHUTDOWN);
-	register_value |= ADT7420_CONFIG_OP_MODE(mode);
-	adt7420_set_register_value(dev, ADT7420_REG_CONFIG, register_value);
+	if (adt7420_is_spi(dev)) {
+		device_reg_address = ADT7320_REG_CONFIG;
+		register_value[0]  = adt7420_get_register_value(dev, device_reg_address);
+	} else {
+		device_reg_address = ADT7420_REG_CONFIG;
+		register_value[0]  = adt7420_get_register_value(dev, device_reg_address);
+	}
+
+	set_register_value(dev, device_reg_address, 2, register_value,
+			   ADT7420_CONFIG_OP_MODE(ADT7420_OP_MODE_SHUTDOWN), ADT7420_CONFIG_OP_MODE(mode));
 }
 
 /***************************************************************************//**
- * @brief Sets the resolution for ADT7420.
+ * @brief Sets the resolution for ADT7420/ADT7320.
  *
  * @param dev        - The device structure.
  * @param resolution - Resolution.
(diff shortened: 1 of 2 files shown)


drivers/temperature/adt7420/adt7420.c
```diff
@@ -3,7 +3,7 @@
  *   @brief  Implementation of ADT7420 Driver.
  *   @author DBogdan (dragos.bogdan@analog.com)
 ********************************************************************************
- * Copyright 2012(c) Analog Devices, Inc.
```
mchindri, line 6: Change the commit title to: `drivers:temperature:adt7420:Add new supported devices` or something similar but with the path for the modifications and the current commit message can go into the description with the lines wrapped to 75 characters.

drivers/temperature/adt7420/adt7420.c
```diff
@@ -40,55 +40,107 @@
...
+	[ID_ADT7422] = {
+		.resolution = 16,
+		.communication = I2C,
+	},
+	[ID_ADT7310] = {
+		.resolution = 16,
+		.communication = SPI,
+	},
+	[ID_ADT7320] = {
+		.resolution = 16,
```
mchindri, line 68: is resolution needed if is the same for all?
Can it go as before in adt7420_dev ?

drivers/temperature/adt7420/adt7420.h
```diff
@@ -42,92 +42,151 @@
...
-#define ADT7420_REG_T_HIGH_MSB		0x04 // Temperature HIGH setpoint MSB
-#define ADT7420_REG_T_HIGH_LSB		0x05 // Temperature HIGH setpoint LSB
-#define ADT7420_REG_T_LOW_MSB		0x06 // Temperature LOW setpoint MSB
-#define ADT7420_REG_T_LOW_LSB		0x07 // Temperature LOW setpoint LSB
-#define ADT7420_REG_T_CRIT_MSB		0x08 // Temperature CRIT setpoint MSB
-#define ADT7420_REG_T_CRIT_LSB		0x09 // Temperature CRIT setpoint LSB
-#define ADT7420_REG_HIST		0x0A // Temperature HYST setpoint
-#define ADT7420_REG_ID			0x0B // ID
-#define ADT7420_REG_RESET		0x2F // Software reset
+#define ADT7420_A0_PIN(x)	            (((x) & 0x1) << 0) // I2C Serial Bus Address Selection Pin
```
mchindri, line 54: Please don't change the spacing is really hard to see the changes. 
Anyway the indentation is not ok, use tab with a width of 8 space

drivers/temperature/adt7420/adt7420.c
```diff
@@ -40,55 +40,107 @@
...
-uint8_t adt7420_get_register_value(struct adt7420_dev *dev,
-				   uint8_t register_address)
+uint16_t adt7420_get_register_value(struct adt7420_dev *dev,
+				    uint8_t register_address)
 {
 	uint8_t register_value = 0;
+	uint8_t data[2] = {0,0xFF};
+
+	if (chip_info[dev->active_device].communication == SPI) {
+		/* Form the SPI register read command byte. Bits [5:3] of command byte
```
mchindri, line 88: What about creating 2 function instead of this if/else:
adt7420_i2c_get_register_value and adt7420_spi_get_register_value.

drivers/temperature/adt7420/adt7420.h
```diff
@@ -147,6 +203,9 @@ void adt7420_set_operation_mode(struct adt7420_dev *dev,
 void adt7420_set_resolution(struct adt7420_dev *dev,
 			    uint8_t resolution);
 
+/*! Resets the SPI or I2C inteface for the ADT7420/ADT7320*/
+void adt7420_reset_interface(struct adt7420_dev *dev);
+
```
mchindri, line 208: Can you do this modification in place? I mean on the adt7420_reset function, not delete that one and create a new one.

mchindri: Anyway changing function name is really needed?

drivers/temperature/adt7420/adt7420.h
```diff
@@ -42,92 +42,151 @@
...
+	uint8_t		    resolution_setting;
+
 };
 
 struct adt7420_init_param {
 	/* I2C */
-	i2c_init_param	i2c_init;
+	i2c_init_param		i2c_init;
+	/* SPI */
+	spi_init_param		spi_init;
```
mchindri, line 158: You can do:
```
i2c_init_param		*i2c_init;
spi_init_param		*spi_init;
```
And create a filed in 
adt7420_dev
`bool is_spi;`
In adt7420_init you can check if  i2c_spi is not NULL and initialized the spi and set is_spi to true. Otherwise initialize i2c.
Then in the code you can just check 
`if (dev->is_spi)`
This way some extra code wont be needed

mchindri (commented):
Please do a forcepush one this branch when you have the changes done. Don't close the PR and open a new one. It is useful for us to see the previous feedback

drivers/temperature/adt7420/adt7420.c
buha: I see many devices from the family that don't have 16-bit resolution so it's best to keep resolution in each entry:
https://eepower.com/new-industry-products/analog-devices-intros-adt7xxx-converter-family/#

drivers/temperature/adt7420/adt7420.c
```diff
@@ -40,55 +40,107 @@
...
+
+
+	data_buffer[1] = data[0];
+	data_buffer[2] = data[1];
+
+	if (chip_info[dev->active_device].communication == SPI) {
+		/*Form the SPI register read command byte. Bits [5:3] of command byte
+		 holds the register address.*/
+		data_buffer[0] = (register_address << 3) & ADT7320_WRITE_MASK_CMD;
+		spi_write_and_read(dev->spi_desc, data_buffer, num_data_bytes);
```
buha, line 135: check return code of each function

drivers/temperature/adt7420/adt7420.c
```diff
@@ -40,55 +40,107 @@
...
+	} else {
+		data_buffer[0] = register_address;
+
+		i2c_write(dev->i2c_desc,
```
buha, line 139: check return code

drivers/temperature/adt7420/adt7420.c
```diff
@@ -40,55 +40,107 @@
...
- * @param register_value   - Value of the register.
+ * @param num_data_bytes   - Number of bytes.
+ * @param data             - Data to be written in input register.
  *
  * @return None.
 *******************************************************************************/
-void adt7420_set_register_value(struct adt7420_dev *dev,
-				uint8_t register_address,
-				uint8_t register_value)
+void set_register_value(struct adt7420_dev *dev,
```
buha, line 119: should return a return code

drivers/temperature/adt7420/adt7420.c
```diff
@@ -205,15 +272,48 @@ void adt7420_set_operation_mode(struct adt7420_dev *dev,
...
+
+	if (chip_info[dev->active_device].communication == SPI) {
+		device_reg_address = ADT7320_REG_CONFIG;
+		register_value[0]  = adt7420_get_register_value(dev, device_reg_address);
+	} else {
+		device_reg_address = ADT7420_REG_CONFIG;
+		register_value[0]  = adt7420_get_register_value(dev, device_reg_address);
+	}
+
+	register_value[0] &= ~ADT7420_CONFIG_RESOLUTION;
```
buha, line 286: All this masking that you do extensively in several functions could be done by calling one function, something like this:
https://github.com/analogdevicesinc/no-OS/blob/master/drivers/adc/ad7606/ad7606.c#L339-L355

drivers/temperature/adt7420/adt7420.c
```diff
@@ -205,15 +272,48 @@ void adt7420_set_operation_mode(struct adt7420_dev *dev,
...
+	if (chip_info[dev->active_device].communication == SPI)
+		spi_write_and_read(dev->spi_desc, data_buffer, sizeof(data_buffer));
+	else {
+		uint8_t register_address = ADT7420_REG_RESET;
+		i2c_write(dev->i2c_desc,
+			  &register_address,
+			  1,
+			  1);//no repeat start
+	}
+	dev->resolution_setting = 0;
```
buha, line 314: you need to check for the error codes above and if all was ok, then consider the reset successful

when you know for sure the reset was successful you probably need to set some device driver parameters to their default - which you do in the case of resolution_setting, but the question is, are there other members that need to be reset to default as well ?

drivers/temperature/adt7420/adt7420.h
```diff
@@ -42,92 +42,151 @@
...
+#define ADT7420_REG_HIST			    0x0A // Temperature HYST setpoint
+#define ADT7420_REG_ID				    0x0B // ID
+#define ADT7420_REG_RESET			    0x2F // Software reset
 
 /* ADT7420_REG_STATUS definition */
-#define ADT7420_STATUS_T_LOW		(1 << 4)
-#define ADT7420_STATUS_T_HIGH		(1 << 5)
-#define ADT7420_STATUS_T_CRIT		(1 << 6)
-#define ADT7420_STATUS_RDY		(1 << 7)
+#define ADT7420_STATUS_T_LOW		    (1 << 4)
```
buha, line 88: Use BIT(4)
Same comment for all the ones below

drivers/temperature/adt7420/adt7420.h
```diff
@@ -42,92 +42,151 @@
...
 #define ADT7420_CONFIG_FAULT_QUEUE(x)	(x & 0x3)
-#define ADT7420_CONFIG_CT_POL		(1 << 2)
-#define ADT7420_CONFIG_INT_POL		(1 << 3)
-#define ADT7420_CONFIG_INT_CT_MODE	(1 << 4)
-#define ADT7420_CONFIG_OP_MODE(x)	((x & 0x3) << 5)
-#define ADT7420_CONFIG_RESOLUTION	(1 << 7)
+#define ADT7420_CONFIG_CT_POL		    (1 << 2)
+#define ADT7420_CONFIG_INT_POL		    (1 << 3)
+#define ADT7420_CONFIG_INT_CT_MODE  	(1 << 4)
+#define ADT7420_CONFIG_OP_MODE(x)	    ((x & 0x3) << 5)
```
buha, line 98: Use GENMASK(6,5) from util.h

drivers/temperature/adt7420/adt7420.h
buha: this would be redundant information that is avoidable

a simpler way would be to have a simple static function like:
```
static bool is_spi(struct adt7420_dev *dev) {
    return chip_info[dev->active_device].communication == SPI;
}
```

It's short enough and doesn't introduce redundant desc attributes:
`if (is_spi(dev)) ...`

drivers/temperature/adt7420/adt7420.h
buha: yeah, i agree, please use the old name, no reason to change to a lenghtier one

drivers/temperature/adt7420/adt7420.c
```diff
@@ -181,14 +154,16 @@ int32_t adt7420_init(struct adt7420_dev **device,
 		dev->resolution_setting = init_param.resolution_setting;
 
 		/* Reset device to default values to ensure all internal circuitry is properly initialised*/
-		adt7420_reset_interface(dev);
+		if (adt7420_reset(dev) != SUCCESS) {
```
buha, line 157: I see the error code is checked now but there is one more thing to do, pass the return code back - unaltered. There is no reason for us to translate the error code here from -ENOMEM or -EBADMSG or anything like that into FAILURE.

See this as an example, by following the `ret` variable:
https://github.com/analogdevicesinc/no-OS/blob/master/drivers/adc/ad400x/ad400x.c#L173

drivers/temperature/adt7420/adt7420.h
```diff
@@ -45,74 +45,76 @@
...
 #define ADT7420_CONFIG_FAULT_QUEUE(x)	(x & 0x3)
-#define ADT7420_CONFIG_CT_POL		    (1 << 2)
-#define ADT7420_CONFIG_INT_POL		    (1 << 3)
-#define ADT7420_CONFIG_INT_CT_MODE  	(1 << 4)
-#define ADT7420_CONFIG_OP_MODE(x)	    ((x & 0x3) << 5)
-#define ADT7420_CONFIG_RESOLUTION	    (1 << 7)
+#define ADT7420_CONFIG_CT_POL		BIT(2)
+#define ADT7420_CONFIG_INT_POL		BIT(3)
+#define ADT7420_CONFIG_INT_CT_MODE	BIT(4)
+#define ADT7420_CONFIG_OP_MODE(x)	(x << 5) & (GENMASK(6,5))
```
buha, line 100: Always put parantheses around macro parameter when referencing it in the expression, this should be:
`((x) << 5) ...`

If you don't put parantheses it may not work like you intend.
Without parantheses, `ADT7420_CONFIG_OP_MODE(a + b)` would work because it translates as `(a + b << 5) ...` and + is more prioritary than <<. But  `ADT7420_CONFIG_OP_MODE(a & b)` would not work because & is less prioritary than << in C operator precedence. https://en.cppreference.com/w/c/language/operator_precedence

drivers/temperature/adt7420/adt7420.c
jansunil: Modified the commit message as suggested

drivers/temperature/adt7420/adt7420.c
jansunil: Resolution field remains unaltered, considering devices that do not have a 16-bit resolution

drivers/temperature/adt7420/adt7420.h
jansunil: Spacing has been restored as in the original file.

drivers/temperature/adt7420/adt7420.c
jansunil: Two new functions:
adt7420_i2c_get_register_value () and
adt7420_spi_get_register_value() 
has been created in place of adt7420_get_register_value()

drivers/temperature/adt7420/adt7420.h
jansunil: Name has been changed back to the original

drivers/temperature/adt7420/adt7420.h
jansunil: A new function adt7420_spi_comm_check() has been created for the same. Used the same function call wherever the check has been made

drivers/temperature/adt7420/adt7420.c
jansunil: Added return code check

drivers/temperature/adt7420/adt7420.c
jansunil: Added return code check

drivers/temperature/adt7420/adt7420.c
jansunil: The function has been re-framed in such a way that it returns an int32_t value. Modified function prototype for the same as well.

drivers/temperature/adt7420/adt7420.c
jansunil: The masking has been implemented as suggested as a part of the set_register_value() function. Following changes made respective to this:
1.) Added new function parameters for the 'mask' and the 'value' parameter
2.)Removed lines preceding the set_register_value() function call that dealt with masking.

drivers/temperature/adt7420/adt7420.c
jansunil: Error code check has been added. 

Regarding the other members to be reset - The only member in the **adt7420_dev** structure that may need a manual reset would be the resolution setting. Other members are the spi/i2c init parameters and the active device, which may get selected upon executing the application. There may not be a relevant default case for the same.

drivers/temperature/adt7420/adt7420.h
jansunil: Implemented the functionality using BIT() function from util.h

drivers/temperature/adt7420/adt7420.h
jansunil: Implemented the same using GENMASK()

drivers/temperature/adt7420/adt7420.c
jansunil: Implemented the return value by a variable holding the actual value returned.

drivers/temperature/adt7420/adt7420.h
jansunil: Thanks for flagging this, added parenthesis, considering the above mentioned case.

buha (approved):
from my side it looks good

buha:
make sure to merge the 2nd and 3rd commits into 1st, and force push this change until Mihail reviews again

buha:
So @jansunil, the way I do it is the following (parent_branch is probably master for you):
# git rebase -i parent_branch

In this case, git will propose you 5 commits, you will have to squash commits 2 and 3 into the 1st and drop the 4th and 5th commit.
Follow this guide to find out how to rewrite history with git interactive rebase.
https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History

drivers/temperature/adt7420/adt7420.h
```diff
@@ -54,7 +57,21 @@
...
+#define ADT7320_REG_T_CRIT	    0x04 // Temperature CRIT setpoint (147'C)
+#define ADT7320_REG_HIST	    0x05 // Temperature HYST setpoint (5'C)
+#define ADT7320_REG_T_HIGH    	    0x06 // Temperature HIGH setpoint (64'C)
+#define ADT7320_REG_T_LOW    	    0x07 // Temperature LOW setpoint (10'C)
+
+/* ADT7320 SPI command byte */
+#define ADT7320_WRITE_MASK_CMD      0b00111000 // SPI write command
+#define ADT7320_READ_CMD            0b01000000 // SPI read command
+
+/* ADT7420 (I2C) registers */
```
mchindri, line 74: Fix the indentation: Makes them be inline with the rest of the defines:
See this view: https://github.com/analogdevicesinc/no-OS/blob/72f0062ebdf4f831a346afa128166f29d62340c0/drivers/temperature/adt7420/adt7420.h?ts=8
How it is displayed here is not really accurate.

drivers/temperature/adt7420/adt7420.h
```diff
@@ -96,38 +113,73 @@
...
 	/* Device Settings */
 	uint8_t		resolution_setting;
+
 };
 
 struct adt7420_init_param {
 	/* I2C */
 	i2c_init_param	i2c_init;
+	/* SPI */
+	spi_init_param	spi_init;
```
mchindri, line 160: Can this be a union with i2c_init param? I think it would make sense because only one of them can be used depending on active_device.

drivers/temperature/adt7420/adt7420.h
```diff
@@ -96,38 +113,73 @@
...
+
 /******************************************************************************/
 /************************ Functions Declarations ******************************/
 /******************************************************************************/
 
-/*! Reads the value of a register. */
-uint8_t adt7420_get_register_value(struct adt7420_dev *dev,
-				   uint8_t register_address);
+/*! Write to input shift register SPI interface. */
+uint16_t set_shift_reg(struct adt7420_dev *dev,
```
mchindri, line 174: I want do any modification in the current API.
When an application calls adt7420_get_register_value or adt7420_set_register_value, the driver will check the dev type and then will choose to call either adt7420_i2c_get_register_value or adt7420_spi_get_register_value

mchindri: If you really think some changes are needed to the API functions, please do it in an other commit an motivate the reason.

drivers/temperature/adt7420/adt7420.c
```diff
@@ -40,55 +40,80 @@
...
- * @param register_value   - Value of the register.
- *
- * @return None.
-*******************************************************************************/
-void adt7420_set_register_value(struct adt7420_dev *dev,
-				uint8_t register_address,
-				uint8_t register_value)
-{
-	uint8_t data_buffer[2] = {0, 0};
+	if (adt7420_spi_comm_check(dev)) {
```
mchindri, line 98: I would rename `adt7420_spi_comm_check ` to `adt7420_is_spi`

drivers/temperature/adt7420/adt7420.c
```diff
@@ -40,55 +40,80 @@
...
+		/*Form the SPI register read command byte. Bits [5:3] of command byte
+		 holds the register address.*/
+		data_buffer[0] = (register_address << 3) & ADT7320_WRITE_MASK_CMD;
+		if (spi_write_and_read(dev->spi_desc, data_buffer, num_data_bytes) != SUCCESS) {
```
mchindri, line 102: Single instructions don't need {}.

mchindri: Same for the rest of the changes

drivers/temperature/adt7420/adt7420.c
```diff
@@ -110,24 +135,54 @@ int32_t adt7420_init(struct adt7420_dev **device,
...
+		if (adt7420_spi_comm_check(dev)) {
+			device_connected_check = adt7420_spi_get_register_value(dev, ADT7320_REG_ID);
+		} else {
+			device_connected_check = adt7420_i2c_get_register_value(dev, ADT7420_REG_ID);
+		}
 
-	*device = dev;
+		device_connected_check >>= 4; // Manufacturer ID
 
+		if(device_connected_check != 0xC) // AD7xxx ID Check
```
mchindri, line 172: Replace 0xC with a define, It will be easy to understand.

jansunil:
@buha , @mchindri -  Do these changes look good?

drivers/temperature/adt7420/adt7420.c
```diff
@@ -51,44 +78,57 @@
...
- * @return None.
+ * @return SUCCESS in case of success, FAILURE otherwise.
 *******************************************************************************/
-void adt7420_set_register_value(struct adt7420_dev *dev,
-				uint8_t register_address,
-				uint8_t register_value)
+int32_t set_register_value(struct adt7420_dev *dev,
+			   uint8_t register_address,
+			   uint8_t num_data_bytes,
+			   uint8_t *data, uint8_t mask, uint8_t value)
```
adrimbarean, line 105: This function seems to be a read-modify-write, but it's not self-contained. Could it work better if it had only the device handler, register address, mask and value as parameters? And read the current value of the register inside, change it with the mask and value, then write it?

This way it will assume some things that are not readily apparent to the user, that for example data is at least 2 bytes long or that it already contains the current value of the register.

drivers/temperature/adt7420/adt7420.c
```diff
@@ -250,3 +345,106 @@ float adt7420_get_temperature(struct adt7420_dev *dev)
...
+
+	if (numBytes == 2) {
+		data_buffer[0] = data[0];
+		data_buffer[1] = data[1] & 0x00FF;
+		spi_write_and_read(dev->spi_desc, data_buffer, 2);
+		read_back_data = (data_buffer[0] & 0xFF00) | data_buffer[1];
+	} else {
+		data_buffer[0] = data[0];
+		data_buffer[1] = data[1] & 0x00FF;
+		data_buffer[2] = (data[2] & 0x00FF) << 8;
```
adrimbarean, line 387: If `data_buffer` is only represented on 8 bits here shifting something by 8 bits and assigning the value to `data_buffer` will truncate the result to probably 0, most of the time. Am I missing something here?

adrimbarean: nitpick: `data` points to 8-bit values. Masking with `0x00FF` seems pointless.

drivers/temperature/adt7420/adt7420.c
jansunil: @adrimbarean - There can be a use case where we don't really need the current value of a register during a set register operation. 

Example:
In a case where we want to write to the one of the set point registers, we may want to write the value as such without any masking being performed. In such a scenario, we don't need to retain the current value held by the register before writing to it.  So should we really do the get register operation within this function?

jansunil: @adrimbarean -  Can we proceed with the existing flow itself? Or do you suggest modifying it?

adrimbarean (approved):

drivers/temperature/adt7420/adt7420.c
jansunil: data[2] is not used anywhere. Hence removed the line. There has been a slight modification made in the function definition to accommodate only the read back value calculation within the if condition

PR #1075, opened 2021-08-23
build_projects.yml: minor build updates

amiclaus:
Similar to `azure_pipelines.yml` update build projects file to support
Pull Requests that target `main` branch and create custom  branches with
`staging/*` prefix that trigger the project build.

### DIFF
--- a/build_projects.yml
+++ b/build_projects.yml
@@ -1,8 +1,10 @@
 trigger:
+- main
 - master
-- dev/*
+- staging/*
 
 pr:
+- main
 - master
 
 variables:


build_projects.yml
```diff
@@ -1,8 +1,10 @@
 trigger:
 - master
 - dev/*
+- staging/*
 
 pr:
+- main
```
mchindri, line 7: If our main branch is master, why adding main?

build_projects.yml
```diff
@@ -1,8 +1,10 @@
 trigger:
 - master
 - dev/*
+- staging/*
```
mchindri, line 4: Why staging if there is dev for triggering the build?

amiclaus: Alignment with: https://github.com/analogdevicesinc/no-OS/commit/07f6e2bc73a5f49e027438d5b91b631853da4ffb
We either remove `staging` from both files or keep them.
If we want custom branch that builds only projects, then a more suggestive prefix should be added.

build_projects.yml
amiclaus: Since github started the migration from `master` to `main`, this is a safety measure more-or-less.
It is also an alignment with:
https://github.com/analogdevicesinc/no-OS/commit/07f6e2bc73a5f49e027438d5b91b631853da4ffb

build_projects.yml
amiclaus: Alignment with: 07f6e2b
We either remove staging from both files or keep them.
If we want custom branch that builds only projects, then a more suggestive prefix should be added.

mchindri: I would replace dev then instead of adding staging and add main here too

amiclaus: Yes, will do so, since linux repo uses same approach (only `staging/*`, no `dev/*`)

amiclaus (commented):
v2:
- keep only `staging` prefix
- add `main` also in the trigger section

mchindri (approved):

PR #1078, opened 2021-08-23
Dev/aducm flash driver

adrimbarean:
Add Flash driver for aducm3029 and an example project.

### DIFF
--- a/drivers/platform/aducm3029/flash.c
+++ b/drivers/platform/aducm3029/flash.c
@@ -1,0 +1,293 @@
+/***************************************************************************//**
+ *   @file   aducm3029/flash.c
+ *   @brief  Implementation of the flash driver for ADuCM302x.
+ *   @author Andrei Drimbarean (adnrei.drimbarean@analog.com)
+********************************************************************************
+ * Copyright 2021(c) Analog Devices, Inc.
+ *
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *  - Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ *  - Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in
+ *    the documentation and/or other materials provided with the
+ *    distribution.
+ *  - Neither the name of Analog Devices, Inc. nor the names of its
+ *    contributors may be used to endorse or promote products derived
+ *    from this software without specific prior written permission.
+ *  - The use of this software may or may not infringe the patent rights
+ *    of one or more patent holders.  This license does not release you
+ *    from the requirement that you obtain separate licenses from these
+ *    patent holders to use this software.
+ *  - Use of the software either in source or binary form, must be run
+ *    on or directly connected to an Analog Devices Inc. component.
+ *
+ * THIS SOFTWARE IS PROVIDED BY ANALOG DEVICES "AS IS" AND ANY EXPRESS OR
+ * IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, NON-INFRINGEMENT,
+ * MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED.
+ * IN NO EVENT SHALL ANALOG DEVICES BE LIABLE FOR ANY DIRECT, INDIRECT,
+ * INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, INTELLECTUAL PROPERTY RIGHTS, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+*******************************************************************************/
+
+/******************************************************************************/
+/************************* Include Files **************************************/
+/******************************************************************************/
+
+#include <drivers/flash/adi_flash.h>
+#include <drivers/dma/adi_dma.h>
+#include <stdlib.h>
+#include <malloc.h>
+#include <string.h>
+#include "flash.h"
+#include "flash_extra.h"
+#include "error.h"
+
+/******************************************************************************/
+/*************************** Types Declarations *******************************/
+/******************************************************************************/
+
+/**
+ * @struct adicup_flash_dev
+ * @brief Aducm3029 flash controller handler
+ */
+struct adicup_flash_dev {
+	/** ADuCM3029 DFP driver handler */
+	ADI_FEE_HANDLE instance;
+	/** Necessary memory for DFP driver */
+	uint8_t dfp_memory[ADI_FEE_MEMORY_SIZE];
+	/** Buffer to read one flash page */
+	uint32_t temp_ptr[FLASH_PAGE_SIZE_WORDS] __attribute__ ((aligned (sizeof(
+				uint32_t))));
+};
+
+/******************************************************************************/
+/************************ Functions Definitions *******************************/
+/******************************************************************************/
+
+/**
+ * Initialize flash controller.
+ *
+ * @param [out] device     - Pointer to the driver handler.
+ * @param [in]  init_param - Pointer to the initialization structure.
+ *
+ * @return 0 in case of success, error code otherwise.
+ */
+int32_t flash_init(struct flash_dev **device,
+		   struct flash_init_param *init_param)
+{
+	int32_t ret;
+	struct flash_dev *dev;
+	struct adicup_flash_dev *adicup_extra;
+
+	dev = (struct flash_dev *)calloc(1, sizeof(*dev));
+	if (!dev)
+		return FAILURE;
+
+	dev->id = init_param->id;
+	adicup_extra = (struct adicup_flash_dev *)calloc(1,
+			sizeof(*adicup_extra));
+	if (!adicup_extra)
+		goto error_dev;
+
+	dev->extra = adicup_extra;
+	dev->flash_size = ADUCM3029_FLASH_SIZE_BYTES;
+	dev->page_size = FLASH_PAGE_SIZE_BYTES;
+
+	/* Open flash device */
+	ret = adi_fee_Open(dev->id, adicup_extra->dfp_memory,
+			   ADI_FEE_MEMORY_SIZE,
+			   &adicup_extra->instance);
+	if(ret != ADI_FEE_SUCCESS)
+		goto error_adicup;
+
+	*device = dev;
+
+	return SUCCESS;
+
+error_adicup:
+	free(adicup_extra);
+error_dev:
+	free(dev);
+
+	return FAILURE;
+}
+
+/**
+ * Free memory allocated by flash_init().
+ *
+ * @param [in] dev - Pointer to the driver handler.
+ *
+ * @return 0 in case of success, error code otherwise.
+ */
+int32_t flash_remove(struct flash_dev *dev)
+{
+	int32_t ret;
+	struct adicup_flash_dev *adicup_extra;
+
+	if (!dev)
+		return FAILURE;
+
+	adicup_extra = dev->extra;
+
+	ret = adi_fee_Close(adicup_extra->instance);
+	if (ret != ADI_FEE_SUCCESS)
+		return FAILURE;
+
+	free(adicup_extra);
+	free(dev);
+
+	return SUCCESS;
+}
+
+/**
+ * Erase a flash page.
+ *
+ * @param [in] dev        - Pointer to the flash device handler.
+ * @param [in] page_no    - Page number.
+ *
+ * @return 0 in case of success, negative error code otherwise.
+ */
+int32_t flash_clear_page(struct flash_dev *dev, int32_t page_no)
+{
+	struct adicup_flash_dev *adicup_extra = dev->extra;
+	uint32_t fee_hw_error;
+	int32_t ret;
+
+	if (page_no > (dev->flash_size / dev->page_size))
+		return FAILURE;
+
+	ret = adi_fee_PageErase(adicup_extra->instance, page_no, page_no,
+				&fee_hw_error);
+	if(ret != ADI_FEE_SUCCESS)
+		return FAILURE;
+
+	return SUCCESS;
+}
+
+/**
+ * Write a page in flash memory.
+ *
+ * @param [in] dev        - Pointer to the flash device handler.
+ * @param [in] page_no    - Page number.
+ * @param [in] data       - Pointer to the data to be written.
+ *                          data length must be FLASH_PAGE_SIZE_WORDS in size.
+ *
+ * @return 0 in case of success, negative error code otherwise.
+ */
+int32_t flash_write_page(struct flash_dev *dev, int32_t page_no, uint32_t *data)
+{
+	ADI_FEE_TRANSACTION transaction;
+	struct adicup_flash_dev *adicup_extra = dev->extra;
+	uint32_t fee_hw_error;
+	int32_t ret;
+
+	if (page_no > (dev->flash_size / dev->page_size))
+		return FAILURE;
+
+	transaction.bUseDma = true;
+	transaction.nSize = FLASH_PAGE_SIZE_BYTES;
+	transaction.pWriteAddr = (uint32_t *)(page_no << FLASH_PAGE_ADDR_SHIFT);
+	transaction.pWriteData = data;
+
+	ret = adi_fee_Write(adicup_extra->instance, &transaction,
+			    &fee_hw_error);
+	if(ret != ADI_FEE_SUCCESS)
+		return FAILURE;
+
+	return SUCCESS;
+}
+
+/**
+ * Read-modify-write helper function for flash_write(). Since the smallest write
+ * unit for the ADuCM3029 is the page this function helps access data smaller than
+ * a page or accross multiple pages.
+ */
+static int32_t flash_read_then_write(struct flash_dev *dev, uint32_t flash_addr,
+				     uint32_t *array, uint32_t array_size)
+{
+	struct adicup_flash_dev *adicup_extra = dev->extra;
+	uint32_t page_nr, fee_hw_error;
+	int32_t ret;
+
+	if ((flash_addr & 0x3) != 0)
+		return FAILURE;
+	if ((FLASH_OFFSET_IN_PAGE(flash_addr) + array_size) >
+	    FLASH_PAGE_SIZE_WORDS)
+		return FAILURE;
+
+	flash_read(dev, FLASH_ADDRESS_PAGE_START(flash_addr), adicup_extra->temp_ptr,
+		   FLASH_PAGE_SIZE_WORDS);
+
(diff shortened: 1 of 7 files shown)


include/flash.h
```diff
@@ -0,0 +1,92 @@
...
+	void *extra;
+};
+
+/**
+ * @struct flash_init_param
+ * @brief Flash controller initialization structure
+ */
+struct flash_init_param {
+	/** Flash Device ID */
+	uint8_t id;
```
mchindri, line 70: Shouldn't extra exist here too?

drivers/platform/aducm3029/flash.c
```diff
@@ -0,0 +1,250 @@
...
+		return FAILURE;
+
+	dev->id = init_param->id;
+	adicup_extra = (struct adicup_flash_dev *)calloc(1,
+			sizeof(*adicup_extra));
+	if (!adicup_extra)
+		goto error_dev;
+
+	dev->extra = adicup_extra;
+	adicup_extra->dfp_memory = calloc(1, ADI_FEE_MEMORY_SIZE);
```
mchindri, line 95: You can do:
'''
struct adicup_flash_dev {
	void *instance;
	uint8_t dfp_memory[ADI_FEE_MEMORY_SIZE]; 
};
'''
And maybe align it if needed.

drivers/platform/aducm3029/flash.c
```diff
@@ -0,0 +1,250 @@
...
+	if (!adicup_extra->dfp_memory)
+		goto error_adicup;
+
+	/* Open flash device */
+	ret = adi_fee_Open(dev->id, adicup_extra->dfp_memory,
+			   ADI_FEE_MEMORY_SIZE,
+			   (ADI_FEE_HANDLE *)&adicup_extra->instance);
```
mchindri, line 102: Just declare adicup_extra->instance as ADI_FEE_HANDLE *

drivers/platform/aducm3029/flash.c
```diff
@@ -0,0 +1,250 @@
...
+	return SUCCESS;
+}
+
+/**
+ * Read-modify-write helper function for flash_write(). Since the smallest write
+ * unit for the ADuCM3029 is the page this function helps access data smaller than
+ * a page or accross multiple pages.
+ */
+static int32_t flash_write_page(struct flash_dev *dev, uint32_t flash_addr,
+				uint32_t *array, uint32_t array_size)
```
mchindri, line 154: If writting a page why array_size is needed?

drivers/platform/aducm3029/flash.c
```diff
@@ -0,0 +1,250 @@
...
+	if ((flash_addr & 0x3) != 0)
+		return FAILURE;
+	if (((flash_addr & 0x7FF) + (array_size * sizeof(uint32_t))) >
+	    0x800)
+		return FAILURE;
+
+	flash_read(dev, (flash_addr & (~0x7FF)), temp_ptr,
+		   FLASH_PAGE_SIZE_WORDS);
+
+	for(i = 0; i < array_size; i++)
```
mchindri, line 171: memcpy(temp_ptr + (flash_addr & 0x7FF) / 4, array, array_size)

drivers/platform/aducm3029/flash.c
```diff
@@ -0,0 +1,250 @@
...
+{
+	ADI_FEE_TRANSACTION transaction;
+	struct adicup_flash_dev *adicup_extra = dev->extra;
+	uint32_t page_nr, fee_hw_error;
+	uint32_t temp_ptr[FLASH_PAGE_SIZE_WORDS] __attribute__ ((aligned (4)));
+	int32_t ret, i;
+
+	if ((flash_addr & 0x3) != 0)
+		return FAILURE;
+	if (((flash_addr & 0x7FF) + (array_size * sizeof(uint32_t))) >
```
mchindri, line 164: can 0x7FF be a define?

mchindri: or use a macro for (flash_addr & 0x7FF), I assume is ADDR_TO_PAGE(addr)

drivers/platform/aducm3029/flash.c
```diff
@@ -0,0 +1,250 @@
...
+	} while (data_index < initial_size);
+
+	return SUCCESS;
+}
+
+/**
+ * Read data from the flash memory.
+ *
+ * @param [in] dev        - Pointer to the flash device handler.
+ * @param [in] flash_addr - Start address of the write.
```
mchindri, line 236: Start address of read

drivers/platform/aducm3029/flash.c
```diff
@@ -0,0 +1,250 @@
...
+ * @param [out] array     - Pointer to the container for the read data.
+ * @param [in]  size      - Size of the read data in 32-bit words.
+ *
+ * @return void
+ */
+void flash_read(struct flash_dev *dev, uint32_t flash_addr, uint32_t *array,
+		uint32_t size)
+{
+	uint32_t *data_in_flash = (uint32_t *)flash_addr;
```
mchindri, line 245: This is the way to do flash readings?

drivers/platform/aducm3029/flash.c
```diff
@@ -0,0 +1,250 @@
...
+	uint32_t *data		    = array;
+
+	for (uint32_t i = 0; i < size; i++)
```
mchindri, line 248: memcpy

drivers/platform/aducm3029/flash.c
```diff
@@ -0,0 +1,250 @@
...
+
+	return SUCCESS;
+}
+
+/**
+ * Read-modify-write helper function for flash_write(). Since the smallest write
+ * unit for the ADuCM3029 is the page this function helps access data smaller than
+ * a page or accross multiple pages.
+ */
+static int32_t flash_write_page(struct flash_dev *dev, uint32_t flash_addr,
```
buha, line 153: i can't really think of a flash driver without having a public function for manipulating pages - so this function should probably be public and perhaps the flash_dev structure should contain some info about the total size and page size of the flash

drivers/platform/aducm3029/flash.c
adrimbarean: Maybe should rename. This takes care of a write within a page, but we still need to know how much we are writing. What this does is:

1. Read one page (in temporary buffer);
2. Change only the values we want inside the buffer (this is where we use the size);
3. Erase page;
4. Write temporary (size of one page) buffer to page.

mchindri: I would do something like: `flash_write_in_page(struct flash_dev *dev, uint32_t page, uin32_t offset, uint32_t *data, uint32_t len)`

drivers/platform/aducm3029/flash.c
adrimbarean: Yes

adrimbarean:
V1:
 - added `extra` pointer to init param;
 - declare instance and dfp_memory as targeted pointers instead of generic in platform implementation;
 - rename `flash_write_page` helper fuction to `flash_read_then_write`;
 - used `memcpy`;
 - added useful defines;
 - fix comments;
 - use `printf`;
 - added flash page handling API and implemented clear and write page APIs.

include/flash.h
adrimbarean: Added

drivers/platform/aducm3029/flash.c
adrimbarean: Changed

drivers/platform/aducm3029/flash.c
adrimbarean: Done

drivers/platform/aducm3029/flash.c
adrimbarean: Renamed the function to something more intuitive.

drivers/platform/aducm3029/flash.c
adrimbarean: Done

drivers/platform/aducm3029/flash.c
adrimbarean: Done

drivers/platform/aducm3029/flash.c
adrimbarean: Fixed

drivers/platform/aducm3029/flash.c
adrimbarean: done

projects/aducm3029_flash_demo/src/main.c
```diff
@@ -0,0 +1,187 @@
...
+	ret = uart_init(&uart_dut, &uart_init_par);
+	if (ret < 0)
+		return ret;
+
+	ret = flash_init(&flash_dut, &flash_init_par);
+	if (ret < 0)
+		return ret;
+
+	msg_len = sprintf(buff, "UART online.\n");
+	ret = uart_write(uart_dut, (uint8_t *)buff, msg_len);
```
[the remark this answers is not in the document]
adrimbarean: Done

drivers/platform/aducm3029/flash.c
```diff
@@ -0,0 +1,303 @@
...
+ * @return 0 in case of success, negative error code otherwise.
+ */
+int32_t flash_write(struct flash_dev *dev, uint32_t flash_addr,
+		    uint32_t *array, uint32_t array_size)
+{
+	uint32_t data_index = 0, temp, initial_size = array_size;
+	int32_t ret;
+
+	do {
+		temp = (FLASH_PAGE_SIZE_BYTES - FLASH_ADDRESS_TO_PAGE(flash_addr)) / 4;
```
mchindri, line 269: It is strange to me that you substract from bytes number of pages. Something seems not to be ok here or am I not understanding correctly?

drivers/platform/aducm3029/flash.c
adrimbarean: Added `flash_clear_page`, `flash_write_page` and 'flash_read_page', but implemented only `flash_clear_page` and `flash_write_page` for aducm3029 because 'flash_read_page' would only lead to duplicated code with only a few restraints added.

drivers/platform/aducm3029/flash.c
```diff
@@ -0,0 +1,303 @@
...
+	if ((flash_addr & 0x3) != 0)
+		return FAILURE;
+	if ((FLASH_ADDRESS_TO_PAGE(flash_addr) + (array_size * sizeof(uint32_t))) >
+	    0x800)
+		return FAILURE;
+
+	flash_read(dev, (flash_addr & (~FLASH_PAGE_START_ADDRESS_MASK)), temp_ptr,
+		   FLASH_PAGE_SIZE_WORDS);
+
+	memcpy(temp_ptr + FLASH_ADDRESS_TO_PAGE(flash_addr) / 4, array, array_size);
```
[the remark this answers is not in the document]
adrimbarean: Yes, this is the offset. `flash_addr % (FLASH_PAGE_SIZE_WORDS * 4)` will give you offest, but in bytes. And we access the data in 4-byte words, so we would need to do another divide by 4 to get the good offset. 

This is exactly what `FLASH_ADDRESS_TO_PAGE(flash_addr) / 4` does. Just written differently.

adrimbarean:
V2:
 - changed `flash_page_no` to `flash_page_size`;
 - place `extra` member last;
 - remove unnecessary comments;
 - include division by word size in macros;
 - use new `write_page` API in `write` API;
 - remove unnecessary middle pointers;
 - fix doxygen complaints.

include/flash.h
```diff
@@ -0,0 +1,112 @@
...
+ */
+struct flash_dev {
+	/** Flash Device ID */
+	uint8_t id;
+	/** Flash extra parameters (device specific) */
+	void *extra;
+	/** Size of flash memory in bytes (device specific) */
+	uint32_t flash_size;
+	/** Number of pages in flash memory (device specific) */
+	uint32_t flash_page_no;
```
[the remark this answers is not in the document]
adrimbarean: done

include/flash.h
```diff
@@ -0,0 +1,112 @@
...
+
+/**
+ * @struct flash_dev
+ * @brief Flash controller device structure
+ */
+struct flash_dev {
+	/** Flash Device ID */
+	uint8_t id;
+	/** Flash extra parameters (device specific) */
+	void *extra;
```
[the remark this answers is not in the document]
adrimbarean: moved

include/flash.h
```diff
@@ -0,0 +1,112 @@
...
+/******************************************************************************/
+
+/**
+ * @struct flash_dev
+ * @brief Flash controller device structure
+ */
+struct flash_dev {
+	/** Flash Device ID */
+	uint8_t id;
+	/** Flash extra parameters (device specific) */
```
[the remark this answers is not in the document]
adrimbarean: removed

drivers/platform/aducm3029/flash.c
```diff
@@ -0,0 +1,303 @@
...
+
+	transaction.bUseDma = true;
```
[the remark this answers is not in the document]
adrimbarean: Used write_page API instead. No more transaction variable.

drivers/platform/aducm3029/flash.c
```diff
@@ -0,0 +1,303 @@
...
+		return FAILURE;
+
+	/* First erase page */
+	ret = adi_fee_PageErase(adicup_extra->instance, page_nr, page_nr,
+				&fee_hw_error);
+	if(ret != ADI_FEE_SUCCESS)
+		return FAILURE;
+
+	/* Then write */
+	ret = adi_fee_Write(adicup_extra->instance, &transaction,
```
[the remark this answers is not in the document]
adrimbarean: ditto

drivers/platform/aducm3029/flash.c
```diff
@@ -0,0 +1,303 @@
...
+ *
+ * @return 0 in case of success, negative error code otherwise.
+ */
+int32_t flash_read(struct flash_dev *dev, uint32_t flash_addr, uint32_t *array,
+		   uint32_t size)
+{
+	uint32_t *data_in_flash = (uint32_t *)flash_addr;
+	uint32_t *data		    = array;
+
+	memcpy(data, data_in_flash, size);
```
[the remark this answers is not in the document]
adrimbarean: done

drivers/platform/aducm3029/flash.c
adrimbarean: Moved the division inside macro and I think I made it overall more consistent. What do you think?

drivers/platform/aducm3029/flash.c
```diff
@@ -0,0 +1,289 @@
...
+{
+	struct adicup_flash_dev *adicup_extra = dev->extra;
+	uint32_t page_nr, fee_hw_error;
+	uint32_t temp_ptr[FLASH_PAGE_SIZE_WORDS] __attribute__ ((aligned (sizeof(
+				uint32_t))));
+	int32_t ret;
+
+	if ((flash_addr & 0x3) != 0)
+		return FAILURE;
+	if (((FLASH_OFFSET_IN_PAGE(flash_addr) + array_size) * sizeof(uint32_t)) >
```
mchindri, line 216: if array is uin32_t  then array_size should be in words then` * sizeof(uint32_t) `shouldn't be needed.

drivers/platform/aducm3029/flash.c
```diff
@@ -0,0 +1,289 @@
...
+	    0x800)
```
mchindri, line 217: Can you use a macro for this?

drivers/platform/aducm3029/flash.c
mchindri: Or use uint8_t for array

mchindri: Same for the other functions.

adrimbarean:
V3:
 - remove unecessary multiplication;
 - use macro for word count.

drivers/platform/aducm3029/flash.c
adrimbarean: done

drivers/platform/aducm3029/flash.c
adrimbarean: done

include/flash.h
```diff
@@ -0,0 +1,112 @@
...
+ * @struct flash_dev
+ * @brief Flash controller device structure
+ */
+struct flash_dev {
+	/** Flash Device ID */
+	uint8_t id;
+	/** Size of flash memory in bytes */
+	uint32_t flash_size;
+	/** Size of flash page in bytes */
+	uint32_t flash_page_size;
```
buha, line 63: i would've simplified to page_size, but as you wish

drivers/platform/aducm3029/flash.c
```diff
@@ -0,0 +1,289 @@
...
+ * Read-modify-write helper function for flash_write(). Since the smallest write
+ * unit for the ADuCM3029 is the page this function helps access data smaller than
+ * a page or accross multiple pages.
+ */
+static int32_t flash_read_then_write(struct flash_dev *dev, uint32_t flash_addr,
+				     uint32_t *array, uint32_t array_size)
+{
+	struct adicup_flash_dev *adicup_extra = dev->extra;
+	uint32_t page_nr, fee_hw_error;
+	uint32_t temp_ptr[FLASH_PAGE_SIZE_WORDS] __attribute__ ((aligned (sizeof(
```
buha, line 210: This should work ok here but given it's quite large (512 * 4), it would be better to move it to the device structure.
The reason is, this gets allocated on the function stack for each call and you could end up with a stack overflow if there isn't that much memory left for stack.

This is a nasty issue because it happens only when certain functions are called in certain order so as to fill up the stack - so not every call to this function may show the error.

Here's a user to whom this had happened:
https://stackoverflow.com/questions/571945/getting-a-stack-overflow-exception-when-declaring-a-large-array

drivers/platform/aducm3029/flash.c
```diff
@@ -0,0 +1,289 @@
...
+int32_t flash_write_page(struct flash_dev *dev, int32_t page_no, uint32_t *data)
+{
+	ADI_FEE_TRANSACTION transaction;
+	struct adicup_flash_dev *adicup_extra = dev->extra;
+	uint32_t fee_hw_error;
+	int32_t ret;
+
+	transaction.bUseDma = true;
+	transaction.nSize = FLASH_PAGE_SIZE_BYTES;
+	transaction.pWriteAddr = (uint32_t *)(page_no << FLASH_PAGE_ADDR_SHIFT);
```
mchindri, line 189: Whould it make sens to check for page_no to be in a valid range?

drivers/platform/aducm3029/flash.c
```diff
@@ -0,0 +1,289 @@
...
+		if (temp > array_size)
+			temp = array_size;
+		ret = flash_read_then_write(dev, flash_addr, (array + data_index),
+					    temp);
+		if (ret != SUCCESS)
+			return FAILURE;
+		flash_addr += temp;
+		array_size -= temp;
+		data_index += temp;
+	} while (data_index < initial_size);
```
mchindri, line 268: nitpcik: I would simplify this to don't have these many variables if it makes sense
```suggestion
        i = 0;
	while (i < array_size) {
		len = FLASH_PAGE_SIZE_WORDS - FLASH_OFFSET_IN_PAGE(flash_addr + i);
		if (len > array_size - i)
			len = array_size - i;
		ret = flash_read_then_write(dev, flash_addr + i, array + i, len);
		if (ret != SUCCESS)
			return FAILURE;
		i += len;
	}
```

mchindri (commented):
LGTM, I have some minor comments but not really needed. I think it can be merged when last @buha suggestions are implemented.

adrimbarean:
V5:
 - remove unused variables.

drivers/platform/aducm3029/flash.c
```diff
@@ -0,0 +1,294 @@
...
+ * @param [in] flash_addr - Start address of the write.
+ * @param [in] array      - Pointer to the data to be written.
+ * @param [in] array_size - Size of the written data in 32-bit words.
+ *
+ * @return 0 in case of success, negative error code otherwise.
+ */
+int32_t flash_write(struct flash_dev *dev, uint32_t flash_addr,
+		    uint32_t *array, uint32_t array_size)
+{
+	uint32_t data_index = 0, temp, initial_size = array_size;
```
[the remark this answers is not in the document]
adrimbarean: done

PR #1083, opened 2021-08-27
Fix some issues found during development

mchindri:
- Fix xilinx timer drivers
 - Improve util/list.c
 - add bswap_constant_16 to util
 - Add windows commands in iio_demo readme

PR #1086, opened 2021-08-30
drivers:platform:xilinx:Timer: Fixes and improvements

mchindri:
These commit where initially in #1083

### DIFF
--- a/drivers/platform/xilinx/timer.c
+++ b/drivers/platform/xilinx/timer.c
@@ -106,7 +106,14 @@ int32_t timer_init(struct timer_desc **desc,
 			free(xdesc->instance);
 			goto error_xdesc;
 		}
-		XScuTimer_LoadTimer((XScuTimer *)xdesc->instance, dev->load_value);
+		ret = timer_count_clk_set(dev, param->freq_hz);
+		if (IS_ERR_VALUE(ret))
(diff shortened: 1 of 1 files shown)


cristiantpop (approved):

amiclaus (approved):

drivers/platform/xilinx/timer.c
```diff
@@ -402,9 +402,9 @@ int32_t timer_count_clk_set(struct timer_desc *desc, uint32_t freq_hz)
 #ifdef XSCUTIMER_H
 		;
 		uint32_t prescaler = CORE_PRIVATE_TIMER_CLOCK / freq_hz;
-		if (prescaler >= 256)
+		if (prescaler == 0 || prescaler >= 256)
```
buha, line 404: In this case it's acceptable to perform clipping instead of returning FAILURE - that's what's typically done on linux and in some no-OS drivers already, for example:
https://github.com/analogdevicesinc/no-OS/blob/master/drivers/platform/xilinx/xilinx_spi.c#L234-L238

dbogdan (approved):

drivers/platform/xilinx/timer.c
dbogdan: Discussed this.

PR #1087, opened 2021-08-30
[DRAFT] Add iio trigger support

mchindri:
First step will review and merge the support from libtinyiiod: https://github.com/analogdevicesinc/libtinyiiod/pull/43 
Then, to choose an approach for how to implement the trigger, I suggested one but is not the better one, it is just a fast one.
 - How the trigger works:
   - An iio trigger is similar to a iio device, it has attributes and an id like "triggerX" instead of "iio:deviceX" for devices.
   - When a trigger is triggered (irq, manually ...) it should call the callback of the devices that are using that trigger.
   - To set a trigger for a device the "SETTRIG iio:deviceX <triggerId>" command is used. 
 -  The proposed implementation:
    - Extend the `struct iio_device`  with is_trigger field to select if it should act as a trigger.
    - Add add_callback, remove_callback fields which needs to be implemented by the trigger. They will be called when a new devices needs to be notified by the trigger.
    - Add a trigger_callback for a device which will be the one that will be called by the trigger.
    - Implement the logic of add/remove_callback notification in iio.c
    - Add an example of trigger implementation. (It is an implementation for linux platform not a generic one)
 - Improvements to the implementation:
   - Create a `struct iio_triggers` but this will need some refactor of the iio_app API which means update all the project that are using it. 
   - How should we implement a trigger with the current generic files?

### DIFF
--- a/iio/iio.c
+++ b/iio/iio.c
@@ -1077,7 +1077,7 @@ ssize_t iio_step(struct iio_desc *desc)
  * If buff_size is 0, no data will be written to buff, but size will be returned
  */
 static uint32_t iio_generate_device_xml(struct iio_device *device, char *name,
-					int32_t id, char *buff,
+					char * id, char *buff,
 					uint32_t buff_size)
 {
 	struct iio_channel	*ch;
@@ -1099,7 +1099,7 @@ static uint32_t iio_generate_device_xml(struct iio_device *device, char *name,
 
 	i = 0;
 	i += snprintf(buff, max(n - i, 0),
-		      "<device id=\"device%"PRIi32"\" name=\"%s\">", id, name);
+		      "<device id=\"%s\" name=\"%s\">", id, name);
 
 	/* Write channels */
 	if (device->channels)
@@ -1272,10 +1272,12 @@ ssize_t iio_register(struct iio_desc *desc, struct iio_device *dev_descriptor,
 	iio_interface->read_buffer = read_buff;
 	iio_interface->write_buffer = write_buff;
 
+	sprintf((char *)iio_interface->dev_id, "iio:device%d",
+		(int)desc->dev_count);
 	/* Get number of bytes needed for the xml of the new device */
 	n = iio_generate_device_xml(iio_interface->dev_descriptor,
 				    (char *)iio_interface->name,
-				    desc->dev_count, NULL, -1);
+				    iio_interface->dev_id, NULL, -1);
 
 	new_size = desc->xml_size + n;
 	aux = realloc(desc->xml_desc, new_size);
@@ -1295,10 +1297,9 @@ ssize_t iio_register(struct iio_desc *desc, struct iio_device *dev_descriptor,
 	/* Print the new device xml at the end of the xml */
 	iio_generate_device_xml(iio_interface->dev_descriptor,
 				(char *)iio_interface->name,
-				desc->dev_count,
+				iio_interface->dev_id,
 				desc->xml_desc + desc->xml_size_to_last_dev,
 				new_size - desc->xml_size_to_last_dev);
-	sprintf((char *)iio_interface->dev_id, "device%d", (int)desc->dev_count);
 	desc->xml_size_to_last_dev += n;
 	desc->xml_size += n;
 	/* Copy end header at the end */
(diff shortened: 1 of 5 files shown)


iio/iio_trigger/iio_timer_trigger.c
```diff
@@ -0,0 +1,255 @@
...
+
+	while(true)
+	{
+		err = iterator_move_to_idx(desc->calls_iter, 0);
+		while (!IS_ERR_VALUE(err)) {
+			err = iterator_read(desc->calls_iter, (void **)&call);
+			call->callback(call->dev);
+			err = iterator_move(desc->calls_iter, 1);
+		}
+		usleep(1000000);
```
adrimbarean, line 109: Can `msleep` be used?

iio/iio_trigger/iio_timer_trigger.c
```diff
@@ -0,0 +1,255 @@
...
+/*****************************************************************************/
+/************************* Functions Definitions *****************************/
+/*****************************************************************************/
+
+struct timer_trigger_callback {
+	void (*callback)(void * dev);
+	void *dev;
+};
+
+static int32_t iio_cmp_interfaces(struct timer_trigger_callback *a,
```
adrimbarean, line 87: nitpick: could be `iio_cmp_callbacks`

iio/iio_trigger/iio_timer_trigger.c
adrimbarean: Also, why do we need to wait here?

iio/iio_trigger/iio_timer_trigger.c
```diff
@@ -0,0 +1,255 @@
...
+
+static ssize_t iio_attr_set(void *device, char *buf, size_t len,
+			    const struct iio_ch_info *channel, intptr_t priv)
+{
+	int ok = 0;
+        if (priv == IIO_SAMPLING_FREQUENCY)
+                ok++;
+
+        return -EINVAL;
+}
```
adrimbarean, line 192: Why do we have these?

iio/iio_trigger/iio_timer_trigger.c
```diff
@@ -0,0 +1,255 @@
...
+		return FAILURE;
+
+	ldesc = (struct iio_timer_trigger_desc *)calloc(1,
+			sizeof(struct iio_timer_trigger_desc));
+	if (!ldesc)
+		return -ENOMEM;
+
+	
+	ret = list_init(&ldesc->callbacks, LIST_DEFAULT,
+			(f_cmp)iio_cmp_interfaces);
```
adrimbarean, line 129: The comparison function can go in the init_param, I think.

iio/iio_app/iio_app.h
```diff
@@ -50,6 +50,18 @@
...
+	IIO_APP_DEVICE(_name, _dev, _dev_descriptor, _read_buff, NULL)
+
+#define IIO_APP_OUTPUT_DEVICE(_name, _dev, _dev_descriptor, _write_buff) \
+	IIO_APP_DEVICE(_name, _dev, _dev_descriptor, NULL, _write_buff)
+
+#define IIO_APP_NO_BUFFER_DEVICE(_name, _dev, _dev_descriptor) \
+	IIO_APP_DEVICE(_name, _dev, _dev_descriptor, NULL, NULL)
+
+#define IIO_APP_TRIGGER(_name, _dev, _dev_descriptor) \
+	IIO_APP_DEVICE(_name, _dev, _dev_descriptor, NULL, NULL)
```
adrimbarean, line 63: Maybe there's a way to have only one of those.  Maybe `IIO_APP_DEVICE_OR_TRIGGER`.

iio/iio.c
```diff
@@ -1077,7 +1077,7 @@ ssize_t iio_step(struct iio_desc *desc)
  * If buff_size is 0, no data will be written to buff, but size will be returned
  */
 static uint32_t iio_generate_device_xml(struct iio_device *device, char *name,
-					int32_t id, char *buff,
+					char * id, char *buff,
```
adrimbarean, line 1127: Why do we need this?

PR #1090, opened 2021-09-06
irq_platform_ops : adding platform ops to irq files

APorumb:
irq_platform_ops : Modified IRQ to support specific platform operations.

Tested with a simple example.

### DIFF
--- a/drivers/irq/irq.c
+++ b/drivers/irq/irq.c
@@ -1,0 +1,153 @@
+/***************************************************************************//**
+ *   @file   irq.c
+ *   @brief  Implementation of the IRQ Interface
+ *   @author Andrei Porumb (andrei.porumb@analog.com)
+********************************************************************************
+ * Copyright 2021(c) Analog Devices, Inc.
+ *
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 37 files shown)


mchindri:
Why is this needed?

amiclaus (commented):
Please make sure that all no-OS projects that use irq interface are working properly.

dbogdan (approved):

buha (approved):
looks good, projects build now, only thing would be to check one or two projects (like iio_demo on aducm3029 and on xilinx) and see if they behave normally

if irq doesn't work, uart doesn't work and the project test will fail
if project test is ok then it means irq works too

dbogdan:
@APorumb , when you think the pull request it's ready to be merged, please remove the draft flag.

APorumb:
Tested iio_demo on both xilinx (my side) and aducm3025 (@adrimbarean 's side).

Because xilinx_irq.c has 9 functions and aducm3025_irq.c has only 8 functions (trigger_level_set not supported on aducm3029),
I added a null pointer check at line 127 in drivers/irq/irq.c.

Ready to merge!

adrimbarean (approved):
Looks good to me.

PR #1095, opened 2021-09-14
Initial implementation of STM32 IRQ

mphalke:
Added Initial implementation of STM32 IRQ (covers only external interrupts). Tested using STM32F769I-DISCO board.

### DIFF
--- a/drivers/platform/stm32/stm32_gpio_irq.c
+++ b/drivers/platform/stm32/stm32_gpio_irq.c
@@ -1,0 +1,250 @@
+/***************************************************************************//**
+ *   @file   stm32/stm32_gpio_irq.c
+ *   @brief  Implementation of stm32 GPIO IRQ functionality.
+********************************************************************************
+ * Copyright 2022(c) Analog Devices, Inc.
+ *
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *  - Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ *  - Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in
+ *    the documentation and/or other materials provided with the
+ *    distribution.
+ *  - Neither the name of Analog Devices, Inc. nor the names of its
+ *    contributors may be used to endorse or promote products derived
+ *    from this software without specific prior written permission.
+ *  - The use of this software may or may not infringe the patent rights
+ *    of one or more patent holders.  This license does not release you
+ *    from the requirement that you obtain separate licenses from these
+ *    patent holders to use this software.
+ *  - Use of the software either in source or binary form, must be run
+ *    on or directly connected to an Analog Devices Inc. component.
+ *
+ * THIS SOFTWARE IS PROVIDED BY ANALOG DEVICES "AS IS" AND ANY EXPRESS OR
+ * IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, NON-INFRINGEMENT,
+ * MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED.
+ * IN NO EVENT SHALL ANALOG DEVICES BE LIABLE FOR ANY DIRECT, INDIRECT,
+ * INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, INTELLECTUAL PROPERTY RIGHTS, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+*******************************************************************************/
+#include <stdlib.h>
+#include <errno.h>
+#include "no-os/util.h"
+#include "no-os/irq.h"
+#include "stm32_gpio_irq.h"
+
+/**
+* @struct stm32_gpio_irq_callback_desc
+* @brief Structure holding the callback functions for STM32 gpio irqs
+* @note  The callback functions are mapped with 'gpio_irq_id' structure
+*/
+typedef struct stm32_gpio_irq_callback_desc {
+	uint16_t GPIO_Pin;
+	struct callback_desc desc;
+} stm32_gpio_irq_callback_desc[NB_GPIO_INTERRUPTS];
+
+stm32_gpio_irq_callback_desc stm32_gpio_irq_callback;
+
+/******************************************************************************/
+/************************ Functions Definitions *******************************/
+/******************************************************************************/
+
+/**
+ * @brief STM32 GPIO external interrupt callback function
+ * @param GPIO_Pin[in] - GPIO pin
+ * @return none
+ */
+void HAL_GPIO_EXTI_Callback(uint16_t GPIO_Pin)
+{
+	uint8_t id;
+
+	/* Invoke the callback function attached to corresponding GPIO pin */
+	for (id = GPIO_IRQ_ID1; id <= GPIO_IRQ_ID5; id++) {
+		if (stm32_gpio_irq_callback[id].GPIO_Pin == GPIO_Pin) {
+			stm32_gpio_irq_callback[id].desc.callback(stm32_gpio_irq_callback[id].desc.ctx,
+					id, NULL);
+			return;
+		}
+	}
+}
+
+/**
+ * @brief Initialize the STM32 GPIO IRQ Controller
+ * @param desc[in,out] - GPIO IRQ descriptor
+ * @param param[in] - GPIO IRQ initialization parameters
+ * @return SUCCESS in case of success, negative error code otherwise
+ */
+int32_t stm32_gpio_irq_ctrl_init(struct irq_ctrl_desc **desc,
+				 const struct irq_init_param *param)
+{
+	struct irq_ctrl_desc *descriptor;
+	struct stm32_gpio_irq_desc *stm32_desc;
+	int32_t ret;
+
+	if (!desc || !param)
+		return -EINVAL;
+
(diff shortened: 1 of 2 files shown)


drivers/platform/stm32/stm32_irq.c
```diff
@@ -0,0 +1,267 @@
...
+
+	descriptor = (struct irq_ctrl_desc *)malloc(sizeof(*descriptor));
+	if (!descriptor) {
+		ret = -ENOMEM;
+		goto error;
+	}
+
+	extra = (struct stm32_irq_desc*)malloc(sizeof(*extra));
+	if (!extra) {
+		ret = -ENOMEM;
```
amiclaus, line 119: same here. no need to `free(extra)`

drivers/platform/stm32/stm32_irq.c
```diff
@@ -0,0 +1,267 @@
...
+	struct stm32_irq_desc *extra;
+	int32_t ret;
+
+	if (!desc || !param)
+		return -EINVAL;
+
+	descriptor = (struct irq_ctrl_desc *)malloc(sizeof(*descriptor));
+	if (!descriptor) {
+		ret = -ENOMEM;
+		goto error;
```
amiclaus, line 114: here you perform `free(extra)` which I think is redundant.

drivers/platform/stm32/stm32_irq.c
```diff
@@ -0,0 +1,267 @@
...
+
+	extra->irq_type = ((stm32_irq_init_param *)(param->extra))->irq_type;
+
+	descriptor->extra = extra;
+	*desc = descriptor;
+
+	return SUCCESS;
+error:
+	free(extra);
+	free(desc);
```
amiclaus, line 144: `free(descriptor)`?

drivers/platform/stm32/stm32_irq.c
```diff
@@ -0,0 +1,267 @@
...
+	case EXTERNAL_INT_ID1:
+	case EXTERNAL_INT_ID2:
+	case EXTERNAL_INT_ID3:
+	case EXTERNAL_INT_ID4:
+	case EXTERNAL_INT_ID5:
+		extra->gpio_num = ((stm32_irq_init_param *)(param->extra))->gpio_num;
+		break;
+
+	default:
+		return -EINVAL;
```
amiclaus, line 133: `goto error;` ?

drivers/platform/stm32/stm32_irq.h
```diff
@@ -0,0 +1,136 @@
...
+ * @brief stm32 platform specific irq platform ops structure
+ */
+extern const struct irq_platform_ops stm32_irq_platform_ops;
+
+/******************************************************************************/
+/************************ Functions Declarations ******************************/
+/******************************************************************************/
+
+/* Initialize a interrupt controller peripheral. */
+int32_t stm32_irq_ctrl_init(struct irq_ctrl_desc **desc,
```
buha, line 111: All these functions could be deleted from .h since the functionality is exposed through ops.

drivers/platform/stm32/stm32_irq.h
```diff
@@ -0,0 +1,136 @@
...
+#include <stdint.h>
+#include <stdbool.h>
+#include "irq.h"
+#include "stm32_hal.h"
+
+/******************************************************************************/
+/***************************** Include Files **********************************/
+/******************************************************************************/
+
+#include <stdbool.h>
```
buha, line 50: Perhaps you could put stdbool next to the other std... headers,

drivers/platform/stm32/stm32_irq.c
```diff
@@ -0,0 +1,267 @@
...
+ * @param desc - IRQ descriptor
+ * @param irq_id - IRQ ID
+ * @return SUCCESS in case of success, FAILURE otherwise
+ */
+int32_t stm32_irq_disable(struct irq_ctrl_desc *desc, uint32_t irq_id)
+{
+	IRQn_Type irq_type;
+
+	irq_type = ((stm32_irq_desc *)(desc->extra))->irq_type;
+	HAL_NVIC_DisableIRQ(irq_type);
```
buha, line 264: irq_id is ignored, is this the intention ?

drivers/platform/stm32/stm32_irq.c
```diff
@@ -0,0 +1,267 @@
...
+ * @param desc - IRQ descriptor
+ * @param irq_id - IRQ ID
+ * @return SUCCESS in case of success, FAILURE otherwise
+ */
+int32_t stm32_irq_enable(struct irq_ctrl_desc *desc, uint32_t irq_id)
+{
+	IRQn_Type irq_type;
+
+	irq_type = ((stm32_irq_desc *)(desc->extra))->irq_type;
+	HAL_NVIC_EnableIRQ(irq_type);
```
buha, line 248: same here, seems like an error that irq_id is ignored

drivers/platform/stm32/stm32_irq.c
```diff
@@ -0,0 +1,267 @@
...
+}
+
+/**
+ * @brief Free the resources allocated by irq_ctrl_init()
+ * @param desc - IRQ descriptor
+ * @return SUCCESS in case of success, FAILURE otherwise
+ */
+int32_t stm32_irq_ctrl_remove(struct irq_ctrl_desc *desc)
+{
+	if (desc) {
```
buha, line 156: i would simplify this by doing:
```
if (!desc)
    return -EINVAL
```

Then the normal behavior code with return SUCCESS at the end.

buha (commented):
one general comment, i think function input parameters could be checked better (pointers against NULL, other inputs between expected boudaries - if any) and as a general rule, we shouldn't use FAILURE anymore but errors from errno.h like -EINVAL, -EFAULT etc. that actually can carry a meaning

drivers/platform/stm32/stm32_irq.c
mphalke: I got your point. I will update the code accordingly.

drivers/platform/stm32/stm32_irq.c
mphalke: I will add a check for validating irq_id

drivers/platform/stm32/stm32_irq.h
mphalke: Make sense, I will delete them. However I need to move 'stm32_irq_platform_ops' structure variable defined in the stm32_irq.c file at the end to have visibility of the stm32 irq functions.

mphalke:
Hi, I have removed the 2nd commit (out of scope). Please verify it.

amiclaus (approved):

buha:
@mphalke can you please squast the two commits into one ? it's hard to review code like this, we should only be seeing the latest code, not a sequence of patches

mphalke:
@buha I have squashed the commits into single one. I though adding multiple commits will make review easy but looks like its opposite of that. Will take care of it in future..

drivers/platform/stm32/stm32_irq.c
buha: it's still strange to me that the irq_id is ignored and i did some research - it seems that for EXTI on stm32 it is not possible to disable a specific interrupt but the type (in this case, all EXTI interrupts).

it would be cool to enable/disable specifi EXTI interrupts by unmasking/masking them but i'm not sure if the HAL has API for that, could you please check ? 

if it's not possible, then i'm ok with the current version, we're somewhat forced to do it this way by the CMSIS HAL

buha: ok, looks like we canțt go round it, stm32 EXTI is a bit special so no need to change anything ;)

drivers/platform/stm32/stm32_irq.c
```diff
@@ -0,0 +1,278 @@
...
+	if (!descriptor)
+		return -ENOMEM;
+
+	extra = (struct stm32_irq_desc*)malloc(sizeof(*extra));
+	if (!extra) {
+		ret = -ENOMEM;
+		goto error_desc;
+	}
+
+	switch (param->irq_ctrl_id) {
```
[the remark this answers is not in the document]
mphalke: Hi @buha 
Sorry for coming very late on this PR.

Actually I was following the ADuCM3029 irq drivers for STM32 irq driver development. 
1.If you look at the 'aducm3029_irq_ctrl_init()' function from 3029, the 'irq_ctrl_id' is referenced from the 'irq_init_param' structure declared in 'irq.h' include file. That means each platform specific 'irq_extra.h' file will contain interrupt number constants (enum irq_id) which will be assigned to 'irq_ctrl_id' param from 'irq_init_param' structure. So code from lines 107-114 make sense to me as 'irq_id' is referenced from high level (generic) irq init param structure and not the stm32 specific irq init extra structure.
2.Also the 'irq_ctrl_id' is copied into 'desc' but is never used again in 3029. The reason being, each irq function from .c file has explicit 'irq_id' argument, so this irq_id is redundant at later stages. Same thing has been done for STM32 irq drivers I believe.

Please let me know if this make sense and if I understood your original comment.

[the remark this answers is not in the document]
buha: Hi @mphalke ,
Sorry, I missed your comment here.
So we've been going through some thinking and rethinking of how irq platform driver should look like and one colleague came up with this: https://github.com/analogdevicesinc/no-OS/pull/1113
This is quite a long PR discussion because the topic is not trivial.
I think this PR should now be the standard for irq platform drivers and here's an example of the usage:
https://github.com/analogdevicesinc/no-OS/blob/58b25b627e56278d5731aaac4e6506ceb3218c70/projects/ad4110/src/app/main.c#L98-L112
https://github.com/analogdevicesinc/no-OS/blob/master/drivers/afe/ad4110/ad4110.c#L871-L886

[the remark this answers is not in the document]
mphalke: Hi @buha 
Thanks for the comment. Sorry it's not clear to me what change is expected in STM32 irq platform drivers. I looked it over and over again and find that all the functions defined in this file are matching to ones defined in drivers/api/irq.c file and platform ops structure except addition of new function: irq_trigger_level_set(), which is not currently implemented in stm32 irq drivers. Is it expected that this function should be implemented into STM32 irq module because by looking at AD4110 drivers, the irq ID is now passed as GPIO number for which trigger needs to be set. So seems like IRQ ID should not be a hardcoded into platform specific extra file and rather it is attached to GPIO instance?
Could you please throw some light on this?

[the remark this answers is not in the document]
APorumb: Hi @mphalke,

To throw some light, `#define GPIO_IRQ_ID 	52U` refers to GPIO's interrupt ID to be used by the generic interrupt controller (used by Xilinx and not by stm32).  GPIO's interrupt ID is not equal to number of the pin. 

image

mphalke:
Hi @buha 
Shall I delete this PR as I saw new PR created with STM32 IRQ changes by you?

buha:
Let's move the discussion to #1272

PR #1100, opened 2021-09-23
jesd204b: add missing prototypes for adxcvr_read/write

aholtzma-am:


drivers/axi_core/jesd204/axi_adxcvr.h
```diff
@@ -94,4 +94,6 @@ int32_t adxcvr_remove(struct adxcvr *xcvr);
 int32_t adxcvr_clk_set_rate(struct adxcvr *xcvr,
 			    uint32_t rate,
 			    uint32_t parent_rate);
+int32_t adxcvr_write(struct adxcvr *xcvr, uint32_t reg_addr, uint32_t reg_val);
```
cristiantpop, line 97: This functions are used only in
https://github.com/analogdevicesinc/no-OS/blob/cbd8405892eeadc4383a6d8ddc3644fb9a6f4743/drivers/axi_core/jesd204/axi_adxcvr.c
file.
Why do you need them? Probably they should be static.

aholtzma-am: Need access to these registers for monitoring and debug.

adrimbarean (approved):
Looks good to me.

PR #1103, opened 2021-09-28
projects:iio_adpd1080:src: add clock calibarion routines

adrimbarean:
The ADPD1080 features two internal clocks that need to be calibrated: a
32kHz one and a 32MHz one. The 32kHz clock needs to be calibrated using an
external reference clock.

In this case a function is added to calibrate the
32kHz clock against an internal timer of the host controller.

Then a function to calibrate the 32MHz clock against the 32kHz clock is added.

### DIFF
--- a/projects/iio_adpd1080/src/main.c
+++ b/projects/iio_adpd1080/src/main.c
@@ -47,12 +47,161 @@
 #include "iio_adpd188.h"
 #include "iio_app.h"
 #include "aducm3029_gpio.h"
+#include "gpio.h"
+#include "timer.h"
+#include "irq_extra.h"
 
 #define MAX_SIZE_BASE_ADDR		1024
 
 static uint8_t in_buff[MAX_SIZE_BASE_ADDR];
 
 #define ADC_DDR_BASEADDR	((uint32_t)in_buff)
+
+static void adpd1080_sync_gpio_cb(void *ctx, uint32_t event, void *config)
+{
+	int32_t *cnt = ctx;
+
+	*cnt += 1;
+}
+
+static int32_t adpd1080pmod_32k_calib(struct adpd188_dev *adpd1080_dev)
+{
+	int32_t status;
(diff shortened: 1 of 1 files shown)


dbogdan (approved):

projects/iio_adpd1080/src/main.c
```diff
@@ -47,13 +47,168 @@
...
+		goto finish;
+
+finish:
+	fin_ret = irq_ctrl_remove(cal_irq);
+	if(fin_ret != SUCCESS)
+		return FAILURE;
+gpio_finish:
+	fin_ret = gpio_remove(sync_gpio);
+	if(fin_ret != SUCCESS)
+		return FAILURE;
```
amiclaus, line 203: if you return here, `cal_timer` memory might remain allocated.

projects/iio_adpd1080/src/main.c
```diff
@@ -47,13 +47,168 @@
...
+		goto finish;
+
+	status = adpd188_reg_write(adpd1080_dev, ADPD188_REG_GPIO_CTRL, 0);
+	if(status != SUCCESS)
+		goto finish;
+
+finish:
+	fin_ret = irq_ctrl_remove(cal_irq);
+	if(fin_ret != SUCCESS)
+		return FAILURE;
```
amiclaus, line 199: if you return here, `cal_timer` and `sync_gpio` memory might remain allocated.

scuciurean: If freeing the memory inside cal_irq's remove() function is failing, probably freeing the memory reserved for the gpio won't work either.

mchindri: I think it wouldn't be wrong to try :)

adrimbarean:
V1:
 - remove returns at the end that might cause memory leaks.

amiclaus (approved):

PR #1106, opened 2021-09-29
Ad3552r draft. Don't review it yet, created in order to have a link for the driver

mchindri:


mchindri:
Continued here: https://github.com/analogdevicesinc/no-OS/pull/1188

PR #1107, opened 2021-09-30
Minor fixes w.r.t. void pointer typecasting

mphalke:
1. Direct operations done on void pointers which causes build failure on IAR compiler. The build works well with GCC and ARMC compiler but not on IAR. Added typecasting to pointers before performing any operations.

### DIFF
--- a/util/list.c
+++ b/util/list.c
@@ -95,7 +95,7 @@ struct _list_desc {
 /** @brief Default function used to compare element in the list ( \ref f_cmp) */
 static int32_t default_comparator(void *data1, void *data2)
 {
-	return (int32_t)(data1 - data2);
+	return (int32_t)((int32_t *)data1 - (int32_t *)data2);
 }
 
 /**
--- a/iio/iio.c
+++ b/iio/iio.c
@@ -970,7 +970,7 @@ static ssize_t iio_read_dev(const char *device, char *pbuf, size_t offset,
 		if (offset + bytes_count > r_buff->size)
 			return -ENOMEM;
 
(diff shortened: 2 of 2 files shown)


amiclaus (approved):

amiclaus (commented):
nit: the PR description would fit also in the commit description.

util/list.c
```diff
@@ -95,7 +95,7 @@ struct _list_desc {
 /** @brief Default function used to compare element in the list ( \ref f_cmp) */
 static int32_t default_comparator(void *data1, void *data2)
 {
-	return (int32_t)(data1 - data2);
+	return (int32_t)((int32_t *)data1 - (int32_t *)data2);
```
buha, line 98: The original code here is weird. @mchindri can you please explain what this default comparator should do ? If it's comparing pointers that point to data of unknown size then I think you should drop this default comparator - it can lead to weird behaviour. I saw it's not even used.

buha: @mphalke I think it's ok for now to cast to int32_t * even though the default comparator doesn't make sense (to me). At least it unblocks building with IAR.

buha (approved):

util/list.c
mchindri: The list implementation make it possible to work as an ordered list. In order to short the items a comparator function is needed. If no comparator is provided, then this default_comparator function is used. My best though was to compare the addresses because anything else can be compared (the data content is unknown).
In this sense, I think the cast used is not relevant so this change seems to be valid.

PR #1114, opened 2021-10-07
Dev/dac dma examples

ranechita:
This pull request uses the adc/dac buffers for the iio code of adrv9001.


ranechita:
@mchindri I was thinking to add this so that the project makes full use of the buffers and do the iio_app update on another branch

PR #1121, opened 2021-10-20
app_iio: Fix missing init for irq

cristiantpop:
Recent changes to irq subsystem have probably introduced this
error, this commit fixes it.


cristiantpop:
Change log V2:
 
-  update project to use iio_app from iio/iio_app

PR #1134, opened 2021-11-01
Minor fixes to adxrs290 driver to prepare it for trigger support

mchindri:


drivers/gyro/adxrs290/iio_adxrs290.h
```diff
@@ -42,116 +42,7 @@
 
 #include "iio_types.h"
 
-ssize_t get_adxrs290_iio_ch_raw(void *device, char *buf, size_t len,
```
amiclaus, line 45: Were the function headers removed on purpose? I see the non-static implementations are still available in the *.c file.

mchindri: They are in the .c I will make them static.

amiclaus: I suggest making a separate commit for this change.

mchindri:
v2
 - Add static where needed

drivers/gyro/adxrs290/iio_adxrs290.h
amiclaus: Because it is unrelated to the commit title/description

mchindri: Adding static is related to this commit. If removing from .h implies make them static.
Also they are moved because struct iio_adxrs290_descriptor is moved so I think to split this neither would make sense

amiclaus: Ack

amiclaus (approved):
Lgtm, as long as the CI issues are fixed

mchindri:
v3
 - Fix astyle

PR #1135, opened 2021-11-02
drivers:ada4250: device structure and wake delay changes.

pmallick-adi:
1) Added changes in functions so that the device structure
   remains in sync with the actual configuration of the device.
2) Modified the wake delay duration according to the power state of the device.

### DIFF
--- a/drivers/amplifiers/ada4250/ada4250.c
+++ b/drivers/amplifiers/ada4250/ada4250.c
@@ -42,6 +42,7 @@
 /******************************************************************************/
 #include <malloc.h>
 #include "ada4250.h"
+#include "delay.h"
 #include "error.h"
 
 /******************************************************************************/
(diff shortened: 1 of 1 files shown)


drivers/amplifiers/ada4250/ada4250.c
```diff
@@ -164,15 +165,22 @@ int32_t ada4250_soft_reset(struct ada4250_dev *dev)
...
+		ret = ada4250_update(dev, ADA4250_REG_REFBUF_EN, ADA4250_REFBUF_MSK,
+				     ADA4250_REFBUF(refbuf));
 	case ADA4230:
-		return gpio_set_value(dev->gpio_bufen, dev->refbuf_en);
+		ret = gpio_set_value(dev->gpio_bufen, dev->refbuf_en);
 	default:
 		return -ENODEV;
 	}
+
+	if (ret == SUCCESS)
```
buha, line 180: The proposed change is good and mergeable but perhaps you could change a bit the phrasing of this code. I noticed the team currently uses this approach:
https://github.com/analogdevicesinc/no-OS/blob/master/drivers/frequency/hmc7044/hmc7044.c#L288-L294

So if there is an error, quit.
Then set any variable for the no-error scenario.
Finally return SUCCESS or return 0.

It's obviously not mandatory, it's more style related. From my side I'm approving as-is but if you wanna get the style homogenous with rest of noOS code it's up to you.

buha (approved):

amiclaus (approved):
lgtm, but @buha 's argument seems to be valid.

drivers/amplifiers/ada4250/ada4250.c
pmallick-adi: Thanks for the information. I'll update the code with the current styling.

PR #1136, opened 2021-11-02
drivers/platform/xilinx:gpio_irq Added retriggerable option

APorumb:
This option allows the user to disable the nested interrupt.
Previously, if the user wanted to disable the irq pin inside the irq handler,
the nested interrupt would reactivate the pin.

### DIFF
--- a/drivers/platform/xilinx/xilinx_gpio_irq.c
+++ b/drivers/platform/xilinx/xilinx_gpio_irq.c
@@ -93,7 +93,8 @@ static void xil_gpio_irq_handler(struct xil_gpio_irq_desc *param)
 			XGpioPs_IntrDisablePin(&param->my_Gpio, callback_desc->pin_nb);
 			XGpioPs_IntrClearPin(&param->my_Gpio, callback_desc->pin_nb);
 			callback_desc->callback.callback(callback_desc->callback.ctx, 0U, NULL);
-			XGpioPs_IntrEnablePin(&param->my_Gpio, callback_desc->pin_nb);
+			if(param->retriggerable)
+				XGpioPs_IntrEnablePin(&param->my_Gpio, callback_desc->pin_nb);
 		}
 	}
 }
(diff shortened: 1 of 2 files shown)


amiclaus (commented):
are the projects that use irq driver affected in any manner by this change?

APorumb:
> 
> 
> are the projects that use irq driver affected in any manner by this change?

As I know, there isn't any project using this driver yet. Except the one I'm working on right now.

amiclaus (approved):

drivers/platform/xilinx/xilinx_gpio_irq.c
```diff
@@ -93,7 +93,8 @@ static void xil_gpio_irq_handler(struct xil_gpio_irq_desc *param)
 			XGpioPs_IntrDisablePin(&param->my_Gpio, callback_desc->pin_nb);
 			XGpioPs_IntrClearPin(&param->my_Gpio, callback_desc->pin_nb);
 			callback_desc->callback.callback(callback_desc->callback.ctx, 0U, NULL);
-			XGpioPs_IntrEnablePin(&param->my_Gpio, callback_desc->pin_nb);
+			if(param->retriggerable)
```
mchindri, line 96: I think the change was not needed. You can disable the interrupt with irq_disable rigth?
Otherwise will be trigger for every event

APorumb: If you want to disable the interrupt within the irq_handler, using irq_disable won't work because, after the function return, XGpioPs_IntrEnablePin will be called (neutralizing irq_disable effect).

mchindri: You can us a flag set in irq_disable in check that.  For example in xil_callback_desc  have a bool: Is enabled.

mchindri: As you are saying, because you also tried :) is to disable the interrupt from the irq_handler so this is what the use will expect too.

PR #1137, opened 2021-11-03
iio_app refactor

mchindri:
Mainly split the main code into separate functions. The main reason for this was that the wifi for aducm3029 was not working and there were a lot of ifdef to handle. I hope this looks better.
Tested:
  - Aducm UART
  - Aducm WIFI
  - Linux userspace
Will be useful if anyone can do a test on xilinx and maybe stm32

### DIFF
--- a/iio/iio_app/iio_app.c
+++ b/iio/iio_app/iio_app.c
@@ -153,6 +153,53 @@ static int32_t print_uart_error_message(struct uart_desc **uart_desc,
 #else
 	return iio_print_uart_info_message(uart_desc, uart_init_par, message,
 					   msglen);
+#endif
+}
+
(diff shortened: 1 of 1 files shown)


iio/iio_app/iio_app.c
```diff
@@ -56,46 +56,28 @@
 #endif
 
 #ifdef USE_TCP_SOCKET
+#include <stdio.h>
```
amiclaus, line 59: maybe just add `#include <stdio.h>` once at the top of includes

mchindri:
v2: 
 - Fix astyle
 - Fix build
 - Move stdio.h

mchindri:
v3
 - Hope for no more problems with build

amiclaus (approved):

PR #1143, opened 2021-11-05
projects:adrv9001: Use iio_app

mchindri:
Update to use iio_app and refactor a little to simplify the defines.
Not tested


mchindri:
v2
 - Fix astyle

amiclaus (approved):
lgtm, as long as this patch was/is tested.

buha (approved):
same comment as @amiclaus, looks good, needs a test on hardware

buha:
I think 2b47764ff71206baaae9ec8dc54a457f597edc1e might be fishy because I can't get this to run properly.
In debugger it fails here https://github.com/analogdevicesinc/no-OS/blob/master/iio/iio_app/iio_app.c#L267 because somewhere within that function it's checking for the PS/PL type and it's value is junk (https://github.com/analogdevicesinc/no-OS/blob/master/iio/iio_app/iio_app.c#L252)

I also think this variable is not needed https://github.com/analogdevicesinc/no-OS/blob/master/iio/iio_app/iio_app.c#L255

So perhaps have a look at that function, do a test on whatever xilinx PS platform / project you have on hand, and when it works let me know to retest.

mchindri:
v2:
 - rebase onto latest master fix

amiclaus (commented):
Some conflicts have appeared in the meantime.

mchindri:
v3: 
 - Rebase onto master

PR #1150, opened 2021-11-15
Aducm i2c_ops added

APorumb:
-added i2c platform ops for Aducm platform.

-fixed projects build errors

### DIFF
--- a/drivers/platform/aducm3029/aducm3029_i2c.c
+++ b/drivers/platform/aducm3029/aducm3029_i2c.c
@@ -1,5 +1,5 @@
 /***************************************************************************//**
- *   @file   aducm3029/i2c.c
+ *   @file   aducm3029/aducm3029_i2c.c
  *   @brief  Implementation of I2C driver for ADuCM302x
  *   @author Mihail Chindris (mihail.chindris@analog.com)
 ********************************************************************************
--- a/drivers/platform/aducm3029/aducm3029_i2c.c
+++ b/drivers/platform/aducm3029/aducm3029_i2c.c
@@ -127,8 +127,8 @@ static uint32_t set_transmission_configuration(struct i2c_desc *desc)
  * it is not used and must be set to NULL.
  * @return \ref SUCCESS in case of success, \ref FAILURE otherwise.
  */
-int32_t i2c_init(struct i2c_desc **desc,
-		 const struct i2c_init_param *param)
+static int32_t aducm3029_i2c_init(struct i2c_desc **desc,
+				  const struct i2c_init_param *param)
 {
 	struct aducm_i2c_extra *aducm_i2c;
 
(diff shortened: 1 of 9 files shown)


amiclaus (commented):
One question on my side: were you able to test the projects with the new i2c ops implementation?

APorumb:
> 
> 
> One question on my side: were you able to test the projects with the new i2c ops implementation?

I created a new project to test it and it worked as expected. I did not test it on older projects.

amiclaus (approved):
lgtm, as long as we test on at least one project that is already on the no-OS repo.

drivers/platform/aducm3029/aducm3029_i2c.c
```diff
@@ -74,17 +75,6 @@ static uint32_t last_bitrate = UNINITIALIZED_BITRATE;
...
- * @struct i2c_extra
- * @brief I2C platform specific parameters.
- */
-struct aducm_i2c_extra {
-	/** Pointer to the prologue data. */
-	uint8_t *prologue_data;
-	/** Prologue size. */
-	uint8_t prologue_size;
-};
-
```
mchindri, line 87: There is no reason to move this to the .h It is only used here. No external code need to access this fields

drivers/platform/aducm3029/aducm3029_i2c.c
```diff
@@ -126,8 +116,8 @@ static uint32_t set_transmission_configuration(struct i2c_desc *desc)
  * it is not used and must be set to NULL.
  * @return \ref SUCCESS in case of success, \ref FAILURE otherwise.
  */
-int32_t i2c_init(struct i2c_desc **desc,
-		 const struct i2c_init_param *param)
+int32_t aducm3029_i2c_init(struct i2c_desc **desc,
```
mchindri, line 119: make functions static. The should only be used through ops.

mchindri: nitpick. If function became static, rename is not needed, but this is not a big issue anyway

drivers/platform/aducm3029/i2c_extra.h
```diff
@@ -0,0 +1,69 @@
...
+*******************************************************************************/
+
+#ifndef I2C_EXTRA_H_
+#define I2C_EXTRA_H_
+
+/******************************************************************************/
+/***************************** Include Files **********************************/
+/******************************************************************************/
+
+#include <stdint.h>
```
mchindri, line 47: If removing aducm_i2c_extra  this won't be needed

drivers/platform/aducm3029/aducm3029_i2c.c
APorumb: The idea was to keep the code consistent (like any other _extra.h file).

drivers/platform/aducm3029/aducm3029_i2c.c
APorumb: Same, wanted to keep the code consistent. Function definitions are not included in the header file so they should be used only through ops.

mchindri: This means that they should be static :)

APorumb:
V2:

-reworked commit history to not have failed tests for each of the commits
-made aducm_i2c functions static

mchindri (approved):

PR #1152, opened 2021-11-19
remove src.mk whitespace

buha:
Some src.mk files use whitespace to align the line break \ character
in the Makefile. Whitespace used is either tab or space or a mixture
of both. This leads to awkward looking Makefiles that have either
missaligned backslashes and are harder to maintain because typing
tab tab tab tab ... backslash is always longer than typing space
backslash.

Command used to to this:
```
cd projects
find . -name "src.mk" -print0 | xargs -0 sed -i 's/\s*\\/ \\/'
```

### DIFF
--- a/projects/ad400x-fmcz/src.mk
+++ b/projects/ad400x-fmcz/src.mk
@@ -10,27 +10,27 @@
 ################################################################################
 
 SRCS += $(PROJECT)/src/ad400x_fmcz.c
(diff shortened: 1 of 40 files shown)


amiclaus (approved):
lgtm

buha:
V2:
- added a second commit that fixes iio_demo on stm32

mchindri (approved):

Issue #1153, opened 2021-11-20
Xilinx GPIO input routine fails for .number>=32


prwang:
https://github.com/analogdevicesinc/no-OS/blob/304003481c3ed2c0c6f45fe66622a35c7d59d315/drivers/platform/xilinx/xilinx_gpio.c#L223

This line permanently reduced desc->number by 32 and all following xil_gpio_get_value will read the wrong GPIO pin.

mchindri:
fixed: https://github.com/analogdevicesinc/no-OS/pull/1154

PR #1154, opened 2021-11-22
drivers:platform:xilinx:xilinx_gpio: Small fix

mchindri:
In xil_gpio_direction_input desc->number
is updated. Use local variable as in 
xil_gpio_direction_output


mchindri:
To fix: https://github.com/analogdevicesinc/no-OS/issues/1153

amiclaus (approved):
I see that this approach is already used in `xil_gpio_direction_output` so lgtm.

PR #1171, opened 2021-12-08
projects:ad9081: Use buffers for iio

mchindri:
Raw addresses were used as buffers. They may overlap with other
program parts and break it. Use buffers instead.


Demon000:
Typo: where -> were?

amiclaus (approved):

PR #1180, opened 2021-12-14
drivers:iio:adc:mykonos: fix ARM state check routine

adrimbarean:
Correctly use a bitwise operation to check for the desired ARM state
instead of logical operation.

### DIFF
--- a/projects/ad9371/src/devices/mykonos/mykonos.c
+++ b/projects/ad9371/src/devices/mykonos/mykonos.c
@@ -13539,7 +13539,7 @@ mykonosErr_t MYKONOS_checkArmState(mykonosDevice_t *device, mykonosArmState_t ar
                 break;
         }
 
-        if (armStateCheck && armStatusMapped)
+        if ((armStateCheck & armStatusMapped) || !(armStateCheck || armStatusMapped))
         {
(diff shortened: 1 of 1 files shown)


amiclaus (commented):
if both` armStateCheck` and `armStatusMapped` are in `MYK_ARM_POWERUP` state , this approach might not work.

adrimbarean:
V1:
 - changed condition checking to use "==" to detect 0 mapped MYK_ARM_POWERUP.

amiclaus (approved):

dbogdan (approved):

amiclaus (commented):
needs to be synced with linux version of the patch :) (https://github.com/analogdevicesinc/linux/pull/1760)

adrimbarean:
V2:
 - changed to `if ((armStateCheck & armStatusMapped) || !(armStateCheck || armStatusMapped))` to keep bitwise operation approach and check the `MYK_ARM_POWERUP` case. (synced with analogdevicesinc/linux#1760)

adrimbarean:
Quick note that the [corresponding linux PR](https://github.com/analogdevicesinc/linux/pull/1760) is mereged.

PR #1190, opened 2021-12-23
Staging/ad5758 fix

APorumb:
Fix ad5758 driver and project gpio output.


mchindri (approved):

amiclaus (commented):
Why 2 commits instead of one? Both are fixing gpio issues. Moreover, the description is the same for both of them.

mchindri:
> Why 2 commits instead of one? Both are fixing gpio issues. Moreover, the description is the same for both of them.

Maybe a better commit description whould help.
"1: Make gpio as output in order to be able to set the value"
"2: Fix gpio offset number, wrong number was used"

APorumb:
V2:

Commit descriptions reworked.

amiclaus (approved):

PR #1140, opened 2021-11-04
Update all projects with iio_app

mchindri:

PR #1195, opened 2022-01-04
iio:Create non-blocking iiod

mchindri:
This is an improvement of libtinyiiod.
https://github.com/analogdevicesinc/libtinyiiod

The main improvments are:
 - Non-blocking processing function.
   This is needed in order to support triggers in single threaded
   enviroment like No-OS.
 - Handle multiple connection instances (e.g. Sockets)
 - Connections and application instance are passed as ctx
   to functions.

It was tested with minor changes in iio on:
  - **zed**:  xilinx **sockets** and **uart** with **ad9361** and **adxrs290**
  - **zc706**: xilinx **sockets** and **uart** with **fmcdaq2**
 Both dac buffer write and plot works.
 
 I will add iio change in an other PR because #1140 needs to be merge before any update

### DIFF
--- a/iio/iiod.c
+++ b/iio/iiod.c
@@ -1,0 +1,830 @@
+/***************************************************************************//**
+ *   @file   iiod.c
+ *   @brief  Nonblockin implementation of libtinyiiod.
+ *           Based on https://github.com/analogdevicesinc/libtinyiiod
+ *   @author Mihail Chindris (mihail.chindris@analog.com)
+********************************************************************************
+ * Copyright 2022(c) Analog Devices, Inc.
+ *
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *  - Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ *  - Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in
+ *    the documentation and/or other materials provided with the
+ *    distribution.
+ *  - Neither the name of Analog Devices, Inc. nor the names of its
+ *    contributors may be used to endorse or promote products derived
+ *    from this software without specific prior written permission.
+ *  - The use of this software may or may not infringe the patent rights
+ *    of one or more patent holders.  This license does not release you
+ *    from the requirement that you obtain separate licenses from these
+ *    patent holders to use this software.
+ *  - Use of the software either in source or binary form, must be run
+ *    on or directly connected to an Analog Devices Inc. component.
+ *
+ * THIS SOFTWARE IS PROVIDED BY ANALOG DEVICES "AS IS" AND ANY EXPRESS OR
+ * IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, NON-INFRINGEMENT,
+ * MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED.
+ * IN NO EVENT SHALL ANALOG DEVICES BE LIABLE FOR ANY DIRECT, INDIRECT,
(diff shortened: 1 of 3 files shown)


mchindri:
v2:
 - Replace iiod_conn_step with while loop and create iiod_run_state. Each state will return and flow is simplified
 - Move static local variable to global scope

mchindri:
> You return SUCCESS on an error case?
> 
> Also - you could do
> 
> ```
> if (IS_ERR_VALUE(ret)) {
>     ....
>     conn->state = IIOD_WRITING_CMD_RESULT;
> } else if (conn->cmd_data.cmd == IIOD_CMD_WRITE) {
>     conn->state = IIOD_READING_WRITE_DATA;
> } else {
>     conn->state = IIOD_RUNNING_CMD;
> }
> 
> return SUCCESS;
> ```

Well, it is an error related to an iiod line that is waiting a response and we need to respond with an error code.
It is not an error regarding the states. I will add a comment to try to explain this.

mchindri:
v3:
 - Refactor ifs
 - Add comment to iiod_run_state

iio/iiod.c
```diff
@@ -0,0 +1,814 @@
...
+	[IIOD_CMD_CLOSE]	= IIOD_STR("CLOSE"),
+	[IIOD_CMD_READ]		= IIOD_STR("READ"),
+	[IIOD_CMD_WRITE]	= IIOD_STR("WRITE"),
+	[IIOD_CMD_READBUF]	= IIOD_STR("READBUF"),
+	[IIOD_CMD_WRITEBUF]	= IIOD_STR("WRITEBUF"),
+	[IIOD_CMD_GETTRIG]	= IIOD_STR("GETTRIG"),
+	[IIOD_CMD_SETTRIG]	= IIOD_STR("SETTRIG"),
+	[IIOD_CMD_SET]		= IIOD_STR("SET")
+};
+static const uint32_t priority_array[ARRAY_SIZE(cmds)] = {
```
pcercuei, line 72: The size of the array will be derived from the content. If you want to make sure both arrays are in sync, I'd suggest to use a `static_assert(ARRAY_SIZE(cmds) == ARRAY_SIZE(priority_array), "Arrays must have the same size");`

iio/iiod.c
```diff
@@ -0,0 +1,814 @@
...
+	IIOD_CMD_TIMEOUT,
+	IIOD_CMD_VERSION,
+	IIOD_CMD_GETTRIG,
+	IIOD_CMD_SETTRIG,
+	IIOD_CMD_HELP,
+	IIOD_CMD_SET
+};
+
+/* Set res->cmd to corresponding cmd and return the processed length of buf */
+static int32_t parse_cmd(char *token, struct comand_desc *res)
```
pcercuei, line 92: `token` could be a `const char *`

iio/iiod.c
```diff
@@ -0,0 +1,814 @@
...
+	token = strtok(NULL, delim);
+	if (!token)
+		return -EINVAL;
+
+	return parse_num(token, &res->count, 10);
+}
+
+static int32_t iiod_parse_rw_attr(char *token, struct comand_desc *res)
+{
+	static char *types_strs[] = {
```
pcercuei, line 171: Static array outside the function would be better.

pcercuei:
Btw - if it is possible in your case, I'd suggest to use `strtok_r` instead of `strtok`.

mchindri:
> Btw - if it is possible in your case, I'd suggest to use `strtok_r` instead of `strtok`.

Very good point, thanks

mchindri:
v3
 - use static assert
 - use const char * where appropiate
 - move types_strs outside funct and rename to attr_types_strs
 - add strtok_ctx in iiod_conn and use strtok_r

mchindri:
v4
 - Fix astyle

pcercuei (approved):

dbogdan (approved):

PR #1196, opened 2022-01-05
"Small" refactor in iio.c

mchindri:
- Remove iio_register and use init param. No project use register during runtime. This simplify a lot structures in desc and xml is generated only once.
 - Some naming changes.

### DIFF
--- a/iio/iio.c
+++ b/iio/iio.c
@@ -342,7 +342,7 @@ static inline struct iio_channel *iio_get_channel(const char *channel,
  * @param iio_dev_privs - List of interfaces.
  * @return Interface pointer if interface is found, NULL otherwise.
  */
-static struct iio_dev_priv *iio_get_interface(const char *device_name)
+static struct iio_dev_priv *get_iio_device(const char *device_name)
 {
 	uint32_t i;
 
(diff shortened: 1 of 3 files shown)


mchindri:
v2:
 - Fix cppcheck

buha:
So I looked at all commits and agree with all changes.
There is just a slight problem of how the code is split into commits I think, for example:
 - Commit no 5 does "Rename iio_interface -> iio_dev_priv" but I see already iio_dev_priv in commit no 2
 - This whole thing about commenting lines in commit no 2 in functions to be removed in commit no 3 doesn't look good, I think you should just merge 2 & 3
 - Also commit 2 does a bit more than what it says

mchindri:
> So I looked at all commits and agree with all changes. There is just a slight problem of how the code is split into commits I think, for example:
> 
> * Commit no 5 does "Rename iio_interface -> iio_dev_priv" but I see already iio_dev_priv in commit no 2
> * This whole thing about commenting lines in commit no 2 in functions to be removed in commit no 3 doesn't look good, I think you should just merge 2 & 3
> * Also commit 2 does a bit more than what it says

I didn't like how the diff of 2 and 3 together was looking that's why I did it this way. I will try it again.
Will do the suggestions soon.

mchindri:
v3:
 - Do a better split of commit 2 and 3
 - move wrong renaming to correct commit

buha (approved):

amiclaus (commented):
Was this PR tested with at least one no-OS project that has IIO_SUPPORT?

mchindri:
> Was this PR tested with at least one no-OS project that has IIO_SUPPORT?

Yes, ad9361

PR #1197, opened 2022-01-05
drivers: dac: ltc2688: Minor fix

pmallick-adi:
Added a missing return statement before the error code segment.

### DIFF
--- a/drivers/dac/ltc2688/ltc2688.c
+++ b/drivers/dac/ltc2688/ltc2688.c
@@ -464,13 +464,12 @@ int32_t ltc2688_init(struct ltc2688_dev **device,
 		goto error;
 
 	*device = dev;
+	printf("LTC2688 successfully initialized\n");
 
-	if (!ret)
-		printf("LTC2688 successfully initialized\n");
-	else
-		printf("LTC2688 initialization error (%d)\n", ret);
+	return ret;
 
 error:
+	printf("LTC2688 initialization error (%d)\n", ret);
 	free(dev);
 	return ret;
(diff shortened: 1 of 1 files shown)


drivers/dac/ltc2688/ltc2688.c
```diff
@@ -469,6 +469,8 @@ int32_t ltc2688_init(struct ltc2688_dev **device,
 		printf("LTC2688 successfully initialized\n");
 	else
 		printf("LTC2688 initialization error (%d)\n", ret);
+	
+	return ret;
```
mchindri, line 473: Put this ret inside if (!ret)
On error dev must be released.

Thanks for the fix.

drivers/dac/ltc2688/ltc2688.c
```diff
@@ -469,6 +469,8 @@ int32_t ltc2688_init(struct ltc2688_dev **device,
 		printf("LTC2688 successfully initialized\n");
 	else
 		printf("LTC2688 initialization error (%d)\n", ret);
+
```
mchindri, line 472: can you do?

mchindri: This way the message is printed at each error.

Thanks for the fix.

pmallick-adi: Sure will do this.

pmallick-adi:
Change Log: 

Added changes suggested in the PR.

mchindri (approved):

PR #1202, opened 2022-01-10
drivers: ada4250: Minor Updates

pmallick-adi:
Apparently the logic is reversed for this one. When in low BW mode, the BW pin should tied to AVDD and when in high BW mode, the BW pin should be tied to AVSS.
The low BW mode is actually the standard response while the high BW mode is more of a BW boost feature.

### DIFF
--- a/drivers/amplifiers/ada4250/ada4250.c
+++ b/drivers/amplifiers/ada4250/ada4250.c
@@ -338,10 +338,10 @@ int32_t ada4250_set_bandwidth(struct ada4250_dev *dev,
 
 	switch (bw) {
 	case ADA4250_BANDWIDTH_LOW:
-		val = GPIO_LOW;
+		val = GPIO_HIGH;
 		break;
 	case ADA4250_BANDWIDTH_HIGH:
-		val = GPIO_HIGH;
+		val = GPIO_LOW;
 		break;
 	default:
 		return -EINVAL;
--- a/drivers/amplifiers/ada4250/ada4250.c
+++ b/drivers/amplifiers/ada4250/ada4250.c
@@ -129,6 +129,40 @@ int32_t ada4250_update(struct ada4250_dev *dev, uint8_t reg_addr,
 }
 
 /**
+ * @brief Update ADA4250 device descriptor.
+ * @param dev - The device structure.
+ * @return Returns SUCCESS in case of success or negative error code otherwise.
+ */
+int32_t ada4250_update_desc(struct ada4250_dev *dev)
+{
+	uint8_t reg_data;
+	int32_t ret;
+
+	ret = ada4250_read(dev, ADA4250_REG_GAIN_MUX, &reg_data);
+	if (ret != SUCCESS)
+		return ret;
+	dev->gain = reg_data & ADA4250_GAIN_MUX_MSK;
+
+	ret = ada4250_read(dev, ADA4250_REG_REFBUF_EN, &reg_data);
+	if (ret != SUCCESS)
+		return ret;
+	dev->refbuf_en = reg_data & ADA4250_REFBUF_MSK;
+
+	ret = ada4250_read(dev, ADA4250_REG_SNSR_CAL_VAL, &reg_data);
+	if (ret != SUCCESS)
+		return ret;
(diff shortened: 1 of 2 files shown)


drivers/amplifiers/ada4250/ada4250.c
```diff
@@ -151,9 +151,33 @@ int32_t ada4250_soft_reset(struct ada4250_dev *dev)
 			return ret;
 
 		if(!(data & ADA4250_RESET(ADA4250_RESET_ENABLE)))
-			return SUCCESS;
+			break;
 	}
 
+	ret = ada4250_read(dev, ADA4250_REG_GAIN_MUX, &data);
+	if (ret != SUCCESS)
+		goto error_read;
```
amiclaus, line 159: nit: no need to do conditional jumps in this situation. Just return directly.

amiclaus (commented):
one nit on my side + please fix CI checks.

pmallick-adi:
Change log: 

Added code to update the device descriptor parameters after soft_reset.

drivers/amplifiers/ada4250/ada4250.c
pmallick-adi: Okay will update this

pmallick-adi:
Change log:

Updated the code to directly return in the soft_reset function.

drivers/amplifiers/ada4250/ada4250.c
```diff
@@ -151,10 +151,31 @@ int32_t ada4250_soft_reset(struct ada4250_dev *dev)
 			return ret;
 
 		if(!(data & ADA4250_RESET(ADA4250_RESET_ENABLE)))
-			return SUCCESS;
+			break;
 	}
 
-	return FAILURE;
```
adrimbarean, line 157: I think we still need to return `FAILURE` in case of timeout. I suggest making a function that does the descriptor update by itself and calling it in the `if` clause above and keep the returns.

drivers/amplifiers/ada4250/ada4250.c
```diff
@@ -463,13 +484,10 @@ int32_t ada4250_set_normal_mode(struct ada4250_dev *dev, bool reconfig)
...
 				return ret;
-		} else { /* Reset driver configurations */
-			dev->gain = ADA4250_GAIN_1;
-			dev->refbuf_en = false;
-			dev->power_mode = ADA4250_POWER_NORMAL;
-			dev->bias = ADA4250_BIAS_DISABLE;
-			dev->offset_range = ADA4250_RANGE1;
-			dev->bandwidth = ADA4250_BANDWIDTH_LOW;
+		} else { /* Reset device configurations */
+			ret = ada4250_soft_reset(dev);
```
adrimbarean, line 488: Question: are we sure we want to do a software reset here? Before it was only a handler update, now it's actually a device command being run.

adrimbarean (changes requested):
Unless we have a good reason to omit the timeout issue described in an earlier comment I think we need to update.

drivers/amplifiers/ada4250/ada4250.c
pmallick-adi: The handler should reflect the actual device parameter and only updating the handler would not not necessarily be in sync with the actual device parameters.

drivers/amplifiers/ada4250/ada4250.c
pmallick-adi: Will follow this approach.

drivers/amplifiers/ada4250/ada4250.c
```diff
@@ -150,11 +184,19 @@ int32_t ada4250_soft_reset(struct ada4250_dev *dev)
...
+			ret = ada4250_update_desc(dev);
+			if (ret != SUCCESS)
+				return ret;
+			else
+				break;
+		}
 	}
 
-	return FAILURE;
+	if (timeout<1)
```
amiclaus, line 196: nit: `(timeout<1)` -> `(timeout < 1)` (with spaces)

pmallick-adi:
Change Log:

Added spaces between the timeout if clause.

adrimbarean (approved):
Looks good to me.

amiclaus (approved):

PR #1207, opened 2022-01-13
drivers:platform:maxim Maxim platform drivers

CiprianRegus:
Initial implementation of peripheral drivers for MAX32660 (UART, RTC,
SPI, GPIO and IRQ).

Signed-off-by: Ciprian Regus <ciprian.regus@analog.com>

### DIFF
--- a/drivers/platform/maxim/gpio_extra.h
+++ b/drivers/platform/maxim/gpio_extra.h
@@ -1,0 +1,81 @@
+/***************************************************************************//**
+ *   @file   maxim/gpio_extra.h
+ *   @brief  Header file for maxim gpio specifics.
+ *   @author Ciprian Regus (ciprian.regus@analog.com)
+********************************************************************************
+ * Copyright 2022(c) Analog Devices, Inc.
+ *
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *  - Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ *  - Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in
+ *    the documentation and/or other materials provided with the
+ *    distribution.
+ *  - Neither the name of Analog Devices, Inc. nor the names of its
+ *    contributors may be used to endorse or promote products derived
+ *    from this software without specific prior written permission.
+ *  - The use of this software may or may not infringe the patent rights
+ *    of one or more patent holders.  This license does not release you
+ *    from the requirement that you obtain separate licenses from these
+ *    patent holders to use this software.
+ *  - Use of the software either in source or binary form, must be run
+ *    on or directly connected to an Analog Devices Inc. component.
+ *
+ * THIS SOFTWARE IS PROVIDED BY ANALOG DEVICES "AS IS" AND ANY EXPRESS OR
+ * IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, NON-INFRINGEMENT,
+ * MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED.
+ * IN NO EVENT SHALL ANALOG DEVICES BE LIABLE FOR ANY DIRECT, INDIRECT,
+ * INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, INTELLECTUAL PROPERTY RIGHTS, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+*******************************************************************************/
+#ifndef GPIO_EXTRA_H_
+#define GPIO_EXTRA_H_
+
+#include <stdbool.h>
+#include <stdint.h>
+#include "no-os/irq.h"
+#include "no-os/gpio.h"
+#include "max32660.h"
+
+#define N_PINS	MXC_CFG_GPIO_PINS_PORT
+#define N_PORTS	MXC_CFG_GPIO_INSTANCES
+
+/**
+ * @brief maxim platform specific gpio platform ops structure
+ */
+extern const struct gpio_platform_ops max_gpio_ops;
+/**
+ * @brief maxim platform specific gpio irq platform ops structure
+ */
+extern const struct irq_platform_ops max_gpio_irq_ops;
+
+/**
+ * @struct gpio_irq_config
+ * @brief maxim configuration descriptor for irq operations
+ */
+struct gpio_irq_config {
+	enum irq_trig_level mode;
+};
+
+struct max_gpio_init_param {
+	/** Input/Output */
+	uint32_t mode;
+	/** Pull_up/Pull_down  */
+	uint32_t pull;
+	uint32_t port;
+};
+
+struct max_gpio_irq_param {
+	struct irq_ctrl_desc *parent;
+	struct gpio_desc *desc
+	};
+
+#endif
--- a/drivers/platform/maxim/maxim_gpio.c
+++ b/drivers/platform/maxim/maxim_gpio.c
@@ -1,0 +1,594 @@
+/***************************************************************************//**
+ *   @file   maxim/maxim_gpio.c
+ *   @brief  Implementation of gpio driver.
+ *   @author Ciprian Regus (ciprian.regus@analog.com)
+********************************************************************************
+ * Copyright 2022(c) Analog Devices, Inc.
+ *
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *  - Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ *  - Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in
+ *    the documentation and/or other materials provided with the
+ *    distribution.
+ *  - Neither the name of Analog Devices, Inc. nor the names of its
+ *    contributors may be used to endorse or promote products derived
+ *    from this software without specific prior written permission.
+ *  - The use of this software may or may not infringe the patent rights
+ *    of one or more patent holders.  This license does not release you
+ *    from the requirement that you obtain separate licenses from these
+ *    patent holders to use this software.
+ *  - Use of the software either in source or binary form, must be run
+ *    on or directly connected to an Analog Devices Inc. component.
+ *
+ * THIS SOFTWARE IS PROVIDED BY ANALOG DEVICES "AS IS" AND ANY EXPRESS OR
+ * IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, NON-INFRINGEMENT,
+ * MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED.
+ * IN NO EVENT SHALL ANALOG DEVICES BE LIABLE FOR ANY DIRECT, INDIRECT,
+ * INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, INTELLECTUAL PROPERTY RIGHTS, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+*******************************************************************************/
+
+#include <stdio.h>
+#include <errno.h>
+#include <stdlib.h>
+#include "no-os/gpio.h"
+#include "no-os/irq.h"
+#include "no-os/util.h"
+#include "gpio.h"
+#include "gpio_regs.h"
+#include "gpio_extra.h"
+#include "irq_extra.h"
+#include "max32660.h"
+#include "mxc_errors.h"
+
+static struct callback_desc *gpio_callback[N_PORTS][N_PINS];
+
+/******************************************************************************/
+/************************ Functions Definitions *******************************/
+/******************************************************************************/
+
+static void _gpio_irq(uint8_t port)
+{
+	uint32_t pin;
+	mxc_gpio_regs_t *gpio_regs = MXC_GPIO_GET_GPIO(port);
+	uint32_t stat_reg = gpio_regs->int_stat;
+	/** Clear interrupt flags for the current port*/
+	gpio_regs->int_clr = stat_reg;
+	while(stat_reg) {
+		pin = find_first_set_bit(stat_reg);
+		if (!gpio_callback[port][pin]) {
+			stat_reg >>= pin + 1;
+			continue;
+		}
+		void *ctx = gpio_callback[port][pin]->ctx;
+		gpio_callback[port][pin]->callback(ctx, pin, NULL);
+
+		stat_reg >>= pin + 1;
+	}
+}
+
+void GPIO0_IRQHandler()
+{
+	_gpio_irq(0);
+}
+
+/**
+ * @brief Obtain the GPIO decriptor.
+ * @param desc - The GPIO descriptor.
+ * @param param - GPIO initialization parameters
+ * @return 0 in case of success, errno error codes otherwise.
+ */
+int32_t max_gpio_get(struct gpio_desc **desc,
+		     const struct gpio_init_param *param)
+{
+	int32_t ret;
+	gpio_cfg_t *g_cfg;
+	struct max_gpio_init_param *pextra;
+	uint32_t m_pad, m_func;
+	struct gpio_desc *descriptor;
+
+	if (!param || !param->extra || param->number >= N_PINS)
+		return -EINVAL;
+
+	descriptor = calloc(1, sizeof(*descriptor));
+	if (!descriptor)
+		return -ENOMEM;
+
+	g_cfg = calloc(1, sizeof(*g_cfg));
+	if (!g_cfg) {
+		ret = -ENOMEM;
+		goto free_descriptor;
+	}
+
+	pextra = param->extra;
+	m_pad = (pextra->pull == 0) ? GPIO_PAD_PULL_UP :  GPIO_PAD_PULL_DOWN;
+	m_func = (pextra->mode == 0) ? GPIO_FUNC_IN : GPIO_FUNC_OUT;
+
+	if (pextra->port >= N_PORTS) {
+		ret = -EINVAL;
+		goto free_g_cfg;
+	}
+
+	g_cfg->port = pextra->port;
+	g_cfg->mask = BIT(param->number);
+	g_cfg->pad = m_pad;
+	g_cfg->func = m_func;
+
+	descriptor->number = param->number;
+	descriptor->platform_ops = param->platform_ops;
+	descriptor->extra = g_cfg;
+
+	GPIO_Config(descriptor->extra);
+
+	*desc = descriptor;
+
+	return 0;
+
+free_g_cfg:
+	free(g_cfg);
+free_descriptor:
+	free(descriptor);
+
+	return ret;
+}
+
+/**
+ * @brief Get the value of an optional GPIO.
+ * @param desc - The GPIO descriptor.
+ * @param param - GPIO initialization parameters
+ * @return 0 in case of success, errno error codes otherwise.
+ */
+int32_t max_gpio_get_optional(struct gpio_desc **desc,
+			      const struct gpio_init_param *param)
+{
+	if (param == NULL) {
+		*desc = NULL;
+		return 0;
+	}
+
+	return max_gpio_get(desc, param);
+}
+
+/**
+ * @brief Free the resources allocated by gpio_get().
+ * @param desc - The GPIO descriptor.
+ * @return 0 in case of success, errno error codes otherwise.
+ */
+int32_t max_gpio_remove(struct gpio_desc *desc)
+{
+	if (!desc)
+		return -EINVAL;
+
+	free(desc->extra);
+	free(desc);
+
+	return 0;
+}
+
+/**
+ * @brief Enable the input direction of the specified GPIO.
+ * @param desc - The GPIO descriptor.
+ * @return 0 in case of success, errno error codes otherwise.
+ */
+int32_t max_gpio_direction_input(struct gpio_desc *desc)
+{
+	int32_t ret;
+	gpio_cfg_t *maxim_extra;
+
+	if (!desc || desc->number >= N_PINS)
+		return -EINVAL;
+
+	maxim_extra = desc->extra;
+
+	if (maxim_extra->port >= N_PORTS)
+		return -EINVAL;
+
+	maxim_extra->func = GPIO_FUNC_IN;
+	GPIO_Config(maxim_extra);
+
+	return 0;
+}
+
+/**
+ * @brief Enable the output direction of the specified GPIO.
+ * @param desc - The GPIO descriptor.
+ * @param value - The value.
+ *                Example: GPIO_HIGH
+ *                         GPIO_LOW
+ * @return 0 in case of success, errno error codes otherwise.
+ */
+int32_t max_gpio_direction_output(struct gpio_desc *desc, uint8_t value)
+{
+	mxc_gpio_regs_t *gpio_regs;
+	gpio_cfg_t *maxim_extra;
+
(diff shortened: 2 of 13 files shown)


mchindri:
Please separate the drivers in different commits

CiprianRegus:
V1:
- Created individual commits for each driver.

drivers/platform/maxim/maxim_uart.c
```diff
@@ -0,0 +1,353 @@
...
+ */
+int32_t uart_write(struct uart_desc *desc, const uint8_t *data,
+		   uint32_t bytes_number)
+{
+	if(!desc || !data || !bytes_number)
+		return -EINVAL;
+
+	uint32_t err = 0;
+
+	int32_t len = bytes_number;
```
mchindri, line 149: I assume you have a maximum length to transfer with UART_Write, if so, I would split the transfer in multiple transfer.

drivers/platform/maxim/maxim_uart.h
```diff
@@ -0,0 +1,74 @@
...
+struct maxim_uart_init_param {
+	uint32_t port;
+	uart_parity_t parity;
+	uart_size_t size;
+	uart_stop_t stop;
+	uart_flow_ctrl_t flow;
+	uart_flow_pol_t pol;
+};
+
+int32_t uart_register_callback(uint8_t, struct callback_desc *);
```
mchindri, line 70: Even though I would put this function here to. Callback should go in irq and use irq_register

drivers/platform/maxim/maxim_uart.c
```diff
@@ -0,0 +1,353 @@
...
+* @brief Callback descriptor that contains a function to be called when an
+* interrupt occurs.
+*/
+static struct callback_desc *cb[2];
+
+/******************************************************************************/
+/************************ Functions Definitions *******************************/
+/******************************************************************************/
+
+void UART0_IRQHandler()
```
mchindri, line 64: Can you read some register to see which UART triggerd the function, instead of having 2 function doing the same thing?

amiclaus (commented):
there are some valid issues detected by the CI, they need to be addressed.

drivers/platform/maxim/maxim_uart.h
CiprianRegus: Will do that.

drivers/platform/maxim/maxim_uart.c
CiprianRegus: UART_Write handles that. It works by adding each byte to the tx fifo, and also waiting (in a loop) if the fifo is full. So in this case I don't think I have to split into multiple transfers.

drivers/platform/maxim/maxim_stdio.c
```diff
@@ -0,0 +1,134 @@
+/**
+ * @file	stdio.c
```
amiclaus, line 2: `@file maxim_stdio.c`

drivers/platform/maxim/maxim_rtc.c
```diff
@@ -0,0 +1,315 @@
...
+	return 0;
+}
+
+/**
+ * @brief Set the current count for the real time clock.
+ * @param desc - The RTC descriptor.
+ * @param tmr_cnt - New value of the timer counter.
+ * @return 0 in case of success, errno codes otherwise.
+ */
+int32_t rtc_set_cnt(struct rtc_desc *desc, uint32_t tmr_cnt)
```
amiclaus, line 201: let's keep consistent, and use `dev` naming as in other functions from this file

drivers/platform/maxim/gpio_extra.h
```diff
@@ -0,0 +1,72 @@
...
+
+int32_t max_gpio_disable_irq(struct irq_ctrl_desc *);
+int32_t max_gpio_enable_irq(struct irq_ctrl_desc *);
+int32_t max_gpio_register_callback(struct irq_ctrl_desc *,
+				   struct callback_desc *);
+int32_t max_gpio_unregister_callback(struct irq_ctrl_desc *);
+int32_t max_gpio_irq_set_trigger_level(struct gpio_desc *, enum irq_trig_level);
+
+#endif
+
```
amiclaus, line 72: nit: extra line

drivers/platform/maxim/maxim_stdio.h
```diff
@@ -0,0 +1,55 @@
+/***************************************************************************//**
+ *   @file   maxim/maxim_stdio.h
+ *   @brief  Header file for UART driver stdout/stdin redirection.
+ *   @author Ciprian Regus (ciprian.regus@analog.com)
+********************************************************************************
+ * Copyright 2021(c) Analog Devices, Inc.
```
amiclaus, line 6: nit: `2022`

drivers/platform/maxim/maxim_stdio.h
```diff
@@ -0,0 +1,55 @@
...
+int _open(const char *, int, int);
+int _isatty(int);
+int _write(int, char *, int);
+int _close(int);
+int _lseek(int, off_t, int);
+int _read(int, char *, int);
+int _fstat(int, struct stat *);
+
+#endif
+
```
amiclaus, line 55: nit: extra line

drivers/platform/maxim/spi_extra.h
```diff
@@ -0,0 +1,49 @@
...
+#ifndef SPI_EXTRA_H_
+#define SPI_EXTRA_H_
+
+/**
+ * @brief maxim specific SPI platform ops structure
+ */
+extern const struct spi_platform_ops max_spi_ops;
+
+#endif
+
```
amiclaus, line 49: nit: extra line

drivers/platform/maxim/maxim_uart.c
buha: Yeah valid point from Mihail above.
I think these UART0_IRQHandler, UART1_IRQHandler etc. are hardcoded with that name and the functions need to be defined. What you could do is find a Maxim HAL define that tells you if UART0 is available and do something like:
```
#ifdef MAXIM_UART0_EXISTS # <--- arbitrary name
void UART0_IRQHandler() {
    uart_irq(0);
}
#endif
#ifdef MAXIM_UART1_EXISTS
... etc.
```
where uart_irq() is a static function that will contain the same code for all uart instances.

drivers/platform/maxim/maxim_uart.h
```diff
@@ -0,0 +1,74 @@
...
+	uart_cfg_t *maxim_desc;
+};
+
+/**
+ * @struct maxim_uart_init_param
+ * @brief maxim platform specific UART descriptor
+ */
+struct maxim_uart_init_param {
+	uint32_t port;
+	uart_parity_t parity;
```
buha, line 63: size, parity and stop configurations already exist in uart.h - they are general for all UARTs, regardless of manufacturer/implementation.
https://github.com/analogdevicesinc/no-OS/blob/master/include/no-os/uart.h#L105-L111

You should only have in maxim_uart_init_param settings that are Maxim specific only.

CiprianRegus: I no longer use that struct for init and will delete it.

drivers/platform/maxim/maxim_spi.c
```diff
@@ -0,0 +1,179 @@
...
+		return -ENOMEM;
+
+	descriptor->device_id = param->device_id;
+	descriptor->max_speed_hz = param->max_speed_hz;
+	descriptor->chip_select = param->chip_select;
+	descriptor->mode = param->mode;
+	descriptor->bit_order = param->bit_order;
+	descriptor->platform_ops = &max_spi_ops;
+
+	ret = SPI_Init(SPI0A, descriptor->mode, param->max_speed_hz);
```
buha, line 79: This SPI0A should be passed in to the driver from the application that initializes the driver (unless all Maxim chips have only 1 SPI interface, which I doubt).

So you should create a maxim_spi_init_param which holds a member named e.g. "interface" or "instance", which gets initialized with SPI0A, gets saved in driver desc->extra and used thereafter.

drivers/platform/maxim/maxim_stdio.c
```diff
@@ -0,0 +1,134 @@
...
+#define MXC_UARTn   MXC_UART_GET_UART(CONSOLE_UART)
+#define UART_FIFO   MXC_UART_GET_FIFO(CONSOLE_UART)
+
+#include <unistd.h>
+#include <sys/stat.h>
+
+#define STDIN_FILENO	0   /**> Definition of stdin */
+#define STDOUT_FILENO   1   /**> Definition of stdout */
+#define STDERR_FILENO   2   /**> Definition of stderr */
+#define EBADF		  -1   /**> Error code for EBADf */
```
buha, line 63: Does Maxim toolchain not defined EBADF ? I see <errno.h> is included above.

CiprianRegus: I looked it up and EBADF is available in errno.h. That #define was part of the original file provided by Maxim, but I think I can delete it.

drivers/platform/maxim/maxim_uart.c
```diff
@@ -0,0 +1,353 @@
...
+{
+	if(!cb[0])
```
adrimbarean, line 66: I think we should clear interrupt flags even if callback is not registered. This way we avoid continuously triggering interrupts and hanging the controller.

drivers/platform/maxim/maxim_uart.c
```diff
@@ -0,0 +1,353 @@
...
+
+	while(reg_int) {
+		if((reg_int & 1) && uart_regs->int_en && BIT(n_int)) {
+			void *ctx = cb[0]->ctx;
+			void *config = cb[0]->config;
+			cb[0]->callback(ctx, n_int, config);
+		}
+		n_int++;
+		reg_int >>= 1;
+	}
```
adrimbarean, line 84: This doesn't seem to work correctly. I propose:
`while (reg_int) {
  n_int = find_first_set_bit(reg_int);
  if (reg_int & (uart_regs->int_en & BIT(n_int))) {
    //call cb
  }
  reg_int >>= n_int + 1;
}`

drivers/platform/maxim/maxim_uart.c
```diff
@@ -0,0 +1,353 @@
...
+ */
+int32_t uart_read(struct uart_desc *desc, uint8_t *data, uint32_t bytes_number)
+{
+	if(!desc || !data || !bytes_number)
+		return -EINVAL;
+
+	int32_t err = 0;
+	uint32_t bytes_read = 0;
+
+	int32_t len = bytes_number;
```
adrimbarean, line 125: If this is done to avoid warnings, I think we can do that by using a cast.

drivers/platform/maxim/maxim_uart.c
```diff
@@ -0,0 +1,353 @@
...
+	err = UART_Read(MXC_UART_GET_UART(desc->device_id), data, len, &bytes_read);
+
```
adrimbarean, line 127: nitpick: extra line

drivers/platform/maxim/maxim_uart.c
```diff
@@ -0,0 +1,353 @@
...
+	uint32_t err = 0;
+	uart_req_t read_req;
+
+	read_req.data = data;
+	read_req.len = bytes_number;
+	read_req.callback = NULL;
+	err = UART_ReadAsync(MXC_UART_GET_UART(desc->device_id), &read_req);
+
+	return err;
```
adrimbarean, line 179: We can return `UART_ReadAsync` directly here.

drivers/platform/maxim/maxim_uart.c
```diff
@@ -0,0 +1,353 @@
...
+	int ret = 0;
+	uint32_t did = param->device_id;
+
+	struct maxim_uart_desc *maxim_uart = calloc(1, sizeof(*maxim_uart));
+	struct uart_desc *descriptor = calloc(1, sizeof(*descriptor));
+
+	uart_cfg_t maxim_desc;
+	sys_cfg_uart_t maxim_desc_sys;
+
+	if(!maxim_uart || !descriptor) {
```
adrimbarean, line 226: I would do something similar to this approach, specifically to check allocation just after it's completed and to only `free` known allocated memory. This is done because `free()` [has undefined behavior on non-null, unallocated pointers](https://stackoverflow.com/questions/5308758/can-a-call-to-free-in-c-ever-fail) and may create a [bohrbug](https://en.wiktionary.org/wiki/bohrbug).

drivers/platform/maxim/maxim_uart.c
```diff
@@ -0,0 +1,353 @@
...
+		ret = -ENOMEM;
+		goto error;
+	}
+
+	NVIC_ClearPendingIRQ(MXC_UART_GET_IRQ(did));
+	NVIC_DisableIRQ(MXC_UART_GET_IRQ(did));
+	NVIC_SetPriority(MXC_UART_GET_IRQ(did), 1);
```
adrimbarean, line 233: Question: why are we setting interrupt priority of 1 here? And is there a way where we could do this in an IRQ driver? Status quo right now is to [include the IRQ driver](https://github.com/analogdevicesinc/no-OS/blob/master/drivers/platform/xilinx/uart.c#L42) in the other ones to access IRQ specific functionality.

drivers/platform/maxim/gpio_extra.h
```diff
@@ -0,0 +1,72 @@
...
+/**
+ * @struct gpio_irq_config
+ * @brief maxim configuration descriptor for irq operations
+ */
+struct gpio_irq_config {
+	struct gpio_desc *desc;
+	enum irq_trig_level mode;
+};
+
+int32_t max_gpio_disable_irq(struct irq_ctrl_desc *);
```
adrimbarean, line 64: Here as well as with UART I think we might want to use IRQ specific methods.

drivers/platform/maxim/maxim_gpio.c
```diff
@@ -0,0 +1,439 @@
...
+	if(!param || param->number >= N_PINS)
+		return -EINVAL;
+
+	struct gpio_desc *descriptor = calloc(1, sizeof(*descriptor));
+	if(!descriptor)
+		return -ENOMEM;
+
+	descriptor->number = param->number;
+	descriptor->platform_ops = param->platform_ops;
+	descriptor->extra = param->extra;
```
adrimbarean, line 96: To use this properly the user will have to know how this driver works as well as the maxim DFP driver and know the name of the `gpio_cfg_t` object the they need to place in the initialization structure. Is there a way to infer it based on the pin number or maybe a port ID? Those things should be more straight forward to use.

drivers/platform/maxim/maxim_gpio.c
```diff
@@ -0,0 +1,439 @@
...
+	if(!desc || !desc->extra || desc->number >= N_PINS)
+		return -EINVAL;
+
+	gpio_cfg_t *maxim_extra = (gpio_cfg_t *)desc->extra;
+
+	maxim_extra->mask = BIT(desc->number);
+	maxim_extra->func = GPIO_FUNC_OUT;
+	GPIO_Config(maxim_extra);
+
+	if(value == 0)
```
adrimbarean, line 179: nitpick: we can use `GPIO_LOW` in the if compare. Also we need a space between if and brackets e.g. `if (condition)`

adrimbarean: If we have `GPIO_HIGH_Z` option in `max_gpio_set_value` should we have it here as well?

CiprianRegus:
V2:
- Implemented a common IRQ handler function for both UART ports.
- Removed maxim_uart_init_param from maxim_uart.h
- SPI interfaces (SPI0A and SPI1A) are initialised based on device id.
- Removed EBADF define from maxim_stdio.c
- Fixed the UART IRQ handler as suggested and updated the other IRQ handlers to use a similar approach.
- Fixed the double free in maxim_uart.c
- Removed the NVIC function calls in maxim_uart.c
- Added max_gpio_init_param in gpio_extra.h to make the gpio init easier (and not use gpio_cfg_t directly as before).
- GPIO_HIGH_Z can now be used when setting a gpio pin as output.
- Fixed doxygen.

drivers/platform/maxim/maxim_uart.c
```diff
@@ -0,0 +1,323 @@
...
+ * @brief Unregister a callback function.
+ * @param port - The UART port.
+ * @return 0 in case of success, errno codes otherwise.
+ */
+int32_t uart_unregister_callback(uint8_t port)
+{
+	if (port >= N_PORTS || !cb[port])
+		return -EINVAL;
+
+	free(cb[port]);
```
buha, line 320: set the cb[port] to NULL after free, it could be accessed again
https://stackoverflow.com/questions/1025589/setting-variable-to-null-after-free

drivers/platform/maxim/gpio_extra.h
```diff
@@ -0,0 +1,79 @@
...
+ * LIMITED TO, INTELLECTUAL PROPERTY RIGHTS, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+*******************************************************************************/
+#ifndef GPIO_EXTRA_H_
+#define GPIO_EXTRA_H_
+
+#include "stdbool.h"
```
buha, line 42: include the standard libraries with the <> characters

drivers/platform/maxim/gpio_extra.h
```diff
@@ -0,0 +1,79 @@
...
+#include <stdint.h>
+#include "no-os/irq.h"
+#include "no-os/gpio.h"
+
+#define N_PINS	14
```
buha, line 47: So I see that you've included these here, but I think you can drop the N_PINS and N_PORTS checks because the Maxim HAL will check these. If there is a bug in Maxim HAL and it doesn't check them, and it crashes the program, then it's fine to add the checks ourselves, but I'm pretty sure that GPIO_Config() will return some meaningful error code if you pass it port -1 and pin 1234. So you simply need to translate that error code into no-os error codes and pass it back.

drivers/platform/maxim/maxim_gpio.c
```diff
@@ -0,0 +1,451 @@
...
+
+	g_cfg->port = pextra->port;
+	g_cfg->mask = BIT(param->number);
+	g_cfg->pad = m_pad;
+	g_cfg->padfunc = m_func;
+
+	descriptor->number = param->number;
+	descriptor->platform_ops = param->platform_ops;
+
+	GPIO_Config(descriptor->extra);
```
buha, line 106: check error code for all these maxim HAL calls

drivers/platform/maxim/maxim_gpio.c
```diff
@@ -0,0 +1,451 @@
...
+#include "no-os/irq.h"
+#include "no-os/util.h"
+#include "gpio.h"
+#include "gpio_regs.h"
+#include "gpio_extra.h"
+#include "max32660.h"
+
+#define N_INT	14
+
+static struct callback_desc *gpio_callback[N_INT];
```
buha, line 52: This global could be problematic when working on maxim chips with multiple ports, I think gpio_callback should account also for ports, not only pins. Otherwise, when working with GPIO1, GPIO2 etc. on top of GPIO0, these values would get overwritten.

drivers/platform/maxim/maxim_rtc.c
```diff
@@ -0,0 +1,315 @@
...
+
+	dev->id = init_param->id;
+	dev->freq = init_param->freq;
+	dev->load = init_param->load;
+
+	TMR_Enable(MXC_TMR0);
+
+	if (RTC_Init(MXC_RTC, dev->load, maxim_init_param->ms_load,
+		     &sys_cfg) != E_NO_ERROR) {
+		ret = -1;
```
buha, line 122: -EINVAL or something like that

drivers/platform/maxim/rtc_extra.h
```diff
@@ -0,0 +1,81 @@
...
+#include <stdint.h>
+#include "no-os/rtc.h"
+
+#define MAX_RTC_MAX_VALUE	(0xFFFFF)
+
+/**
+ * @struct rtc_init_maxim
+ * @brief maxim specific init parameter
+ */
+struct rtc_init_maxim {
```
buha, line 52: max_rtc_init ?

drivers/platform/maxim/rtc_extra.h
```diff
@@ -0,0 +1,81 @@
...
+enum rtc_interrupt_id {
+	RTC_TIMEOFDAY_INT,
+	RTC_SUBSEC_INT
+};
+
+/**
+ * @struct maxim_rtc_alarm_desc
+ * @brief maxim specific interrupt descriptor
+ */
+struct maxim_rtc_alarm_desc {
```
buha, line 70: also here, max_rtc_alarm_desc

drivers/platform/maxim/rtc_extra.h
```diff
@@ -0,0 +1,81 @@
...
+	enum rtc_interrupt_id int_id;
+	uint32_t irq_time;
+};
+
+int32_t rtc_get_time(uint32_t *, uint32_t *);
```
buha, line 75: https://github.com/analogdevicesinc/no-OS/blob/master/include/no-os/rtc.h#L100
I think you should simply implement that one and not create another function.

drivers/platform/maxim/rtc_extra.h
```diff
@@ -0,0 +1,81 @@
...
+int32_t rtc_enable_irq(enum rtc_interrupt_id, uint32_t);
```
buha, line 76: For these irq/callback functions, again, I think we need to discuss as a team what to do with them, I'm not sure if this is a good way to make them public to the rest of the code.

drivers/platform/maxim/maxim_spi.c
```diff
@@ -0,0 +1,186 @@
...
+	req.tx_data = tx_buffer;
+	req.tx_num = 0;
+	req.rx_data = rx_buffer;
+	req.rx_num = 0;
+	req.deass = 0;
+	req.width = SPI17Y_WIDTH_1;
+	req.len = bytes_number;
+
+	if (desc->device_id == 0)
+		SPI_MasterTrans(SPI0A, &req);
```
buha, line 139: check error codes for all HAL functions in spi.c and translate them into no-os error codes

drivers/platform/maxim/gpio_extra.h
CiprianRegus: I checked that, and GPIO_Config() doesn't check for either the port or pin number (passing an invalid port results in a null pointer being dereferenced). So, in this case, I'll keep those defines.

drivers/platform/maxim/rtc_extra.h
CiprianRegus: That function is no longer needed, so I'll delete it.

drivers/platform/maxim/maxim_rtc.c
```diff
@@ -0,0 +1,315 @@
...
+ * @param dev - The RTC descriptor.
+ * @return 0 in case of success, errno codes otherwise.
+ */
+int32_t rtc_start(struct rtc_desc *dev)
+{
+	RTC_EnableRTCE(MXC_RTC);
+
+	/** Wait for synchronization */
+	if (RTC_CheckBusy())
+		return -EBUSY;
```
adrimbarean, line 162: nitpick: I think the comment is a bit misleading since it doesn't do any actual waiting, just returns an error if the RTC is busy.

CiprianRegus: RTC_CheckBusy actually blocks the execution until the RTC's value can be read. It returns EBUSY if it waits for too long.

drivers/platform/maxim/maxim_rtc.c
```diff
@@ -0,0 +1,315 @@
...
+	if (!dev)
+		return -ENOMEM;
+
+	dev->id = init_param->id;
+	dev->freq = init_param->freq;
+	dev->load = init_param->load;
+
+	TMR_Enable(MXC_TMR0);
+
+	if (RTC_Init(MXC_RTC, dev->load, maxim_init_param->ms_load,
```
adrimbarean, line 120: We could use an extra rtc descriptor for maxim to store the `MXC_RTC` element. This way the driver is scalable to maxim devices with more than one RTC.

drivers/platform/maxim/maxim_spi.c
```diff
@@ -0,0 +1,186 @@
...
+{
+	if (!desc || !data)
+		return -EINVAL;
+
+	uint8_t *tx_buffer = data;
+	uint8_t *rx_buffer = data;
+	spi_req_t req;
+
+	req.bits = 8;
+	req.ssel = 0;
```
adrimbarean, line 129: We need this to be `desc->chip_select`.

drivers/platform/maxim/maxim_spi.c
```diff
@@ -0,0 +1,186 @@
...
+	req.tx_data = tx_buffer;
+	req.tx_num = 0;
+	req.rx_data = rx_buffer;
+	req.rx_num = 0;
+	req.deass = 0;
```
adrimbarean, line 134: I think we need this to be non-zero to release the slave after transaction.

drivers/platform/maxim/gpio_extra.h
buha: ok that's fine, it seems they have an issue in their HAL so it's fine to keep the checks, although you might have a problem then trying to keep up with all devices' available pins, you'll have to do something like:
```
#ifdef __MAX12345
#define N_PINS 14
#elif defined(__MAX54321)
#define N_PINS 16
...
```

CiprianRegus:
V4

- UART, GPIO and RTC irq functions are now called through platform ops.
- `maxim_irq.c` now implements a NVIC controller. Each peripheral has it's own IRQ controller.
- SPI now releases the CS line after the end of transfer.
- The CS pin can now be selected using the chip_select field.
- Added some checks for SDK function calls return values. In some places those are ignored because the functions cannot return an error (the parameters are already validated in the caller).
- Included an #ifdef to guarantee that the code compiles even though the chip doesn't have an SPI1A port.
- Renamed `maxim_rtc_alarm_desc` to `max_rtc_alarm_desc`.
- `gpio_callback` is now a 2D array to account for maxim chips with multiple GPIO ports.
- Fixed uart_read_nonblocking not reading from RX fifo.
- Fixed a few minor bugs.

buha (approved):
looks good!

CiprianRegus:
V5:

- SPI CS is now active low.

PR #1208, opened 2022-01-13
Fix new iiod for aducm3029

mchindri:
I didn't added this fixes in latest PR because it was already approved.
Tested with ad9361(xilinx) and adxrs290(aducm3029) on uart


mchindri:
v2
 - Fix build

buha (approved):

PR #1209, opened 2022-01-13
Create iio buffer

mchindri:
Last commit was tested only for ad9361 on xilinx.


mchindri:
v2-
 - Rework commit modifying all projects. Implement backwards compatibility in iio. This way don't need to do test and the commit
is cleaner
 - Add commit to fail in iio_read/write_all_attrs (details in commit msg)

mchindri:
v3
 - Fix astyle

PR #1210, opened 2022-01-14
Add iio trigger support

mchindri:
At the moment won't work. Need to make read and write functions nonblocking
 - Aducm uart: use some buffer and use uart_read_nonblocking
 - Aducm wifi: I think it should work but there may be memory issues
 - Xilinx uart with the folloing change it may:
 
image

- Xilinx eth: Yes, it should work, but not merged yet

### DIFF
--- a/drivers/gyro/adxrs290/iio_adxrs290_trig.c
+++ b/drivers/gyro/adxrs290/iio_adxrs290_trig.c
@@ -1,9 +1,9 @@
 /***************************************************************************//**
- *   @file   iio_adxrs290.h
- *   @brief  Implementation of ADXRS290 iio.
- *   @author Kister Genesis Jimenez (kister.jimenez@analog.com)
+ *   @file   iio_adxrs290_trig.c
+ *   @brief  Implementation of ADXRS290 iio trigger.
+ *   @author Mihail Chindris (Mihail.Chindris@analog.com)
 ********************************************************************************
- * Copyright 2020(c) Analog Devices, Inc.
+ * Copyright 2022(c) Analog Devices, Inc.
  *
  * All rights reserved.
  *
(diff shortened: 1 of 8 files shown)


mchindri:
v2
 - Fix astyle
 - Compile fix

iio/iio.c
```diff
@@ -1345,11 +1486,32 @@ static int32_t iio_init_devs(struct iio_desc *desc,
...
+	return SUCCESS;
+}
 
-	return ret;
+static int32_t iio_init_trigs(struct iio_desc *desc,
+			      struct iio_trigger_init *trigs, int32_t n)
+{
+	uint32_t i;
+	struct iio_trig_priv *ltrig;
+	struct iio_trigger_init *ntrig;
```
adrimbarean, line 1497: I'd rename `ltrig` and `ntrig` to something more explicit and less likely to be confused with one-another. `ntrig` could be `init_iter` and `ltrig` could be `priv_iter`.

iio/iio.h
```diff
@@ -101,6 +111,10 @@ int iio_init(struct iio_desc **desc, struct iio_init_param *init_param);
 int iio_remove(struct iio_desc *desc);
 /* Execut an iio step. */
 int iio_step(struct iio_desc *desc);
+/* Signal iio that a trigger has been triggered.
+ * This can be called in interrupt context. But application callback will
+ * be called when from iio_step */
```
adrimbarean, line 116: nitpick: be called only when iio_step is called

iio/iio_app/iio_app.h
```diff
@@ -73,4 +80,8 @@ struct iio_app_device {
  */
 int32_t iio_app_run(struct iio_app_device *devices, int32_t len);
 
+/* This function won't initialize irq_desc */
+int32_t iio_app_run2(struct iio_app_device *devices, int32_t len,
```
adrimbarean, line 84: Question: does this need to be an API?

iio/iio_attr_trig.c
```diff
@@ -0,0 +1,114 @@
...
+		return -ENOMEM;
+
+	ltrig->iio_desc = param->iio_desc;
+	strncpy(ltrig->name, param->name, MAX_NAME_SIZE);
+
+	*trig = ltrig;
+
+	return SUCCESS;
+error:
+//TODO cleanup
```
adrimbarean, line 86: As it is right now function can't fail, so I think we can omit clean-up.

iio/iio_attr_trig.c
```diff
@@ -0,0 +1,114 @@
...
+	return ret;
+}
+
+static int iio_attr_trigger_now(void *device, char *buf, uint32_t len,
+				const struct iio_ch_info *channel,
+				intptr_t priv)
+{
+	struct iio_attr_trig *trig = device;
+
+	//Should it be triggered only when in buf is the number 1 ?
```
adrimbarean, line 96: I'd say the `buf` parameter doesn't need to contain anything. Unless we want this attribute to do anything beside trigger now we shouldn't request for any inputs besides calling the attribute itself.

drivers/gyro/adxrs290/iio_adxrs290_trig.c
```diff
@@ -0,0 +1,120 @@
...
+#include <string.h>
+#include "iio_adxrs290.h"
+#include "iio_types.h"
+#include "iio.h"
+#include "no-os/error.h"
+
+#define MAX_NAME_SIZE 20
+
+struct iio_adxrs290_trig {
+	struct iio_desc		**iio_desc;
```
adrimbarean, line 50: Question: is there a need to pass a double pointer here?

iio/iio_app/iio_app.c
```diff
@@ -342,21 +335,38 @@ int32_t iio_app_run(struct iio_app_device *devices, int32_t len)
...
+		status = iio_step(*iio_desc);
 	} while (true);
 error:
 	status = print_uart_error_message(&uart_desc, uart_init_par, status);
 	return status;
 }
 
+int32_t iio_app_run(struct iio_app_device *devices, int32_t len)
+{
+	struct iio_desc	*iio_desc;
```
adrimbarean, line 359: We may not need to pass this along. I don't think we need it outside `iio_app_run`.

adrimbarean (commented):
Added some comments that I think we should discuss.

adrimbarean:
Work in progress is done here: https://github.com/analogdevicesinc/no-OS/pull/1321

PR #1211, opened 2022-01-14
Example of use adxrs290 with trigger

mchindri:
This PR was not tested, are just some guideline of changes I thing the should be made.
Only the last commit is relevant, the rest are from #1211


adrimbarean:
Work in progress is done here: https://github.com/analogdevicesinc/no-OS/pull/1321

PR #1214, opened 2022-01-17
Dev/axi dmac init fix

adrimbarean:
A small fix and an enhancement to the AXI DMAC driver.

### DIFF
--- a/drivers/axi_core/axi_dmac/axi_dmac.c
+++ b/drivers/axi_core/axi_dmac/axi_dmac.c
@@ -294,7 +294,7 @@ int32_t axi_dmac_init(struct axi_dmac **dmac_core,
 {
 	struct axi_dmac *dmac;
 
-	dmac = (struct axi_dmac *)malloc(sizeof(*dmac));
+	dmac = (struct axi_dmac *)calloc(1, sizeof(*dmac));
 	if (!dmac)
 		return FAILURE;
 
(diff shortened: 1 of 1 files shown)


drivers/axi_core/axi_dmac/axi_dmac.c
```diff
@@ -303,10 +303,6 @@ int32_t axi_dmac_init(struct axi_dmac **dmac_core,
 	dmac->direction = init->direction;
 	dmac->flags = init->flags;
 	dmac->transfer_max_size = -1;
-	dmac->big_transfer.address = 0;
```
amiclaus, line 306: the lines were deleted by mistake, i guess? You are adding `dmac->big_transfer.transfer_done = false;` in previous commit

adrimbarean: The deletion is not a mistake. The change could have been in one commit, but I thought it would have been harder to explain what I tried to do so I separated it. These lines are not needed anymore because `calloc` already initializes them to 0. Now I realize it might be easier to just drop the first commit and just solve the bug as a byproduct of the second. What do you think? Should we even mention the bug anymore since it can be solved sneakily?

adrimbarean:
V1:
 - solve `calloc` missing parameter build error.

drivers/axi_core/axi_dmac/axi_dmac.c
amiclaus: Got it! Adding and then removing the same line got me confused. Well, I would just do the `calloc` and remove the explicit initializations with 0 in a single commit. And mention also the transfer issue that was fixed in the commit description.

amiclaus (approved):

adrimbarean:
V2:
 - merge the two commits into one.

dbogdan (approved):

PR #1222, opened 2022-01-20
Makefile update

danmois:
Implementation of first makefile features regarding the use of a specific SDK.

### DIFF
--- a/tools/scripts/aducm.mk
+++ b/tools/scripts/aducm.mk
@@ -184,7 +184,7 @@ debug: all $(BINARY).gdb
 
 PROJECT_BINARY = $(PROJECT_BUILD)/Release/$(PROJECT_NAME)
 
-develop:
+$(PLATFORM)_sdkopen:
 	cces -data $(WORKSPACE)
 
 #This way will not work if the rest button is press or if a printf is executed
--- a/tools/scripts/aducm.mk
+++ b/tools/scripts/aducm.mk
@@ -283,4 +283,12 @@ clean_project:
  		-project $(PROJECT_NAME) \
  		-cleanOnly all
 
+PHONY += $(PLATFORM)_sdkclean
+$(PLATFORM)_sdkclean:
+	$(call print,[Delete] SDK artefacts from $(BUILD_DIR))
+	$(MUTE) $(CCES) -nosplash -application com.analog.crosscore.headlesstools \
+		-data $(WORKSPACE) \
+		-project $(CCES_PROJ_NAME) \
+		-cleanOnly all $(HIDE)
+
 .PHONY: $(PHONY)
--- a/tools/scripts/aducm.mk
+++ b/tools/scripts/aducm.mk
@@ -271,7 +271,7 @@ $(PROJECT_TARGET):
 copy_pinmux:
 	$(MUTE) $(call copy_file,$(PIN_MUX),$(PROJECT_PIN_MUX)) $(HIDE)
 
-update_srcs: copy_pinmux
+update: copy_pinmux
 
 PHONY += clean_project
 clean_project:
--- a/tools/scripts/aducm.mk
+++ b/tools/scripts/aducm.mk
@@ -210,8 +210,9 @@ project_run: build_project
 	-c "resume" \
 	-c exit $(HIDE)
 
-PHONY += aducm3029_project_build
-aducm3029_project_build: aducm3029_project $(LIB_TARGETS)
+PHONY += $(PLATFORM)_sdkbuild
+$(PLATFORM)_sdkbuild: $(LIB_TARGETS)
+	$(call print,[Build] using CCES)
 	$(MUTE) $(CCES) -nosplash -application com.analog.crosscore.headlesstools \
 		-data $(WORKSPACE) \
 		-project $(CCES_PROJ_NAME) \
@@ -265,7 +266,7 @@ $(PROJECT_TARGET):
 	$(MUTE) $(call copy_file\
 	,$(PLATFORM_TOOLS)/startup_ADuCM3029_patch.c,$(PROJECT_BUILD)/RTE/Device/ADuCM3029/startup_ADuCM3029.c) $(HIDE)
 #Remove default files from projectsrc
-	$(MUTE) $(call remove_dir,$(PROJECT_BUILD)/src) $(HIDE)
+	$(MUTE) $(call remove_dir_action,$(PROJECT_BUILD)/src) $(HIDE)
 	$(MUTE) $(call set_one_time_rule,$@)
 
 copy_pinmux:
--- a/tools/scripts/build_projects.py
+++ b/tools/scripts/build_projects.py
@@ -192,14 +192,17 @@ class BuildConfig:
 				new_hdf = True
 		
 		if new_hdf:
-			err = run_cmd(cmd + ' clean_all')
+			err = run_cmd(cmd + ' reset')
 			if err != 0:
 				return err
 
-		err = run_cmd(cmd + ' update_srcs')
+		err = run_cmd(cmd + ' update')
 		if err != 0:
 			return err
-		err = run_cmd(cmd + ' -j%d re' % (multiprocessing.cpu_count() - 1))
+		err = run_cmd(cmd + ' clean')
+		if err != 0:
+			return err
+		err = run_cmd(cmd + ' -j%d all' % (multiprocessing.cpu_count() - 1))
 		if err != 0:
 			return err
 		
--- a/tools/scripts/generic.mk
+++ b/tools/scripts/generic.mk
@@ -46,6 +46,7 @@ TIMESTAMP = $(shell date +"%T")
 copy_file = cp $1 $2
 copy_dir = cp -r $1 $2
 remove_file = rm -rf $1
+remove_dir_action = rm -rf $1
 remove_dir = rm -rf $1
 mk_dir = mkdir -p $1
 read_file = cat $1 2> /dev/null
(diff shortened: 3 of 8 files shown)


amiclaus (commented):
It looks like these changes broke the `Build Projects` pipeline.
If you add `staging/` prefix to your branch: https://github.com/analogdevicesinc/no-OS/blob/f2c0ffea809196425300d2cdc39174959c54493f/build_projects.yml#L4
 the builds will be triggered on your custom branch, prior to doing a PR so you can debug the issues easier and double check that everything works fine :).

tools/scripts/build_projects.py
```diff
@@ -192,11 +192,11 @@ def build(self):
...
-			err = run_cmd(cmd + ' clean_all')
+			err = run_cmd(cmd + ' reset')
 			if err != 0:
 				return err
 
-		err = run_cmd(cmd + ' update_srcs')
+		err = run_cmd(cmd + ' update')
 		if err != 0:
 			return err
 		err = run_cmd(cmd + ' -j%d re' % (multiprocessing.cpu_count() - 1))
```
amiclaus, line 202: `make re` was removed with d71c363baa61b574042e58f3a037ef32d350c02e

tools/scripts/generic.mk
```diff
@@ -379,18 +380,11 @@ clean_all: clean_libs
 	@$(call print,[Delete] $(BUILD_DIR))
 	$(MUTE) $(call remove_dir,$(BUILD_DIR)) $(HIDE)
 
-PHONY += ca
```
buha, line 382: I would rephrase the commit title to "...: remove obsolete/unneeded rules" or something like that

tools/scripts/build_projects.py
buha: yes, true, we need to make sure all the changes in this PR don't break our CI builds

tools/scripts/generic.mk
```diff
@@ -366,7 +367,10 @@ standalone:
 	$(MUTE) $(MAKE) --no-print-directory project LINK_SRCS=n MAKEFLAGS=$(MAKEOVERRIDES)
 
 # Build project using SDK
-project_build: $(PLATFORM)_project_build
+sdkbuild: $(PLATFORM)_project_build
```
buha, line 370: Let's also rename $(PLATFORM)_project_build ---> $(PLATFORM)_sdkbuild so that it's easier to follow the script. This is internal-only, users wouldn't care, but it makes these scripts more maintainable... same for $(PLATFORM)_project_clean ---> $(PLATFORM)_sdkclean

buha (commented):
I also think we should do the changes stm32.mk and aducm3029.mk platforms in this same PR, define the rules $(PLATFORM)_sdkbuild, $(PLATFORM)_sdkclean, $(PLATFORM)_sdkopen in all of them.
If we only do it for xilinx it will fail when you `make sdkbuild` on any other platform.

danmois:
V2

- Removed also the ra rule

tools/scripts/generic.mk
danmois: Done

tools/scripts/build_projects.py
danmois: Done

danmois:
V3

- renamed $(PLATFORM)_project_build to $(PLATFORM)_sdkbuild
- renamed $(PLATFORM)_project_clean to $(PLATFORM)_sdkclean

tools/scripts/generic.mk
danmois: Done

tools/scripts/generic.mk
```diff
@@ -375,8 +375,8 @@ clean:
 	$(MUTE) $(call remove_file,$(BINARY) $(OBJS) $(ASM_OBJS)) $(HIDE)
 
 # Remove the whole build directory
-PHONY += clean_all
-clean_all: clean_libs
+PHONY += reset
```
adrimbarean, line 382: Is "reset" the correct name here? To me "reset" implies that after the clean operation is done the build starts again. Would "purge" be a better choice?

tools/scripts/platform/xilinx/build_project.tcl
```diff
@@ -5,6 +5,6 @@ if { $argc != 1 } {
 
 sdk setws [lindex $argv 0]
 
-projects -build -name app -type app
+app build -name app
```
adrimbarean, line 8: I would either make a visible note about this change in the commit description or place it in a separate commit. It probably is best if placed in a separate commit because, if I'm not mistaken, this it an update to the make process following an update to the Xilinx tools. I'd say it should special notice.

adrimbarean (commented):
A couple of questions.

tools/scripts/generic.mk
danmois: I would say that "purge" is a term that is usual for the removal of sensitive data, while "reset" implies the return to the initial state. However, I would be happy to organize a poll on this matter, and choose the proper naming.

tools/scripts/platform/xilinx/build_project.tcl
danmois: Yes, this is true. It is the current recommended (by the Xilinx tools) way of dealing with projects.

danmois:
V4

- addition of the sdkclean rule for the STM32 platform
- correction of echo message of remove file function in genetic.mk
- the creation of SDK-related rules for the ADuCM3029 platform

tools/scripts/build_projects.py
```diff
@@ -192,14 +192,17 @@ def build(self):
 				new_hdf = True
 		
 		if new_hdf:
-			err = run_cmd(cmd + ' clean_all')
+			err = run_cmd(cmd + ' reset')
```
buha, line 195: the changes in build_projects should probably go in the 2nd commit, that's the one that deals with reset and update

tools/scripts/stm32.mk
```diff
@@ -163,8 +163,8 @@ post_build: $(HEX)
 	$(MUTE) $(foreach inc, $(EXTRA_INC_PATHS), sed -i '/Core\/Inc"\/>/a <listOptionValue builtIn="false" value="$(inc)"\/>' $(PROJECT_BUILDROOT)/.cproject;) $(HIDE)
 	$(MUTE) $(foreach flag, $(CPROJECTFLAGS), sed -i '/USE_HAL_DRIVER"\/>/a <listOptionValue builtIn="false" value="$(flag)"\/>' $(PROJECT_BUILDROOT)/.cproject;) $(HIDE)
 
-PHONY += $(PLATFORM)_project_build
-$(PLATFORM)_project_build: $(PROJECT_TARGET) $(LIB_TARGETS)
+PHONY += $(PLATFORM)_sdkbuild
```
buha, line 168: commit title suggests this is for Xilinx only but you have a stm32 related change here.

buha: Actually I think you should rather do each rule like you did it in the following commit:
`tools: scripts: *.mk: sdkopen rule `
Put the top-level generic.mk rule altogether with all platform specific rules that it uses in the same commit. That way, when somebody checks-out this commit, the rule will work for any platform.

danmois:
V5

- moved the modifications performed in build_projects.py to the second commit

tools/scripts/build_projects.py
danmois: Done

danmois:
V6

- reorganized modifications in commits dealing with specific rules on all applicable targets (as recommended in https://github.com/analogdevicesinc/no-OS/pull/1222#discussion_r795409130)

tools/scripts/generic.mk
adrimbarean: No need. Whatever we choose it's gonna be great.

adrimbarean (approved):
Looks good to me.

buha (approved):
Looks good, but we need to do some testing

tools/scripts/aducm.mk
```diff
@@ -283,4 +283,12 @@ clean_project:
  		-project $(PROJECT_NAME) \
  		-cleanOnly all
 
+PHONY += $(PLATFORM)_sdkclean
+$(PLATFORM)_sdkclean:
+	$(call print,[Delete] SDK created artefacts from $(BUILD_DIR))
```
buha, line 288: Could you please delete the "created" word ?
It generates this output and I find it very confusing whether something was created or deleted, even though by careful reading one may understand the proper meaning of it.
```
[iio_demo]$ make sdkclean
[09:52:42] [Delete] SDK created artefacts from /home/dari/Work/no-OS/projects/iio_demo/build/app/Debug
[09:52:42] [Delete] SDK created artefacts from /home/dari/Work/no-OS/projects/iio_demo/build/app/Release
```

tools/scripts/stm32.mk
```diff
@@ -168,6 +170,13 @@ $(PLATFORM)_sdkbuild:
 	$(MUTE) $(STM32CUBEIDE)/$(IDE) -nosplash -application org.eclipse.cdt.managedbuilder.core.headlessbuild \
 		-import $(PROJECT_BUILDROOT) -data $(BUILD_DIR) -build app $(HIDE)
 
+PHONY += $(PLATFORM)_sdkclean
+$(PLATFORM)_sdkclean:
+	$(call print,[Delete] SDK created artefacts from $(DEBUG_DIR))
```
buha, line 175: also here, "created"
and everywhere else

buha:
So I've tested this on stm32 and works as expected

danmois:
V7

- Fixed the message displayed by the sdkclean rule (removal of the "created" work)

buha:
remove_dir_action on linux is missing

buha:
let's also implement `make build`

danmois:
V8

- added the build rule
- fixed a bug regarding the sdkbuild for STM32 on Linux (the addition of the remove_dir_action function in generic.mk - required for consistency with the Win implementation)

APorumb:
I've tested building xilinx project on windows machine and works as expected.

PR #1225, opened 2022-01-22
drivers:platform:mbed: SPI platform drivers implementation

mphalke:
Initial implementation of the Mbed SPI platform drivers.
 
### DIFF
--- a/drivers/platform/mbed/mbed_spi.cpp
+++ b/drivers/platform/mbed/mbed_spi.cpp
@@ -1,0 +1,273 @@
+/***************************************************************************//**
+ *   @file   mbed_spi.cpp
+ *   @brief  Implementation of SPI Mbed platform driver interfaces
+********************************************************************************
+ * Copyright (c) 2019-2022 Analog Devices, Inc.
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *  - Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ *  - Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in
+ *    the documentation and/or other materials provided with the
+ *    distribution.
+ *  - Neither the name of Analog Devices, Inc. nor the names of its
+ *    contributors may be used to endorse or promote products derived
+ *    from this software without specific prior written permission.
+ *  - The use of this software may or may not infringe the patent rights
+ *    of one or more patent holders.  This license does not release you
+ *    from the requirement that you obtain separate licenses from these
+ *    patent holders to use this software.
+ *  - Use of the software either in source or binary form, must be run
+ *    on or directly connected to an Analog Devices Inc. component.
+ *
+ * THIS SOFTWARE IS PROVIDED BY ANALOG DEVICES "AS IS" AND ANY EXPRESS OR
+ * IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, NON-INFRINGEMENT,
+ * MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED.
+ * IN NO EVENT SHALL ANALOG DEVICES BE LIABLE FOR ANY DIRECT, INDIRECT,
+ * INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, INTELLECTUAL PROPERTY RIGHTS, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+*******************************************************************************/
+
+/******************************************************************************/
+/***************************** Include Files **********************************/
+/******************************************************************************/
+
+#include <stdio.h>
+#include <mbed.h>
+
+// Platform drivers needs to be C-compatible to work with other drivers
+#ifdef __cplusplus
+extern "C"
+{
+#endif //  _cplusplus
+
+#include "no-os/error.h"
+#include "no-os/spi.h"
+#include "no-os/gpio.h"
+#include "mbed_spi.h"
+
+#define		SPI_8_BIT_FRAME			8		// SPI 8-bit frame size
+
+/******************************************************************************/
+/************************ Functions Definitions *******************************/
+/******************************************************************************/
+
+/**
+ * @brief Initialize the Mbed SPI communication peripheral.
+ * @param desc[in,out] - The SPI descriptor.
+ * @param param[in] - The structure that contains the SPI parameters.
+ * @return SUCCESS in case of success, negative error code otherwise.
+ */
+int32_t mbed_spi_init(struct spi_desc **desc,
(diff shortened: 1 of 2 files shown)


adrimbarean:
Can you split each driver in a separate commit, please? It will help in the review process.

mphalke:
Sure. I will put each driver module as a separate commit to make review process simple.
I'll first start with SPI and then one by one other modules.

drivers/platform/mbed/mbed_spi.cpp
```diff
@@ -0,0 +1,292 @@
...
+
+/******************************************************************************/
+/********************** Macros and Constants Definitions **********************/
+/******************************************************************************/
+
+#define		SPI_8_BIT_FRAME			8		// SPI 8-bit frame size
+
+/******************************************************************************/
+/********************** Variables and User defined data types *****************/
+/******************************************************************************/
```
amiclaus, line 64: nit: these headers are not needed

drivers/platform/mbed/mbed_spi.h
```diff
@@ -0,0 +1,103 @@
...
+/******************************************************************************/
+/***************************** Include Files **********************************/
+/******************************************************************************/
+#include <stdio.h>
+#include <stdbool.h>
+
+/******************************************************************************/
+/********************** Macros and Constants Definitions **********************/
+/******************************************************************************/
+
```
amiclaus, line 56: same here

buha (approved):
looks good to me

buha:
just make sure to fix the things cppcheck is complaining about

mphalke:
> 

The cpp check is complaining about memory leak for the structure variable 'mbed_spi_desc' in mbed_spi_init() function. 
This instance is freed during error on line 152, so not sure why it is complaining about. Any suggestions?

drivers/platform/mbed/mbed_spi.h
mphalke: I have removed the unnecessary headers from .cpp and .h file. Please verify.

drivers/platform/mbed/mbed_spi.cpp
```diff
@@ -0,0 +1,280 @@
...
+		SPI frequency is required, consult your device documentation.
+		**/
+	spi->frequency(param->max_speed_hz);
+	spi->format(SPI_8_BIT_FRAME, param->mode);   // data write/read format
+	spi->set_default_write_value(0x00);          // code to write when reading back
+
+	return SUCCESS;
+
+err_spi:
+	if (((struct mbed_spi_init_param *)param->extra)->use_sw_csb) {
```
adrimbarean, line 147: We should check if `csb` is defined here as well, because if the SPI constructor didn't succeed and `use_sw_csb` is true we can end up freeing unallocated memory.

drivers/platform/mbed/mbed_spi.cpp
```diff
@@ -0,0 +1,280 @@
...
+		mbed_spi_desc->csb_gpio = NULL;
+	}
+
+	if (!spi) {
+		goto err_spi;
+	}
+
+	mbed_spi_desc->spi_port = (SPI *)spi;
+
+	spi_desc->extra = (struct mbed_spi_desc *)mbed_spi_desc;
```
adrimbarean, line 132: These pointers should be cast to `void *` to be allocated to a void pointer.But I think we can skip casting here.

drivers/platform/mbed/mbed_spi.cpp
```diff
@@ -0,0 +1,280 @@
...
+			delete((DigitalOut *)(((struct mbed_spi_desc *)(desc->extra))->csb_gpio));
+		}
+	}
+
+	/* Free the SPI port object */
+	if ((SPI *)(((struct mbed_spi_desc *)(desc->extra))->spi_port))
+		delete((SPI *)(((struct mbed_spi_desc *)(desc->extra))->spi_port));
+
+	/* Free the SPI extra descriptor object */
+	if ((struct mbed_spi_desc *)(desc->extra))
```
adrimbarean, line 181: Since the driver doesn't function without the `desc->extra` parameter allocated, don't include this check. This can also be the reason for the memleak error.

drivers/platform/mbed/mbed_spi.cpp
```diff
@@ -0,0 +1,280 @@
...
+		free((struct mbed_spi_desc *)(desc->extra));
```
adrimbarean, line 182: No need to cast here.

drivers/platform/mbed/mbed_spi.cpp
```diff
@@ -0,0 +1,280 @@
...
+
+	if (((struct mbed_spi_desc *)desc->extra)->use_sw_csb) {
+		csb = (DigitalOut *)(((struct mbed_spi_desc *)(desc->extra))->csb_gpio);
+		csb->write(GPIO_LOW);
+	}
+
+	/* Perform synchronous SPI write and read */
+	spi->write((const char *)data, bytes_number, (char *)data, bytes_number);
+
+	if (((struct mbed_spi_desc *)desc->extra)->use_sw_csb) {
```
adrimbarean, line 217: nitpick: no need for brackets in a single expression if.

drivers/platform/mbed/mbed_spi.cpp
```diff
@@ -0,0 +1,280 @@
...
+		csb->write(GPIO_LOW);
+
+		/* Perform synchronous SPI write and read */
+		if (!msgs[msg_cnt].tx_buff) {
+			spi->write(NULL, 0,
+				   (char *)msgs[msg_cnt].rx_buff, msgs[msg_cnt].bytes_number);
+		} else {
+			spi->write((const char *)msgs[msg_cnt].tx_buff, msgs[msg_cnt].bytes_number,
+				   (char *)msgs[msg_cnt].rx_buff, msgs[msg_cnt].bytes_number);
+		}
```
adrimbarean, line 258: Is this check necessary? If we use just the code in the else branch all possibilities should be covered, if I'm not mistaken.

drivers/platform/mbed/mbed_spi.cpp
```diff
@@ -0,0 +1,280 @@
...
+}
+
+/**
+* @brief Mbed platform specific SPI platform ops structure
+*/
+const struct spi_platform_ops mbed_spi_ops = {
+	.init = &mbed_spi_init,
+	.write_and_read = &mbed_spi_write_and_read,
+	.remove = &mbed_spi_remove
+};
```
adrimbarean, line 276: We need to provide the transfer function as well.

mphalke: Thanks for the catch.

drivers/platform/mbed/mbed_spi.cpp
mphalke: Make sense. Added check for 'csb' before freeing memory.

drivers/platform/mbed/mbed_spi.cpp
mphalke: Make sense. Removed casting here.

drivers/platform/mbed/mbed_spi.cpp
mphalke: Make sense. Removed this check.

drivers/platform/mbed/mbed_spi.cpp
mphalke: We need to pass transmit byte count as 0 and transmit buffer NULL if just intended to read the data from SPI. The function overloading is used for spi->write() function in Mbed layer.
Also the 2nd argument to this function i.e. transmit byte count is not explicitly passed to spi_transfer() function and hence some sort of condition is needed for setting transmit count as 0.

adrimbarean (approved):
I'd add "Initial implementation of the MBED SPI platform driver." It's customary to have a non-empty commit message body and I think it's pertinent information. Feel free to include anything extra you think it's necessary.

amiclaus (approved):

mphalke:
> > I'd add "Initial implementation of the MBED SPI platform driver." It's customary to have a non-empty commit message body and I think it's pertinent information. Feel free to include anything extra you think it's necessary.
> 
> Apologies, I meant if you could add it in the commit description. So it would need another force push. Not only the PR description.

Sorry got confused. I have added the commit description and forced push the commit. Also updating the PR description to be more specific.

PR #1229, opened 2022-01-28
IIOD attribute write and debug register read/write issue fix

mphalke:
The non-blocking version of IIOD doesn't work well with IIO client (e.g. IIO oscilloscope) to write the IIO attribute value and perform debug register read/write operations. The IIO attribute write is a special case where two consecutive commands are sent from IIO client over serial link: 1) Attribute write command with size of data byte  2) Attribute write value
Existing code had bug in reading the attribute write value (i.e. 2nd command was not handled correctly). So updated code to handle this command by initializing payload buffer and length in IIOD command run state.
Verified other operation (attr read/write, data capture, debug register read/write) after this change and nothing is broken.

### DIFF
--- a/iio/iiod.c
+++ b/iio/iiod.c
@@ -634,6 +634,7 @@ static int32_t iiod_run_cmd(struct iiod_desc *desc,
 		ret = desc->ops.write_attr(&ctx, data->device, &attr,
 					   conn->payload_buf,
 					   data->bytes_count);
+		conn->nb_buf.len = 0;
 		conn->res.val = ret;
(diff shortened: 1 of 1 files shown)


buha (approved):
I've just tested this by writing attributes and indeed it fixes the problem.
I was also on this issue but you found it first :)
Could you please just add all that PR description text in the commit message and then I will merge.

adrimbarean (approved):
Also make the title of the commit a bit shorter so that it looks better on Github.

amiclaus (approved):

PR #1231, opened 2022-01-31
Update Chip Select data member in the SPI headers

jansunil:
include: no-os: Changed the chip select data member type to 16 bit wide

Certain mbed supported MCUs require a 16 Bit wide data member to store
the pin name. Since the SPI chip select data member utilizes a GPIO name
to be stored as its value, it has been updated to uint16_t in the SPI
init param and in the SPI descriptor.

### DIFF
--- a/include/no-os/spi.h
+++ b/include/no-os/spi.h
@@ -115,7 +115,7 @@ typedef struct spi_init_param {
 	/** maximum transfer speed */
 	uint32_t	max_speed_hz;
 	/** SPI chip select */
-	uint8_t		chip_select;
+	uint16_t		chip_select;
 	/** SPI mode */
 	enum spi_mode	mode;
 	/** SPI bit order */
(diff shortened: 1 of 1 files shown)


amiclaus (approved):

buha:
Hmm... I think the whole point of `include/no-os` headers is to have a homogeneous interface across all platforms.
As such, the `chip_select` member should only contain positive integer numbers, Chip Select 0, 1, 2, 3 ..... (currently up to) 255.
If you try to store in it something platform specific like "GPIO name to be stored as its value", this kind of goes against the purpose of `include/no-os` generic and homogeneous interface... 
I would advise instead a translation from integers like 0, 1, 2.... to whatever "GPIO name" is required - this translation should go in spi_init() function of mbed platform.

amiclaus (commented):
I haven't read properly the reason of this change, my bad on this ... I think @buha has a good point.

jansunil:
Deleting the PR to implement the changes as suggested on the mbed side.

PR #1232, opened 2022-01-31
drivers:adc:ad7799_iio: read raw data

amiclaus:
Insead of processing the channel data in the IIO layer of the driver,
just send the raw data in the upper layers to be processed.


adrimbarean (approved):
Is `ad7799_read_channel` made redundant by this change? Looks good to me either way.

amiclaus (commented):
> Is `ad7799_read_channel` made redundant by this change? Looks good to me either way.

Not really, this particular change was requested by the production team that uses the IIO layer.

The `ad7799_read_channel` should remain a valid option for the driver usage.

PR #1234, opened 2022-02-01
drivers:platform:mbed: I2C platform driver implementation

mphalke:
Initial implementation of the Mbed I2C platform drivers.

### DIFF
--- a/drivers/platform/mbed/mbed_i2c.cpp
+++ b/drivers/platform/mbed/mbed_i2c.cpp
@@ -1,0 +1,213 @@
+/***************************************************************************//**
+ *   @file   mbed_i2c.cpp
+ *   @brief  Implementation of I2C Mbed platform driver interfaces
+********************************************************************************
+ * Copyright (c) 2019-2022 Analog Devices, Inc.
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *  - Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ *  - Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in
+ *    the documentation and/or other materials provided with the
+ *    distribution.
+ *  - Neither the name of Analog Devices, Inc. nor the names of its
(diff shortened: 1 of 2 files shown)


drivers/platform/mbed/mbed_i2c.cpp
```diff
@@ -0,0 +1,215 @@
...
+
+	i2c = (I2C *)(((mbed_i2c_desc *)(desc->extra))->i2c_port);
+
+	/**
+	    The MBED I2C API is reversed for parameter 4
+	    Instead of stop_bit - it has
+	    @param repeated   - Repeated start, true - don't send stop at end default value is false.
+	    Inverting here to keep the no-OS/platform_drivers API
+	 */
+	if (!(i2c->write(desc->slave_address, (char *)data, bytes_number, !stop_bit))) {
```
amiclaus, line 161: nit: no need for brackets in this `if ... else` statement. actually you can also get rid of `else`:
```
if ((i2c->write(desc->slave_address, (char *)data, bytes_number, !stop_bit)))
     return -EINVAL;

return SUCCESS;
```

drivers/platform/mbed/mbed_i2c.cpp
```diff
@@ -0,0 +1,215 @@
...
+	i2c = (I2C *)(((mbed_i2c_desc *)(desc->extra))->i2c_port);
+
+	/**
+	    The MBED I2C API is reversed for parameter 4
+	    Instead of stop_bit - it has
+	    @param repeated   - Repeated start, true - don't send stop at end default value is false.
+	    Inverting here to keep the no-OS/platform_drivers API
+	 */
+	if (!(i2c->read(desc->slave_address, (char *)data, bytes_number, !stop_bit))) {
+		return SUCCESS;
```
amiclaus, line 197: ditto

amiclaus (commented):
Also please provide commit description, not only the title

drivers/platform/mbed/mbed_i2c.cpp
mphalke: I have updated this code as per review comment.

adrimbarean (approved):
Looks good to me.

amiclaus (approved):

PR #1235, opened 2022-02-01
drivers:platform:mbed: GPIO platform driver implementation

mphalke:
Implemented the GPIO read/write functionality for Mbed platform.

### DIFF
--- a/drivers/platform/mbed/mbed_gpio.cpp
+++ b/drivers/platform/mbed/mbed_gpio.cpp
@@ -1,0 +1,322 @@
+/***************************************************************************//**
+ *   @file   mbed_gpio.cpp
+ *   @brief  Implementation of GPIO Mbed platform driver interfaces
+********************************************************************************
+ * Copyright (c) 2019-2022 Analog Devices, Inc.
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *  - Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ *  - Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in
+ *    the documentation and/or other materials provided with the
+ *    distribution.
+ *  - Neither the name of Analog Devices, Inc. nor the names of its
+ *    contributors may be used to endorse or promote products derived
+ *    from this software without specific prior written permission.
+ *  - The use of this software may or may not infringe the patent rights
+ *    of one or more patent holders.  This license does not release you
+ *    from the requirement that you obtain separate licenses from these
+ *    patent holders to use this software.
+ *  - Use of the software either in source or binary form, must be run
+ *    on or directly connected to an Analog Devices Inc. component.
+ *
+ * THIS SOFTWARE IS PROVIDED BY ANALOG DEVICES "AS IS" AND ANY EXPRESS OR
+ * IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, NON-INFRINGEMENT,
+ * MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED.
+ * IN NO EVENT SHALL ANALOG DEVICES BE LIABLE FOR ANY DIRECT, INDIRECT,
+ * INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, INTELLECTUAL PROPERTY RIGHTS, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+*******************************************************************************/
+
+/******************************************************************************/
+/***************************** Include Files **********************************/
+/******************************************************************************/
+
+#include <stdio.h>
+#include <stdlib.h>
+#include <mbed.h>
+
+// Platform drivers needs to be C-compatible to work with other drivers
+#ifdef __cplusplus
+extern "C"
+{
+#endif //  _cplusplus
+
+#include "no-os/error.h"
+#include "no-os/gpio.h"
+#include "mbed_gpio.h"
+
+/******************************************************************************/
+/************************ Functions Definitions *******************************/
+/******************************************************************************/
+
+/**
+ * @brief Obtain the GPIO descriptor.
+ * @param desc[in,out] - The GPIO descriptor.
+ * @param param[in] - The structure that contains the GPIO init parameters.
+ * @return SUCCESS in case of success, negative error code otherwise.
+ */
+int32_t mbed_gpio_get(struct gpio_desc **desc,
+		      const struct gpio_init_param *param)
+{
+	struct gpio_desc *gpio_desc;
+	struct mbed_gpio_desc *mbed_gpio_desc;
+
+	if (!desc || !param)
+		return -EINVAL;
+
+	/* Create gpio description object for the device */
+	gpio_desc = (struct gpio_desc *)calloc(1, sizeof(*gpio_desc));
+	if (!gpio_desc)
+		return -ENOMEM;
+
+	gpio_desc->number = param->number;
+
+	/* Create the gpio extra descriptor object to store extra mbed gpio info */
+	mbed_gpio_desc = (struct mbed_gpio_desc *)calloc(1,
+			 sizeof(*mbed_gpio_desc));
+	if (!mbed_gpio_desc)
+		goto err_mbed_gpio_desc;
+
+	mbed_gpio_desc->direction = GPIO_IN;
+	mbed_gpio_desc->gpio_pin = NULL;
+
+	if (param->extra)
+		mbed_gpio_desc->pin_mode = ((struct mbed_gpio_init_param *)
+					    param->extra)->pin_mode;
+	else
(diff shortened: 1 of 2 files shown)


drivers/platform/mbed/mbed_gpio.cpp
```diff
@@ -0,0 +1,308 @@
...
+		gpio_output = new DigitalOut((PinName)desc->number);
+		if (!gpio_output)
+			return -ENOMEM;
+
+		gpio_desc_extra->gpio_pin = (struct mbed_gpio_desc *)gpio_output;
+		gpio_desc_extra->direction = GPIO_OUT;
+
+		/* Set the GPIO value */
+		if (gpio_set_value(desc, value) != SUCCESS) {
+			free(gpio_output);
```
adrimbarean, line 210: Should we use `delete` to correctly pair with `new`?

drivers/platform/mbed/mbed_gpio.cpp
```diff
@@ -0,0 +1,308 @@
...
+		/* Configure and instantiate GPIO pin as output */
+		gpio_output = new DigitalOut((PinName)desc->number);
+		if (!gpio_output)
+			return -ENOMEM;
+
+		gpio_desc_extra->gpio_pin = (struct mbed_gpio_desc *)gpio_output;
+		gpio_desc_extra->direction = GPIO_OUT;
+
+		/* Set the GPIO value */
+		if (gpio_set_value(desc, value) != SUCCESS) {
```
adrimbarean, line 209: Would it be better to use `gpio_output->write(value)`? `gpio_set_value` seems a bit unclear what it will do.

drivers/platform/mbed/mbed_gpio.cpp
```diff
@@ -0,0 +1,308 @@
...
+{
+	DigitalOut *gpio_output; 		// pointer to gpio output object
+	mbed_gpio_desc *gpio_desc_extra;    // pointer to gpio desc extra parameters
+
+	if (!desc|| !desc->extra)
+		return -EINVAL;
+
+	gpio_desc_extra = (struct mbed_gpio_desc *)(desc->extra);
+
+	if (gpio_desc_extra->gpio_pin) {
```
adrimbarean, line 261: Is it worthwhile to get the direction first to make sure the pin is output? What happens if we try to write an input pin?

drivers/platform/mbed/mbed_gpio.cpp
```diff
@@ -0,0 +1,308 @@
...
+{
+	DigitalIn *gpio_input;    	// pointer to gpio input object
+	int32_t ret;
+
+	if (!desc || !desc->extra || !value)
+		return -EINVAL;
+
+	gpio_input = (DigitalIn *)((struct mbed_gpio_desc *)desc->extra)->gpio_pin;
+	*value = (uint8_t)gpio_input->read();
+	ret = gpio_input->is_connected() ? SUCCESS : -EIO;
```
adrimbarean, line 287: Question: should this be done before trying to read the GPIO value? And should we do this in every function before this?

drivers/platform/mbed/mbed_gpio.cpp
```diff
@@ -0,0 +1,308 @@
...
+int32_t mbed_gpio_get_direction(struct gpio_desc *desc, uint8_t *direction)
+{
+	mbed_gpio_desc *gpio_desc_extra;     // pointer to gpio desc extra parameters
+
+	if (!desc || !desc->extra || !direction)
+		return -EINVAL;
+
+	gpio_desc_extra = (struct mbed_gpio_desc *)(desc->extra);
+
+	if (gpio_desc_extra->gpio_pin)
```
amiclaus, line 237: I think you should do something like:
```
if (!(gpio_desc_extra->gpio_pin))
    return -ENODEV (or -EINVAL);

*direction = gpio_desc_extra->direction;

return SUCCESS;
```

drivers/platform/mbed/mbed_gpio.cpp
```diff
@@ -0,0 +1,308 @@
...
+{
+	DigitalIn *gpio_input;				// pointer to gpio input object
+	struct mbed_gpio_desc
+		*gpio_desc_extra; 	// pointer to gpio desc extra parameters
+
+	if (!desc || !desc->extra)
+		return -EINVAL;
+
+	gpio_desc_extra = (struct mbed_gpio_desc *)(desc->extra);
+	if (gpio_desc_extra->gpio_pin == NULL) {
```
adrimbarean, line 162: Suggestion: I think we could improve the code if we check `if (gpio_desc_extra->gpio_pin != NULL)` and do a `delete gpio_desc_extra->gpio_pin;` if we have instantiated it before. This can make the code easier to read, in my opinion, and can improve functionality by allowing us to change the direction of the pin after we set it once.

This change is not necessary if you think it doesn't make sense. Let me know.

drivers/platform/mbed/mbed_gpio.cpp
```diff
@@ -0,0 +1,308 @@
...
+ * @return SUCCESS in case of success, negative error code otherwise.
+ */
+int32_t mbed_gpio_remove(struct gpio_desc *desc)
+{
+	if (!desc || !desc->extra)
+		return -EINVAL;
+
+	/* Free the gpio object */
+	if (((struct mbed_gpio_desc *)(desc->extra))->gpio_pin)
+		free(((struct mbed_gpio_desc *)(desc->extra))->gpio_pin);
```
adrimbarean, line 137: I believe this is allocated with `new` always so we should use `delete` to clean it up. It will have undefined behavior otherwise.

drivers/platform/mbed/mbed_gpio.cpp
adrimbarean: May also apply to `mbed_gpio_direction_output`.

adrimbarean (commented):
A few questions and comments for now.

drivers/platform/mbed/mbed_gpio.cpp
mphalke: This is good point. We can use write() function directly.

drivers/platform/mbed/mbed_gpio.cpp
mphalke: Valid point. Added check for GPIO direction.

drivers/platform/mbed/mbed_gpio.cpp
mphalke: Good point. I think it doesn't make any harm to use this check for every function as it just checks if pin is mapped to valid GPIO and not NC.

drivers/platform/mbed/mbed_gpio.cpp
mphalke: Agreed. I will update code to simplify it.

drivers/platform/mbed/mbed_gpio.cpp
mphalke: Make sense. Updated the code accordingly.

drivers/platform/mbed/mbed_gpio.cpp
mphalke: gpio_pin can be either digital_in Or digital_out instance. Use of delete needs specific typecast to the object and compiler gives warning if not typecasted, so using free here. Is this fine to use free or should we use delete anyway?

mphalke:
Hi, I have pushed the changes. Do you want me to squash the last 2 commits Or two separate commits are okay?

amiclaus (commented):
> Hi, I have pushed the changes. Do you want me to squash the last 2 commits Or two separate commits are okay?

We should not have commits like `Fix review comments` in the history of our master branch.
Therefore I'd suggest squashing them

mphalke:
> > Hi, I have pushed the changes. Do you want me to squash the last 2 commits Or two separate commits are okay?
> 
> We should not have commits like `Fix review comments` in the history of our master branch. Therefore I'd suggest squashing them

I got it, thanks. I have squashed the commits and forced pushed.

drivers/platform/mbed/mbed_gpio.cpp
buha: I also think this should be done with `delete`, it's the proper way to free resources allocated with `new` (performs something more than `free`)

amiclaus (commented):
> > > Hi, I have pushed the changes. Do you want me to squash the last 2 commits Or two separate commits are okay?
> > 
> > 
> > We should not have commits like `Fix review comments` in the history of our master branch. Therefore I'd suggest squashing them
> 
> I got it, thanks. I have squashed the commits and forced pushed.

But we still need to keep track of the changes that were made based on the review. Therefore a changelog should be added. See example: https://github.com/analogdevicesinc/no-OS/pull/1240#issuecomment-1029845317

mphalke:
> > > > Hi, I have pushed the changes. Do you want me to squash the last 2 commits Or two separate commits are okay?
> > > 
> > > 
> > > We should not have commits like `Fix review comments` in the history of our master branch. Therefore I'd suggest squashing them
> > 
> > 
> > I got it, thanks. I have squashed the commits and forced pushed.
> 
> But we still need to keep track of the changes that were made based on the review. Therefore a changelog should be added. See example: [#1240 (comment)](https://github.com/analogdevicesinc/no-OS/pull/1240#issuecomment-1029845317)

How do I generate the change logs as I have already squashed them. Any suggestions?

amiclaus (commented):
> > > > > Hi, I have pushed the changes. Do you want me to squash the last 2 commits Or two separate commits are okay?
> > > > 
> > > > 
> > > > We should not have commits like `Fix review comments` in the history of our master branch. Therefore I'd suggest squashing them
> > > 
> > > 
> > > I got it, thanks. I have squashed the commits and forced pushed.
> > 
> > 
> > But we still need to keep track of the changes that were made based on the review. Therefore a changelog should be added. See example: [#1240 (comment)](https://github.com/analogdevicesinc/no-OS/pull/1240#issuecomment-1029845317)
> 
> How do I generate the change logs as I have already squashed them. Any suggestions?

You don't generate them. You mention what changed after your last force push.

mphalke:
V2:

- Improved the code readability by restructuring the negative error check. Negative error checks are added at the start of all functions.
- Added GPIO mapping validation using 'is_connected()' function in all the GPIO read/write functions.
- Added check for GPIO direction before writing to input GPIO.

mphalke:
V3: 
- Using 'delete' to free the memory allocated for mbed GPIO descriptor instead of 'free' in gpio_remove() function.

amiclaus (approved):

buha (approved):

PR #1236, opened 2022-02-02
stm32 gpio: remove MODER access

### LINKED ISSUE #1228: ADXRS453 Driver compile error on stm32f1xx
It looks like ADXRS453 Driver is not suitable for stm32f1xx since GPIO_TypeDef struct on this ucontroller does not have MODER member. [Context.](https://github.com/Bodmer/TFT_eSPI/issues/589)

`arm-none-eabi-gcc "../Core/Src/stm32_gpio.c" -mcpu=cortex-m3 -std=gnu11 -g3 -DDEBUG -DUSE_HAL_DRIVER -DSTM32F103xB -c -I../Core/Inc -I../Drivers/STM32F1xx_HAL_Driver/Inc -I../Drivers/STM32F1xx_HAL_Driver/Inc/Legacy -I../Drivers/CMSIS/Device/ST/STM32F1xx/Include -I../Drivers/CMSIS/Include -O0 -ffunction-sections -fdata-sections -Wall -fstack-usage -MMD -MP -MF"Core/Src/stm32_gpio.d" -MT"Core/Src/stm32_gpio.o" --specs=nano.specs -mfloat-abi=soft -mthumb -o "Core/Src/stm32_gpio.o"`

`../Core/Src/stm32_gpio.c: In function 'stm32_gpio_get_direction':`

`../Core/Src/stm32_gpio.c:280:29: error: 'GPIO_TypeDef' {aka 'struct <anonymous>'} has no member named 'MODER'; did you mean 'ODR'?
  280 |  *direction = (extra->port->MODER >> (2 * pin)) & 0x3;
      |                             ^~~~~
      |                             ODR`

buha:
MODER register seems to not be present on all stm32 mcu's so
avoid accessing it directly.

This commit changes the gpio_get_direction implementation on stm32
by reading the direction configuration set using the initialization
functions. So it doesn't perform a register access.

Closes #1228

### DIFF
--- a/drivers/platform/stm32/stm32_gpio.c
+++ b/drivers/platform/stm32/stm32_gpio.c
@@ -97,6 +97,14 @@ static int32_t _gpio_init(struct gpio_desc *desc,
 	else
 		return -EINVAL;
 
+	switch (pextra->mode & GPIO_MODE) {
+	case MODE_INPUT:
+	case MODE_OUTPUT:
+		break;
+	default:
+		return -EINVAL;
+	}
+
 	/* copy the settings to gpio descriptor */
 	desc->number = param->number;
 	extra->port = pextra->port;
@@ -272,12 +280,7 @@ int32_t stm32_gpio_get_direction(struct gpio_desc *desc,
 		return -EINVAL;
 
 	struct stm32_gpio_desc *extra = desc->extra;
-	uint8_t pin = desc->number;
-
-	/* MODER = 0x00 - input */
-	/* MODER = 0x01 - general purpose output mode */
-	*direction = (extra->port->MODER >> (2 * pin)) & 0x3;
-
+	*direction = (extra->mode & GPIO_MODE) ? GPIO_OUT : GPIO_IN;
 	return 0;
 }
 


drivers/platform/stm32/stm32_gpio.c
```diff
@@ -97,6 +97,14 @@ static int32_t _gpio_init(struct gpio_desc *desc,
 	else
 		return -EINVAL;
 
+	switch (pextra->mode & GPIO_MODE) {
```
amiclaus, line 100: Is this check really required? Since we have multiple `pextra` members that are directly assigned to the `extra` struct without any prior checks. (although it's good to have this kind of checks for all init params if possible)

amiclaus (approved):
LGTM, one minor question.

drivers/platform/stm32/stm32_gpio.c
buha: On stm32, (pextra->mode & GPIO_MODE) can have the following values:
```
#define MODE_INPUT                              (0x0UL << GPIO_MODE_Pos)
#define MODE_OUTPUT                             (0x1UL << GPIO_MODE_Pos)
#define MODE_AF                                 (0x2UL << GPIO_MODE_Pos)
#define MODE_ANALOG                             (0x3UL << GPIO_MODE_Pos)
```
It doesn't make sense for the GPIO driver to allow anything else than input and output. AF is alternate function (like SPI, I2C etc,) and ANALOG is for ADC/DAC pins.

This is why I put the check, I think it makes sense to have it there.

PR #1237, opened 2022-02-02
drivers:platform:mbed: Delay platform driver implementation

mphalke:
Implemented the usec and msec delay functions for mbed platform

### DIFF
--- a/drivers/platform/mbed/mbed_delay.cpp
+++ b/drivers/platform/mbed/mbed_delay.cpp
@@ -1,0 +1,82 @@
+/***************************************************************************//**
+ *   @file   mbed_delay.cpp
+ *   @brief  Implementation of Mbed specific delay functionality
+********************************************************************************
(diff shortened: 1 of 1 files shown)


drivers/platform/mbed/mbed_delay.cpp
```diff
@@ -0,0 +1,82 @@
...
+
+/**
+ * @brief Generate miliseconds delay.
+ * @param msecs[in] - Delay in miliseconds.
+ * @return None.
+ */
+void mdelay(uint32_t msecs)
+{
+	if (msecs)
+		thread_sleep_for(msecs);
```
amiclaus, line 77: why not `wait_us`?

mphalke: We can use it but already there is a thread_sleep function which accepts value in msec directly. Else we have to convert msec into usec by multiplying msec value by 1000.

amiclaus (approved):

buha (approved):

PR #1238, opened 2022-02-02
drivers:platform:mbed: UART/VCOM platform drivers implementation

mphalke:
Implemented the UART and USBCDC (VirtualCOM) functionality for Mbed platform to read/write data over serial link.
USBCDC (VCOM) is supported/tested fully on SDP-K1 platform.

### DIFF
--- a/drivers/platform/mbed/mbed_uart.cpp
+++ b/drivers/platform/mbed/mbed_uart.cpp
@@ -1,0 +1,478 @@
+/***************************************************************************//**
+ * @file  mbed_uart.cpp
+ * @brief Implementation of UART Mbed platform driver interfaces
+********************************************************************************
+ * Copyright (c) 2021-22 Analog Devices, Inc.
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *  - Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ *  - Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in
+ *    the documentation and/or other materials provided with the
+ *    distribution.
+ *  - Neither the name of Analog Devices, Inc. nor the names of its
+ *    contributors may be used to endorse or promote products derived
+ *    from this software without specific prior written permission.
+ *  - The use of this software may or may not infringe the patent rights
+ *    of one or more patent holders.  This license does not release you
+ *    from the requirement that you obtain separate licenses from these
+ *    patent holders to use this software.
+ *  - Use of the software either in source or binary form, must be run
+ *    on or directly connected to an Analog Devices Inc. component.
+ *
+ * THIS SOFTWARE IS PROVIDED BY ANALOG DEVICES "AS IS" AND ANY EXPRESS OR
+ * IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, NON-INFRINGEMENT,
+ * MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED.
+ * IN NO EVENT SHALL ANALOG DEVICES BE LIABLE FOR ANY DIRECT, INDIRECT,
+ * INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, INTELLECTUAL PROPERTY RIGHTS, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+*******************************************************************************/
+
+/******************************************************************************/
+/***************************** Include Files **********************************/
+/******************************************************************************/
+#include <stdio.h>
+#include <mbed.h>
+#include <USBCDC.h>
+
+// Platform drivers needs to be C-compatible to work with other drivers
+#ifdef __cplusplus
+extern "C"
+{
+#endif //  _cplusplus
+
+#include "no-os/error.h"
+#include "no-os/delay.h"
+#include "no-os/uart.h"
+#include "mbed_uart.h"
+
+/* Max size for USB CDC packet during transmit/receive */
+#define USB_CDC_MAX_PACKET_SIZE		64
+
+/* Max allowed length of USB serial number in characters */
+#define USB_SERIAL_NUM_MAX_LENGTH	100
+
+/* Derived USBCDC class to access protected members of USBCDC class */
+class platform_usbcdc :public USBCDC
+{
+private:
+	uint8_t usb_iserial_descriptor[(USB_SERIAL_NUM_MAX_LENGTH * 2) + 2];
+
+public :
+	/* Call parent class (USBCDC) constructor explicitly */
+	platform_usbcdc(bool connect_blocking, uint16_t vendor_id, uint16_t product_id,
+			const char *serial_number)
+		: USBCDC(connect_blocking, vendor_id, product_id)
+	{
+		uint8_t usb_iserial_len;	// USB serial number length
+		uint8_t i, j = 0;
+
+		usb_iserial_len = strlen(serial_number);
+		if (usb_iserial_len > USB_SERIAL_NUM_MAX_LENGTH) {
+			usb_iserial_len = USB_SERIAL_NUM_MAX_LENGTH;
+		}
+
+		this->usb_iserial_descriptor[j++] = (usb_iserial_len * 2) + 2;	/* bLength */
+		this->usb_iserial_descriptor[j++] = STRING_DESCRIPTOR;			/* bDescriptorType */
+
+		/* bString iSerial */
+		for (i = 0; i < usb_iserial_len; i++) {
+			this->usb_iserial_descriptor[j++] = serial_number[i];
+			this->usb_iserial_descriptor[j++] = 0;
+		}
+	}
+
+	/* Override the virtual function that sets the USB serial number
+	 * The custom (user provided) USB serial number is passed through this function */
+	virtual const uint8_t *string_iserial_desc()
+	{
+		return this->usb_iserial_descriptor;
+	}
+
+	void change_terminal_connection(bool connect_status);
+	bool data_received(uint32_t rx_size);
+	bool data_transmited(void);
+};
+
+/******************************************************************************/
+/************************ Functions Definitions *******************************/
+/******************************************************************************/
+
+/**
+ * @brief Change the terminal connection status (for non-terminal) based USB interface
+ * @param connect_status- new connection status
+ * @note  This functions is used to change the terminal connection status of USB client
+ *        interface which is different than the 'console terminal'. The console terminals acknowledge
+ *        back to USB host when USB connection is opened on the console terminal and Mbed USBCDC
+ *        class automatically changes the '_terminal_connected' status accordingly. However, for
+ *        custom PC applications (non terminal), the terminal connection status needs to be changed
+ *        manually. The '_terminal_connected' is protected member of USBCDC parent class and thus can
+ *        be accessed through 'platform_usbcdc' derived class using below function.
+ */
+void platform_usbcdc::change_terminal_connection(bool connect_status)
+{
+	_terminal_connected = connect_status;
+}
+
+/**
+ * @brief	Check if new USB data is received/available
+ * @param	bytes[in] - Number of expected bytes to be received
+ * @return	true if expected number of bytes received, else false
+ */
+bool platform_usbcdc::data_received(uint32_t bytes)
+{
+	volatile static uint32_t *rx_size = &_rx_size;
+
+	if (*rx_size >= bytes)
+		return true;
+	else
+		return false;
+}
+
+/**
+ * @brief  Check if old USB data was transmitted
+ * @return true if transmit not in progress, else false
+ */
+bool platform_usbcdc::data_transmited(void)
+{
+	volatile static bool *tx_in_progress = &_tx_in_progress;
+	return !(*tx_in_progress);
+}
+
+/**
+ * @brief Read data from UART device.
+ * @param desc[in, out] - Instance of UART.
+ * @param data[out] - Pointer to buffer containing data.
+ * @param bytes_number[in] - Number of bytes to read.
+ * @return SUCCESS in case of success, negative error code otherwise.
+ */
+int32_t uart_read(struct uart_desc *desc, uint8_t *data,
+		  uint32_t bytes_number)
+{
+	mbed::BufferedSerial *uart;	// pointer to BufferedSerial/UART instance
+	platform_usbcdc *usb_cdc_dev;	// Pointer to usb cdc device class instance
+	uint32_t size_rd;
+
+	if (!desc || !desc->extra || !data)
+		return -EINVAL;
+
+	if (!((struct mbed_uart_desc *)(desc->extra))->uart_port)
+		return -EINVAL;
+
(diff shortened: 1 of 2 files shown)


drivers/platform/mbed/mbed_uart.cpp
```diff
@@ -0,0 +1,420 @@
...
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+*******************************************************************************/
+
+/******************************************************************************/
+/***************************** Include Files **********************************/
+/******************************************************************************/
+#include <stdio.h>
+#include <mbed.h>
+#include <USBCDC.h>
```
buha, line 43: where dos this file come from ? it it a separate package dependency ?

drivers/platform/mbed/mbed_uart.cpp
```diff
@@ -0,0 +1,420 @@
...
+		return - EINVAL;
+
+	if (!((struct mbed_uart_desc *)(desc->extra))->uart_port)
+		return -EINVAL;
+
+	uart = (UnbufferedSerial *)(((struct mbed_uart_desc *)(
+					     desc->extra))->uart_port);
+
+	for (i = 0; i < bytes_number; i++)
+		uart->write(&data[i], 1);
```
buha, line 309: Looks like you are using the same ->write() function as in the uart_write() blocking function so this is not actually nonblocking ? Maybe I'm missing something

drivers/platform/mbed/mbed_uart.cpp
```diff
@@ -0,0 +1,420 @@
...
+
+	if (!desc || !param)
+		return -EINVAL;
+
+	// Create the UART description object for the device
+	uart_desc = (struct uart_desc *)calloc(1, sizeof(*uart_desc));
+	if (!uart_desc)
+		return -ENOMEM;
+
+	uart_desc->baud_rate = param->baud_rate;
```
buha, line 335: I find it interesting that you don't use the uart_init_param fields, except for this one - baud_rate. The unused ones are:
```
/** UART Device ID */
uint8_t	device_id;
/** UART number of data bits */
enum uart_size		size;
/** UART parity */
enum uart_parity	parity;
/** UART number of stop bits */
enum uart_stop		stop;
```

mphalke: Thanks for pointing this. We were always using default values for these parameters however it make sense to update them in the mbed layer for custom needs. I will updated the drivers accordingly.

drivers/platform/mbed/mbed_uart.cpp
mphalke: Looks like there is a "writable()" function that should have been called before calling "write()" similar to non-blocking read.
I will correct the code.

drivers/platform/mbed/mbed_uart.cpp
mphalke: It comes from the mbed-os library however it is not included in the standard mbed drivers directory and therefor mbed.h file does not include this file. 
The usb drivers are part of separate directory under drivers and added into the project during project creation if mbed_app.json file explicitly includes it.

image

mphalke:
Also the UnbufferedSerial object needs to be changed to BufferedSerial as buffered serial uses internal UART buffer to store the received data, which prevents flushing the old data if new data arrives immediately. I have made this change which was missing in the first commit.

drivers/platform/mbed/mbed_uart.cpp
buha: I'm just wondering if a user just wants basic UART without caring about UART over USB (USBCDC). The driver file will include the USBCDC.h file anyway which is probably not what the user wants. I don't see how else we could go around this other than #ifdefs

buha (approved):
Looks good, my only gripe is about that USBCDC.h, which is debatable.

drivers/platform/mbed/mbed_uart.cpp
mphalke: That's true and only way seems to be using #if-def which doesn't seem to be feasible as selection b/w UART and VCOM is done from init_param structure at run time.

drivers/platform/mbed/mbed_uart.cpp
```diff
@@ -0,0 +1,496 @@
...
+{
+#endif //  _cplusplus
+
+#include "no-os/error.h"
+#include "no-os/delay.h"
+#include "no-os/uart.h"
+#include "mbed_uart.h"
+
+/* Max size for USB CDC packet during transmit/receive */
+#define USB_CDC_MAX_PACKET_SIZE		(64)
```
amiclaus, line 57: nit: are `()` really needed?

drivers/platform/mbed/mbed_uart.cpp
```diff
@@ -0,0 +1,496 @@
...
+
+/* Max allowed length of USB serial number in characters */
+#define USB_SERIAL_NUM_MAX_LENGTH	(100)
```
amiclaus, line 60: ditto

drivers/platform/mbed/mbed_uart.cpp
mphalke: Don't think they are needed. I'll remove them.

mphalke:
V3:

- Removed the brackets around macros

amiclaus (approved):

drivers/platform/mbed/mbed_uart.cpp
```diff
@@ -0,0 +1,496 @@
...
+ * @return true if transmit not in progress, else false
+ */
+bool platform_usbcdc::data_transmited(void)
+{
+	volatile static bool *tx_in_progress = &_tx_in_progress;
+
+	if (*tx_in_progress)
+		return false;
+	else
+		return true;
```
adrimbarean, line 150: `return !(*tx_in_progress);` might be easier here.

drivers/platform/mbed/mbed_uart.cpp
```diff
@@ -0,0 +1,496 @@
...
+	if (!((struct mbed_uart_desc *)(desc->extra))->uart_port)
+		return -EINVAL;
+
+	if (((struct mbed_uart_desc *)desc->extra)->virtual_com_enable) {
+		usb_cdc_dev = (platform_usbcdc *)((struct mbed_uart_desc *)(
+				desc->extra))->uart_port;
+
+		while (!usb_cdc_dev->data_received(bytes_number)) {
+			/* Wait until new data is available */
+		}
```
adrimbarean, line 180: Not necessary right now, but it might be good to have a timeout.

mphalke: This is blocking read so not sure if we need timeout? Also it is difficult to estimate timeout value as it depends upon target MCU clock frequency and instruction set, so may be put it as MAX_INT32 if needed at all?

mphalke:
V4:

- Modified the return logic of data_transmited() method.

drivers/platform/mbed/mbed_uart.cpp
```diff
@@ -0,0 +1,492 @@
...
+		return -EINVAL;
+
+	if (!((struct mbed_uart_desc *)(desc->extra))->uart_port)
+		return -EINVAL;
+
+	if (((struct mbed_uart_desc *)desc->extra)->virtual_com_enable) {
+		usb_cdc_dev = (platform_usbcdc *)((struct mbed_uart_desc *)(
+				desc->extra))->uart_port;
+
+		while (bytes_number) {
```
adrimbarean, line 219: nitpick: `while (true)` is more correct, if what we want to do is wait for a previous transfer to complete. We never know where bugs might arise.

drivers/platform/mbed/mbed_uart.cpp
```diff
@@ -0,0 +1,492 @@
...
+		/* Change terminal connection status manually */
+		usb_cdc_dev->change_terminal_connection(true);
+
+		usb_cdc_dev->receive_nb(data, bytes_number, &size_rd);
+	} else {
+		uart = (BufferedSerial *)(((struct mbed_uart_desc *)(
+						   desc->extra))->uart_port);
+
+		for (cnt = 0; cnt < bytes_number; cnt++)
+			uart->read(data + cnt, 1);
```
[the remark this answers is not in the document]
mphalke: I think that's right, we can use that way. I was facing some issue earlier while reading all bytes simultaneously so used 1 byte read approach. But I will test and confirm this again.

drivers/platform/mbed/mbed_uart.cpp
```diff
@@ -0,0 +1,492 @@
...
+		return -EINVAL;
+
+	if (!((struct mbed_uart_desc *)(desc->extra))->uart_port)
+		return -EINVAL;
+
+	uart = (BufferedSerial *)(((struct mbed_uart_desc *)(
+					   desc->extra))->uart_port);
+
+	for (i = 0; i < bytes_number; i++) {
+		if (uart->readable() > 0)
```
adrimbarean, line 272: If `uart->readable()` returns number of bytes available it should be queried first and if not enough, then exit directly with return 0, in my opinion. The way it is now it will try to read a number of times equal to the number requested, read an arbitrary number of bytes and return that it's done. It should only read and return a a positive, > 0 number if it has actually done so.

drivers/platform/mbed/mbed_uart.cpp
```diff
@@ -0,0 +1,492 @@
...
+	mbed::BufferedSerial *uart;	// pointer to BufferedSerial/UART instance
+	uint32_t i;
+
+	if (!desc || !desc->extra || !data)
+		return -EINVAL;
+
+	if (!((struct mbed_uart_desc *)(desc->extra))->uart_port)
+		return -EINVAL;
+
+	uart = (BufferedSerial *)(((struct mbed_uart_desc *)(
```
adrimbarean, line 268: Is there a virtual COM option for this one?

drivers/platform/mbed/mbed_uart.cpp
```diff
@@ -0,0 +1,492 @@
...
+		return - EINVAL;
+
+	if (!((struct mbed_uart_desc *)(desc->extra))->uart_port)
+		return -EINVAL;
+
+	uart = (BufferedSerial *)(((struct mbed_uart_desc *)(
+					   desc->extra))->uart_port);
+
+	for (i = 0; i < bytes_number; i++) {
+		if (uart ->writable() > 0)
```
adrimbarean, line 305: Same as the read: we should query once at the beginning if it's writable, and, if yes, `uart->write(data, bytes_number);`. There may be some checks here that we can actually transfer that many.

drivers/platform/mbed/mbed_uart.cpp
```diff
@@ -0,0 +1,492 @@
...
+		if (!uart)
+			goto err_uart;
+
+		/* Set the UART format */
+		if (mbed_uart_set_format(param, uart) != SUCCESS)
+			goto err_uart;
+	}
+
+	// Create a new mbed descriptor to store new UART instances
+	mbed_uart_desc = (struct mbed_uart_desc *)calloc(1, sizeof(*mbed_uart_desc));
```
adrimbarean, line 430: Can we do this before the previous `if` and include the following `if` in the previous one? If it makes sense it will make for a smoother flow, in my opinion.

drivers/platform/mbed/mbed_uart.cpp
```diff
@@ -0,0 +1,492 @@
...
+ */
+int32_t uart_remove(struct uart_desc *desc)
+{
+	if (!desc || !desc->extra)
+		return -EINVAL;
+
+	/* Free the UART object */
+	if (((struct mbed_uart_desc *)desc->extra)->virtual_com_enable) {
+		if ((platform_usbcdc *)((struct mbed_uart_desc *)desc->extra)->uart_port)
+			delete((platform_usbcdc *)(platform_usbcdc *)((struct mbed_uart_desc *)
```
adrimbarean, line 475: Double cast here.

adrimbarean (commented):
Some questions and suggestions for now.

drivers/platform/mbed/mbed_uart.cpp
mphalke: What readable() mbed implementation says is, it will "return true if there is something to read else return false". It does not return number of bytes. Return type of readable is bool. I think I made a mistake there. I should check for true or false instead of > 0.
It does not explicitly mention or allow to pass number of bytes available to read. It's bit tricky I guess.

drivers/platform/mbed/mbed_uart.cpp
mphalke: We haven't figured that out yet Or not implemented. I have mentioned in the function description.

drivers/platform/mbed/mbed_uart.cpp
mphalke: Yes. I will update the code.

mphalke:
V5:

- Corrected the readable() and writable() funcion calling logic. Error checks are added for the same.
- Corrected blocking read logic. Calling read(data, cnt) to read all bytes instead of single byte at a time.
- Corrected descriptor/instances memory freeing order in the uart_init() function.
-  Other minor edits as per code review comments.

mphalke:
Hi All,
Are we good with the changes or are there any more comments which I need to address?

drivers/platform/mbed/mbed_uart.cpp
```diff
@@ -0,0 +1,492 @@
...
+	uart = (BufferedSerial *)(((struct mbed_uart_desc *)(
+					   desc->extra))->uart_port);
+
+	/* Return 0 bytes if nothing to read */
+	if (!uart->readable())
+		return 0;
+
+	for (i = 0; i < bytes_number; i++) {
+		if (uart->readable())
+			uart->read(&data[i], 1);
```
adrimbarean, line 273: I'd say we should `break` on the `else` path and return `i` instead of `bytes_number`. Because the danger here is that the user will ask for, let's say 5 bytes, and the buffer has 2 bytes. The function will read 2 bytes and then stay a little while more in the function and return that it has read 5 bytes, which might cause errors.

drivers/platform/mbed/mbed_uart.cpp
```diff
@@ -0,0 +1,492 @@
...
+	uart = (BufferedSerial *)(((struct mbed_uart_desc *)(
+					   desc->extra))->uart_port);
+
+	/* Return error if unable to write */
+	if (!uart->writable())
+		return -EIO;
+
+	for (i = 0; i < bytes_number; i++) {
+		if (uart ->writable())
+			uart->write(&data[i], 1);
```
adrimbarean, line 310: I'd say things are easier here. We just have to account for the maximum number of bytes that can be queued to be written at once. If `writable`, compare `bytes_number` with the maximum that can be queued at once and queue the smallest number. Then return the number of bytes queued. The for shouldn't be necessary here.

adrimbarean (commented):
I'd say we still need to address two things before we can properly merge here.

drivers/platform/mbed/mbed_uart.cpp
mphalke: Make sense. I'll update the code.

drivers/platform/mbed/mbed_uart.cpp
mphalke: Not sure if I get you correctly here. 
Do you mean:
if (uart->writable())
		return uart->write(data, bytes_number);
else
		return 0;

adrimbarean: Yes, this ca work, but does `uart->write` have a maximum number of bytes that can be transmitted at once? From the documentation it seems that it will automatically write a partial write if the requested transfer is too big. So we can do just:

if (uart->writable())
return uart->write(data, bytes_number);

Or it might even work without `if` because if it can't write anything it will return `-EAGAIN`. I'd say both can work.
Note: `uart->write` will return the number of bytes it transmitted, or queued, even if it's just a partial transfer.

mphalke: Okay. I will implement the first one. I was also thinking, we can do same for non-blocking read. If ```readable()``` is true, we will just return ```read()``` with number of bytes else will return 0.
e.g. 
```
if (readable())
    return read(data, bytes)
else
   return 0
```
Let me know if this is okay to implement

adrimbarean: This seems good. Thank you!

mphalke:
V6:

- Updated the UART non-blocking read and write functions. ```readable() or writable()``` conditional check is added before ```read/write``` functions. The number of read/write bytes are returned based on the available UART buffer size without blocking.

adrimbarean (approved):
Looks good to me.

PR #1243, opened 2022-02-09
drivers:platform:mbed: Timer platform driver implementation

mphalke:
1. Implemented the timer init, timer start, timer stop and timer remove functions.
2. Mbed timer has very limited scope. User can't load the timer with specific value neither can set the
timer frequency as these parameters are dependent upon the low level hardware. The functions such as
timer_clk_get and timer_cnt_get are not supported and implemented in the mbed drivers.
3. New get_elapsed_time_in_nsec() function is created in timer.cpp file to get the elapsed time in nsec
as mbed doesn't have any function to get the elapsed timer count in decimal.

### DIFF
--- a/drivers/platform/mbed/mbed_timer.cpp
+++ b/drivers/platform/mbed/mbed_timer.cpp
@@ -1,0 +1,191 @@
+/**************************************************************************//**
+ *   @file   mbed_timer.cpp
+ *   @brief  Implementation of Timer Mbed Platform Driver Interfaces
+*******************************************************************************
+ * Copyright (c) 2021-22 Analog Devices, Inc.
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *  - Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ *  - Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in
+ *    the documentation and/or other materials provided with the
+ *    distribution.
+ *  - Neither the name of Analog Devices, Inc. nor the names of its
+ *    contributors may be used to endorse or promote products derived
+ *    from this software without specific prior written permission.
+ *  - The use of this software may or may not infringe the patent rights
+ *    of one or more patent holders.  This license does not release you
+ *    from the requirement that you obtain separate licenses from these
+ *    patent holders to use this software.
+ *  - Use of the software either in source or binary form, must be run
+ *    on or directly connected to an Analog Devices Inc. component.
+ *
+ * THIS SOFTWARE IS PROVIDED BY ANALOG DEVICES "AS IS" AND ANY EXPRESS OR
+ * IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, NON-INFRINGEMENT,
+ * MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED.
+ * IN NO EVENT SHALL ANALOG DEVICES BE LIABLE FOR ANY DIRECT, INDIRECT,
+ * INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, INTELLECTUAL PROPERTY RIGHTS, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+*******************************************************************************/
+
+/*****************************************************************************/
+/***************************** Include Files *********************************/
+/*****************************************************************************/
+
+#include <stdio.h>
+#include <mbed.h>
+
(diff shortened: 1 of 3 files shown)


drivers/platform/mbed/mbed_timer.cpp
```diff
@@ -0,0 +1,191 @@
...
+int32_t timer_start(struct timer_desc *desc)
+{
+	mbed::Timer *timer;
+
+	if (!desc || !desc->extra)
+		return -EINVAL;
+
+	timer = (Timer *)(((struct mbed_timer_desc *)(desc->extra))->timer);
+	if (!timer)
+		return -ENOMEM;
```
adrimbarean, line 117: nitpick: since this isn't an allocation EINVAL would be more appropriate.

drivers/platform/mbed/mbed_timer.cpp
```diff
@@ -0,0 +1,191 @@
...
+* @return SUCCESS in case of success, negative error code otherwise.
+*/
+int32_t timer_stop(struct timer_desc *desc)
+{
+	mbed::Timer *timer;
+
+	if (!desc || !desc->extra)
+		return -EINVAL;
+
+	timer = (Timer *)(((struct mbed_timer_desc *)(desc->extra))->timer);
```
amiclaus, line 136: ditto as in https://github.com/analogdevicesinc/no-OS/pull/1243/files#r802329865

drivers/platform/mbed/mbed_timer.cpp
```diff
@@ -0,0 +1,191 @@
...
+*/
+int32_t get_elapsed_time_in_nsec(struct timer_desc *desc,
+				 uint64_t *elapsed_time)
+{
+	mbed::Timer *timer;
+
+	if (!desc || !desc->extra || !elapsed_time)
+		return -EINVAL;
+
+	timer = (Timer *)(((struct mbed_timer_desc *)(desc->extra))->timer);
```
amiclaus, line 180: ditto as in https://github.com/analogdevicesinc/no-OS/pull/1243/files#r802329865

drivers/platform/mbed/mbed_timer.cpp
```diff
@@ -0,0 +1,191 @@
...
+	return SUCCESS;
+}
+
+/**
+* @brief Get the elapsed time in nanoseconds.
+* @param desc[in] - The Timer descriptor.
+* @param elapsed_time[in, out] - Pointer where the elapsed time value is stored.
+* @return SUCCESS in case of success, negative error code otherwise.
+*/
+int32_t get_elapsed_time_in_nsec(struct timer_desc *desc,
```
adrimbarean, line 172: Is there a difference between this method and `timer_counter_get`? Could we use the existing API here?

mphalke: This method returns the elapsed time in usec however I think timer_counter_get method expects the difference between previous timer count and current timer count, which would be then used to calculate the actual elapsed time in time units based on timer frequency (*Please correct me if my understanding is wrong here).

So basically mbed does not have function that gives us the difference b/w timer counts. However, we could use timer_counter_get method and treat it as an API which returns value as elapsed time directly instead of elapsed count.
Only problem is this mbed method uses INT64 type for holding timer elapsed value and timer_counter_get uses uint32 argument for return value, so we have to typecast return value from mbed function to uint32. See below:

*elapsed_time = (uint32_t )(duration_cast<nanoseconds>(timer->elapsed_time()).count());

It may loose some resolution though and result may be incorrect. So not a very good idea to do.
Let me know what you think.

mphalke:
V2:

- Replaced ENOMEM with EINVAL return code in timer start, stop and get_count functions.

drivers/platform/mbed/mbed_timer.cpp
amiclaus: I agree with @adrimbarean that we should stick to the existing API. I'm wondering if we can use the ticker functions directly, like `timestamp_t ticker_read(const ticker_data_t *const ticker);` from https://os.mbed.com/docs/mbed-os/v6.15/mbed-os-api-doxy/ticker__api_8h_source.html since if I am not mistaken the Ticker API is included in https://os.mbed.com/docs/mbed-os/v6.6/mbed-os-api-doxy/_timer_8h_source.html

mphalke: Hi All,
Looks like lack of mbed API that directly returns timer count in decimal, makes it difficult to make mbed timer code compatible with no-os driver timer include API's. I understand 'timer_counter_get()' method is mandatory to use mbed timer module with other no-os projects. 

Do you want me to delete this PR (that means we will not add Timer support in mbed platform drivers) Or do you want to keep the changes those are done so far (without timer_counter_get() method)?

adrimbarean: Hello,

Let's keep this implementation. But could you add the `get_elapsed_time_in_nsec()` in the [generic timer header](https://github.com/analogdevicesinc/no-OS/blob/master/include/no-os/timer.h)? And call it something with `timer_` at the beginning. We should avoid defining API just for one platform.

mphalke: Sure. I will update the changes.

mphalke:
V3:

- Added timer_get_elapsed_time_nsec() method in no-os/timer.h file and removed existing get_elapsed_time_nsec() method from mbed_timer.h file.

include/no-os/timer.h
```diff
@@ -109,5 +109,9 @@ int32_t timer_count_clk_get(struct timer_desc *desc, uint32_t *freq_hz);
 /* Set the timer clock frequency. */
 int32_t timer_count_clk_set(struct timer_desc *desc, uint32_t freq_hz);
 
+/* Get the elapsed time in nsec for the timer. */
+int32_t timer_get_elapsed_time_nsec(struct timer_desc *desc,
```
adrimbarean, line 113: Can we have this in a separate commit, but same pull-request?

mphalke: I have added separate commit for this.

mphalke:
V4:

- Added separate commit for timer.h include file changes. No change done in the source code.

adrimbarean (approved):
Looks good to me.

PR #1246, opened 2022-02-10
ad719x: Fix ID mask

APorumb:
Previously, the ID mask was set only for the last nibble. AD7195 has a full byte
ID so we need to consider the whole byte. Mask changed to 0xFF for AD7195 case.

### DIFF
--- a/drivers/adc/ad719x/ad719x.h
+++ b/drivers/adc/ad719x/ad719x.h
@@ -127,8 +127,10 @@
 #define AD719X_CH_TEMP   8
 #define AD719X_CH_SHORT  9
 
+
 /* ID Register Bit Designations (AD7193_REG_ID) */
-#define AD719X_ID_MASK          0x0F
+#define AD7190_4_ID_MASK			0x0F
+#define AD7195_ID_MASK				0xFF
 
 /* GPOCON Register Bit Designations (AD719X_REG_GPOCON) */
 #define AD719X_GPOCON_BPDSW     (1 << 6) // Bridge power-down switch enable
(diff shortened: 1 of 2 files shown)


drivers/adc/ad719x/ad719x.h
```diff
@@ -127,8 +127,8 @@
 #define AD719X_CH_TEMP   8
 #define AD719X_CH_SHORT  9
 
-/* ID Register Bit Designations (AD7193_REG_ID) */
-#define AD719X_ID_MASK          0x0F
```
amiclaus, line 131: for AD7193, for example, there are only 4 bits for the chip id. the rest of them are undetermined:
image

This means that the 4 MSB can be either 0 or 1, depending on the chip default configuration.
That might cause some issues at initialization if you expand the mask for all the chips supported by this driver.
I'd rather change the mask inline, depending on the part name.

amiclaus: Or you can leave it as it is (`0x0F`), since for ad7195 the 4LSB are different from any other supported part.

APorumb: > Or you can leave it as it is (`0x0F`), since for ad7195 the 4LSB are different from any other supported part.

What if, in the near future, another ad719x chip will be released with ID 0x06?

Fixed ID mask inline.

amiclaus: > > Or you can leave it as it is (`0x0F`), since for ad7195 the 4LSB are different from any other supported part.
> 
> What if, in the near future, another ad719x chip will be released with ID 0x06?
> 
> Fixed ID mask inline.

I agree, separate masks is a better solution

APorumb:
Edit 2: Added separate mask for AD7195.

PR #1248, opened 2022-02-11
drivers:platform:mbed: IRQ platform drivers implementation

mphalke:
1. Implemented the IRQ handling for External (GPIO), Ticker and UART Rx IRQs.
2. Callback registration and invokement implementaion is slightly different than the conventional
IRQ platform drivers. Dedicated callback is needed for each IRQ. Callback function in mbed
expects no argument and void return type.

**gpio irq example:**

```
// -------------------------- IRQ callback function ----------------------------------
void ext_int_callback(void *ctx, uint32_t event, void *extra)
{
}

// -------------------------- Platform specific ----------------------------------
/* External interrupt Mbed platform specific parameters */
struct mbed_irq_init_param mbed_ext_int_extra_init_params = {
	.int_mode = EXT_IRQ_RISE,
	.ext_int_pin = CONV_MON_GPIO,
	.int_obj_type = NULL
};

// -------------------------- Generic IRQ API stuff ----------------------------------
/* External interrupt generic init parameters */
static struct irq_init_param ext_int_init_params = {
	.irq_ctrl_id = 0,
	.extra = &mbed_ext_int_extra_init_params,
	.platform_ops = &irq_ops
};

/* External interrupt callback descriptor */
static struct callback_desc ext_int_callback_desc = {
	ext_int_callback,
	NULL,
	NULL,
};

/* External interrupt descriptor */
struct irq_ctrl_desc *external_int_desc;

/* Initialize IRQ */
if (irq_ctrl_init(&external_int_desc, &ext_int_init_params) != SUCCESS) {
	return FAILURE;
}

/* Register a callback function for external interrupt */
if (irq_register_callback(external_int_desc,
		  EXTERNAL_INT_ID1,
		  &ext_int_callback_desc) != SUCCESS) {
      return FAILURE;
}

/* Enable external interrupt */
if (irq_enable(external_int_desc, EXTERNAL_INT_ID1) != SUCCESS) {
	return FAILURE;
}
```

**gpio uart_rx example:**
```
// -------------------------- UART Specific ----------------------------------
/* Initialize the UART */
struct uart_desc *uart_desc;
uart_init(&uart_desc, &uart_init_params);

// -------------------------- IRQ callback function ----------------------------------
void uart_rx_callback(void *ctx, uint32_t event, void *extra)
{
}

// -------------------------- Platform specific ----------------------------------
/* UART Rx interrupt Mbed platform specific parameters */
struct mbed_irq_init_param mbed_uart_rx_int_extra_init_params = {
        .irq_ctrl_id = 0,
	.int_obj_type = (((struct mbed_uart_desc *)(uart_desc->extra))->uart_port);
};

// -------------------------- Generic IRQ API stuff ----------------------------------
/* UART Rx interrupt generic init parameters */
static struct irq_init_param uart_rx_int_init_params = {
	.irq_ctrl_id = UART_RX_INT_ID1,
	.extra = &mbed_uart_rx_int_extra_init_params,
	.platform_ops = &irq_ops
};

/* UART Rx interrupt callback descriptor */
static struct callback_desc uart_rx_int_callback_desc = {
	uart_rx_callback,
	NULL,
	NULL,
};

/* UART Rx interrupt descriptor */
struct irq_ctrl_desc *uart_rx_int_desc;

/* Initialize IRQ */
if (irq_ctrl_init(&uart_rx_int_desc, &uart_rx_int_init_params) != SUCCESS) {
	return FAILURE;
}

/* Register a callback function for external interrupt */
if (irq_register_callback(uart_rx_int_desc,
		  UART_RX_INT_ID1,
		  &uart_rx_int_callback_desc) != SUCCESS) {
      return FAILURE;
}

/* Enable external interrupt */
if (irq_enable(uart_rx_int_desc, UART_RX_INT_ID1) != SUCCESS) {
	return FAILURE;
}
```

### DIFF
--- a/drivers/platform/mbed/mbed_irq.cpp
+++ b/drivers/platform/mbed/mbed_irq.cpp
@@ -1,0 +1,508 @@
+/***************************************************************************//**
+ * @file   mbed_irq.cpp
+ * @brief  Implementation of Interrupt Mbed platform driver interfaces
+********************************************************************************
+ * Copyright (c) 2020-2022 Analog Devices, Inc.
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *  - Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ *  - Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in
+ *    the documentation and/or other materials provided with the
+ *    distribution.
+ *  - Neither the name of Analog Devices, Inc. nor the names of its
+ *    contributors may be used to endorse or promote products derived
+ *    from this software without specific prior written permission.
+ *  - The use of this software may or may not infringe the patent rights
+ *    of one or more patent holders.  This license does not release you
+ *    from the requirement that you obtain separate licenses from these
+ *    patent holders to use this software.
+ *  - Use of the software either in source or binary form, must be run
+ *    on or directly connected to an Analog Devices Inc. component.
+ *
+ * THIS SOFTWARE IS PROVIDED BY ANALOG DEVICES "AS IS" AND ANY EXPRESS OR
+ * IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, NON-INFRINGEMENT,
+ * MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED.
+ * IN NO EVENT SHALL ANALOG DEVICES BE LIABLE FOR ANY DIRECT, INDIRECT,
+ * INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, INTELLECTUAL PROPERTY RIGHTS, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+*******************************************************************************/
+
+/******************************************************************************/
+/***************************** Include Files **********************************/
+/******************************************************************************/
+
+#include <stdio.h>
+#include <mbed.h>
+
+// Platform support needs to be C-compatible to work with other drivers
+#ifdef __cplusplus
+extern "C"
+{
+#endif
+
+#include "no-os/error.h"
+#include "no-os/irq.h"
+#include "mbed_irq.h"
+#include "mbed_uart.h"
+
+using namespace std::chrono;
+
+/**
+* @struct mbed_irq_callback_desc.
+* @brief Structure holding the callback functions for mbed irqs.
+* @note  The callback functions are directly mapped with 'irq_id' enum.
+* Mbed doesn't provide a single callback function for all external/GPIO
+* IRQ events. A seperate callback function is required to be registered
+* for every IRQ instance. Also, the 'callback_function' required for
+* mbed layer expects void argument and void return type.
+*/
+struct mbed_irq_callback_desc {
+	struct callback_desc desc;
+	void (*mbed_callback)(void);
+};
+
+/* Mbed irq callback structure variable */
+static struct mbed_irq_callback_desc mbed_irq_callback[NB_INTERRUPTS];
+
+/******************************************************************************/
+/************************ Functions Definitions *******************************/
+/******************************************************************************/
+
+/**
+ * @brief	Mbed callback function for external interrupt ID1 event.
+ * @return	none.
+ * @note	This function is called when external IRQ event mapped to
+ *			EXTERNAL_INT_ID1 is triggered.
+ */
+static void mbed_ext_int_id1_callback(void)
+{
+	if (mbed_irq_callback[EXTERNAL_INT_ID1].desc.callback)
+		/* Invoke the application registered callback function */
+		mbed_irq_callback[EXTERNAL_INT_ID1].desc.callback(
+			mbed_irq_callback[EXTERNAL_INT_ID1].desc.ctx, EXTERNAL_INT_ID1, NULL);
+}
+
+/**
+ * @brief	Mbed callback function for external interrupt ID2 event.
+ * @return	none.
+ * @note	This function is called when external IRQ event mapped to
+ *			EXTERNAL_INT_ID2 is triggered.
+ */
+static void mbed_ext_int_id2_callback(void)
+{
+	if (mbed_irq_callback[EXTERNAL_INT_ID2].desc.callback)
+		/* Invoke the application registered callback function */
+		mbed_irq_callback[EXTERNAL_INT_ID2].desc.callback(
+			mbed_irq_callback[EXTERNAL_INT_ID2].desc.ctx, EXTERNAL_INT_ID2, NULL);
+}
+
+/**
+ * @brief	Mbed callback function for external interrupt ID3 event.
+ * @return	none.
+ * @note	This function is called when external IRQ event mapped to
+ *			EXTERNAL_INT_ID3 is triggered.
+ */
+static void mbed_ext_int_id3_callback(void)
+{
+	if (mbed_irq_callback[EXTERNAL_INT_ID3].desc.callback)
+		/* Invoke the application registered callback function */
+		mbed_irq_callback[EXTERNAL_INT_ID3].desc.callback(
+			mbed_irq_callback[EXTERNAL_INT_ID3].desc.ctx, EXTERNAL_INT_ID3, NULL);
+}
+
+/**
+ * @brief	Mbed callback function for external interrupt ID4 event.
+ * @return	none.
+ * @note	This function is called when external IRQ event mapped to
+ *			EXTERNAL_INT_ID4 is triggered.
+ */
+static void mbed_ext_int_id4_callback(void)
+{
+	if (mbed_irq_callback[EXTERNAL_INT_ID4].desc.callback)
+		/* Invoke the application registered callback function */
+		mbed_irq_callback[EXTERNAL_INT_ID4].desc.callback(
+			mbed_irq_callback[EXTERNAL_INT_ID4].desc.ctx, EXTERNAL_INT_ID4, NULL);
+}
+
+/**
+ * @brief	Mbed callback function for external interrupt ID5 event.
+ * @return	none.
+ * @note	This function is called when external IRQ event mapped to
+ *			EXTERNAL_INT_ID5 is triggered.
+ */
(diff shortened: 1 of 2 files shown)


drivers/platform/mbed/mbed_irq.cpp
```diff
@@ -0,0 +1,569 @@
...
+/**
+* @struct mbed_irq_callback_desc.
+* @brief Structure holding the callback functions for mbed irqs.
+* @note  The callback functions are directly mapped with 'irq_id' enum.
+* Mbed doesn't provide a single callback function for all external/GPIO
+* IRQ events. A seperate callback function is required to be registered
+* for every IRQ instance. Also, the 'callback_function' required for
+* mbed layer expects void argument and void return type.
+*/
+typedef struct {
```
adrimbarean, line 67: We shouldn't use typedef in this case, according to the [Linux kernel coding style](https://www.kernel.org/doc/html/v4.11/process/coding-style.html#typedefs).

drivers/platform/mbed/mbed_irq.cpp
```diff
@@ -0,0 +1,569 @@
...
+	struct callback_desc callback_ext_int_id1;
+	struct callback_desc callback_ext_int_id2;
+	struct callback_desc callback_ext_int_id3;
+	struct callback_desc callback_ext_int_id4;
+	struct callback_desc callback_ext_int_id5;
```
adrimbarean, line 72: We could make those an array `struct callback_desc callback_ext_int_id[5];` and refer to them by index. I think this will help with the double `switch` scenario in the register function.

drivers/platform/mbed/mbed_irq.cpp
```diff
@@ -0,0 +1,569 @@
...
+	if (!mbed_irq_desc)
+		goto err_mbed_irq_desc;
+
+	irq_desc->irq_ctrl_id = param->irq_ctrl_id;
+	mbed_irq_desc->int_mode = ((struct mbed_irq_init_param *)
+				   param->extra)->int_mode;
+	mbed_irq_desc->ext_int_pin = ((struct mbed_irq_init_param *)
+				      param->extra)->ext_int_pin;
+
+	switch (param->irq_ctrl_id) {
```
adrimbarean, line 217: Here, the `irq_ctrl_id` member stands for the ID of the interrupt controller, not the ID of the interrupt. This is so that we can have multiple interrupt controllers in the same system, possibly cascading into each other. We should just save this in the `irq_desc` and do this `switch case` later, in the register interrupt method. Is it possible to do this?

drivers/platform/mbed/mbed_irq.cpp
```diff
@@ -0,0 +1,569 @@
...
+ * @return	SUCCESS in case of success, negative error code otherwise.
+ */
+int32_t mbed_irq_unregister(struct irq_ctrl_desc *desc, uint32_t irq_id)
+{
+	if (!desc || !desc->extra)
+		return -EINVAL;
+
+	switch (irq_id) {
+	case  EXTERNAL_INT_ID1:
+		mbed_irq_callbacks.callback_ext_int_id1.callback = NULL;
```
adrimbarean, line 376: We should call `ext_interrupt->rise/fall(NULL);` here to unregister in the DFP as well.

drivers/platform/mbed/mbed_irq.cpp
```diff
@@ -0,0 +1,569 @@
...
+
+	if (!desc || !desc->extra)
+		return -EINVAL;
+
+	if (!(((struct mbed_irq_desc *)desc->extra)->int_obj))
+		return -EINVAL;
+
+	/* Unregister all callbacks */
+	for (irq_id = 0; irq_id < (uint8_t)NB_INTERRUPTS; irq_id++) {
+		if (mbed_irq_unregister(desc, irq_id) != SUCCESS)
```
adrimbarean, line 518: Should we call `disable` here as well?

drivers/platform/mbed/mbed_irq.cpp
```diff
@@ -0,0 +1,569 @@
...
+	case EXTERNAL_INT_ID4:
+	case EXTERNAL_INT_ID5:
+		delete((InterruptIn *)(((struct mbed_irq_desc *)desc->extra)->int_obj));
+		break;
+
+	case TICKER_INT_ID:
+		/* Ticker doesn't have destructor implementation so only detach is used */
+		ticker = (mbed::Ticker *)(((struct mbed_irq_desc *)desc->extra)->int_obj);
+		ticker->detach();
+		break;
```
adrimbarean, line 535: We should `delete(((struct mbed_irq_desc *)desc->extra)->int_obj);` here as well.

adrimbarean (commented):
Looks good overall. A few comments and questions.

drivers/platform/mbed/mbed_irq.cpp
mphalke: Sorry not sure if I got your comment correctly, so just clarifying it.

Do you mean to say that ```irq_id``` passed as an argument to other functions in this module is not the same as ```irq_ctrl_id```? That means the ```irq_id enum``` value from ```mbed_irq.h``` file will be passed as argument to ```irq_id``` parameter for all other functions except this init function. The code under this ```switch()``` case would be then part of ```mbed_irq_register_callback()``` function.  
If this the case, then it make sense. I was also wondering why ```irq_id``` is passed again to all the functions if it is already passed to init function via irq init structure, but both seems to be different. 

Please confirm my understanding.

adrimbarean: Yes, you are right. `irq_ctrl_id` is the ID of the interrupt controller that is used, and is set in the init function, and `irq_id` is the ID of the interrupt within the interrupt controller.

mphalke: Thanks for the confirmation. I will implement the changes accordingly.

drivers/platform/mbed/mbed_irq.cpp
mphalke: Make sense. I have restructured this bit differently (using array of callback structure) to simplify some of the later code.

mphalke:
V1:

- Created an array of structure for storing irq general callback structure parameters and mbed irq callback function.
- Removed the IRQ object creation from ```irq_ctr_init()``` function and moved into ```mbed_irq_callback_register()``` function as per review comment.
- Any attach (enabling) of interrupt is done from ```mbed_irq_enable()``` function as attaching interrupt automatically enabled it.
- Detach (disabling) of interrupt is also called from ```mbed_irq_unregister_callback()``` function to disable active interrupt callbacks.
- Other restructuring and code cleanup for better readability of code.

mphalke:
Hi All,
Any comments on the recent (V1) changes?

mphalke:
Hi,
Can we merge this PR if no further comments?

buha:
Also, it would help a lot to have a usage example of how to set up a gpio and a uart interrupt on mbed using the code in this PR. I think we can even better discuss on irq setup example than on the driver itself.

mphalke:
> Also, it would help a lot to have a usage example of how to set up a gpio and a uart interrupt on mbed using the code in this PR. I think we can even better discuss on irq setup example than on the driver itself.

@buha I have added usage example on how to setup GPIO and UART Rx interrupts on mbed platform with current implementation.

buha:
Ok, so in light of the irq.c/irq.h changes which hopefully will get accepted:

```
/* UART Rx interrupt callback descriptor */
static struct callback_desc uart_rx_int_callback_desc = {
	uart_rx_callback,
	NULL,                     // <----------------- unused
	NULL,                    // <----------------- unused
};
```
Given that the ctx and config members of callback_desc are not needed, I suggest you implement a platform specific struct mbed_callback that only holds the actual callback function. In my opinion, struct callback_desc should be abolished because it is impossible to have even the same callback signature (some platforms require void func(void), others void func(p1), others int func(p1, p2) etc.).

buha:
Also,

```

/* Register a callback function for external interrupt */
if (irq_register_callback(uart_rx_int_desc,
		  UART_RX_INT_ID1,                                             // <--------------- inq_id
		  &uart_rx_int_callback_desc) != SUCCESS) {
      return FAILURE;
}
```
irq_id should not be a custom defined enum, but it should be the actual interrupt id as taken from the interrupt controller documentation.

In this case, I think `enum irq_id` should be abolished, and we should try to use directly the interrupt ID as exported by the mbed libraries. These irq_* functions are intended to work as interrupt controller control functions, so consequently, the irq_id parameter should be precisely the interrupt number as the interrupt controller understands it on each platform.

drivers/platform/mbed/mbed_irq.h
```diff
@@ -0,0 +1,149 @@
...
+	NB_INTERRUPTS
+};
+
+/*
+ * @enum External/GPIO IRQ events to be set into 'int_mode' parameter.
+ * Example: struct mbed_irq_init_param my_ext_mbed_irq_init_param = {
+ *				.int_mode = EXT_IRQ_RISE
+ *			}
+ * */
+enum ext_irq_event {
```
buha, line 100: In my opinion, this event should go together into the `struct mbed_callback` i mentioned above.
When you configure a GPIO interrupt, you may want to have two callbacks:
```
void rising_edge_irq(void) {
    // do something on rising edge
}

void falling_edge_irq(void) {
    // do something on falling edge
}
```
This should be configurable with `irq_register_callback(desc, irq_id, callback)`.
I think something like:
```
struct mbed_callback rising {
    .callback = rising_edge_irq,
    .event = MBED_RISING // invented macro name, look up the correct one in mbed libraries
}

struct mbed_callback falling {
    .callback = falling_edge_irq,
    .event = MBED_FALLING // invented macro name, look up the correct one in mbed libraries
}

irq_register_callback(desc, irq_id, rising)
irq_register_callback(desc, irq_id, falling)
```

mphalke:
> Ok, so in light of the irq.c/irq.h changes which hopefully will get accepted:
> 
> ```
> /* UART Rx interrupt callback descriptor */
> static struct callback_desc uart_rx_int_callback_desc = {
> 	uart_rx_callback,
> 	NULL,                     // <----------------- unused
> 	NULL,                    // <----------------- unused
> };
> ```
> 
> Given that the ctx and config members of callback_desc are not needed, I suggest you implement a platform specific struct mbed_callback that only holds the actual callback function. In my opinion, struct callback_desc should be abolished because it is impossible to have even the same callback signature (some platforms require void func(void), others void func(p1), others int func(p1, p2) etc.).

This make sense. If 'callback_desc' is abolished, things will be simpler and no need to define callback functions in the mbed_irq.c file as it is done currently. 
So I will wait for the other PR to get approved before doing this change.

mphalke:
> Also,
> 
> ```
> 
> /* Register a callback function for external interrupt */
> if (irq_register_callback(uart_rx_int_desc,
> 		  UART_RX_INT_ID1,                                             // <--------------- irq_id
> 		  &uart_rx_int_callback_desc) != SUCCESS) {
>       return FAILURE;
> }
> ```
> 
> irq_id should not be a custom defined enum, but it should be the actual interrupt id as taken from the interrupt controller documentation.
> 
> In this case, I think `enum irq_id` should be abolished, and we should try to use directly the interrupt ID as exported by the mbed libraries. These irq_* functions are intended to work as interrupt controller control functions, so consequently, the irq_id parameter should be precisely the interrupt number as the interrupt controller understands it on each platform.

The mbed library does not expose the interrupt controller ID/number for each interrupt. Mbed is an abstracted layer that can support multiple MCUs, so details of controller IDs, names, etc is hidden in the HAL layer (vendor specific) residing below Mbed layer, which we can't/shouldn't access directly. I understand other platform drivers are specific to certain MCU Or series of MCU, so it is easy to get interrupt ID for those.

buha (approved):

PR #1252, opened 2022-02-11
AD413x driver

APorumb:
AD413X driver created based on AD4130-8 chip. PR still in progress and marked as draft. Basic functions tested on hardware.

### DIFF
--- a/drivers/afe/ad413x/ad413x.c
+++ b/drivers/afe/ad413x/ad413x.c
@@ -1,0 +1,816 @@
+/***************************************************************************//**
+ *   @file   ad413x.c
+ *   @brief  Implementation of AD413X Driver.
+ *   @author Andrei Porumb (andrei.porumb@analog.com)
+********************************************************************************
+ * Copyright 2022(c) Analog Devices, Inc.
+ *
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *  - Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ *  - Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in
+ *    the documentation and/or other materials provided with the
+ *    distribution.
+ *  - Neither the name of Analog Devices, Inc. nor the names of its
+ *    contributors may be used to endorse or promote products derived
+ *    from this software without specific prior written permission.
+ *  - The use of this software may or may not infringe the patent rights
+ *    of one or more patent holders.  This license does not release you
+ *    from the requirement that you obtain separate licenses from these
+ *    patent holders to use this software.
+ *  - Use of the software either in source or binary form, must be run
+ *    on or directly connected to an Analog Devices Inc. component.
+ *
+ * THIS SOFTWARE IS PROVIDED BY ANALOG DEVICES "AS IS" AND ANY EXPRESS OR
+ * IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, NON-INFRINGEMENT,
+ * MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED.
+ * IN NO EVENT SHALL ANALOG DEVICES BE LIABLE FOR ANY DIRECT, INDIRECT,
+ * INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, INTELLECTUAL PROPERTY RIGHTS, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+*******************************************************************************/
+
+/******************************************************************************/
+/***************************** Include Files **********************************/
+/******************************************************************************/
+#include <stdio.h>
+#include <stdlib.h>
+#include "ad413x.h"
+#include "no-os/error.h"
+#include "no-os/irq.h"
+#include "no-os/print_log.h"
+#include "no-os/delay.h"
+
+/******************************************************************************/
+/************************** Functions Implementation **************************/
+/******************************************************************************/
+/***************************************************************************//**
+ * Compute CRC8 checksum.
+ *
+ * @param data      - The data buffer.
+ * @param data_size - The size of the data buffer.
+ *
+ * @return CRC8 checksum.
+*******************************************************************************/
+static uint8_t ad413x_compute_crc8(uint8_t *data, uint8_t data_size)
+{
+	uint8_t i;
+	uint8_t crc = 0;
+
+	while (data_size) {
+		for (i = 0x80; i != 0; i >>= 1) {
+			if (((crc & 0x80) != 0) != ((*data & i) != 0)) {
+				crc <<= 1;
+				crc ^= AD413X_CRC8_POLY;
+			} else
+				crc <<= 1;
+		}
+		data++;
+		data_size--;
+	}
+	return crc;
+}
+
+/***************************************************************************//**
+ * SPI DATA register continuous read from device.
+ *
+ * @param dev      - The device structure.
+ * @param reg_data - The register data.
+ *
+ * @return SUCCESS in case of success, negative error code otherwise.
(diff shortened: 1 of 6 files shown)


drivers/afe/ad413x/ad413x.c
```diff
@@ -0,0 +1,816 @@
...
+/******************************************************************************/
+/***************************************************************************//**
+ * Compute CRC8 checksum.
+ *
+ * @param data      - The data buffer.
+ * @param data_size - The size of the data buffer.
+ *
+ * @return CRC8 checksum.
+*******************************************************************************/
+static uint8_t ad413x_compute_crc8(uint8_t *data, uint8_t data_size)
```
amiclaus, line 62: maybe we can use https://github.com/analogdevicesinc/no-OS/blob/master/util/crc8.c ?

drivers/afe/ad413x/ad413x.h
```diff
@@ -0,0 +1,417 @@
...
+
+/*************************** ADC Register Lengths *****************************/
+#define AD413X_R1B					(1 << 8)
+#define AD413X_R2B					(2 << 8)
+#define AD413X_R3B					(3 << 8)
+#define AD413X_TRANSF_LEN(x)		(((x) >> 8) & 0xFF)
+
+/***************************** ADC Register Map *******************************/
+#define AD413X_REG_STATUS 			(AD413X_R1B | 0x0)
+#define AD413X_REG_ADC_CTRL			(AD413X_R2B | 0x1)
```
amiclaus, line 66: odd indentations throughout the defines. see raw file

projects/ad413x/src/app/main.c
```diff
@@ -0,0 +1,141 @@
...
+		.ch[6] = { 0, 1, 3, 0x13}, //ch 6, preset 0, enable, AIN3, DGND
+		.ch[12] = { 2, 0, 4, 0x13}, //ch 12, preset 2, disabled, AIN4, DGND
+		.bipolar = 0,
+		.int_ref = 1
+	};
+
+	ret = ad413x_init(&ad413x_dev, ad413x_dev_ip);
+	if(ret)
+		return FAILURE;
+
```
amiclaus, line 139: I'd expand this minimal application and read some data (using `ad413x_spi_data_cont_read`) and print it.

drivers/afe/ad413x/ad413x.c
```diff
@@ -0,0 +1,816 @@
...
+
+	ret = spi_write_and_read(dev->spi_dev, buf, 3);
+	if (ret)
+		return ret;
+
+	data = (buf[0] << 16) | (buf[1] << 8) | buf[2];
+
+	*reg_data = data;
+
+	return SUCCESS;
```
amiclaus, line 106: not sure if it is the case, but I think we want to transit slowly to `return 0;` instead of `return SUCCESS;`

drivers/afe/ad413x/ad413x.c
```diff
@@ -0,0 +1,816 @@
...
+int32_t ad413x_ch_en(struct ad413x_dev *dev, uint8_t ch_nb, uint8_t enable)
+{
+	int32_t ret;
+
+	if (enable) {
+		ret = ad413x_spi_int_reg_write_msk(dev,
+						   AD413X_REG_CHN(ch_nb),
+						   AD413X_ENABLE_M,
+						   AD413X_ENABLE_M);
+		if (ret)
```
amiclaus, line 343: this can be condensed to:
```
if (enable)
      ret = ad413x_spi_int_reg_write_msk(dev,
				AD413X_REG_CHN(ch_nb),
				AD413X_ENABLE_M,
				AD413X_ENABLE_M);
else
       ret = ad413x_spi_int_reg_write_msk(dev,
				AD413X_REG_CHN(ch_nb),
				0x0U,
				AD413X_ENABLE_M);
if (ret)
			return ret;

```

drivers/afe/ad413x/ad413x.c
```diff
@@ -0,0 +1,816 @@
...
+int32_t ad413x_adc_bipolar(struct ad413x_dev *dev, uint8_t enable)
+{
+	int32_t ret;
+
+	if (enable) {
+		ret = ad413x_spi_int_reg_write_msk(dev,
+						   AD413X_REG_ADC_CTRL,
+						   AD413X_ADC_BIPOLAR,
+						   AD413X_ADC_BIPOLAR);
+		if (ret)
```
amiclaus, line 377: ditto: `ret` check only after the `if/else` statement is enough

drivers/afe/ad413x/ad413x.c
```diff
@@ -0,0 +1,816 @@
...
+	if (ret)
+		goto err_dev;
+
+	/* Device Settings */
+	ret = ad413x_spi_do_soft_reset(dev);
+	if (ret)
+		goto err_spi;
+
+	/* Change SPI to 4 wire*/
+	ret = ad413x_spi_int_reg_write_msk(dev,
```
amiclaus, line 701: it think we should expose a device parameter `spi3wire` as boolean, and give the user the option to selecte between 3-wire configuration and 4-wire configuration

amiclaus (commented):
some more remarks that might be considered for the latest version of the driver/project.

drivers/afe/ad413x/ad413x.c
```diff
@@ -0,0 +1,816 @@
...
+
+/***************************************************************************//**
+ * SPI DATA register continuous read from device.
+ *
+ * @param dev      - The device structure.
+ * @param reg_data - The register data.
+ *
+ * @return SUCCESS in case of success, negative error code otherwise.
+*******************************************************************************/
+static int32_t ad413x_spi_data_cont_read(struct ad413x_dev *dev,
```
adrimbarean, line 89: nitpick: we could lose the SPI denominator here since the chip doesn't support other digital interfaces. Shortens the name.

drivers/afe/ad413x/ad413x.c
```diff
@@ -0,0 +1,816 @@
...
+}
+
+/***************************************************************************//**
+ * IRQ handler for continuously ADC read.
+*******************************************************************************/
+static void irq_adc_cont_read(struct ad413x_callback_ctx *ctx)
+{
+	struct ad413x_dev *dev = ctx->dev;
+	if(ctx->buffer_size > 0) {
+		if(ad413x_spi_data_cont_read(ctx->dev, ctx->buffer))
```
adrimbarean, line 134: Since the code here is almost identical with `irq_adc_single_read` maybe we could include a flag in the ctx handler and call different functions here instead of almost duplicating the code. Do you think this makes sense or would functions would be generally different usually?

drivers/afe/ad413x/ad413x.c
adrimbarean: Even more condensation:
`ret = ad413x_spi_int_reg_write_msk(dev, AD413X_REG_CHN(ch_nb), enable ? AD413X_ENABLE_M : 0x0U, AD413X_ENABLE_M);`

drivers/afe/ad413x/ad413x.c
```diff
@@ -0,0 +1,816 @@
...
+	for (i = 0; i < 8; i++) {
+		ret = ad413x_set_gain(dev, init_param.preset[i].gain, i);
+		if (ret)
+			goto err_spi;
+		ret = ad413x_set_ref(dev, init_param.preset[i].ref_sel, i);
+		if (ret)
+			goto err_spi;
+		ret = ad413x_set_filter(dev, init_param.preset[i].filter, i);
+		if (ret)
+			goto err_spi;
```
adrimbarean, line 740: Should we use `ad413x_preset_store` here?

drivers/afe/ad413x/ad413x.h
```diff
@@ -0,0 +1,417 @@
...
+#include "no-os/irq.h"
+
+/******************************************************************************/
+/********************** Macros and Constants Definitions **********************/
+/******************************************************************************/
+#define AD413X_CMD_WR_COM_REG(x)	(0x00 | ((x) & 0x3F)) // Write to Register x
+#define AD413X_CMD_RD_COM_REG(x) 	(0x40 | ((x) & 0x3F)) // Read from Register x
+
+/*************************** ADC Register Lengths *****************************/
+#define AD413X_R1B					(1 << 8)
```
adrimbarean, line 59: Can use [BIT](https://github.com/analogdevicesinc/no-OS/blob/master/include/no-os/util.h#L51) define.

drivers/afe/ad413x/ad413x.h
```diff
@@ -0,0 +1,417 @@
...
+#define AD413X_ADC_DATA_STATUS			(1 << 10)
+#define AD413X_ADC_CSB_EN				(1 <<  9)
+#define AD413X_ADC_REF_EN				(1 <<  8)
+#define AD413X_ADC_DUTY_CYC_RATIO		(1 <<  6)
+#define AD413X_ADC_MODE(x)				(((x) & 0xF) << 2)
+#define AD413X_ADC_CNTRL_MCLK(x)		((x) & 0x3)
+
+/* IO_CONTROL Register */
+#define AD413X_SYNCB_CLEAR 				(1 << 10)
+#define AD413X_INT_PIN_SEL(x) 			(((x) & 0x3) << 8)
```
adrimbarean, line 100: Can use [GENMASK](https://github.com/analogdevicesinc/no-OS/blob/master/include/no-os/util.h#L86).

PR #1258, opened 2022-02-17
ADXL355 no-os driver exemplified with eval-adxl355-pmdz on SDP-K1 eval board

rbolboac:
- No-os driver for ADXL355 device tested with eval-adxl355-pmdz.
- Contains small example for SDP-K1 evaluation board.


projects/eval-adxl355-pmdz/sdp-ck1z.ioc
```diff
@@ -0,0 +1,147 @@
+#MicroXplorer Configuration settings - do not modify
```
amiclaus, line 1: do we need this file added?

drivers/accel/adxl355/adxl355.h
```diff
@@ -0,0 +1,313 @@
...
+
+/*! Reads fifo data and returns the values converted in g. */
+int32_t adxl355_get_available_fifo_data(struct adxl355_dev *dev,
+					uint8_t *fifo_entries,
+					uint8_t *data_buffer,
+					float *x,
+					float *y,
+					float *z);
+
+#endif /* __ADXL355_H__ */
```
amiclaus, line 313: nit: new line at the end of the file

projects/eval-adxl355-pmdz/src/main.c
```diff
@@ -0,0 +1,115 @@
...
+		printf("Temp = %f \n", temp);
+		HAL_Delay(1000);
+	}
+	return 0;
+error:
+	printf("Error!\n");
+}
+
+
+
```
amiclaus, line 114: remove extra lines.

drivers/accel/adxl355/adxl355.h
```diff
@@ -0,0 +1,313 @@
...
+/******************************************************************************/
+/***************************** Include Files **********************************/
+/******************************************************************************/
+#include <stdint.h>
+#include <string.h>
+#include "no-os/i2c.h"
+#include "no-os/spi.h"
+
+/******************************************************************************/
+/******************************** ADXL355 *************************************/
```
amiclaus, line 51: `ADXL355` -> `Macro Definitions`

drivers/accel/adxl355/adxl355.h
```diff
@@ -0,0 +1,313 @@
+/***************************************************************************//**
+ *   @file   adxl355.h
+ *   @brief  Header file of ADXL355 Driver.
+ *   @author RBolboac (ramona.bolboaca@analog.com)
+********************************************************************************
+ * Copyright 2012(c) Analog Devices, Inc.
```
amiclaus, line 6: 2022

drivers/accel/adxl355/adxl355.c
```diff
@@ -0,0 +1,801 @@
...
+
+/***************************************************************************//**
+ * @brief Reads the number of FIFO entries register value.
+ *
+ * @param dev       - The device structure.
+ * @param reg_value - Register value.
+ *
+ * @return status   - Result of the reading procedure.
+*******************************************************************************/
+int32_t adxl355_get_nb_of_fifo_entries(struct adxl355_dev *dev,
```
amiclaus, line 415: is this function needed? `adxl355_get_register_value(dev, ADXL355_FIFO_ENTRIES, reg_value);` can be used inline where needed.

projects/eval-adxl355-pmdz/src/main.c
```diff
@@ -0,0 +1,115 @@
...
+	while(1) {
+
+		adxl355_get_available_fifo_data(adxl355,
+						&fifo_entries,
+						data_buffer,
+						&x[0],
+						&y[0],
+						&z[0]);
+		printf("Number of read entries from the FIFO %d \n", fifo_entries);
```
amiclaus, line 88: `print_log.h` can be used in the main function.
https://github.com/analogdevicesinc/no-OS/blob/master/include/no-os/print_log.h

amiclaus (commented):
looks good overall, some comments on my side.

amiclaus (commented):
Also, please consider squashing the commits: one for the driver and one for the application.

drivers/accel/adxl355/adxl355.h
```diff
@@ -0,0 +1,254 @@
...
+#include "no-os/spi.h"
+
+/******************************************************************************/
+/******************************** ADXL355 *************************************/
+/******************************************************************************/
+
+/* SPI commands */
+#define ADXL355_SPI_READ         0x01
+#define ADXL355_SPI_WRITE        0x00
+
```
buha: try to remove these double/triple empty lines, there are several

drivers/accel/adxl355/adxl355.h
```diff
@@ -0,0 +1,254 @@
...
+#define ADXL355_XDATA_BYTE_SIZE    0x03
+#define ADXL355_YDATA_BYTE_SIZE    0x03
+#define ADXL355_ZDATA_BYTE_SIZE    0x03
+#define ADXL355_TEMPDATA_BYTE_SIZE 0x02
+
+#define ADXL355_SELF_TEST_TRIGGER_VAL 0x03
+#define ADXL355_ACCEL_DATA_SIGN_MASK 0x80000
+#define ADXL355_ACCEL_DATA_NEG_VAL_MASK 0xFFF00000
+
+#define ADXL355_TEMP_DATA_MASK 0x0F
```
buha, line 114: for all the _MASK fields, try to use the GENMASK() macro in util.h, it's easier to read the code like that and draw correspondence with the datasheet.

Here, it'd be GENMASK(3, 0)

drivers/accel/adxl355/adxl355.h
```diff
@@ -0,0 +1,254 @@
...
+#define ADXL355_TEMP_SCALE_FACTOR   -2111.25
+#define ADXL355_TEMP_OFFSET         -0.11049723
+
+
+
+/******************************************************************************/
+/*************************** Types Declarations *******************************/
+/******************************************************************************/
+
+typedef enum
```
buha: let's not use typedefs, they can be confusing, linux coding rules recommend against them too
just leave it as `enum adxl355_op_mode {};` and  use it in code using `enum adxl355_op_mode mode`

projects/eval-adxl355-pmdz/builds.json
```diff
@@ -0,0 +1,10 @@
+{
```
buha, line 1: make sure to edit this file in the final version to match your project

projects/eval-adxl355-pmdz/sdp-ck1z.ioc
buha: yes i think we need it for stm32 builds
if we had a repository with such files hosted somewhere, we wouldn't need it, but since we don't...

drivers/accel/adxl355/adxl355.h
```diff
@@ -0,0 +1,254 @@
...
+#define ADXL355_SHADOW_REGISTER_BASE_ADDR  0x50
+#define ADXL355_NB_OF_SHADOW_REGISTERS     0x05
+
+#define ADXL355_DEVID_AD_VAL	0xAD
+#define ADXL355_DEVID_MST_VAL	0x1D
+#define ADXL355_PARTID_VAL	    0xED
+#define ADXL355_REVID_VAL	    0x01
+#define ADXL355_RESET_CODE      0x52
+
+#define ADXL355_XDATA_BYTE_SIZE    0x03
```
buha: there is an easier way to do this. Please have a look at the code here:
https://github.com/analogdevicesinc/no-OS/blob/119ba490c1d5762e7092332c3a8abe24e671a922/drivers/adc/ad9250/ad9250.h#L59-L62

Basically, your chip has a fixed number of registers that you know. This number can be coded to fit on 8 bits, or 16 or whatever. In any circumstance, it's generally not bigger than 2^16, that would be a huge number of registers. Since we always use uint32_t for the register address, the top 8 or 16-bits lay unused. So we code some info in the topmost bits regarding the register size in bytes when we define the register address.

This you can use in your adxl355_get_register_value() and adxl355_set_register_value() functions and drop the  adxl355_get_n_consec_reg_val() function.

We have many such examples on no-OS/linux, just search for "_R2B" or "_2B", you'll find them.

drivers/accel/adxl355/adxl355.c
```diff
@@ -0,0 +1,652 @@
...
+	// Standby.
+
+	current_op_mode = dev->op_mode;
+
+	if (current_op_mode != ADXL355_STANDBY)
+	{
+		status = adxl355_set_op_mode(dev, ADXL355_STANDBY);
+	}
+
+	status |= adxl355_get_register_value(dev, ADXL355_FILTER, &reg_value);
```
buha, line 497: You should not `status |= function()` but simply `status = function()` and return immediately on error.

drivers/accel/adxl355/adxl355.h
```diff
@@ -0,0 +1,303 @@
...
+	ADXL355_ODR_125HZ,
+	ADXL355_ODR_62_5HZ,
+	ADXL355_ODR_31_25HZ,
+	ADXL355_ODR_15_625HZ,
+	ADXL355_ODR_7_813HZ,
+	ADXL355_ODR_3_906HZ
+};
+
+enum adxl355_range {
+	ADXL355_RANGE_INVALID,
```
buha, line 148: This is a good habit, having invalid values in enums ;)
But we tend to prefer the other method where we match these enums to register field values.
So if the device register where range is set defaults to value zero (which means range 2G), then we simply start the enum at ADXL355_RANGE_255 which takes this default  zero value.

The advantage is simpler code, with some implicit defaults, therefore, you can also remove:
```

	// Default range is set
	dev->range = ADXL355_RANGE_2G;

	// Default value for FIFO SAMPLES
	dev->fifo_samples =  ADXL355_MAX_FIFO_SAMPLES_VAL;

	// Default value for POWER_CTL
	dev->op_mode = ADXL355_STANDBY_TEMP_ON_DRDY_ON;
```

drivers/accel/adxl355/adxl355.c
```diff
@@ -0,0 +1,804 @@
...
+
+/***************************************************************************//**
+ * @brief Gets the current operation mode of the device
+ *
+ * @param dev     - The device structure.
+ * @param op_mode - Read operation mode.
+ *
+ * @return status - Result of the reading operation procedure.
+*******************************************************************************/
+int32_t adxl355_get_op_mode(struct adxl355_dev *dev,
```
buha, line 257: Change all `status` and al functions that return a return code from `int32_t` to `int`. You were likely not aware of this, we decided at some point to just use `int` due to it being of native size for the platform (compiler takes care of that).

drivers/accel/adxl355/adxl355.h
rbolboac: I still have to initialize the driver with the value of range, just in case the user does not set the range manually, because dev->range is actually being used inside adxl355_accel_conversion function. 
The same goes for dev->op_mode which is used in adxl355_set_odr_lpf.
I would also like to keep the initialization of dev->fifo_samples (which is different from 0) for debugging purposes, such that the user can see the actual configuration of the device.  

However, some improvement could be made and this is how I modified the code section you mentioned:

```
        // Default range is set
	dev->range = (enum adxl355_range)(GET_ADXL355_RESET_VAL(
			ADXL355_RANGE) & GENMASK(1,0));
	// Default value for FIFO SAMPLES
	dev->fifo_samples = GET_ADXL355_RESET_VAL(ADXL355_FIFO_SAMPLES);

	// Default value for POWER_CTL
	dev->op_mode = (enum adxl355_op_mode)(GET_ADXL355_RESET_VAL(ADXL355_POWER_CTL));
```

Thank you for your feedback!

drivers/accel/adxl355/adxl355.c
```diff
@@ -0,0 +1,814 @@
...
+				struct adxl355_status_register_flags *status_flags)
+{
+	int status = 0;
+	uint8_t reg_value = 0;
+
+	status = adxl355_read_device_data(dev, ADXL355_ADDR(ADXL355_STATUS),
+					  GET_ADXL355_TRANSF_LEN(ADXL355_STATUS), &reg_value);
+
+	if (status == 0) {
+		status_flags->NVM_BUSY  = (reg_value >> 4) & 1;
```
scuciurean, line 394: If you're using bitfields, why not pack the flag structure and an uint8_t in a union and assign the reg_value directly instead of assigning each member sequentially

drivers/accel/adxl355/adxl355.c
```diff
@@ -0,0 +1,814 @@
...
+*******************************************************************************/
+int adxl355_init(struct adxl355_dev **device,
+		 struct adxl355_init_param init_param)
+{
+	struct adxl355_dev *dev;
+	int status = 0;
+	uint8_t reg_value = 0;
+
+	dev = (struct adxl355_dev *)calloc(1, sizeof(*dev));
+	if (!dev)
```
CiprianRegus, line 153: We're either using 0 or errno(3) error codes as function return values (this way the user can get more info about what went wrong). In this case you could return -ENOMEM (notice the -).

drivers/accel/adxl355/adxl355.c
```diff
@@ -0,0 +1,814 @@
...
+ * @param dev     - The device structure.
+ *
+ * @return status - Result of the writing procedure.
+*******************************************************************************/
+int adxl355_set_self_test(struct adxl355_dev *dev)
+{
+	int status = 0;
+	uint8_t data = ADXL355_SELF_TEST_TRIGGER_VAL;
+
+	status = adxl355_write_device_data(dev, ADXL355_ADDR(ADXL355_SELF_TEST),
```
CiprianRegus, line 565: You could return directly.

amiclaus (commented):
> Also, please consider squashing the commits: one for the driver and one for the application.

As mentioned before, I'd suggest that the first commit should be for the driver and the second one for the application.

drivers/accel/adxl355/adxl355.c
rbolboac: I modified the code implementation such that it uses bit fields now.
I made these changes on the assumption that the compiler will define bitfields from the least significant bit to the most significant bit.
@buha , could you offer your input here?

drivers/accel/adxl355/adxl355.c
```diff
@@ -0,0 +1,824 @@
...
+	} else {
+		status = i2c_init(&dev->i2c_desc, &init_param.i2c_init);
+		if (status != 0)
+			goto error;
+	}
+
+	status = adxl355_read_device_data(dev, ADXL355_ADDR(ADXL355_DEVID_AD),
+					  GET_ADXL355_TRANSF_LEN(ADXL355_DEVID_AD), &reg_value);
+	if ((status != 0) || (reg_value != GET_ADXL355_RESET_VAL(ADXL355_DEVID_AD)))
+		goto error;
```
amiclaus, line 168: memory leak: `i2c_desc`/`spi_desc`

drivers/accel/adxl355/adxl355.c
```diff
@@ -0,0 +1,824 @@
...
+ * @param reg_value - Register value.
+ *
+ * @return status   - Result of the reading procedure.
+*******************************************************************************/
+int adxl355_get_nb_of_fifo_entries(struct adxl355_dev *dev,
+				   uint8_t *reg_value)
+{
+	int status = 0;
+
+	status = adxl355_read_device_data(dev, ADXL355_ADDR(ADXL355_FIFO_ENTRIES),
```
amiclaus, line 410: there's no point here to add another local variable, return directly. Or even remove this function and use `adxl355_read_device_data` inline.

drivers/accel/adxl355/adxl355.c
```diff
@@ -0,0 +1,824 @@
...
+		return status;
+
+	reg_value = (reg_value & 0x30) | odr_lpf_val;
+	status = adxl355_write_device_data(dev, ADXL355_ADDR(ADXL355_FILTER),
+					   GET_ADXL355_TRANSF_LEN(ADXL355_FILTER), &reg_value);
+	if (status != 0)
+		return status;
+	dev->odr_lpf = odr_lpf_val;
+
+	status = adxl355_set_op_mode(dev, current_op_mode);
```
amiclaus, line 510: return directly

drivers/accel/adxl355/adxl355.c
```diff
@@ -0,0 +1,824 @@
...
+ * @param dev     - The device structure.
+ *
+ * @return status - Result of the writing procedure.
+*******************************************************************************/
+int adxl355_set_self_test(struct adxl355_dev *dev)
+{
+	int status = 0;
+	uint8_t data = ADXL355_SELF_TEST_TRIGGER_VAL;
+
+	status = adxl355_write_device_data(dev, ADXL355_ADDR(ADXL355_SELF_TEST),
```
amiclaus, line 558: ditto: `status` variable is redundant.

drivers/accel/adxl355/adxl355.c
```diff
@@ -0,0 +1,824 @@
...
+ *
+ * @return status  - Result of the configuration procedure.
+*******************************************************************************/
+int adxl355_configure_interrupt_pins (struct adxl355_dev *dev,
+				      union adxl355_interrupt_mask interrupt_config)
+{
+	int status = 0;
+	uint8_t reg_val = interrupt_config.value;
+
+	status = adxl355_write_device_data(dev, ADXL355_ADDR(ADXL355_INT_MAP),
```
amiclaus, line 621: ditto: `status` variable is redundant.

drivers/accel/adxl355/adxl355.c
```diff
@@ -0,0 +1,824 @@
...
+ * @param act_count - Register value.
+ *
+ * @return status   - Result of the reading procedure.
+*******************************************************************************/
+int adxl355_get_activity_count_register(struct adxl355_dev *dev,
+					uint8_t *act_count)
+{
+	int status = 0;
+
+	status = adxl355_read_device_data(dev, ADXL355_ADDR(ADXL355_ACT_COUNT),
```
amiclaus, line 754: ditto: remove `status` variable or use `adxl355_read_device_data` inline without wrapper.

drivers/accel/adxl355/adxl355.c
```diff
@@ -0,0 +1,824 @@
...
+	// raw_data[1] bits [7-0]: DATA bits [11: 4]
+	// raw_data[2] bits [7-4]: DATA bits [ 3: 0]
+	// raw_data[2] bits i[3-0]: reserved
+	raw_accel = (raw_data[0] << 16) | (raw_data[1] << 8) | raw_data[2];
+	raw_accel = raw_accel >> 4;
+
+	// Raw acceleration is in two's complement
+	// Convert from two's complement to int
+
+	if ((raw_accel & GENMASK(19,19)) == GENMASK(19,19))
```
amiclaus, line 785: `GENMAS(19, 19) = BIT(19)`
You can rework this `if (raw_accel & BIT(19))`

drivers/accel/adxl355/adxl355.c
```diff
@@ -0,0 +1,824 @@
...
+		accel_data = raw_accel | GENMASK(31,20);
```
amiclaus, line 786: what does `GENMASK(31,20)` mean? add a macro for it

drivers/accel/adxl355/adxl355.h
```diff
@@ -0,0 +1,337 @@
...
+
+#define ADXL355_RESET_CODE 0x52
+
+/*
+ * At +/- 2g with 20-bit resolution, scale is given in datasheet as
+ * 3.9ug/LSB = 0.0000039 * 9.80665 = 0.00003824593 m/s^2.
+ * For +/- 4g range a multiplier with value 2 is used.
+ * For +/-8g range, a multiplier with value 4 is used.
+ */
+#define ADXL355_ACC_SCALE_FACTOR 0.00003824593
```
amiclaus, line 103: avoid floating point operations

projects/eval-adxl355-pmdz/src/main.c
```diff
@@ -0,0 +1,113 @@
...
+#include "no-os/util.h"
+#include "no-os/print_log.h"
+#include "stm32_spi.h"
+#include "stm32_gpio.h"
+#include "stm32_uart.h"
+#include "stm32_uart_stdio.h"
+#include "adxl355.h"
+
+#define SPI_DEVICE_ID	1
+#define SPI_CS			15
```
amiclaus, line 17: indentation needs to be fixed.

drivers/accel/adxl355/adxl355.c
```diff
@@ -0,0 +1,818 @@
...
+ *
+ * @param dev      - The device structure.
+ * @param x_offset - X-axis's offset.
+ * @param y_offset - Y-axis's offset.
+ * @param z_offset - Z-axis's offset.
+ *
+ * @return ret     - Result of the writing procedure.
+*******************************************************************************/
+int adxl355_set_offset(struct adxl355_dev *dev,uint16_t x_offset,
+		       uint16_t y_offset,uint16_t z_offset)
```
amiclaus, line 450: nit: space after `uint16_t y_offset,`

drivers/accel/adxl355/adxl355.c
```diff
@@ -0,0 +1,925 @@
...
+static uint32_t adxl355_accel_array_conv(struct adxl355_dev *dev,
+		uint8_t *raw_array)
+{
+	uint32_t raw_accel;
+
+	// raw_array[0] bits [7-0]: DATA bits [19:12]
+	// raw_array[1] bits [7-0]: DATA bits [11: 4]
+	// raw_array[2] bits [7-4]: DATA bits [ 3: 0]
+	// raw_array[2] bits i[3-0]: reserved
+	raw_accel = (raw_array[0] << 16) | (raw_array[1] << 8) | raw_array[2];
```
amiclaus, line 874: After https://github.com/analogdevicesinc/no-OS/pull/1260 is merged into the master branch you can use here `get_unaligned_be24`

drivers/accel/adxl355/adxl355.c
rbolboac: Starting with C51 Version 5 and later the order of allocation of bit-fields within a unit is LSB to MSB.

drivers/accel/adxl355/adxl355.c
```diff
@@ -0,0 +1,934 @@
...
+ * @param write_data   - The data which is going to be written.
+ *
+ * @return ret         - Result of the writing procedure.
+*******************************************************************************/
+int adxl355_write_device_data(struct adxl355_dev *dev, uint8_t base_address,
+			      uint16_t size, uint8_t *write_data)
+{
+	int ret;
+	uint8_t *comm_buffer;
+	comm_buffer = (uint8_t *)calloc(1,1 + size);
```
buha, line 223: i think we should work directly on user buffer and not create the temporary buffer using malloc which can only cause troubles with a function like this one that may be called in critical code areas like intensive loops or interrupts

user is responsible for providing a big enough buffer

same comment for adxl355_read_device_data

drivers/accel/adxl355/adxl355.c
```diff
@@ -0,0 +1,934 @@
...
+	if (!comm_buffer)
+		return -ENOMEM;
+	for (uint16_t idx = 0; idx < size; idx++)
+		comm_buffer[1+idx] = write_data[idx];
+
+	if (dev->communication_type == ADXL355_SPI_COMM) {
+		comm_buffer[0] = ADXL355_SPI_WRITE | (base_address << 1);
+		ret = spi_write_and_read(dev->spi_desc, comm_buffer, size + 1);
```
buha, line 231: also pay attention to endianess, multi-byte reads/writes may be influenced by endianess match/mismatch between the adxl355 and the MCU

rbolboac: For multi-byte write, which is the case for the following APIs:
- adxl355_set_offset
- adxl355_configure_activity_threshold

The written data is split in bytes using shift operations so the functionality is not affected by endianess. 

For multi-byte read, which is the case for the following data:
- shadow registers value
- measured acceleration data
- temperature data
- fifo data

The read data is received in bytes, which are then assembled using shift operations, so the functionality is not affected by endianess.

buha (approved):

drivers/accel/adxl355/adxl355.h
```diff
@@ -0,0 +1,372 @@
...
+/*
+ * At +/- 2g with 20-bit resolution, scale is given in datasheet as
+ * 3.9ug/LSB = 0.0000039 * 9.80665 = 0.00003824593 m/s^2.
+ * For +/- 4g range a multiplier with value 2 is used.
+ * For +/-8g range, a multiplier with value 4 is used.
+ */
+#define ADXL355_ACC_SCALE_FACTOR_MUL  (int64_t)     38245
+#define ADXL355_ACC_SCALE_FACTOR_DIV  (int32_t)1000000000
+
+
```
amiclaus, line 104: nit: extra line

drivers/accel/adxl355/adxl355.h
```diff
@@ -0,0 +1,372 @@
...
+/* SCALE = ADXL355_TEMP_SCALE_FACTOR / ADXL355_TEMP_SCALE_FACTOR_DIV
+   We want to return the temperature in millidegrees Celsius so instead of
+   0.110497238, we are going to use 110.497238 value for scale
+ */
+#define ADXL355_TEMP_SCALE_FACTOR     (int64_t)-110497238
+#define ADXL355_TEMP_SCALE_FACTOR_DIV (int32_t)   1000000
+
+#define ADXL355_NEG_ACC_MSK                 GENMASK(31, 20)
+#define ADXL355_RANGE_FIELD_MSK    (uint8_t)GENMASK( 1,  0)
```
amiclaus, line 125: are these casts required?

projects/eval-adxl355-pmdz/src/main.c
```diff
@@ -0,0 +1,122 @@
...
+		pr_info("NVM_BUSY flag = %d \n", (uint8_t)(status_flags.fields.NVM_BUSY));
+		pr_info("===========================================================\n");
+
+		adxl355_get_temp(adxl355, &temp);
+		pr_info(" Temp =%d"".%09u millidegress Celsius \n", (int)temp.integer,
+			(abs)(temp.fractional));
+
+		HAL_Delay(1000);
+	}
+	return 0;
```
amiclaus, line 119: This line won't be reached.

drivers/accel/adxl355/adxl355.c
```diff
@@ -0,0 +1,913 @@
...
+
+	// The hpf settings can be changed when the device is in measurement mode.
+	// There is no need to modify the device's operation mode to Standby.
+
+	ret = adxl355_read_device_data(dev, ADXL355_ADDR(ADXL355_FILTER),
+				       GET_ADXL355_TRANSF_LEN(ADXL355_FILTER), &reg_value);
+	if (ret)
+		return ret;
+	reg_value &= ~(ADXL355_HPF_FIELD_MSK);
+	reg_value |= ((uint8_t)hpf_corner_val) & ADXL355_HPF_FIELD_MSK;
```
amiclaus, line 431: is the uint8_t cast required?

drivers/accel/adxl355/adxl355.c
```diff
@@ -0,0 +1,913 @@
...
+	// Get shadow register values
+	ret = adxl355_read_device_data(dev,
+				       ADXL355_ADDR(ADXL355_SHADOW_REGISTER_BASE_ADDR),
+				       GET_ADXL355_TRANSF_LEN(ADXL355_SHADOW_REGISTER_BASE_ADDR),
+				       &shadow_register_values[0]);
+	if (ret)
+		goto error_com;
+
+	// Default range is set
+	dev->range = (enum adxl355_range)(GET_ADXL355_RESET_VAL(ADXL355_RANGE)
```
amiclaus, line 183: is this cast required?

amiclaus (commented):
some minor comments regarding the latest version.

scuciurean (approved):

amiclaus (approved):

CiprianRegus (approved):

PR #1260, opened 2022-02-23
util: extend unordered bytes functions

amiclaus:
Add support for 24/32-bit put/get unordered bytes, similar with what we
have on linux.


amiclaus (commented):
> Looks good to me. And I see implementation is different from Linux kernel. Still, do we have to make extra sure that we can use this with ADI license?

Since we already have the functions implemented for 16-bit, I think we are fine.

PR #1261, opened 2022-02-23
projects:ad3552r_fmcz: add IIO driver conditionally

adrimbarean:
If IIO example is not desired the IIO library is not used, but the AD3552R
IIO driver is linked and built. Since the IIO driver for AD3552R depends on
the IIO library, the build throws an error.


amiclaus (approved):
lgtm

PR #1270, opened 2022-03-04
Add IIO support for ADXL355

rbolboac:


drivers/accel/adxl355/iio_adxl355.c
```diff
@@ -0,0 +1,1117 @@
...
+		return -EINVAL;
+
+	iio_adxl355 = (struct adxl355_iio_dev *)dev;
+
+	switch (channel->type) {
+	case IIO_ACCEL:
+		// Remove data existing in the buffer
+		strcpy(buf, "");
+
+		// Go through the values in HPF table for te current ODR and add them to the buffer
```
adrimbarean, line 902: nitpick: typo `te current ODR` -> `the current ODR`

projects/eval-adxl355-pmdz/src/app_config.h
```diff
@@ -0,0 +1,52 @@
+/***************************************************************************//**
+ *   @file   app_config.h
+ *   @brief  Config file of IIO Demo project.
+ *   @author DBogdan (dragos.bogdan@analog.com)
+********************************************************************************
+ * Copyright 2015(c) Analog Devices, Inc.
```
adrimbarean, line 6: Probably need to update those.

projects/eval-adxl355-pmdz/src/parameters.h
```diff
@@ -0,0 +1,121 @@
+/***************************************************************************//**
+ *   @file   iio_demo/src/app/parameters.h
+ *   @brief  Parameters Definitions.
+ *   @author DBogdan (dragos.bogdan@analog.com)
+********************************************************************************
+ * Copyright 2013(c) Analog Devices, Inc.
```
adrimbarean, line 6: Update here as well.

drivers/accel/adxl355/iio_adxl355.c
```diff
@@ -0,0 +1,1117 @@
...
+static struct iio_channel adxl355_channels[] = {
+	ADXL355_ACCEL_CHANNEL(0, chan_x, X),
+	ADXL355_ACCEL_CHANNEL(1, chan_y, Y),
+	ADXL355_ACCEL_CHANNEL(2, chan_z, Z),
+	{
+		.ch_type = IIO_TEMP,
+		.scan_type = &adxl355_iio_temp_scan_type,
+		.scan_index = 3,
+		.attributes = adxl355_iio_temp_attrs,
+		.ch_out = false,
```
adrimbarean, line 241: Is it worthwhile to initialize the `.channel` member here as well? By default it will be zero, coinciding with the X axis channel. Does it have to be the same or different?

drivers/accel/adxl355/iio_adxl355.c
```diff
@@ -0,0 +1,1117 @@
...
+ * @param dev     - The iio device structure.
+ * @param buf	  - Command buffer to be filled with requested data.
+ * @param len     - Length of the received command buffer in bytes.
+ * @param channel - Command channel info.
+ * @param priv    - Command attribute id.
+ *
+ * @return ret    - Result of the reading procedure.
+ * 					In case of success, the size of the read data is returned.
+*******************************************************************************/
+static int adxl355_iio_read_converted(void *dev, char *buf, uint32_t len,
```
adrimbarean, line 440: I'm not sure here, but is this redundant? Do IIO clients calculate this themselves using the raw, offset and scale values?

adrimbarean (approved):
Looks good to me. A few questions and comments, but I think it's good to go.

rbolboac:
- Implemented review findings

rbolboac:
- Fixed astyle

rbolboac:
- Rebase master

drivers/accel/adxl355/iio_adxl355.c
```diff
@@ -0,0 +1,1032 @@
...
+static int adxl355_iio_update_channels(void* dev, uint32_t mask);
+
+/******************************************************************************/
+/************************ Variable Declarations ******************************/
+/******************************************************************************/
+static struct iio_attribute adxl355_iio_temp_attrs[] = {
+	{
+		.name = "offset",
+		.show = adxl355_iio_read_offset,
+		.store = NULL,
```
buha, line 138: I know there are a lot of examples in no-OS where unused pointers are initialized like this, but let's just not initialize them from now on, they get initialized to NULL by the compiler

drivers/accel/adxl355/iio_adxl355.c
```diff
@@ -0,0 +1,1032 @@
...
+	case IIO_ACCEL:
+		// Remove data existing in the buffer
+		strcpy(buf, "");
+
+		// Go through the values in ODR table and add them to the buffer
+		for (uint8_t i = 0; i < NO_OS_ARRAY_SIZE(adxl355_iio_odr_table); i++) {
+			vals[0] = adxl355_iio_odr_table[i][0];
+			vals[1] = adxl355_iio_odr_table[i][1];
+			iio_format_value(buffer, sizeof(buffer), IIO_VAL_INT_PLUS_MICRO, 2, vals);
+			strcat(buf, buffer);
```
buha, line 860: why do we need to add this whitespace after iio_format_value ? should we add it inside iio_format_value perhaps ?

rbolboac: The whitespace is added because, in this case,  there is more than one element element returned in the buffer. 
The buffer will contain the list of available sampling frequencies, which are added one by one in the returned buffer and are delimitated by the white space. I don't think it is needed to be added to iio_format_value just for this list usage scenario.

rbolboac:
- Removed NULL pointer initialization for unused pointers.

drivers/accel/adxl355/iio_adxl355.c
buha: ok, makes sense

buha (approved):

PR #1271, opened 2022-03-04
Staging/maxim bug fixes

CiprianRegus:
Some GPIO bug fixes and stdio rework.

Signed-off-by: Ciprian Regus <ciprian.regus@analog.com>

### DIFF
--- a/drivers/platform/maxim/gpio_extra.h
+++ b/drivers/platform/maxim/gpio_extra.h
@@ -57,25 +57,20 @@ extern const struct gpio_platform_ops max_gpio_ops;
  */
 extern const struct irq_platform_ops max_gpio_irq_ops;
 
-/**
- * @struct gpio_irq_config
- * @brief maxim configuration descriptor for irq operations
- */
-struct gpio_irq_config {
-	enum irq_trig_level mode;
+enum max_pad_config {
+	NO_OS_PAD_NONE,
+	/** Strong pull up */
+	NO_OS_PAD_PULL_UP,
+	/** Strong pull down */
+	NO_OS_PAD_PULL_DOWN,
 };
 
 struct max_gpio_init_param {
 	/** Input/Output */
 	uint32_t mode;
 	/** Pull_up/Pull_down  */
-	uint32_t pull;
+	enum max_pad_config pull;
 	uint32_t port;
-};
-
-struct max_gpio_irq_param {
-	struct irq_ctrl_desc *parent;
-	struct gpio_desc *desc;
 };
 
 #endif
--- a/drivers/platform/maxim/maxim_gpio.c
+++ b/drivers/platform/maxim/maxim_gpio.c
@@ -59,13 +59,16 @@ static struct callback_desc *gpio_callback[N_PORTS][N_PINS];
 static void _gpio_irq(uint8_t port)
 {
 	uint32_t pin;
+	uint32_t shifted = 0;
 	mxc_gpio_regs_t *gpio_regs = MXC_GPIO_GET_GPIO(port);
 	uint32_t stat_reg = gpio_regs->int_stat;
 	/** Clear interrupt flags for the current port*/
 	gpio_regs->int_clr = stat_reg;
 	while(stat_reg) {
 		pin = find_first_set_bit(stat_reg);
+		pin += shifted;
 		if (!gpio_callback[port][pin]) {
+			shifted += pin + 1;
 			stat_reg >>= pin + 1;
 			continue;
 		}
@@ -73,6 +76,7 @@ static void _gpio_irq(uint8_t port)
 		gpio_callback[port][pin]->callback(ctx, pin, NULL);
 
 		stat_reg >>= pin + 1;
+		shifted += pin + 1;
 	}
 }
 
@@ -80,6 +84,24 @@ void GPIO0_IRQHandler()
 {
 	_gpio_irq(0);
 }
+#ifdef MXC_GPIO1
+void GPIO1_IRQHandler()
+{
+	_gpio_irq(1);
+}
+#endif
+#ifdef MXC_GPIO2
+void GPIO2_IRQHandler()
+{
+	_gpio_irq(2);
+}
+#endif
+#ifdef MXC_GPIO3
+void GPIO3_IRQHandler()
+{
+	_gpio_irq(3);
+}
+#endif
 
 /**
  * @brief Obtain the GPIO decriptor.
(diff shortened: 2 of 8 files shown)


CiprianRegus:
V2:
- fixed stdio read length.

drivers/platform/maxim/gpio_extra.h
```diff
@@ -57,25 +57,20 @@ extern const struct gpio_platform_ops max_gpio_ops;
...
 extern const struct irq_platform_ops max_gpio_irq_ops;
 
-/**
- * @struct gpio_irq_config
- * @brief maxim configuration descriptor for irq operations
- */
-struct gpio_irq_config {
-	enum irq_trig_level mode;
+enum max_pad_config {
+	NO_OS_PAD_NONE,
```
buha, line 61: I think the naming is a bit weird.
NO_OS_* would be the prefix for something in no-os/gpio.h but here I guess the prefix should be MAX_*

buha: On the other hand, I agree we should have a NO_OS_GPIO_PULLUP / PULLDOWN etc. enum somewhere in gpio.h.

CiprianRegus:
V3:
- Moved pad_config enum in gpio.h
- Astyled

CiprianRegus:
V4:
- Replaced some register use with API functions. 
- Renamed pad_config to gpio_pull_up.

include/no-os/gpio.h
```diff
@@ -103,6 +103,20 @@ enum gpio_values {
 	GPIO_HIGH_Z
 };
 
+/**
+ * @enum gpio_pull_up
+ * @brief Enum that holds the possible pull up/ pull down resistor configuration.
+ */
+enum gpio_pull_up {
+	NO_OS_NONE,
```
buha, line 111: To avoid confusions, I'd go with:

NO_OS_PULL_NONE,
NO_OS_PULL_DOWN,
NO_OS_PULL_UP,
NO_OS_PULL_DOWN_WEAK,
NO_OS_PULL_UP_WEAK,

CiprianRegus:
V5:
- Renamed pull up fields as @buha suggested.

drivers/platform/maxim/gpio_extra.h
```diff
@@ -57,25 +57,12 @@ extern const struct gpio_platform_ops max_gpio_ops;
...
-struct gpio_irq_config {
-	enum irq_trig_level mode;
-};
-
 struct max_gpio_init_param {
 	/** Input/Output */
 	uint32_t mode;
 	/** Pull_up/Pull_down  */
-	uint32_t pull;
+	enum gpio_pull_up pull;
```
buha, line 64: Since enum gpio_pull_up pull is generic, it should go in gpio_init_param and gpio_desc, not here.
But I'm wondering now, if it'd be better to create a function for this because this setting is only applicable to inputs.

CiprianRegus:
V6:
- Added pull field to gpio_init_param and gpio_desc, removed it from maxim extra param.

buha (approved):

drivers/platform/maxim/maxim_irq.c
```diff
@@ -100,7 +100,7 @@ int32_t max_irq_ctrl_remove(struct irq_ctrl_desc *desc)
 /**
  * @brief Not implemented.
  * @param desc - The IRQ controller descriptor.
- * @param irq_id - Interrupt identifier.
```
amiclaus, line 103: nit: unrelated change / character added by mistake.

drivers/platform/maxim/maxim_spi.c
```diff
@@ -76,6 +76,7 @@ int32_t max_spi_init(struct spi_desc **desc, const struct spi_init_param *param)
 		return -ENOMEM;
 
 	eparam = param->extra;
+
```
amiclaus, line 79: nit: unrelated change with respect to the commit subject.

CiprianRegus:
Removed unrelated changes.

amiclaus (approved):

drivers/platform/maxim/maxim_stdio.c
```diff
@@ -45,89 +45,94 @@
...
 
 void maxim_uart_stdio(struct uart_desc *desc)
 {
 	if (!desc)
 		return;
-	mdesc = desc;
-}
+	guart = desc;
 
-int _open(const char *name, int flags, int mode)
```
adrimbarean, line 74: The "Works correctly." part in the commit description is not exactly needed. I would prefer it mention why the `_open` API is removed.

CiprianRegus:
Updated the commit message.

drivers/platform/maxim/maxim_gpio.c
```diff
@@ -390,60 +415,38 @@ static int32_t max_gpio_irq_set_trigger_level(struct irq_ctrl_desc *desc,
...
-	is_enabled = gpio_regs->int_en & BIT(irq_id);
-
-	/** Disable interrupts for pin desc->number */
-	gpio_regs->int_en &= ~(BIT(irq_id));
-	/** Clear pending interrupts for pin desc->number */
-	gpio_regs->int_clr |= BIT(irq_id);
-	/** Disable dual edge interrupts for pin desc->number */
-	gpio_regs->int_dual_edge &= ~BIT(irq_id);
+	mask = max_gpio_cfg->mask;
+	max_gpio_cfg->mask = BIT(irq_id);
```
adrimbarean, line 426: What does the mask represent here? Does it contain the enabled pins in a port? And if so, should this operation be `max_gpio_cfg->mask &= BIT(irq_id);` to also check that the bit is active?

adrimbarean (approved):
Looks good to me. A couple of questions.

drivers/platform/maxim/maxim_gpio.c
CiprianRegus: This struct (max_gpio_cfg) is used to specify a GPIO port configuration (so the mask field holds the pins on which the configuration should be done). However, the struct is not used to store the entire port state, meaning that you can't check if a pin is enabled by reading the mask field. For that you would have to read the en0 register. Also, I'm not sure the check is necessary because I imagine the interrupt setting sequence as following:

`irq_register_callback()
irq_trigger_level_set()
irq_enable()
`
The `irq_register_callback` call configures the pin as input (which means that it's enabled).

CiprianRegus: Also, the user has to provide a `gpio_desc` as extra for the irq descriptor (for the pin on which the interrupt would be set), so I assume that the GPIO pin has been configured already.

PR #1272, opened 2022-03-04
stm32 updates

buha:
- open the sdk in background
- patching of the _it.c file (preparation for interrupts)
- stm32_init() common function instead of having the user call SystemClock_Config() and HAL_Init() (& fix broken projects)
- change the way stm32_uart_init_param looks like (& fix broken projects)


buha:
V2:
- astyle

adrimbarean (approved):
Looks good to me.

amiclaus (approved):

PR #1276, opened 2022-03-04
drivers:dac:ad9144:ad9144.c: Fix broken check of NCO phase offset

kylex2:
Call to abs() has no effect. Data sheet for AD9144 rev. C pg. 70 says
allowed range is: -180 <= phase < 180.

Signed-off-by: Alexander Vickberg <wickbergster@gmail.com>


dbogdan (approved):

dbogdan:
@kylex2 , thank you for the fix.

PR #1279, opened 2022-03-07
drivers:platform:mbed: Error code handling for ARMC compiler

mphalke:
The standard "errno.h" file comes with ARMC compiler does not include all the error codes defined
in the "errno.h" file comes with GCC compiler. Some of the error codes used in no-os drivers
are not present in the errno.h file from ARMC compiler. The additional codes are therefore
added in the local errno.h file created inside mbed platform drivers folder based on assumption that
build system will locate this file first and then #include_next will include the standard library
errno.h file.

### DIFF
--- a/drivers/platform/mbed/errno.h
+++ b/drivers/platform/mbed/errno.h
@@ -1,0 +1,110 @@
+/***************************************************************************//**
+ *   @file   errno.h
+ *   @brief  Error macro definition for ARM Compiler
+********************************************************************************
+ * Copyright (c) 2021-2022 Analog Devices, Inc.
+ * All rights reserved.
+ *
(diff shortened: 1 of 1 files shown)


buha (approved):

amiclaus (approved):
lgtm, but maybe we can somehow integrate this in our generic `error.h`

mphalke:
> lgtm, but maybe we can somehow integrate this in our generic `error.h`

May be useful but just adding concern- We then have to include "error.h" file in the relevant places than directly using "errno.h".
I will wait for your decision on this..

mphalke:
Hi All,
Is this PR good to merge or are there any other comments?

PR #1284, opened 2022-03-08
drivers:axi_core:axi_dac_core:axi_dac_core.c: Fix DAC init crash

kylex2:
I have problems getting the no-OS code running on AD9364. It seems there's an issue with the axi_dac_init() code which gave me CPU exceptions (division by 0).

Only tested on AD9364 (ZedBoard+fmcomms4) and there it doesn't crash anymore.

> dac->clock_hz needs to be set before axi_dac_dds_set_frequency()
> is run. axi_dac_data_setup() is calling axi_dac_dds_set_frequency().
> Hence, we need to set dac->clock_hz before axi_dac_data_setup().
> 
> Move mdelay(100) so the frequency register value has time to
> stabilize after reset before reading.

### DIFF
--- a/drivers/axi_core/axi_dac_core/axi_dac_core.c
+++ b/drivers/axi_core/axi_dac_core/axi_dac_core.c
@@ -826,19 +826,12 @@ int32_t axi_dac_init_begin(struct axi_dac **dac_core,
 int32_t axi_dac_init_finish(struct axi_dac *dac)
 {
 	uint32_t reg_data;
-	uint32_t freq;
-	uint32_t ratio;
 
 	axi_dac_read(dac, AXI_DAC_REG_STATUS, &reg_data);
 	if(reg_data == 0x0) {
 		printf("%s: Status errors\n", dac->name);
 		return -1;
 	}
-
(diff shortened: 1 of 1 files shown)


kylex2:
Ok, so I just checked the commit history. This bug seems to be introduced in commit 1bf45a05e50d13f0c73a6aa1bbe7e425e45f87f4. Before that commit the reset was done in axi_dac_init_begin(), the delay was after that call. Also the assigning of dac->clock_hz in axi_dac_init_finish() was made before calling axi_dac_data_setup().

I don't know in which functions you prefer to do what.

amiclaus (approved):
lgtm, but I think it might need a bit more testing.

kylex2:
Sure, please test it as much as possible. It should be safe to merge though since it basically restores behavior to before 1bf45a05e50d13f0c73a6aa1bbe7e425e45f87f4.

kylex2:
Tried to resolve conflict with github tool but still said it had conflicts so I rebased to master and force pushed.

buha:

PR #1286, opened 2022-03-08
drivers: ada4250: offset computation updated

pmallick-adi:
The drivers now accepts avdd voltage and offset in float values.
The offset computation is done using the float values. 
The device descriptor now hold the raw offset value.

### DIFF
--- a/drivers/amplifiers/ada4250/ada4250.c
+++ b/drivers/amplifiers/ada4250/ada4250.c
@@ -151,7 +151,7 @@ int32_t ada4250_update_desc(struct ada4250_dev *dev)
 	ret = ada4250_read(dev, ADA4250_REG_SNSR_CAL_VAL, &reg_data);
 	if (ret != SUCCESS)
 		return ret;
-	dev->offset_uv = reg_data & ADA4250_SNSR_CAL_VAL_MSK;
+	dev->offset_raw = reg_data & ADA4250_SNSR_CAL_VAL_MSK;
 
 	ret = ada4250_read(dev, ADA4250_REG_SNSR_CAL_CNFG, &reg_data);
 	if (ret != SUCCESS)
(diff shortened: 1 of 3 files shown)


amiclaus (commented):
the reason why the offset was treated inside the driver in uV in the first place was exactly to avoid float types. We usually avoid this type of variables in our drivers so the drivers are more flexible and usable also with processors that do not support floating point arithmetic.

Pratyush-Mallick:
> the reason why the offset was treated inside the driver in uV in the first place was exactly to avoid float types. We usually avoid this type of variables so the drivers are more flexible and usable also with processors that do not support floating point arithmetic.

But then the AVDD for offset computation can range any 0-5V, which might also include voltages such as 3.3V or 1.8V. So then the AVDD input should also be represented in millivolts. According to the datasheet, the formula for [x[1]](https://github.com/analogdevicesinc/no-OS/blob/master/drivers/amplifiers/ada4250/ada4250.c#L328-L330) (if reprensted in millivoltas)should contain 0.126 as the multiplier and not just 126. And the AVDD in the X1 = 0.126*(AVDD-1) is represented in  Volts.

amiclaus (commented):
> > the reason why the offset was treated inside the driver in uV in the first place was exactly to avoid float types. We usually avoid this type of variables so the drivers are more flexible and usable also with processors that do not support floating point arithmetic.
> 
> But then the AVDD for offset computation can range any 0-5V, which might also include voltages such as 3.3V or 1.8V. So then the AVDD input should also be represented in millivolts. According to the datasheet, the formula for [x[1]](https://github.com/analogdevicesinc/no-OS/blob/master/drivers/amplifiers/ada4250/ada4250.c#L328-L330) (if reprensted in millivoltas)should contain 0.126 as the multiplier and not just 126. And the AVDD in the X1 = 0.126*(AVDD-1) is represented in Volts.

You are right, for avdd values like 1.8V/3.3V you won't obtain accurate values for x[0], x[1], x[2], etc. I'd suggest finding a different workaround than using float types. (This is usually a request also when developing linux drivers: no floating types, and we try to stick to that rule also for the no-OS drivers)

pmallick-adi:
No longer needed. Will create a different one for the changes

PR #1287, opened 2022-03-09
Use ARRAY_SIZE where appropriate

kylex2:
Use ARRAY_SIZE instead of sizeof(var)/sizeof(var[0]). Also remove two extra defines of ARRAY_SIZE.

I didn't touch navassa/talise/mykonos since to me it looks like this could be code from some other source which is imported in bulk to this repo. Navassa code uses its own macro ADI_ARRAY_LEN.


amiclaus (approved):

projects/ad7768-1fmcz/src/ad77681evb.c
```diff
@@ -141,7 +141,7 @@ int main()
 			ad77681_spi_read_adc_data(adc_dev, adc_data,
 						  AD77681_CONTINUOUS_DATA_READ);
 			printf("[ADC DATA]: 0x");
-			for(i = 0; i < sizeof(adc_data) / sizeof(uint8_t); i++) {
+			for(i = 0; i < sizeof(adc_data); i++) {
```
buha, line 144: shouldn't this be ARRAY_SIZE ?

buha: ok i checked the code, it doesn't really matter here, fine by me

buha (approved):

projects/ad7768-1fmcz/src/ad77681evb.c
kylex2: Looking at it now this change seems unnecessary. It doesn't add or change anything meaningful. I would propose to either change it to ARRAY_SIZE, like most other for loop limits, or remove the change.

PR #1288, opened 2022-03-09
afe:ad4110:Function additions, struct update

jansunil:
Added new functions to configure the device
Updated the init structure and device descriptor
Corrected the bit shift value for AD4110_REG_ADC_CONFIG_AIN_BUFF

### DIFF
--- a/drivers/afe/ad4110/ad4110.c
+++ b/drivers/afe/ad4110/ad4110.c
@@ -710,6 +710,171 @@ static void irq_adc_read(struct ad4110_callback_ctx *ctx)
 }
 
 /***************************************************************************//**
+ * @brief Enable/Disable channel
+ * @param dev - The device structure.
+ * @param chan_id - Channel ID (number)
+ * @param status - Channel Status (True for Enable and False for Disable)
+ * @return SUCCESS in case of success, negative error code otherwise.
+*******************************************************************************/
+int ad4110_set_channel_status(struct ad4110_dev *dev, uint8_t chan_id,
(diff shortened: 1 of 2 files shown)


drivers/afe/ad4110/ad4110.c
```diff
@@ -830,6 +966,26 @@ int32_t ad4110_setup(struct ad4110_dev **device,
 	if (ret)
 		goto err_spi;
 
+	ret = ad4110_set_bipolar(dev, init_param.bipolar);
+	if (ret)
+		goto err_spi;
+	dev->bipolar = init_param.bipolar;
```
amiclaus, line 972: why not making the assignment `dev->bipolar=init_param.bipolar;` inside the setter function if the write is completed successfully?
Same applies for all the other setter functions.

In this manner the users can benefit of these functions outside the driver and the devices structure is updated accordingly.

jansunil: Moved the device descriptor update inside the setter function.

jansunil:
Following changed as a part of addressing review comments:
 - Moved the device desc update inside the setter function.

jansunil:
Change made:
Added missing function prototype for ad4110_wait_for_rdy_low() in the ad4110.h file

jansunil:
No changes made in the previous push. Edited the commit message and force pushed to re-trigger a build.

amiclaus (approved):

buha (approved):

PR #1289, opened 2022-03-09
iio: Introduce iio_snprintf to harden printing to buffers

kylex2:
Hi,

I have introduced iio_snprintf which should detect when buffer is too small and return error.

What do you think about it?

> Quote from 'man snprintf':
> 
> The functions snprintf() and vsnprintf() do not write more than size bytes
> (including the terminating null byte ('\0')). If the output was truncated due
> to this limit then the return value is the number of characters (excluding the
> terminating null byte) which would have been written to the final string if
> enough space had been available. Thus, a return value of size or more means
> that the output was truncated.
> 
> Signed-off-by: Alexander Vickberg <wickbergster@gmail.com>

### DIFF
--- a/drivers/adc/ad7124/iio_ad7124.c
+++ b/drivers/adc/ad7124/iio_ad7124.c
@@ -196,7 +196,7 @@ static int ad7124_iio_read_offset_chan(void *device, char *buf, uint32_t len,
 	if (ret != SUCCESS)
 		return ret;
 
-	return snprintf(buf, len, "%"PRIX32"", value);
+	return iio_snprintf(buf, len, "%"PRIX32, value);
 }
 
 /**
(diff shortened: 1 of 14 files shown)


kylex2:
Added attribute for format checking

kylex2:
Add function description. Handle case where incoming buffer length is zero and no characters written. Remove faulty null termination.

buha:
Hi @kylex2 ,
This looks good in principle, but how about using iio_format_value ?
Example usage:
https://github.com/analogdevicesinc/no-OS/blob/d6b3591b5af2969818a55795681332f39ac29e1f/drivers/cdc/ad7746/iio_ad7746.c#L258

The occurrences of snprintf that you replaced with iio_snprintf should in my opinion be replaced with something like iio_format_value, this way the logic of how formatting is done is hidden from the iio driver writer and is common for all drivers.

I should say that I agree with the change myself (let's see other opinions) but if we merge this PR, then each iio_snprintf occurrence should gradually disappear in favor of iio_format_value.

kylex2:
iio_format_value could make sense in many cases. I can see if I get some time to look into it, if so I can add commits to this pull request.

buha:

PR #1290, opened 2022-03-09
iio:iio.c: Rewrite iio_read/write_all_attr

kylex2:
Hi,

IIO-Oscilloscope doesn't fill in values on start. I have rewritten the iio_read_all_attr() and iio_write_all_attr() functions. With these in place the values are populated correctly. My libiio/iio-scope version doesn't send values for WRITE command so I can't really test that at the moment. (Buffer sent is just NULL bytes, checked with Wireshark). Any ideas about that?

If accepted, this should preferably be merged after the iio_snprintf pull-request so that attr->show will return error on too small buffer.

> Not all devices have all types of attributes, this lead to attributes argument
> passed sometimes beeing NULL. (e.g. ad9361-phy has IIO_ATTR_TYPE_DEVICE attri-
> butes but cf-ad9361-lpc and cf-ad9361-dds-core-lpc don't). Add check for this.
> 
> Signed-off-by: Alexander Vickberg <wickbergster@gmail.com>

### DIFF
--- a/iio/iio.c
+++ b/iio/iio.c
@@ -57,6 +57,13 @@
 #include <inttypes.h>
 #include <stdio.h>
 #include <string.h>
+#if defined LINUX_PLATFORM
+#include <arpa/inet.h>
+#else
+#include <machine/endian.h>
+#define	htonl(_x) __htonl(_x)
+#define	ntohl(_x) __ntohl(_x)
+#endif
 
 #ifdef ENABLE_IIO_NETWORK
 #include "no-os/delay.h"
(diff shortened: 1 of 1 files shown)


kylex2:
Force pushed build fix for htonl/ntohl under newlib.

iio/iio.c
```diff
@@ -353,30 +367,41 @@ static int iio_read_all_attr(struct attr_fun_params *params,
...
-			j = ((j >> 2) + 1) << 2;
-		i++;
-	}
+	for (struct iio_attribute *attr = attributes; attr->name; attr++) {
+		uint32_t attr_len;
 
-	if (params->len == 0)
-		return -ENOENT;
+		// Allign to next multiple of 4
+		while (read & 0x3)
```
adrimbarean, line 382: Do we know that data here comes in aligned in this particular way? Do we risk reading the data wrong by doing this?

iio/iio.c
```diff
@@ -300,45 +307,52 @@ static struct iio_dev_priv *get_iio_device(struct iio_desc *desc,
...
+				      params->buf+written, params->len-written,
+				      params->ch_info, attr->priv);
+
+		if (attr_len < 0)
+			return -EINVAL;
+
+		attr_len += 1; // Add '\0' to the count
+
+		attr_len_be = htonl(attr_len);
+		memcpy(pattr_len, &attr_len_be, sizeof(attr_len_be));
```
adrimbarean, line 347: I am not sure here, but I think length should come before data. We have to check this.

kylex2: pattr_len points to place in buffer before data.

iio/iio.c
kylex2: You tell me, the previous code aligned to multiples of 4 too. Is the protocol specification public in documentation other than source code?

iio/iio.c
```diff
@@ -300,45 +307,52 @@ static struct iio_dev_priv *get_iio_device(struct iio_desc *desc,
...
+		// reserve space for next attribute length
+		written += sizeof(attr_len_be);
+		if (written >= params->len)
+			return -ENOBUFS;
+
+		attr_len = attr->show(params->dev_instance,
+				      params->buf+written, params->len-written,
+				      params->ch_info, attr->priv);
+
+		if (attr_len < 0)
```
kylex2, line 341: The previous code printed the error code of show() into the buffer. To keep this functionality here we should check if return is != -ENOBUFS (which should be returned if out of buf space after/if iio_snprintf PR is accepted).

buha:

PR #1291, opened 2022-03-10
iio: ability to set IIO context attributes from application code

mphalke:
Implemented the changes in no-os IIO library to allow user to set the context attributes and values
from an application code.

### DIFF
--- a/iio/iio.c
+++ b/iio/iio.c
@@ -1143,7 +1143,22 @@ static uint32_t iio_generate_device_xml(struct iio_device *device, char *name,
 		buff = ch_id;
 
 	i = 0;
-	i += snprintf(buff, no_os_max(n - i, 0),
+
+	/* Write context attributes */
+	if (device->context_attributes)
+		for (j = 0; device->context_attributes[j].name; j++) {
+			i += snprintf(buff + i,
+				      no_os_max(n - i, 0),
+				      "<context-attribute name=\"%s\" ",
+				      device->context_attributes[j].name);
+
+			i += snprintf(buff + i,
+				      no_os_max(n - i, 0),
+				      "value=\"%s\" />",
+				      device->context_attributes[j].value);
+		}
+
+	i += snprintf(buff + i, no_os_max(n - i, 0),
 		      "<device id=\"%s\" name=\"%s\">", id, name);
 
 	/* Write channels */
--- a/iio/iio_types.h
+++ b/iio/iio_types.h
@@ -115,6 +115,17 @@ enum iio_attribute_shared {
 };
 
 /**
+ * @struct iio_context_attribute
+ * @brief Structure holding the context attribute members
+ */
+struct iio_context_attribute {
+	/** Attribute name */
+	const char *name;
+	/** Attribute value */
+	const char *value;
+};
+
+/**
  * @struct iio_attribute
  * @brief Structure holding pointers to show and store functions.
  */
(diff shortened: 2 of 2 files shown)


iio/iio.c
```diff
@@ -1141,7 +1141,22 @@ static uint32_t iio_generate_device_xml(struct iio_device *device, char *name,
...
 	i = 0;
-	i += snprintf(buff, max(n - i, 0),
+
+	/* Write context attributes */
+	if (device->context_attributes)
+		for (j = 0; device->context_attributes[j].name; j++) {
+			i += snprintf(buff + i,
+				      max(n - i, 0),
+				      "<context-attribute name=\"%s\" ",
+				      device->context_attributes[j].name);
```
buha, line 1151: I'm wondering if it'd make sense to put some default values for .name and .value but I guess it's also fine without any defaults

mphalke: @buha - sorry, is this question for me or other reviewers? Where do you suggest default values, I didn't get you?

buha: I was just thinking out loud. I think currently the name defaults to "no-OS" and the value is the git version that defaults to something like "1.0.0-g0000000". If device->context_attributes[j].name is NULL, I think this is undefined behaviour in C:
https://stackoverflow.com/questions/11589342/what-is-the-behavior-of-printing-null-with-printfs-s-specifier

So maybe just initialize device->context_attributes[j].name with "no-OS" if it is NULL, prior to the snprintf call.

mphalke: Looks like NULL check done in the for loop. Would that be good?
```
for (j = 0; device->context_attributes[j].name; j++)
```

adrimbarean (approved):
Looks good to me.

mphalke:
Hi,
Are we good with this PR?

amiclaus (approved):

iio/iio.c
buha: seems ok then, just make sure to fix conflicts

mphalke:
Not exactly sure from where the conflict is coming from. Any clue?
Do I need to merge the latest changes?

amiclaus (commented):
image

`max` function was renamed in the meantime to `no_os_max`
https://github.com/analogdevicesinc/no-OS/blob/a56c44f093f2764f9350dc1120e96425dde01d88/include/no_os_util.h#L68

amiclaus (commented):
> Not exactly sure from where the conflict is coming from. Any clue? Do I need to merge the latest changes?

So you will have to rebase your branch with latest master branch. And rename in your patch the `max` macro.

mphalke:
> > Not exactly sure from where the conflict is coming from. Any clue? Do I need to merge the latest changes?
> 
> So you will have to rebase your branch with latest master branch. And rename in your patch the `max` macro.

Thanks. I'll rebase with master and make this change.

iio/iio.c
```diff
@@ -1143,7 +1143,22 @@ static uint32_t iio_generate_device_xml(struct iio_device *device, char *name,
 		buff = ch_id;
 
 	i = 0;
-	i += snprintf(buff, no_os_max(n - i, 0),
+
+	/* Write context attributes */
+	if (device->context_attributes)
+		for (j = 0; device->context_attributes[j].name; j++) {
+			i += snprintf(buff + i,
+				      max(n - i, 0),
```
rbolboac, line 1151: Please use no_os_max instead of max.

rbolboac (changes requested):
Besides these 2 small remarks, lgtm.

mphalke:
V1.0:

- Replaced max() with no_os_max() in iio.c file

rbolboac (approved):
lgtm

amiclaus (approved):

PR #1297, opened 2022-03-16
adxrs453: Fix value returned by register read

ryandavid:
Value returned by register read was actually the result from the previous command.

Issue is repeatable:
 - Attempt to read the part ID register (assuming it wasn't the last register read since power on)
 - Result back will not match.
 - Subsequent read of the part ID register will return the correct result, but its really the value from the previous request.

image

### DIFF
--- a/drivers/gyro/adxrs453/adxrs453.c
+++ b/drivers/gyro/adxrs453/adxrs453.c
@@ -131,9 +131,9 @@ uint16_t adxrs453_get_register_value(struct adxrs453_dev *dev,
 	data_buffer[7] = data_buffer[3];
 	no_os_spi_write_and_read(dev->spi_desc, data_buffer, 4);
 	no_os_spi_write_and_read(dev->spi_desc, &data_buffer[4], 4);
-	register_value = ((uint16_t)data_buffer[1] << 11) |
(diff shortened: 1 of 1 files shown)


amiclaus (approved):

buha:
Good catch.
But could you please put the text from the PR description in the commit message ?
And perhaps amend with `git --amend -s` and then `git push --force` ?

amiclaus (changes requested):
> Good catch. But could you please put the text from the PR description in the commit message ? And perhaps amend with `git --amend -s` and then `git push --force` ?



> Good catch. But could you please put the text from the PR description in the commit message ? And perhaps amend with `git --amend -s` and then `git push --force` ?

Indeed, @buha  is right. Also please make sure that the commit contains the `Signed-off by:` tag

amiclaus (commented):
@ryandavid any feedback on the requested changes?

amiclaus (commented):
@ryandavid due to long response time and wanting to reduce the nr. of PR opened on the no-OS repository, I created a PR with the required changes here:  https://github.com/analogdevicesinc/no-OS/pull/1315

On that note, closing this PR.

PR #1298, opened 2022-03-16
drivers: ada4250: offset computation update

pmallick-adi:
The vlsb and input offset is now represented in nanovolts. The AVDD is now represented in millivolts. This would enable the user to input AVDD in 3.3v and 1.8V.

### DIFF
--- a/drivers/amplifiers/ada4250/ada4250.c
+++ b/drivers/amplifiers/ada4250/ada4250.c
@@ -148,11 +148,6 @@ int32_t ada4250_update_desc(struct ada4250_dev *dev)
 		return ret;
 	dev->refbuf_en = reg_data & ADA4250_REFBUF_MSK;
 
-	ret = ada4250_read(dev, ADA4250_REG_SNSR_CAL_VAL, &reg_data);
-	if (ret != 0)
-		return ret;
-	dev->offset_uv = reg_data & ADA4250_SNSR_CAL_VAL_MSK;
-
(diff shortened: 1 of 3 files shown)


pmallick-adi:
Any updates on this?

drivers/amplifiers/ada4250/ada4250.c
```diff
@@ -148,11 +148,6 @@ int32_t ada4250_update_desc(struct ada4250_dev *dev)
 		return ret;
 	dev->refbuf_en = reg_data & ADA4250_REFBUF_MSK;
 
-	ret = ada4250_read(dev, ADA4250_REG_SNSR_CAL_VAL, &reg_data);
-	if (ret != 0)
-		return ret;
-	dev->offset_uv = reg_data & ADA4250_SNSR_CAL_VAL_MSK;
```
adrimbarean, line 154: Is this not needed anymore?

adrimbarean (approved):
Looks good to me. Just a question.

amiclaus (approved):
lgtm, after the question of @adrimbarean is addressed.

drivers/amplifiers/ada4250/ada4250.c
pmallick-adi: I don't think so. The only way the dev->offset_nv is going to be updated is through the ada4250_set_offset() function. Or else we need to maintain a sperate function for just calculating the offset in nanovolts.

PR #1301, opened 2022-03-17
Fix iio_app no_os naming

APorumb:
Fix for commit cc5d097b3a1a598919cf8092757c8cd2afe40b67


buha (approved):
yep. looks like this is a leftover
i checked, it'd the only one, so I'm merging

PR #1303, opened 2022-03-21
Staging/delay

danmois:
The activation of instruction cache for Xilinx projects. This is necessary for the correct operation of the usleep() and msleep() functions, that call usleep_MB(). The usleep_MB() function requires the enabling of the instruction cache on Xilinx platforms.

### DIFF
--- a/projects/ad9371/src/app/headless.c
+++ b/projects/ad9371/src/app/headless.c
@@ -314,6 +314,13 @@ int main(void)
 
 	printf("Please wait...\n");
 
+#ifdef XILINX_PLATFORM
+	/* Enable the instruction cache. */
+	Xil_ICacheEnable();
+	/* Enable the instruction cache. */
+	Xil_DCacheEnable();
+#endif //XILINX_PLATFORM
+
 	ret = platform_init();
 	if (ret != 0) {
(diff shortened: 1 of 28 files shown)


projects/ad9371/src/app/headless.c
```diff
@@ -314,6 +314,12 @@ int main(void)
 
 	printf("Please wait...\n");
 
+	/* Instruction cache should be enabled for usleep functions to work. */
+	/* Enable the instruction cache. */
+	Xil_ICacheEnable();
```
buha, line 320: this ad9371 change should probably go in the 2nd commit together with all the other such changes

danmois:
V2

- reorganized commits so that they deal with the same modifications (1. delay functions. 2. activation of instruction/data cache)

dbogdan (approved):

adrimbarean (approved):
Looks good to me.

dbogdan:
@danmois , I've just realized that it was not mentioned the main reason for enabling the Instruction Cache: it should be enabled for usleep_MB() to work.

danmois:
V3

- the addition of text that describes the main reason for the modifications: Instruction Cache is enabled for the correct operation of usleep_MB().

PR #1306, opened 2022-03-28
iio: fix fractional number interpretation

adrimbarean:
Multiplication with 100000 is only true if only one decimal precision is
used. If the user input more than one decimal precision the resulting
number does not represent micro/nano subdivision anymore.


amiclaus (approved):
lgtm. But a fixes tag would be nice :)

PR #1308, opened 2022-03-29
ADXL355 fix hpf corner value setting 

rbolboac:
- The hpf corner value has to be left shifted 4 times because hpf corner value bits are [6:4] in filter settings reg.
- Even though the hpf settings can be changed when the device is in measurement mode, it has been observed that the measurements are not correct when doing so. Because of this, the device will be set in standby mode also when changing the hpf value.

### DIFF
--- a/drivers/accel/adxl355/adxl355.c
+++ b/drivers/accel/adxl355/adxl355.c
@@ -419,9 +419,27 @@ int adxl355_set_hpf_corner(struct adxl355_dev *dev,
 {
 	int ret;
 	uint8_t reg_value;
+	enum adxl355_op_mode current_op_mode;
 
-	// The hpf settings can be changed when the device is in measurement mode.
-	// There is no need to modify the device's operation mode to Standby.
+	// Even though the hpf settings can be changed when the device is in
+	// measurement mode, it has been observed that the measurements are not
+	// correct when doing so. Because of this, the device will be set in standby
+	// mode also when changing the hpf value.
+
+	current_op_mode = dev->op_mode;
+
+	switch(current_op_mode) {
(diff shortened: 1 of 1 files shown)


drivers/accel/adxl355/adxl355.c
```diff
@@ -419,9 +419,23 @@ int adxl355_set_hpf_corner(struct adxl355_dev *dev,
...
+	// Even though the hpf settings can be changed when the device is in
+	// measurement mode, it has been observed that the measurements are not
+	// correct when doing so. Because of this, the device will be set in standby
+	// mode also when changing the hpf value.
+
+	current_op_mode = dev->op_mode;
 
-	// The hpf settings can be changed when the device is in measurement mode.
-	// There is no need to modify the device's operation mode to Standby.
+	if ((current_op_mode != ADXL355_STDBY_TEMP_ON_DRDY_ON)
```
amiclaus, line 431: this can be condensed by checking if the op_mode is equal with the rest of 3 elements in the  `enum adxl355_op_mode`

amiclaus (approved):
One minor comment. Otherwise, lgtm.

drivers/accel/adxl355/adxl355.c
amiclaus: A switch...case might fit too.

rbolboac: enum adxl355_op_mode contains 8 elements in total, 4 of them are specific to standby state and 4 of them are specific to measurement state so 4 checks have to be performed either way. I made the changes with switch case.

dbogdan (approved):

PR #1310, opened 2022-03-30
util: add proper support for hweigth8/16/32

amiclaus:
The current hweigth8 function was inconcludent since it was implemented
to support 32-bit size words.

Add separate functions for hweight8/hweigth16/hweight32 similar with the
linux implementation.

Update drivers/projects accordingly.

### DIFF
--- a/include/no_os_util.h
+++ b/include/no_os_util.h
@@ -143,8 +143,12 @@ void no_os_rational_best_approximation(uint32_t given_numerator,
 				       uint32_t max_denominator,
 				       uint32_t *best_numerator,
 				       uint32_t *best_denominator);
-/* Calculate the number of set bits. */
-uint32_t no_os_hweight8(uint32_t word);
+/* Calculate the number of set bits (8-bit size). */
+unsigned int no_os_hweight8(uint8_t word);
+/* Calculate the number of set bits (16-bit size). */
+unsigned int no_os_hweight16(uint16_t word);
+/* Calculate the number of set bits (32-bit size). */
+unsigned int no_os_hweight32(uint32_t word);
 /* Calculate the quotient and the remainder of an integer division. */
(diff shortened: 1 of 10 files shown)


include/no_os_util.h
```diff
@@ -143,8 +143,12 @@ void no_os_rational_best_approximation(uint32_t given_numerator,
 				       uint32_t max_denominator,
 				       uint32_t *best_numerator,
 				       uint32_t *best_denominator);
-/* Calculate the number of set bits. */
-uint32_t no_os_hweight8(uint32_t word);
+/* Calculate the number of set bits (8-bit size). */
+uint8_t no_os_hweight8(uint8_t word);
+/* Calculate the number of set bits (16-bit size). */
+uint16_t no_os_hweight16(uint16_t word);
```
rbolboac, line 149: do we really need return type uint16_t? won't uint8_t still suffice?

include/no_os_util.h
```diff
@@ -143,8 +143,12 @@ void no_os_rational_best_approximation(uint32_t given_numerator,
...
+/* Calculate the number of set bits (32-bit size). */
+uint32_t no_os_hweight32(uint32_t word);
```
rbolboac, line 151: ditto: do we really need return type uint32_t? won't uint8_t still suffice?

amiclaus (commented):
changes in v2:
- user `uint8_t` return type for all `hweight` functions

buha (approved):
LGTM although I think the return type should've been native "unsigned int" for all of these functions. We might run into cast warnings if we don't use flat return type for all these.

amiclaus (commented):
changes in v3:
- use `unsigned int` as return type

rbolboac (approved):

Issue #1294, opened 2022-03-11
Is that a mistake about AD469x driver configure standard sequencer channels?


Jimwang9018:
In line 487

~~~
/**
 * @brief Configure standard sequencer channels
 * @param [in] dev - ad469x_dev device handler.
 * @param [in] ch_mask - Extra channels to activate.
 * @return \ref SUCCESS in case of success, \ref FAILURE otherwise.
 */
int32_t ad469x_std_sequence_ch(struct ad469x_dev *dev, uint16_t ch_mask)
{
	int32_t ret;

	ret = ad469x_spi_reg_write(dev,
				   AD469x_REG_STD_SEQ_CONFIG,
				   0x0f & ch_mask);
	if (ret != SUCCESS)
		return ret;

	ret = ad469x_spi_reg_write(dev,
				   AD469x_REG_STD_SEQ_CONFIG + 1,
				   ch_mask >> 8);
	if (ret != SUCCESS)
		return ret;

	dev->num_slots = hweight8(ch_mask);

	return ret;
}
~~~
That “0x0f & ch_mask” 0x0f? I think that is 0xff. 
Please help me !

amiclaus:
Hello,

Thanks for pointing this out. 

We will do a fix soon.

PR #1311, opened 2022-03-30
drivers: adc: ad469x: fix channel mask

amiclaus:
Fixes mask for the number of active channels when writing the
AD469x_REG_STD_SEQ_CONFIG register.

Fix issue reported at #1294.

Fixes: 0db1a92 ("ad469x: Add channel sequencer implementation")


rbolboac (approved):
lgtm

PR #1320, opened 2022-04-05
Maxim IRQ update

CiprianRegus:
Updated the way interrupts are handled on Maxim platform, according to the new standard. The supported drivers are GPIO, RTC and UART.


buha:
yeah so as discussed, let's also do some changes in top level irq:
- new field in struct no_os_callback_desc .handle (arbitrary name)
- perhaps irq_desc should contain a list of such .handle for all the irq_id's configured. Every time irq_register_callback is called, a handle is added to the list (together with its corresponding irq_id). This way, when irq_trigger_level_set is called, one may search the list by irq_id and get the necessary handle.

drivers/platform/maxim/maxim_irq.c
```diff
@@ -44,15 +44,158 @@
...
+	else if (req->txLen == req->txCnt && req->txLen != 0)
+		ee = &_events[NO_OS_EVT_UART_TX_COMPLETE];
+	else
+		ee = &_events[NO_OS_EVT_UART_RX_COMPLETE];
+
+	ret = no_os_list_read_find(ee->actions, (void **)&a, &key);
+	if (ret)
+		return ret;
+	if (a->callback)
+		a->callback(a->ctx);
```
adrimbarean, line 194: The callbacks are added in a list, as far as I can tell. Is this so that there can be more, even on the same event? If so, should we make sure to cal them all?

adrimbarean (commented):
Looks good to me. I think this isn't yet the final form, but I do have one question.

drivers/platform/maxim/maxim_irq.c
CiprianRegus: There can be multiple callbacks for the same event, but there should be only one for each peripheral (e.g TX_COMPLETE for UART0 and UART1). In this case, it doesn't make sense to call all of them (the peripheral is selected based on a handler in the same function). But, as you pointed out there might be a problem if the user registers multiple callbacks on the same event/peripheral instance, since only the first would get called.

CiprianRegus:
V2:
- Added a different IRQ controller for GPIO handling, since multiplexing GPIO IRQ lines was not possible with just 1 irq_id.
- Added the handles list in no_os_irq_ctrl_desc. This is used to store irq_handles (irq_id, handle). There are 2 reasons for this change: to have a way of getting a handle based on an irq_id, since a handle is not always provided (e.g trigger_level_set function) and to be able to use the "high level" IRQ controller for multiple peripherals (the handles are no longer passed in the irq_ctrl_desc's extra field). However, this might not be a good solution, because it requires to have the irq_handles and actions lists synchronized, meaning that the same handles in both lists. Also, the use of no_os_list.h header in no_os_irq.h makes some projects that use IRQs to fail. A better alternative would be to have the irq_id field in the irq_action struct.

Example of use:

```
// NVIC init

struct no_os_irq_ctrl_desc *nvic;
struct no_os_irq_init_param nvic_param = {
    .irq_ctrl_id = 0,
    .platform_ops = &max_irq_ops,
    .extra = NULL
};
ret = no_os_irq_ctrl_init(&nvic, &nvic_param);

// UART (uses only the NVIC)
ret = no_os_uart_init(&uart, &uart_init);

struct no_os_callback_desc uart_cb_desc = {
      .callback = uart_cb,
      .ctx = uart,
      .event = NO_OS_EVT_UART_TX_COMPLETE,
      .peripheral = NO_OS_UART_IRQ,
      .handle = uart, // the handle is no longer in nvic->extra
};

ret = no_os_irq_register_callback(nvic, UART0_IRQn, &uart_cb_desc);
ret = no_os_irq_enable(nvic, UART0_IRQn);

// GPIO (uses NVIC and the GPIO IRQ controller)

ret = no_os_gpio_get(&desc1, &gpio1_init);
ret = no_os_gpio_get(&desc2, &gpio2_init);

struct no_os_irq_ctrl_desc *gpio_irq; // 1 gpio irq controller per gpio port
struct no_os_irq_init_param gpio_irq_param = {
    .irq_ctrl_id = 0,
    .platform_ops = &max_gpio_irq_ops,
    .extra = NULL
};

struct no_os_callback_desc gpio1_cb = {
    .callback = gpio1_callback,
    .ctx = desc1,
    .event = NO_OS_EVT_GPIO,
    .peripheral = NO_OS_GPIO_IRQ,
    .handle = desc1,
};
struct no_os_callback_desc gpio2_cb = {
    .callback = gpio2_callback,
    .ctx = desc2,
    .event = NO_OS_EVT_GPIO,
    .peripheral = NO_OS_GPIO_IRQ,
    .handle = desc2,
};
ret = no_os_irq_ctrl_init(&gpio_irq, &gpio_irq_param);

ret = no_os_irq_register_callback(gpio_irq, 2, &gpio1_cb); // The irq_id is the pin number of the port irq_ctrl_id
ret = no_os_irq_trigger_level_set(gpio_irq, 2, NO_OS_IRQ_EDGE_RISING);
ret = no_os_irq_enable(gpio_irq, 2);

ret = no_os_irq_register_callback(gpio_irq, 2, &gpio2_cb); // Multiple GPIOs on the same port use the same irq controller
ret = no_os_irq_register_callback(gpio_irq, 2, &gpio2_cb); // This overwrites the callback (only 1 action for a handle at a time)
ret = no_os_irq_trigger_level_set(gpio_irq, 3, NO_OS_IRQ_EDGE_BOTH);
ret = no_os_irq_enable(gpio_irq, 3);

ret = no_os_irq_enable(nvic, GPIO0_IRQn);
```

CiprianRegus:
V3:
- Removed the handles list from no_os_irq_ctrl_desc, and  instead added an irq_id field to the action struct.
- Added the no_os_rtc_set_irq_time function, which will be used to set the "time" at which an interrupt will occur.

Example of use for RTC:

```
ret = no_os_rtc_set_irq_time(rtc, 1);
ret = no_os_irq_register_callback(nvic, RTC_IRQn, &rtc_cb);
ret = no_os_irq_enable(nvic, RTC_IRQn);
ret = no_os_rtc_start(rtc);

```

CiprianRegus:
Added doxygen for maxim_gpio_irq.c

drivers/platform/maxim/maxim_rtc.c
```diff
@@ -199,3 +199,16 @@ int32_t no_os_rtc_set_cnt(struct no_os_rtc_desc *dev, uint32_t tmr_cnt)
 
 	return 0;
 }
+
+/**
+ * @brief Set the current count for the real time clock.
+ * @param dev - The RTC descriptor.
+ * @param irq_time - The number of seconds at which the interrupt must occur.
```
buha, line 206: is this an interval ? or is it the number of seconds that should pass from the function call to the interrupt ?

buha:
I think you should squash commits 3 & 4 with the last one, otherwise it's hard to follow the changes.

CiprianRegus:
Squashed IRQ delete commits with the IRQ refactoring one.

CiprianRegus:
Removed unused include

CiprianRegus:
Implemented a mechanism that will allow for registering a new nonblocking operation from inside of a callback. The reason for this change is that the Maxim's UART AsyncHandler function stops all the async transactions after the callback, thus making the following sequence to not work properly:
```
void uart_cb_fn(void *context)
{
	struct no_os_uart_desc *uart_desc = context;
	no_os_uart_read_nonblocking(uart_desc, &c, 1);
}
```

CiprianRegus:
Fixed some warnings.

drivers/platform/maxim/gpio_irq_extra.h
```diff
@@ -0,0 +1,58 @@
...
+
+#ifndef GPIO_IRQ_EXTRA_H_
+#define GPIO_IRQ_EXTRA_H
+
+/******************************************************************************/
+/***************************** Include Files **********************************/
+/******************************************************************************/
+
+#include "no_os_irq.h"
+
```
rbolboac, line 48: nitpick: a comment would be nice

drivers/platform/maxim/irq_extra.h
```diff
@@ -41,10 +41,33 @@
 
 #include "maxim_hal.h"
 #include "no_os_irq.h"
+#include "uart.h"
+
```
rbolboac, line 45: nitpick: a comment would be nice

drivers/platform/maxim/irq_extra.h
```diff
@@ -41,10 +41,33 @@
...
+struct irq_handle {
+	uint32_t irq_id;
+	void *handle;
+};
+
```
rbolboac, line 50: nitpick: a comment would be nice

drivers/platform/maxim/irq_extra.h
```diff
@@ -41,10 +41,33 @@
...
+struct irq_action {
+	uint32_t irq_id;
+	void *handle;
+	void (*callback)(void *context);
+	void *ctx;
+};
+
```
rbolboac, line 57: nitpick: a comment would be nice

drivers/platform/maxim/maxim_irq.c
```diff
@@ -43,16 +43,136 @@
...
+	{.event = NO_OS_EVT_RTC},
+};
+
+extern mxc_uart_req_t *uart_irq_state[MXC_UART_INSTANCES];
 
 /******************************************************************************/
 /************************ Functions Definitions *******************************/
 /******************************************************************************/
 
+static bool is_gpio_irq_id(uint32_t irq_id)
```
rbolboac, line 73: nitpick: function comment is missing.

drivers/platform/maxim/maxim_irq.c
```diff
@@ -43,16 +43,136 @@
...
+{
+	for (uint32_t i = 0; i < MXC_CFG_GPIO_INSTANCES; i++) {
+		if (irq_id == MXC_GPIO_GET_IRQ(i))
+			return true;
+	}
+
+	return false;
+}
+
```
rbolboac, line 82: nitpick: function comment is missing.

drivers/platform/maxim/maxim_irq.c
```diff
@@ -43,16 +43,136 @@
...
+int32_t irq_action_cmp(void *data1, void *data2)
+{
+	return ((struct irq_action *)data1)->irq_id -
+	       ((struct irq_action *)data2)->irq_id;
+}
+
+static void _uart_common_handler(mxc_uart_regs_t *uart)
```
rbolboac, line 89: nitpick: function comment is missing.

CiprianRegus:
Added function comments.

CiprianRegus:
- Changed `uart_irq_state` array to be statically allocated, which improves the latency a bit.
- Added is_callback variable. This mechanism is used to skip the actual nonblocking request registration if it's made in a callback function (because it will be removed anyway by the HAL driver after the callback finishes). This also improves latency.

buha (approved):
looks good! but perhaps we should make it as clear as possible in the doxygen that no_os_irq_ctrl_desc->extra is not intended as a peripheral handle, but is intended to hold irq controller specific stuff.

drivers/platform/maxim/maxim_gpio_irq.c
```diff
@@ -0,0 +1,381 @@
...
+/************************ Functions Definitions *******************************/
+/******************************************************************************/
+
+/**
+ * @brief Action comparator function
+ * @param data1 - List element
+ * @param data2 - Key
+ * @return 0 if the two are equal, any other integer otherwise
+ */
+static int32_t actions_cmp(void *data1, void *data2)
```
rbolboac, line 73: nit: I see there is another function: irq_action_cmp. are they doing the same thing? if so, can this one be removed?

CiprianRegus:
- Removed the `actions_cmp` function and used `irq_action_cmp` instead, since those were identical.
- Updated doxygen comments for `no_os_irq_ctrl_desc`, `no_os_irq_init_param` (`extra` field) and `callaback_desc`'s `handle` to better describe what is the intended use.

rbolboac (approved):

PR #1322, opened 2022-04-06
ADXL367 driver and project addition

APorumb:
-adxl367 driver created
-adxl367 project created

### DIFF
--- a/drivers/accel/adxl367/adxl367.c
+++ b/drivers/accel/adxl367/adxl367.c
@@ -1,0 +1,1322 @@
+/***************************************************************************//**
+ *   @file   adxl367.c
+ *   @brief  Implementation of ADXL367 Driver.
+ *   @author Andrei Porumb (andrei.porumb@analog.com)
+********************************************************************************
+ * Copyright 2022(c) Analog Devices, Inc.
+ *
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *  - Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ *  - Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in
+ *    the documentation and/or other materials provided with the
+ *    distribution.
+ *  - Neither the name of Analog Devices, Inc. nor the names of its
+ *    contributors may be used to endorse or promote products derived
+ *    from this software without specific prior written permission.
+ *  - The use of this software may or may not infringe the patent rights
+ *    of one or more patent holders.  This license does not release you
+ *    from the requirement that you obtain separate licenses from these
+ *    patent holders to use this software.
+ *  - Use of the software either in source or binary form, must be run
+ *    on or directly connected to an Analog Devices Inc. component.
+ *
+ * THIS SOFTWARE IS PROVIDED BY ANALOG DEVICES "AS IS" AND ANY EXPRESS OR
+ * IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, NON-INFRINGEMENT,
+ * MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED.
+ * IN NO EVENT SHALL ANALOG DEVICES BE LIABLE FOR ANY DIRECT, INDIRECT,
+ * INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, INTELLECTUAL PROPERTY RIGHTS, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+*******************************************************************************/
+
+/******************************************************************************/
+/***************************** Include Files **********************************/
+/******************************************************************************/
+#include <stdlib.h>
+#include "adxl367.h"
+#include "no_os_delay.h"
+#include "no_os_util.h"
+#include "no_os_print_log.h"
+
+/******************************************************************************/
+/************************ Variable Declarations ******************************/
+/******************************************************************************/
+static const uint8_t adxl367_scale_mul[3] = {1, 2, 4};
+static uint8_t samples_per_set = 0;
+
+/******************************************************************************/
+/************************ Functions Definitions *******************************/
+/******************************************************************************/
+
+/***************************************************************************//**
+ * @brief Initializes communication with the device and checks if the part is
+ *        present by reading the device id.
+ *
+ * @param device     - The device structure.
+ * @param init_param - The structure that contains the device initial
+ * 		       parameters.
+ *
+ * @return 0 in case of success, negative error code otherwise.
+*******************************************************************************/
+int adxl367_init(struct adxl367_dev **device,
+		 struct adxl367_init_param init_param)
+{
+	struct adxl367_dev *dev;
+	uint8_t reg_value;
+	int status;
+
+	dev = (struct adxl367_dev *)calloc(1, sizeof(*dev));
+	if (!dev)
+		return -1;
+
+	dev->comm_type = init_param.comm_type;
+	if (dev->comm_type == ADXL367_SPI_COMM) {
+		/* SPI */
+		status = no_os_spi_init(&dev->spi_desc, &init_param.spi_init);
+		if (status)
+			goto comm_err;
+	} else if (dev->comm_type == ADXL367_I2C_COMM) {
+		/* I2C */
+		status = no_os_i2c_init(&dev->i2c_desc, &init_param.i2c_init);
+		if (status)
+			goto comm_err;
+		dev->i2c_slave_address = init_param.i2c_slave_address;
+	} else {
+		/* Communication type not defined.*/
+		goto comm_err;
+	}
+
+	/* Perform soft reset */
+	status = adxl367_software_reset(dev);
+	if (status)
+		goto err;
+
+	/* Check for device ID */
+	status = adxl367_get_register_value(dev, &reg_value, ADXL367_REG_DEVID_AD, 1);
+	if (status)
+		goto err;
+	if(reg_value != ADXL367_DEVICE_AD)
+		goto err;
+	status = adxl367_get_register_value(dev, &reg_value, ADXL367_REG_DEVID_MST, 1);
+	if (status)
+		goto err;
+	if(reg_value != ADXL367_DEVICE_MST)
+		goto err;
+	status = adxl367_get_register_value(dev, &reg_value, ADXL367_REG_PARTID, 1);
+	if (status)
+		goto err;
+	if(reg_value != ADXL367_PART_ID)
+		goto err;
+
+	*device = dev;
+
+	pr_info("ADXL367 successfully initialized\n");
+	return 0;
+
+err:
+	if (dev->comm_type == ADXL367_SPI_COMM)
+		no_os_spi_remove(dev->spi_desc);
+	else
+		no_os_i2c_remove(dev->i2c_desc);
+comm_err:
+	free(dev);
+	pr_err("%s: Failed initialization.\n", __func__);
+	return -1;
+}
+
+/***************************************************************************//**
+ * @brief Frees the resources allocated by adxl367_init().
+ *
+ * @param dev - The device structure.
+ *
+ * @return 0 in case of success, negative error code otherwise.
+*******************************************************************************/
+int adxl367_remove(struct adxl367_dev *dev)
+{
+	int ret;
+
+	if (dev->comm_type == ADXL367_SPI_COMM)
+		ret = no_os_spi_remove(dev->spi_desc);
+	else
+		ret = no_os_i2c_remove(dev->i2c_desc);
+
+	free(dev);
+
+	return ret;
+}
+
+/***************************************************************************//**
+ * @brief Performs self test.
+ *
+ * @param dev - The device structure.
+ *
+ * @return 0 in case of success, negative error code otherwise.
+*******************************************************************************/
+int adxl367_self_test(struct adxl367_dev *dev)
+{
+	int ret;
+	int16_t x_axis_1, x_axis_2, dif, min, max;
+	uint8_t read_val;
+
+	ret = adxl367_set_power_mode(dev, ADXL367_OP_MEASURE);
+	if (ret)
+		return ret;
+
+	ret = adxl367_reg_write_msk(dev, ADXL367_REG_SELF_TEST, ADXL367_SELF_TEST_ST,
+				    ADXL367_SELF_TEST_ST);
+	if (ret)
+		return ret;
+	// 4 / lowest ODR = 80ms
+	no_os_mdelay(80);
+	ret = adxl367_get_register_value(dev, &read_val, ADXL367_REG_XDATA_H, 1);
+	if (ret)
+		return ret;
+	x_axis_1 = read_val << 6;
+	ret = adxl367_get_register_value(dev, &read_val, ADXL367_REG_XDATA_L, 1);
+	if (ret)
+		return ret;
+	x_axis_1 += read_val >> 2;
+	// extend sign to 16 bits
+	if (x_axis_1 & NO_OS_BIT(13))
+		x_axis_1 |= NO_OS_GENMASK(15, 14);
+
+	ret = adxl367_reg_write_msk(dev, ADXL367_REG_SELF_TEST,
+				    ADXL367_SELF_TEST_ST_FORCE, ADXL367_SELF_TEST_ST_FORCE);
+	if (ret)
+		return ret;
+	// 4 / lowest ODR = 80ms
+	no_os_mdelay(80);
+	ret = adxl367_get_register_value(dev, &read_val, ADXL367_REG_XDATA_H, 1);
+	if (ret)
+		return ret;
+	x_axis_2 = read_val << 6;
+	ret = adxl367_get_register_value(dev, &read_val, ADXL367_REG_XDATA_L, 1);
+	if (ret)
+		return ret;
+	x_axis_2 += read_val >> 2;
+	// extend sign to 16 bits
+	if (x_axis_1 & NO_OS_BIT(13))
+		x_axis_1 |= NO_OS_GENMASK(15, 14);
+
+	ret = adxl367_set_power_mode(dev, ADXL367_OP_STANDBY);
+	if (ret)
+		return ret;
+
+	ret = adxl367_reg_write_msk(dev, ADXL367_REG_SELF_TEST, 0,
+				    ADXL367_SELF_TEST_ST_FORCE | ADXL367_SELF_TEST_ST);
+	if (ret)
+		return ret;
+
+	dif = x_axis_2 - x_axis_1;
+	min = ADXL367_SELF_TEST_MIN * adxl367_scale_mul[dev->range];
+	max = ADXL367_SELF_TEST_MAX * adxl367_scale_mul[dev->range];
+
+	if ((dif >= min) && (dif <= max)) {
+		pr_info("ADXL367 passed self-test\n");
+		return 0;
+	} else {
+		pr_err("%s: Failed self-test.\n", __func__);
+		return -1;
+	}
+}
+
+/***************************************************************************//**
+ * @brief Writes data into a register.
+ *
+ * @param dev              - The device structure.
+ * @param register_value   - Data value to write.
+ * @param register_address - Address of the register.
+ *
+ * @return 0 in case of success, negative error code otherwise.
+*******************************************************************************/
+int adxl367_set_register_value(struct adxl367_dev *dev,
+			       uint8_t register_value,
+			       uint8_t register_address)
+{
+	uint8_t buffer[3] = {0};
+	buffer[1] = register_address;
+	buffer[2] = register_value;
+
+	if (dev->comm_type == ADXL367_SPI_COMM) {
+		buffer[0] = ADXL367_WRITE_REG;
+		return no_os_spi_write_and_read(dev->spi_desc,
+						buffer, 3);
+	} else {
+		buffer[0] = dev->i2c_slave_address + ADXL367_I2C_WRITE;
+		return no_os_i2c_write(dev->i2c_desc, buffer, 3, 1);
+	}
+
+}
+
+/***************************************************************************//**
+ * @brief Performs a burst read of a specified number of registers.
+ *
+ * @param dev              - The device structure.
+ * @param read_data        - The read values are stored in this buffer.
+ * @param register_address - The start address of the burst read.
+ * @param register_nb	   - Number of registers to be read in burst.
+ *
+ * @return 0 in case of success, negative error code otherwise.
+*******************************************************************************/
+int adxl367_get_register_value(struct adxl367_dev *dev,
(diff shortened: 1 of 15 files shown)


rbolboac:
Since this is a new driver & project, would it be a good idea to implement the new project structure following the example of eval-adxl355-pmdz project?

drivers/accel/adxl367/adxl367.c
```diff
@@ -0,0 +1,1310 @@
...
+				   uint8_t register_value,
+				   uint8_t register_address)
+{
+	uint8_t buffer[3] = {0, 0, 0};
+
+	if (dev->comm_type == ADXL367_SPI_COMM) {
+		buffer[0] = ADXL367_WRITE_REG;
+		buffer[1] = register_address;
+		buffer[2] = register_value;
+		return no_os_spi_write_and_read(dev->spi_desc,
```
amiclaus, line 249: nitpick: this function can be condensed if you move `buffer[1]` and `buffer[2]` above the if..else statement.

drivers/accel/adxl367/adxl367.c
```diff
@@ -0,0 +1,1310 @@
...
+*******************************************************************************/
+int32_t adxl367_get_register_value(struct adxl367_dev *dev,
+				   uint8_t* read_data,
+				   uint8_t  register_address,
+				   uint8_t  register_nb)
+{
+	int32_t ret;
+	uint8_t buffer[0x45];
+	uint8_t index;
+
```
amiclaus, line 278: nit: maybe an initial check to see if register_nb is <= 0x45 would fit here. Also, why not using the already implemented macro corresponding to the latest register? `ADXL367_REG_STATUS_2`. Declaring buffer sized in hexadecimal also looks a bit unusual to me.

drivers/accel/adxl367/adxl367.c
```diff
@@ -0,0 +1,1310 @@
...
+	memset(buffer, 0, 0x45);
+
+	if (dev->comm_type == ADXL367_SPI_COMM) {
+		buffer[0] = ADXL367_READ_REG;
+		buffer[1] = register_address;
```
amiclaus, line 283: ditto: buffer[1] can be moved outside the if else statement

drivers/accel/adxl367/adxl367.c
```diff
@@ -0,0 +1,1310 @@
...
+ * @param mask     - The mask.
+ *
+ * @return 0 in case of success, negative error code otherwise.
+*******************************************************************************/
+int32_t adxl367_reg_write_msk(struct adxl367_dev *dev,
+			      uint8_t reg_addr,
+			      uint8_t data,
+			      uint8_t mask)
+{
+	int32_t ret;
```
amiclaus, line 363: nit: `int32_t` -> `int`. Check throughout the entire code.

drivers/accel/adxl367/adxl367.c
```diff
@@ -0,0 +1,1310 @@
...
+	// ODR: 100Hz (reset default).
+	dev->odr = ADXL367_ODR_100HZ;
+	// FIFO: FIFO disabled (reset default).
+	dev->fifo_mode = ADXL367_FIFO_DISABLED;
+	// FIFO Format: All axis (reset default).
+	dev->fifo_format = ADXL367_FIFO_FORMAT_XYZ;
+	// FIFO Read Mode : 14 bits + CH ID (reset default).
+	dev->fifo_read_mode = ADXL367_14B_CHID;
+
+	no_os_mdelay(20);
```
amiclaus, line 403: Add a comment please regarding the delay value. Is it from datasheet?

drivers/accel/adxl367/adxl367.c
```diff
@@ -0,0 +1,1310 @@
...
+ * @return 0 in case of success, negative error code otherwise.
+*******************************************************************************/
+static int32_t adxl367_temp_conv(struct adxl367_dev *dev,
+				 uint16_t raw_temp, struct adxl367_fractional_val* temp)
+{
+	int64_t temp_data;
+
+	temp_data = (int16_t)raw_temp;
+
+	temp_data += ADXL367_TEMP_OFFSET;
```
amiclaus, line 730: can condense to: `temp_data = (temp_data + ADXL367_TEMP_OFFSET) * ADXL367_TEMP_SCALE`

drivers/accel/adxl367/adxl367.c
```diff
@@ -0,0 +1,1310 @@
...
+	// allocate buffer for local storage, each entry has 2 bytes
+	uint8_t buffer[entries_nb * 2], id;
+	memset(buffer, 0, entries_nb * 2);
+
+	//wait for FIFO to have at least sets_nb * samples_per_set entries
+	while (stored_entr < entries_nb) {
+		ret = adxl367_get_nb_of_fifo_entries(dev, &stored_entr);
+		if (ret)
+			return ret;
+		no_os_mdelay(5);
```
amiclaus, line 1050: why 5ms?

drivers/accel/adxl367/adxl367.c
```diff
@@ -0,0 +1,1310 @@
...
+
+	ret = adxl367_get_fifo_value(dev, buffer, entries_nb * 2);
+	if (ret)
+		return -1;
+
+	// MSB = 6 data bits + 2 bits for CH ID
+	for (i = 0; i < (entries_nb * 2); i += 2 ) {
+		id = buffer[i] >> 6;
+
+		if (id == ADXL367_FIFO_X_ID) {
```
amiclaus, line 1061: maybe `switch...case`?

drivers/accel/adxl367/adxl367.c
```diff
@@ -0,0 +1,1310 @@
...
+ * @return 0 in case of success, negative error code otherwise.
+*******************************************************************************/
+int32_t adxl367_setup_inactivity_detection(struct adxl367_dev *dev,
+		uint8_t  ref_or_abs,
+		uint16_t threshold,
+		uint16_t  time)
+{
+	int32_t ret;
+
+	if (!ref_or_abs) {
```
amiclaus, line 1277: nit: maybe switch...case + afferent macros

drivers/accel/adxl367/adxl367.h
```diff
@@ -0,0 +1,589 @@
...
+#define ADXL367_STATUS_DATA_RDY         NO_OS_BIT(0)
+
+/* ADXL367_REG_THRESH_H mask */
+#define ADXL367_THRESH_H		0x7F
+
+/* ADXL367_REG_THRESH_L mask */
+#define ADXL367_THRESH_L		0xFC
+
+/* ADXL367_REG_ACT_INACT_CTL definitions */
+#define ADXL367_ACT_INACT_CTL_LINKLOOP(x)	(((x) & 0x3) << 4)
```
amiclaus, line 144: use NO_OS_GENMASK

drivers/accel/adxl367/adxl367.h
```diff
@@ -0,0 +1,589 @@
...
+#define ADXL367_REFERENCED_INACTIVITY_ENABLE		0x3
+
+/* ADXL367_ACT_INACT_CTL_ACT_EN(x) options */
+#define ADXL367_NO_ACTIVITY_DETECTION		0x0
+#define ADXL367_ACTIVITY_ENABLE			0x1
+#define ADXL367_NO_ACTIVITY_DETECTION_2		0x2
+#define ADXL367_REFERENCED_ACTIVITY_ENABLE	0x3
+
+/* ADXL367_REG_FIFO_CONTROL */
+#define ADXL367_FIFO_CONTROL_FIFO_CHANNEL(x)	(((x) & 0xF) << 3)
```
amiclaus, line 161: NO_OS_GENMASK

drivers/accel/adxl367/adxl367.h
```diff
@@ -0,0 +1,589 @@
...
+#define ADXL367_FIFO_CONTROL_FIFO_SAMPLES	NO_OS_BIT(2)
+#define ADXL367_FIFO_CONTROL_FIFO_MODE(x)	(((x) & 0x3) << 0)
```
amiclaus, line 163: NO_OS_GENMASK

drivers/accel/adxl367/adxl367.h
```diff
@@ -0,0 +1,589 @@
...
+#define ADXL367_SELF_TEST_ST            NO_OS_BIT(0)
+
+/* XYZ_AXIS_OFFSET MASK */
+#define ADXL367_XYZ_AXIS_OFFSET_MASK	0x1F
+
+/* ADXL367_REG_INTMAPX_UPPER MASK */
+#define ADXL367_INTMAPX_UPPER_MASK	0xDF
+
+/* ADXL367_REG_ADC_CTL definitions. */
+#define ADXL367_FIFO_8_12BIT(x)			(((x) & 0x3) << 6)
```
amiclaus, line 241: ditto

drivers/accel/adxl367/adxl367.h
```diff
@@ -0,0 +1,589 @@
...
+#define ADXL367_FIFO_Z_ID		0x02
+#define ADXL367_FIFO_TEMP_ADC_ID	0x03
+
+/*
+ * At +/- 2g with 14-bit resolution, scale is given in datasheet as
+ * 250ug/LSB = 0.0002500 * 9.80665 = 0.0024516625 m/s^2.
+ * For +/- 4g range a multiplier with value 2 is used.
+ * For +/-8g range, a multiplier with value 4 is used.
+ */
+#define ADXL367_ACC_SCALE_FACTOR_MUL  (int64_t)    245166
```
amiclaus, line 271: nit: `245166ULL`

drivers/accel/adxl367/adxl367.h
```diff
@@ -0,0 +1,589 @@
...
+#define ADXL367_ACC_SCALE_FACTOR_DIV  (int32_t)1000000000
```
amiclaus, line 272: is this cast required?

amiclaus (commented):
lgtm overall, some minor comments on my side.

drivers/accel/adxl367/adxl367.c
```diff
@@ -0,0 +1,1310 @@
...
+			nready = 0U;
+	}
+
+	ret = adxl367_get_register_value(dev,
+					 xyz_values,
+					 ADXL367_REG_XDATA_H,
+					 6);
+
+	// result is 14 bits long, ignore last 2 bits from low byte
+	*x = ((int16_t)xyz_values[0] << 6) + (xyz_values[1] >> 2);
```
amiclaus, line 561: wondering if all this computing can't be done with GENMASK + FIELD_PREP/FIELD_GET? Just a question

drivers/accel/adxl367/adxl367.c
```diff
@@ -0,0 +1,1310 @@
...
+		return ret;
+
+	ret = adxl367_reg_write_msk(dev, ADXL367_REG_SELF_TEST, 0,
+				    ADXL367_SELF_TEST_ST_FORCE | ADXL367_SELF_TEST_ST);
+	if (ret)
+		return ret;
+
+	dif = x_axis_2 - x_axis_1;
+	min = 90 * 100 / 25 * adxl367_scale_mul[dev->range];
+	max = 270 * 100 / 25 * adxl367_scale_mul[dev->range];
```
danmois, line 219: Should he have macros here instead of magic numbers?

drivers/accel/adxl367/adxl367.c
```diff
@@ -0,0 +1,1310 @@
...
+ * @param enable - 1 - ENABLE
+ * 				   2 - DISABLE
+ *
+ * @return 0 in case of success, negative error code otherwise.
+*******************************************************************************/
+int32_t adxl367_temp_read_en(struct adxl367_dev *dev, uint8_t enable)
+{
+	return adxl367_reg_write_msk(dev,
+				     ADXL367_REG_TEMP_CTL,
+				     enable ? ADXL367_TEMP_EN : 0U,
```
danmois, line 645: I was wondering whether having a conditional statement within a function call is appropriate. Doesn't it affect code readability?

drivers/accel/adxl367/adxl367.c
```diff
@@ -0,0 +1,1310 @@
...
+	ret = adxl367_reg_write_msk(dev,
+				    ADXL367_REG_TEMP_CTL,
+				    0U,
+				    ADXL367_TEMP_EN);
+	if (ret)
+		return ret;
+
+	return adxl367_reg_write_msk(dev,
+				     ADXL367_REG_ADC_CTL,
+				     enable ? ADXL367_ADC_EN : 0U,
```
danmois, line 672: Same as my previous comment.

drivers/accel/adxl367/adxl367.c
```diff
@@ -0,0 +1,1310 @@
...
+	}
+
+	ret = adxl367_get_register_value(dev,
+					 raw,
+					 ADXL367_REG_EX_ADC_H,
+					 2);
+	if (ret)
+		return ret;
+
+	*data = ((int16_t)raw[0] << 6) + (raw[1] >> 2);
```
danmois, line 792: Would it be an idea to consider  using a function for this operation, since it is used in several places?

drivers/accel/adxl367/adxl367.c
```diff
@@ -0,0 +1,1310 @@
...
+ * @param register_value   - Data value to write.
+ * @param register_address - Address of the register.
+ *
+ * @return 0 in case of success, negative error code otherwise.
+*******************************************************************************/
+int32_t adxl367_set_register_value(struct adxl367_dev *dev,
+				   uint8_t register_value,
+				   uint8_t register_address)
+{
+	uint8_t buffer[3] = {0, 0, 0};
```
CiprianRegus, line 243: You could just do `uint8_t buffer[3] = {0}` and the whole array would be initialized to 0. This way, you also avoid using memset.

adrimbarean (commented):
Solid driver and example project. A few comments before merging.

drivers/accel/adxl367/adxl367.c
APorumb: Datasheet specifies the default turn on time ~9ms. With 9ms sometimes the chip won't properly start-up so I was told to use 15-20ms delay. (on EZ)

drivers/accel/adxl367/adxl367.c
APorumb: At 50Hz odr, 400 samples will take 8 seconds. That delay will slow down the polling. I can remove it but I think it has a good reason to be there, taking the long conversion time in consideration.

drivers/accel/adxl367/adxl367.c
```diff
@@ -0,0 +1,1310 @@
...
+/***************************************************************************//**
+ * @brief Performs a burst read of FIFO buffer.
+ *
+ * @param dev        - The device structure.
+ * @param read_data  - The read values are stored in this buffer.
+ * @param bytes_nb	 - Number of bytes to be read in burst.
+ *
+ * @return 0 in case of success, negative error code otherwise.
+*******************************************************************************/
+int32_t adxl367_get_fifo_value(struct adxl367_dev *dev,
```
rbolboac, line 314: I would make this function static. If the user uses this function with an invalid bytes_nb he will receive invalid data.

drivers/accel/adxl367/adxl367.c
```diff
@@ -0,0 +1,1310 @@
...
+ * @param x  	    - X axis raw data buffer. If not used, assign NULL.
+ * @param y  	    - Y axis raw data buffer. If not used, assign NULL.
+ * @param z  	    - Z axis raw data buffer. If not used, assign NULL.
+ * @param temp_adc  - Data buffer for temperature or adc raw data. If not
+ * 				used assign NULL.
+ * @param sets_nb   - Number of sample sets to be read.
+ *
+ * @return 0 in case of success, negative error code otherwise.
+*******************************************************************************/
+int32_t adxl367_read_raw_fifo(struct adxl367_dev *dev, int16_t *x, int16_t *y,
```
rbolboac, line 1034: Just an idea here. I would implement this function differently. The hardware FIFO is used to store data after the measurement, without losing any data by making single reads for example. With that in mind, I would expect the user to be more interested in all data in the FIFO rather than a specific number of samples. Using FIFO_ENTRIES you can find out how many data sets are stored in the buffer and return the data together with the fifo entries number such that the user is aware how many data sets are valid from a maximum of 512. 
Using this procedure, you won't have to wait for the data to be available. The user can manage that based on the selected sampling frequency. 

I would not use the delay at all in this case.

drivers/accel/adxl367/adxl367.c
```diff
@@ -0,0 +1,1310 @@
...
+	int32_t ret;
+	uint16_t entries_nb, i, stored_entr = 0;
+	entries_nb = sets_nb * samples_per_set;
+
+	// allocate buffer for local storage, each entry has 2 bytes
+	uint8_t buffer[entries_nb * 2], id;
+	memset(buffer, 0, entries_nb * 2);
+
+	//wait for FIFO to have at least sets_nb * samples_per_set entries
+	while (stored_entr < entries_nb) {
```
rbolboac, line 1046: I would check sets_nb value before. If the value is invalid, it will enter an infinite loop here.

drivers/accel/adxl367/adxl367.c
APorumb: I think the code is easier to read this way.

drivers/accel/adxl367/adxl367.c
APorumb: Ternary operator has been used a lot in other projects, I don't see a reason why not to use it here. I think the code remains consistent this way.

drivers/accel/adxl367/adxl367.c
```diff
@@ -0,0 +1,1310 @@
...
+ *
+ * @return 0 in case of success, negative error code otherwise.
+*******************************************************************************/
+int32_t adxl367_get_register_value(struct adxl367_dev *dev,
+				   uint8_t* read_data,
+				   uint8_t  register_address,
+				   uint8_t  register_nb)
+{
+	int32_t ret;
+	uint8_t buffer[0x45];
```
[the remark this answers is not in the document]
APorumb: Even if the last register address is 0x45, in total there aren't 0x45 registers. 

For example : 
image

drivers/accel/adxl367/adxl367.c
```diff
@@ -0,0 +1,1310 @@
...
+	if (dev->comm_type == ADXL367_SPI_COMM) {
+		buffer[0] = ADXL367_READ_REG;
+		buffer[1] = register_address;
+		ret = no_os_spi_write_and_read(dev->spi_desc, buffer, register_nb + 2);
+		if (ret)
+			return ret;
+		for (index = 0; index < register_nb; index++)
+			read_data[index] = buffer[index + 2];
+	} else {
+		buffer[0] = dev->i2c_slave_address + ADXL367_I2C_READ;
```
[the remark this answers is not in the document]
APorumb: ADXL367 has predefined slave addresses. 
Depending on ASEL pin, it can be 0x53 or 0x1D.

drivers/accel/adxl367/adxl367.c
```diff
@@ -0,0 +1,1310 @@
...
+*******************************************************************************/
+int32_t adxl367_read_raw_fifo(struct adxl367_dev *dev, int16_t *x, int16_t *y,
+			      int16_t *z, int16_t *temp_adc, uint16_t sets_nb)
+{
+	int32_t ret;
+	uint16_t entries_nb, i, stored_entr = 0;
+	entries_nb = sets_nb * samples_per_set;
+
+	// allocate buffer for local storage, each entry has 2 bytes
+	uint8_t buffer[entries_nb * 2], id;
```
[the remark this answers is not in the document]
APorumb: Buffer is used only locally.

APorumb:
Edit 1:
-project restructured to new template
-driver modified as suggested

projects/eval-adxl367z/src/common/common_data.c
```diff
@@ -0,0 +1,61 @@
...
+/******************************************************************************/
+/********************** Macros and Constants Definitions **********************/
+/******************************************************************************/
+const struct no_os_spi_init_param spi_ip = {
+	.device_id = SPI_DEVICE_ID,
+	.max_speed_hz = 100000,
+	.mode = NO_OS_SPI_MODE_0,
+	.chip_select = 0U,
+	.bit_order = NO_OS_SPI_BIT_ORDER_MSB_FIRST,
+	.platform_ops = &xil_spi_ops,
```
rbolboac, line 54: should be platform independent. you could define a macro
#define SPI_OPS &xil_spi_ops in parameters.h for XILINX platform.

rbolboac:
I think you should also have a build.json file for xilinx.

APorumb:
Edit 2:
-removed some extra lines
-defined SPI_OPS as &xil_platform_ops and moved it to platform.h
-.json added

projects/eval-adxl367z/builds.json
```diff
@@ -0,0 +1,10 @@
+{
+      "xilinx": {
+            "demo": {
+                  "flags": "DUMMY_EXAMPLE",
```
amiclaus, line 4: `"flags" : "NEW_CFLAGS=-DDUMMY_EXAMPLE",`
otherwise, it won't work :)

rbolboac: This is what I used for stm32 platform:
```
{
  "stm32": {
    "dummy_example": {
      "flags" : "DUMMY_EXAMPLE=y IIO_EXAMPLE=n"
    },
    "iio_example": {
      "flags" : "DUMMY_EXAMPLE=n IIO_EXAMPLE=y"
    }
  }
}
```
I would suggest using something like this (seeing how we are using y or n values):

```
{
  "xilinx": {
    "dummy_example": {
      "flags" : "DUMMY_EXAMPLE=y",
     ...
    }
}
```

amiclaus: I guess both approaches are valid:
For example, cn0561 (recent project merged into the master branch):
https://github.com/analogdevicesinc/no-OS/blob/master/projects/cn0561/builds.json

drivers/accel/adxl367/adxl367.c
```diff
@@ -0,0 +1,1326 @@
...
+*******************************************************************************/
+
+/******************************************************************************/
+/***************************** Include Files **********************************/
+/******************************************************************************/
+#include <stdlib.h>
+#include "adxl367.h"
+#include "no_os_delay.h"
+#include "no_os_util.h"
+#include "no_os_print_log.h"
```
rbolboac, line 47: I know that this library is used to show useful information but I am wondering if it is a good idea to make the driver dependent also on this library.

drivers/accel/adxl367/adxl367.c
```diff
@@ -0,0 +1,1326 @@
...
+ * @return 0 in case of success, negative error code otherwise.
+*******************************************************************************/
+int adxl367_init(struct adxl367_dev **device,
+		 struct adxl367_init_param init_param)
+{
+	struct adxl367_dev *dev;
+	uint8_t reg_value;
+	int status;
+
+	dev = (struct adxl367_dev *)malloc(sizeof(*dev));
```
rbolboac, line 76: I think the usage of calloc is preferred.

drivers/accel/adxl367/adxl367.c
rbolboac: I think a comment in the code about the delay usage would be useful.

drivers/accel/adxl367/adxl367.c
```diff
@@ -0,0 +1,1326 @@
...
+ *
+ * @return 0 in case of success, negative error code otherwise.
+*******************************************************************************/
+int adxl367_read_converted_fifo(struct adxl367_dev *dev,
+				struct adxl367_fractional_val *x, struct adxl367_fractional_val *y,
+				struct adxl367_fractional_val *z, struct adxl367_fractional_val *temp_adc,
+				uint16_t *entries)
+{
+	int ret;
+	int16_t x_raw[513], y_raw[513], z_raw[513], temp_adc_raw[513];
```
rbolboac, line 1126: I am wondering if this can be avoided. For example, in adxl355 driver, the buffer allocation is done at init time, in adxl355_dev structure, and then the same buffer is used for managing fifo reads.

APorumb:
Edit 3:
-switched from malloc to calloc
-added some comments
-added fifo_buffer to dev structure and use it withing fifo reading function

APorumb:
Edit 4:
-adxl367_get_fifo_value now uses dev->fifo_buffer

rbolboac (approved):
lgtm

PR #1326, opened 2022-04-11
axi_core: add function documentation

amiclaus:
add function documentation for the following drivers:
- axi_adxcvr
- axi_jesd204_tx
- axi_jesd204_rx

No functional changes.

### DIFF
--- a/drivers/axi_core/jesd204/axi_adxcvr.c
+++ b/drivers/axi_core/jesd204/axi_adxcvr.c
@@ -87,7 +87,11 @@
 #define ADXCVR_BROADCAST				0xff
 
 /**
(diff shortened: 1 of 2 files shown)


amiclaus (commented):
> lgtm.
> 
> Is it worth to have a discussion about documentation structure? I know some developers use /** and some prefer using /********************************************//** for function headers. Should we all use the same format?

I agree on using the same format. The approach in this PR was used when doxygen support was added to the no-OS repository.

A solution would be to add here: https://github.com/analogdevicesinc/no-OS/wiki a page on how the code should be documeted. Something similar to: https://www.doxygen.nl/manual/docblocks.html but particular for this repo.

PR #1344, opened 2022-04-19
ADXL367 iio support + minor driver update

APorumb:
-adxl367 iio support added
-adxl367 minor driver update

### DIFF
--- a/drivers/accel/adxl367/adxl367.c
+++ b/drivers/accel/adxl367/adxl367.c
@@ -394,6 +394,10 @@ int adxl367_software_reset(struct adxl367_dev *dev)
 	dev->fifo_format = ADXL367_FIFO_FORMAT_XYZ;
 	// FIFO Read Mode : 14 bits + CH ID (reset default).
 	dev->fifo_read_mode = ADXL367_14B_CHID;
+	//Axis offset = 0 (reset default)
+	dev->x_offset = 0;
+	dev->y_offset = 0;
+	dev->z_offset = 0;
 
 	//initialization delay
 	no_os_mdelay(20);
@@ -511,14 +515,23 @@ int adxl367_set_offset(struct adxl367_dev *dev, uint16_t x_offset,
 				    ADXL367_XYZ_AXIS_OFFSET_MASK);
 	if (ret)
 		return ret;
+	dev->x_offset = x_offset;
+
 	ret = adxl367_reg_write_msk(dev, ADXL367_REG_Y_OFFSET,
 				    y_offset & ADXL367_XYZ_AXIS_OFFSET_MASK,
 				    ADXL367_XYZ_AXIS_OFFSET_MASK);
 	if (ret)
 		return ret;
-	return adxl367_reg_write_msk(dev, ADXL367_REG_Z_OFFSET,
-				     z_offset & ADXL367_XYZ_AXIS_OFFSET_MASK,
-				     ADXL367_XYZ_AXIS_OFFSET_MASK);
+	dev->y_offset = y_offset;
+
+	ret = adxl367_reg_write_msk(dev, ADXL367_REG_Z_OFFSET,
+				    z_offset & ADXL367_XYZ_AXIS_OFFSET_MASK,
+				    ADXL367_XYZ_AXIS_OFFSET_MASK);
+	if (ret)
+		return ret;
+	dev->z_offset = z_offset;
+
+	return 0;
 }
 
 /***************************************************************************//**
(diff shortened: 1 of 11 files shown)


drivers/accel/adxl367/adxl367.c
```diff
@@ -511,14 +515,23 @@ int adxl367_set_offset(struct adxl367_dev *dev, uint16_t x_offset,
 				    ADXL367_XYZ_AXIS_OFFSET_MASK);
 	if (ret)
 		return ret;
+	dev->x_offset = x_offset;
```
buha, line 518: updating of this field should happen after the reg write, only if reg write succeeded (same for y, z)

drivers/accel/adxl367/iio_adxl367.c
```diff
@@ -0,0 +1,936 @@
...
+ * @param channel - Command channel info.
+ * @param priv    - Command attribute id.
+ *
+ * @return ret    - Result of the reading procedure.
+ * 					In case of success, the size of the read data is returned.
+*******************************************************************************/
+static int adxl367_iio_read_offset(void *dev, char *buf, uint32_t len,
+				   const struct iio_ch_info *channel, intptr_t priv)
+{
+	int32_t vals[2];
```
rbolboac, line 378: it does not look like you need an array here seeing how you are using only vals[0]

drivers/accel/adxl367/iio_adxl367.c
```diff
@@ -0,0 +1,936 @@
...
+
+	return 0;
+
+error_adxl367_init:
+	free(desc);
+	return ret;
+error_config:
+	adxl367_remove(desc->adxl367_dev);
+	free(desc);
+	return ret;
```
rbolboac, line 915: it doesn't look like this label is used. can it be removed?

APorumb: If one function fails, ret is used. 

ret = adxl367_init(&desc->adxl367_dev, *(init_param->adxl367_initial_param));
	if (ret)
		goto error_adxl367_init;

error_adxl367_init:
	free(desc);
	return ret;

APorumb:
Edit 1:
-edited buffer from iio_example.c
-corrected label from iio_adxl367.c iio_init function
-corrected makefile, as @rbolboac suggested

APorumb:
Edit 2:
-dummy example label corrected

drivers/accel/adxl367/iio_adxl367.c
```diff
@@ -0,0 +1,936 @@
+/***************************************************************************//**
+ *   @file   iio_adxl367.c
+ *   @brief  Implementation of IIO ADXL367 Driver.
```
buha, line 3: nitpick, phrasing: _ADXL367 IIO Driver_
or _IIO part of the ADXL367 driver_

drivers/accel/adxl367/iio_adxl367.c
```diff
@@ -0,0 +1,936 @@
...
+static const int adxl367_iio_scale_table[3][2] = {
+	{0, 2394493},
+	{0, 4788987},
+	{0, 9577975}
+};
+
+/******************************************************************************/
+/************************ Functions Declarations ******************************/
+/******************************************************************************/
+static int adxl367_iio_read_reg(struct adxl367_iio_dev *dev, uint32_t reg,
```
buha, line 71: I think you can get rid of most (if not all) these declarations by simply moving the functions around in the file so that they are defined and declared before calling them, akin how it's done on linux

drivers/accel/adxl367/iio_adxl367.c
```diff
@@ -0,0 +1,936 @@
...
+	.is_big_endian = false
+};
+
+#define ADXL367_ACCEL_CHANNEL(index, axis) { 	  \
+	.name = axis,								  \
+	.ch_type = IIO_ACCEL,                         \
+	.channel = index,                             \
+	.scan_type = &adxl367_iio_accel_scan_type,    \
+	.scan_index = index,                          \
+	.attributes = adxl367_iio_accel_attrs,        \
```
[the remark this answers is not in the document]
buha: good point, also question for @APorumb , have you done an IIO xml comparison between linux and no-OS IIO drivers ? That comparison should point you differences like @rbolboac just mentioned, and you can easily fix them like that. (xml's should be identical)

APorumb:
Edit 3:
-removed function declarations
-matched xml with linux driver
-implemented other minor remarks from above

amiclaus (approved):
lgtm, some minor stuff (commit history related).

PR #1347, opened 2022-04-21
Input parameter conditional checks always true

RaymondWeber:
Super minor typos where some conditional statements are always true.

Fixes:
- On the AD9081, it will mask bits the LSB's and will set a new value, even if the input is out of range.
- On the AD9467, the jitter mux bit cannot be set and will always return API_CMS_ERROR_INVALID_PARAM.


buha (approved):

rbolboac (approved):

amiclaus (commented):
Please add `Signed-off by:` tags to your commits.

amiclaus (commented):
Also your commit descriptions should fit the 50/71 commit message rule.

RaymondWeber:
Thanks, I think that should fix my commit comment issues....

amiclaus (approved):

PR #1348, opened 2022-04-21
drivers:filter: add admv8818 support

amiclaus:
The ADMV8818-EP is a fully monolithic microwave integrated
circuit (MMIC) that features a digitally selectable frequency of
operation. The device features four independently controlled high-
pass filters (HPFs) and four independently controlled low-pass
filters (LPFs) that span the 2 GHz to 18 GHz frequency range.

Datasheet: https://www.analog.com/en/products/admv8818.html

### DIFF
--- a/drivers/filter/admv8818/admv8818.c
+++ b/drivers/filter/admv8818/admv8818.c
@@ -1,0 +1,427 @@
+/***************************************************************************//**
+ *   @file   admv8818.c
+ *   @brief  Implementation of admv8818 Driver.
+ *   @author Antoniu Miclaus (antoniu.miclaus@analog.com)
+********************************************************************************
+ * Copyright 2022(c) Analog Devices, Inc.
+ *
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *  - Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ *  - Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in
+ *    the documentation and/or other materials provided with the
+ *    distribution.
+ *  - Neither the name of Analog Devices, Inc. nor the names of its
+ *    contributors may be used to endorse or promote products derived
+ *    from this software without specific prior written permission.
+ *  - The use of this software may or may not infringe the patent rights
+ *    of one or more patent holders.  This license does not release you
+ *    from the requirement that you obtain separate licenses from these
+ *    patent holders to use this software.
+ *  - Use of the software either in source or binary form, must be run
+ *    on or directly connected to an Analog Devices Inc. component.
+ *
+ * THIS SOFTWARE IS PROVIDED BY ANALOG DEVICES "AS IS" AND ANY EXPRESS OR
+ * IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, NON-INFRINGEMENT,
+ * MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED.
+ * IN NO EVENT SHALL ANALOG DEVICES BE LIABLE FOR ANY DIRECT, INDIRECT,
+ * INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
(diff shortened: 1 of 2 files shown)


drivers/filter/admv8818/admv8818.c
```diff
@@ -0,0 +1,423 @@
...
+		*freq= 0;
+		return ret;
+	}
+
+	ret = admv8818_spi_read(dev, ADMV8818_REG_WR0_FILTER, &data);
+	if (ret)
+		return ret;
+
+	lpf_state = no_os_field_get(ADMV8818_LPF_WR0_MSK, data);
+
```
rbolboac, line 308: nitpick: would it be useful to add a comment here which explains these operations?

drivers/filter/admv8818/admv8818.c
```diff
@@ -0,0 +1,423 @@
...
+	ret = admv8818_spi_read(dev, ADMV8818_REG_WR0_FILTER, &data);
+	if (ret)
+		return ret;
+
+	hpf_state = no_os_field_get(ADMV8818_HPF_WR0_MSK, data);
+
+	*freq = (freq_range_hpf[hpf_band - 1][1] - freq_range_hpf[hpf_band - 1][0]) /
+		15;
+	*freq = freq_range_hpf[hpf_band - 1][0] + (*freq * hpf_state);
```
rbolboac, line 226: nitpick: would it be useful to add a comment here which explains these operations?

drivers/filter/admv8818/admv8818.c
```diff
@@ -0,0 +1,423 @@
...
+
+			for (j = 0; j <= 15; j++) {
+				if (freq < (freq_range_lpf[i][0] + (freq_step * j))) {
+					lpf_step = j;
+					break;
+				}
+			}
+			break;
+		}
+	}
```
rbolboac, line 265: nitpick: would it be useful to add a comment here which explains these operations?

drivers/filter/admv8818/admv8818.c
```diff
@@ -0,0 +1,423 @@
...
+
+			for (j = 1; j <= 16; j++) {
+				if (freq < (freq_range_hpf[i][0] + (freq_step * j))) {
+					hpf_step = j - 1;
+					break;
+				}
+			}
+			break;
+		}
+	}
```
rbolboac, line 174: nitpick: would it be useful to add a comment here which explains these operations?

drivers/filter/admv8818/admv8818.h
```diff
@@ -0,0 +1,184 @@
...
+/******************************************************************************/
+/********************** Macros and Constants Definitions **********************/
+/******************************************************************************/
+/* ADMV8818 Register Map */
+#define ADMV8818_REG_SPI_CONFIG_A		0x0
+#define ADMV8818_REG_SPI_CONFIG_B		0x1
+#define ADMV8818_REG_CHIPTYPE			0x3
+#define ADMV8818_REG_PRODUCT_ID_L		0x4
+#define ADMV8818_REG_PRODUCT_ID_H		0x5
+#define ADMV8818_REG_FAST_LATCH_POINTER		0x10
```
rbolboac, line 60: nitpick: not sure if here (see throughout the whole file) is an alignment issue or just how github displays it. It would be nice to have the macros aligned if github is used to view the file.

rbolboac (commented):
overall lgtm, I have just a few small remarks which might help code readability.

drivers/filter/admv8818/admv8818.h
amiclaus: You can check if it is indeed an issue by viewing the file in raw format. In this PR view, github usually messes up the alignments.
I checked the Raw file and the alignments look fine.

drivers/filter/admv8818/admv8818.c
amiclaus: Sure, I can add, but we usually add comments only is special cases, like datasheet not public / particular uncommon operation. In this case, and in most of the other cases the datasheet is bound to the commit description of the driver implementation so that the reviewers can analyze the code based on the documentation available (https://github.com/torvalds/linux/commit/f34fe888ad0546dacf678aa604435d442934984f). And if there are discrepancies between the code and the datasheet documentation, then we should intervene with inline comments.

drivers/filter/admv8818/admv8818.h
rbolboac: Thanks for clearing this up!

amiclaus (commented):
v2:
- add inline comments as requested.

rbolboac (approved):

dbogdan (approved):

PR #1352, opened 2022-04-27
drivers: frequency: admv1013: add support

amiclaus:
The ADMV1013 is a wideband, microwave upconverter optimized
for point to point microwave radio designs operating in the
24 GHz to 44 GHz radio frequency (RF) range.

Datasheet: https://www.analog.com/en/products/admv1013.html

### DIFF
--- a/drivers/frequency/admv1013/admv1013.c
+++ b/drivers/frequency/admv1013/admv1013.c
@@ -1,0 +1,398 @@
+/***************************************************************************//**
+ *   @file   admv1013.c
+ *   @brief  Implementation of admv1013 Driver.
+ *   @author Antoniu Miclaus (antoniu.miclaus@analog.com)
+********************************************************************************
+ * Copyright 2022(c) Analog Devices, Inc.
+ *
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *  - Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ *  - Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in
+ *    the documentation and/or other materials provided with the
+ *    distribution.
+ *  - Neither the name of Analog Devices, Inc. nor the names of its
+ *    contributors may be used to endorse or promote products derived
+ *    from this software without specific prior written permission.
+ *  - The use of this software may or may not infringe the patent rights
+ *    of one or more patent holders.  This license does not release you
+ *    from the requirement that you obtain separate licenses from these
+ *    patent holders to use this software.
+ *  - Use of the software either in source or binary form, must be run
+ *    on or directly connected to an Analog Devices Inc. component.
+ *
+ * THIS SOFTWARE IS PROVIDED BY ANALOG DEVICES "AS IS" AND ANY EXPRESS OR
+ * IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, NON-INFRINGEMENT,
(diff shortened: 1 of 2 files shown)


drivers/frequency/admv1013/admv1013.c
```diff
@@ -0,0 +1,395 @@
...
+ * @return Returns 0 in case of success or negative error code otherwise.
+ */
+static int admv1013_update_mixer_vgate(struct admv1013_dev *dev)
+{
+	unsigned int mixer_vgate;
+
+	if (dev->vcm_uv < 1800000)
+		mixer_vgate = (2389 * dev->vcm_uv / 1000000 + 8100) / 100;
+	else if (dev->vcm_uv > 1800000 && dev->vcm_uv < 2600000)
+		mixer_vgate = (2375 * dev->vcm_uv / 1000000 + 125) / 100;
```
rbolboac, line 141: would it make sense here to use macros instead of magic values?

amiclaus: Well, I can invent some macros, but since the datasheet contains only the "magic" values without any further explanation, I decided to stick to that approach.


image
image

Any suggestions are welcome :)

amiclaus: Same for mixer_vgate:
image

rbolboac: maybe something like LO_BAND_#_#_TO_#_#_GHZ? e.g. with value: LO_BAND_8_62_TO_10_25_GHZ

amiclaus: got it, will add an enum for the quad_filter modes then :)

rbolboac: and for mixer gate, maybe create directly a macro with the formula?
something like:
MIXER_GATE_0_to_1_8_V(x) ((2389 * x/ 1000000 + 8100) / 100)
MIXER_GATE_1_8_to_2_6_V(x) ((2375 * x/ 1000000 + 125) / 100)

amiclaus: Sounds good. Thanks.

amiclaus (commented):
V2:
- add requested macros
- fix if else statement brackets

drivers/frequency/admv1013/admv1013.c
```diff
@@ -0,0 +1,395 @@
...
+	unsigned int filt_raw;
+
+	if ((dev->lo_in >= 5400000000) && (dev->lo_in <= 7000000000))
+		filt_raw = LO_BAND_5_4_TO_7_GHZ;
+	else if ((dev->lo_in >= 5400000000) && (dev->lo_in <= 8000000000))
+		filt_raw = LO_BAND_5_4_TO_8_GHZ;
+	else if ((dev->lo_in >= 6600000000) && (dev->lo_in <= 9200000000))
+		filt_raw = LO_BAND_6_6_TO_9_2_GHZ;
+	else
+		filt_raw = LO_BAND_8_62_TO_10_25_GHZ;
```
rbolboac, line 166: this would execute when dev->lo_in is < 5400000000 or is > 9200000000. dev->lo_in is given in the init param by the user. would it make sense to and another if to check whether dev->lo_in is between 8620000000 and 10250000000 and to add an else which would simply return -1 in case dev->lo_in is not in the expected range?

amiclaus: Agree. The out of spec usage of the driver should be independent on the filter selection. I will add a separate condition prior to filter selection.

amiclaus (commented):
v3:
- add condition to check if LO_IN is out of spec.

rbolboac (approved):

dbogdan (approved):

PR #1358, opened 2022-05-03
drivers:dac:ltc2688: Minor Update

Pratyush-Mallick:
Added an extra parameter in device descriptor to store the computed dac code.

### DIFF
--- a/drivers/dac/ltc2688/ltc2688.c
+++ b/drivers/dac/ltc2688/ltc2688.c
@@ -395,6 +395,8 @@ int32_t ltc2688_set_voltage(struct ltc2688_dev *dev, uint8_t channel,
 	if(code > 0xFFFF)
 		code = 0xFFFF;
 
+	dev->dac_code[channel] = code;
+
 	/* Write to the Data Register of the DAC. */
 	return _ltc2688_spi_write(dev, LTC2688_CMD_CH_CODE_UPDATE(channel), code);
 }
--- a/drivers/dac/ltc2688/ltc2688.c
+++ b/drivers/dac/ltc2688/ltc2688.c
@@ -140,8 +140,16 @@ static int32_t _ltc2688_spi_update_bits(struct ltc2688_dev *dev, uint8_t reg,
  */
 int32_t ltc2688_set_pwr_dac(struct ltc2688_dev *dev, uint16_t setting)
 {
-	return _ltc2688_spi_write(dev, LTC2688_CMD_POWERDOWN_REG,
-				  setting);
+	int32_t ret;
+
+	ret = _ltc2688_spi_write(dev, LTC2688_CMD_POWERDOWN_REG,
+				 setting);
+	if (ret < 0)
+		return ret;
+
+	dev->pwd_dac_setting = setting;
+
+	return 0;
 }
 
 /**
--- a/drivers/dac/ltc2688/ltc2688.h
+++ b/drivers/dac/ltc2688/ltc2688.h
@@ -139,6 +139,7 @@ struct ltc2688_dev {
 	uint16_t			pwd_dac_setting;
 	uint16_t			dither_toggle_en;
 	bool				dither_mode[16];
+	uint16_t			dac_code[16];
 	enum ltc2688_voltage_range 	crt_range[16];
 	enum ltc2688_dither_phase	dither_phase[16];
 	enum ltc2688_dither_period	dither_period[16];


amiclaus (approved):

adrimbarean (approved):

drivers/dac/ltc2688/ltc2688.c
```diff
@@ -387,6 +387,8 @@ int32_t ltc2688_set_voltage(struct ltc2688_dev *dev, uint8_t channel,
 	if(code > 0xFFFF)
 		code = 0xFFFF;
 
+	dev->dac_code[channel] = code;
```
buha, line 390: what is the purpose of this change ? I expect the code to make it somehow to the registers of the chip but the code now simply stores it...

pmallick-adi: To let the user know of what the actual code that is being written into the device register after taking gain and offset into consideration.

Pratyush-Mallick: To let the user know of what the actual code that is being written into the device register after taking gain and offset into consideration.

Pratyush-Mallick:
Changelog:
Storing the dac power settings in the ltc2688_set_pwr_dac() function.

pmallick-adi:
Do the changes look good to you?

drivers/dac/ltc2688/ltc2688.c
buha: It seems to me then that this is a debug feature. I would advise against implementing it this way and suggest using `pr_debug` from this file to achieve what you want.
https://github.com/analogdevicesinc/no-OS/blob/master/include/no_os_print_log.h

This computed code is information that the driver doesn't need to store in the device descriptor (proof: it's not used elsewhere than the function it's computed in: ltc2688_set_voltage).

drivers/dac/ltc2688/ltc2688.c
```diff
@@ -140,8 +140,16 @@ static int32_t _ltc2688_spi_update_bits(struct ltc2688_dev *dev, uint8_t reg,
  */
 int32_t ltc2688_set_pwr_dac(struct ltc2688_dev *dev, uint16_t setting)
 {
-	return _ltc2688_spi_write(dev, LTC2688_CMD_POWERDOWN_REG,
-				  setting);
+	int32_t ret;
```
buha, line 143: this should be in a separate commit, it seems unrelated to the dac code change

drivers/dac/ltc2688/ltc2688.c
pmallick-adi: It's not about just printing the value, it about getting the dac codes back to application level. I believe the device descriptor should provide information about what's the current configuration of DAC registers. In this [drivers ](https://github.com/analogdevicesinc/no-OS/blob/master/drivers/dac/ad5770r/ad5770r.h#L317) as well, the dac value is being stored in the descriptor itself and it is not being used anywhere in the drivers except for setting the dac voltage

If _ltc2688_spi_read() function were not declared static, it would have been easier to read contents of the dac registers directly.
Or else I can create a global variable to store these values in the drivers itself.

drivers/dac/ltc2688/ltc2688.c
pmallick-adi: Will add a separate commit for this.

drivers/dac/ltc2688/ltc2688.c
Pratyush-Mallick: It's not about just printing the value; it's about getting the DAC codes back to the application level. I believe the device descriptor should provide information about the current configuration of DAC registers. In this [drivers ](https://github.com/analogdevicesinc/no-OS/blob/master/drivers/dac/ad5770r/ad5770r.h#L317) as well, the DAC value is being stored in the descriptor itself, and it is not being used anywhere in the drivers except for setting the DAC voltage 

If the _ltc2688_spi_read() function were not declared static, it would have been easier to read the contents of the DAC registers directly. Or else I can create a global variable to store these values in the drivers themselves.

drivers/dac/ltc2688/ltc2688.c
Pratyush-Mallick: Will do it.

drivers/dac/ltc2688/ltc2688.c
buha: Ok, fine for me then if you say this should work as a cached version of the value one would read using _ltc2688_spi_read()

buha (approved):
looks good, waiting for splitting in two commits and will merge then

pmallick-adi:
Can you please merge this ASAP. There is an urgent requirement for these drivers.

PR #1360, opened 2022-05-04
Maxim ARM toolchain path update

CiprianRegus:
Updated the path to the ARM toolchain provided by the
Maxim's SDK because it changed in the latest release.


tools/scripts/maxim.mk
```diff
@@ -3,7 +3,7 @@ $(error MAXIM_LIBRARIES not defined.$(ENDL))
 endif
 
 UNIX_TOOLS_PATH := $(MAXIM_LIBRARIES)/../Tools/MinGW/msys/1.0/bin
-ARM_COMPILER_PATH := $(MAXIM_LIBRARIES)/../Tools/GNUTools/bin
+ARM_COMPILER_PATH := $(MAXIM_LIBRARIES)/../Tools/GNUTools/10.3/bin
```
buha, line 6: This is going to keep changing in the future.
How about you do something like this:
https://github.com/analogdevicesinc/no-OS/blob/master/tools/scripts/stm32.mk#L27-L28

Basically you search for a path with a certain pattern (like *bin/arm-none-eabi-gcc) and then resolve just the /full/path/to/bin using Make libraries functions and avoid having to maintain this file

buha: I'm actually now wondering how this explicit version in makefile would work for users who have different versions installed

CiprianRegus:
- Used `rwildcard` function to search for the compiler path instead of hardcoding it. This will allow to use this makefile for future SDK releases, without breaking the compatibility with older ones.

buha (approved):

PR #1363, opened 2022-05-09
iio:iio_trigger: Add generic trigger APIs

rbolboac:
Add generic trigger APIs for hardware and software triggers which shall be used by all IIO devices with trigger capabilities.

There are two types of triggers which can be instantiated:

1. software trigger: 
- does not require any interrupt mapping
- is generated by writing the attribute of the trigger device
- are not supported over serial communication
Example of software trigger initialization for a device:

```
static struct iio_attribute trig_attr[] = {
	{
		.name = "trigger_now",
		.store = iio_sw_trig_handler
	},
	END_ATTRIBUTES_ARRAY
};

struct iio_trigger sw_trig_desc = {
	.is_synchronous = true,
	.attributes = trig_attr,
};
```
In case of a software trigger, the is_synchronous flag specifies whether the measurement is performed from the context of the software trigger handler (.is_synchronous = true), or from the context of the iio application (.is_synchronous = false). In case is_synchronous is set to false, depending on the frequency of the software trigger, some triggers might be missed and the number of data samples might be less than the number of generated triggers. 
2. hardware trigger:
- requires an interrupt mapping, at the moment no-os supports only GPIO interrupts used as hardware triggers
- is generated when a per-configured GPIO interrupt is taking place.
Example of hardware trigger initialization for a device:
```
struct iio_trigger hw_trig_desc = {
	.is_synchronous = true,
	.enable = iio_trig_enable,
	.disable = iio_trig_disable
};
```
In case of a hardware trigger, the is_synchronous flag specifies whether the measurement is performed from the context of the interrupt handler (.is_synchronous = true), or from the context of the iio application (.is_synchronous = false). In case is_synchronous is set to false, depending on the frequency of the trigger, some triggers might be missed and the number of data samples might be less than the number of generated triggers.

### DIFF
--- a/iio/iio_trigger.c
+++ b/iio/iio_trigger.c
@@ -1,0 +1,238 @@
+/***************************************************************************//**
+ *   @file   iio_trigger.c
+ *   @brief  Implementation of generic iio trigger.
+ *   @author RBolboac (ramona.bolboaca@analog.com)
+********************************************************************************
+ * Copyright 2022(c) Analog Devices, Inc.
+ *
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *  - Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ *  - Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in
+ *    the documentation and/or other materials provided with the
+ *    distribution.
+ *  - Neither the name of Analog Devices, Inc. nor the names of its
+ *    contributors may be used to endorse or promote products derived
+ *    from this software without specific prior written permission.
+ *  - The use of this software may or may not infringe the patent rights
+ *    of one or more patent holders.  This license does not release you
+ *    from the requirement that you obtain separate licenses from these
+ *    patent holders to use this software.
+ *  - Use of the software either in source or binary form, must be run
+ *    on or directly connected to an Analog Devices Inc. component.
+ *
+ * THIS SOFTWARE IS PROVIDED BY ANALOG DEVICES "AS IS" AND ANY EXPRESS OR
+ * IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, NON-INFRINGEMENT,
+ * MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED.
+ * IN NO EVENT SHALL ANALOG DEVICES BE LIABLE FOR ANY DIRECT, INDIRECT,
+ * INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, INTELLECTUAL PROPERTY RIGHTS, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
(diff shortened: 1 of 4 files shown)


iio/iio_trigger.h
```diff
@@ -0,0 +1,153 @@
...
+	const char *name;
+};
+
+/**
+ * @struct iio_sw_trig
+ * @brief IIO software trigger structure
+ */
+struct iio_sw_trig {
+	/** IIO descriptor */
+	struct iio_desc **iio_desc;
```
adrimbarean, line 113: Question: do we need a double pointer for the `iio_desc`? Could it work with only a simple pointer?

adrimbarean (approved):
Looks good to me. Only one question, but I think we can move forward.

iio/iio_trigger.h
rbolboac: a simple pointer will do the job, thanks! I made the updates.

rbolboac: Actually, after executing some tests, a double pointer is needed for iio_desc.

Reason: iio_desc is used as a double pointer in iio_app_run_with_trigs, because the memory allocation is taking place in iio_init function from iio_app.c. Here, a new value is given to *iio_desc, see line 1821 from iio_app `*desc = ldesc`.
The new value, is no longer equal to the value we passed to trigger init functions leading to erroneous behavior.

However, the value of **iio_desc remains constant and can be used safely.

iio/iio_trigger.c
```diff
@@ -0,0 +1,208 @@
...
+ * @return ret       - Result of the initialization procedure.
+ */
+int iio_hw_trig_init(struct iio_hw_trig **iio_trig,
+		     struct iio_hw_trig_init_param *init_param)
+{
+	struct iio_hw_trig *trig_desc;
+	int ret;
+
+	if (!init_param->iio_desc || !init_param->name)
+		return  -EINVAL;
```
buha, line 68: nitpick, 2 whitespaces

iio/iio_trigger.c
```diff
@@ -0,0 +1,208 @@
...
+
+/**
+ * @brief Enable system interrupt which is linked to the given trigger.
+ *
+ * @param trig  - Trigger structure.
+*/
+void iio_trig_enable(void *trig)
+{
+	struct iio_hw_trig *desc = trig;
+	no_os_irq_enable(desc->irq_ctrl, desc->irq_id);
```
buha, line 114: function should return a value

iio/iio_trigger.c
```diff
@@ -0,0 +1,208 @@
...
+}
+
+/**
+ * @brief Disable system interrupt which is linked to the given trigger.
+ *
+ * @param trig  - Trigger structure.
+*/
+void iio_trig_disable(void *trig)
```
buha, line 122: return status code

iio/iio_trigger.c
```diff
@@ -0,0 +1,208 @@
...
+	no_os_irq_disable(desc->irq_ctrl, desc->irq_id);
+}
+
+/**
+ * @brief Trigger interrupt handler. This function will be called when a system
+ * interrupt is asserted for the configured trigger.
+ *
+ * @param trig  - Trigger structure which is linked to this handler.
+*/
+void iio_hw_trig_handler(void *trig)
```
buha, line 134: return status code

iio/iio_trigger.c
```diff
@@ -0,0 +1,208 @@
...
+	struct iio_hw_trig *desc = trig;
+	iio_process_trigger_type(*desc->iio_desc, desc->name);
+}
+
+/**
+ * @brief Free the resources allocated by iio_hw_trig_init().
+ *
+ * @param trig - The trigger structure.
+*/
+void iio_hw_trig_remove(struct iio_hw_trig *trig)
```
buha, line 145: return status code

rbolboac:
- added return values to all functions

PR #1368, opened 2022-05-11
iio_demo: number of samples platform specific & adc_demo from 12bit to 16bit data

rbolboac:
-  The number of samples has been increased and moved to parameters.h such that writings to dac buffer are working with default data files from IIO osc.
- Modify adc_demo from 12bit to 16bit data such that both adc and dac devices are using 16-bit format. It is useful when using loop-back feature (the data written to the dac is transferred to the buffer used by the adc). Also, in case iio_readdev is used without loop-back feature, the data will be stored in 2's complement with sign bit on bit16, instead of bit12, allowing gnuplot to render the data correctly.


projects/iio_demo/src/platform/xilinx/parameters.h
```diff
@@ -65,8 +65,7 @@
 #define UART_IRQ_ID		XPAR_AXI_INTC_AXI_UART_INTERRUPT_INTR
 #endif // _XPARAMETERS_PS_H_
 
-/* 400 * 8 * 2 = 6400 Default number of samples requested on a capture */
-#define MAX_SIZE_BASE_ADDR	10000
-#define UART_BAUDRATE	115200
+#define SAMPLES_PER_CHANNEL_PLATFORM    2000
```
amiclaus, line 68: some inline comments might fit for why these values were chosen.

amiclaus (commented):
one nitpick on my side

rbolboac:
- added comment for SAMPLES_PER_CHANNEL_PLATFORM macro

projects/iio_demo/src/platform/stm32/parameters.h
```diff
@@ -53,8 +53,10 @@
 /******************************************************************************/
 /********************** Macros and Constants Definitions **********************/
 /******************************************************************************/
-#define MAX_SIZE_BASE_ADDR	(SAMPLES_PER_CHANNEL * DEMO_CHANNELS * \
-					sizeof(uint16_t))
+/* This value can be moddiffied based on the number
+of samples needed to be stored in the device buffer
+and based on the available RAM memory of the platform */
+#define SAMPLES_PER_CHANNEL_PLATFORM 2000
```
CiprianRegus, line 59: nitpick: moddiffied -> modified, same for the other comment

amiclaus (approved):

drivers/adc/adc_demo/adc_demo.c
```diff
@@ -51,22 +51,14 @@
...
-	0xA58, 0xAA0, 0xAED, 0xB3C, 0xB8E, 0xBE3, 0xC3A, 0xC94,
-	0xCF0, 0xD4E, 0xDAD, 0xE0E, 0xE70, 0xED3, 0xF37, 0xF9B
+	0x0000, 0xFF9B, 0xFF37, 0xFED3, 0xFE70, 0xFE0E, 0xFDAD, 0xFD4E,
+	0xFCF0, 0xFC94, 0xFC3A, 0xFBE3, 0xFB8E, 0xFB3C, 0xFAED, 0xFAA0,
+	0xFA58, 0xFA12, 0xF9D1, 0xF993, 0xF959, 0xF923, 0xF8F2, 0xF8C5,
+	0xF89C, 0xF878, 0xF858, 0xF83D, 0xF827, 0xF816, 0xF80A, 0xF802,
+	0xF800, 0xF802, 0xF80A, 0xF816, 0xF827, 0xF83D, 0xF858, 0xF878,
+	0xF89C, 0xF8C5, 0xF8F2, 0xF923, 0xF959, 0xF993, 0xF9D1, 0xFA12,
+	0xFA58, 0xFAA0, 0xFAED, 0xFB3C, 0xFB8E, 0xFBE3, 0xFC3A, 0xFC94,
+	0xFCF0, 0xFD4E, 0xFDAD, 0xFE0E, 0xFE70, 0xFED3, 0xFF37, 0xFF9B
```
adrimbarean, line 61: Question: This seems to be only the negative part of a sine wave, is this intended? Also, there are only 64 values.

adrimbarean (approved):
Looks good to me. One question.

drivers/adc/adc_demo/adc_demo.c
rbolboac: Thank you for pointing this out. This was not intended. I made the necessary changes.

PR #1377, opened 2022-05-18
Dev/stm timer irq

rbolboac:
- Added no_os_timer to drivers/api
- Added support for timers on stm32 platform
- Added timer trigger support for adc_demo with example
- Added timer trigger support for dac_demo with example: data will be written in the buffer when the trigger is asserted (the writings to the buffer emulate the writings to the dac device).


drivers/api/no_os_timer.c
```diff
@@ -0,0 +1,181 @@
...
+ * @param [in] param - Initialization structure.
+ * @return 0 in case of success, negative error code otherwise
+ */
+int32_t no_os_timer_init(struct no_os_timer_desc **desc,
+			 struct no_os_timer_init_param *param)
+{
+	if (!param)
+		return -EINVAL;
+
+	if ((param->platform_ops->init(desc, param)))
```
buha, line 62: wouldn't it be better to pass along the return code of ops->init() ?

drivers/platform/stm32/stm32_timer.c
```diff
@@ -0,0 +1,343 @@
...
+{
+	if (!desc)
+		return -EINVAL;
+
+	if (!HAL_TIM_Base_DeInit(desc->extra)) {
+		free(desc);
+		return 0;
+	}
+
+	return -1;
```
buha, line 212: I think we need to do a better error reporting here, either:
- return -EIO
- or do some translation between STM32 HAL error codes and no-OS error codes

drivers/platform/stm32/stm32_timer.c
```diff
@@ -0,0 +1,343 @@
...
+ */
+int32_t stm32_timer_start(struct no_os_timer_desc *desc)
+{
+	if (!desc)
+		return -EINVAL;
+
+	if (!HAL_TIM_Base_Start_IT((TIM_HandleTypeDef*)desc->extra))
+		return 0;
+
+	return -1;
```
buha, line 229: also here

drivers/platform/stm32/stm32_timer.c
```diff
@@ -0,0 +1,343 @@
...
+ */
+int32_t stm32_timer_stop(struct no_os_timer_desc *desc)
+{
+	if (!desc)
+		return -EINVAL;
+
+	if (!HAL_TIM_Base_Stop_IT((TIM_HandleTypeDef*)desc->extra))
+		return 0;
+
+	return -1;
```
buha, line 246: same here

util/no_os_circular_buffer.c
```diff
@@ -275,6 +278,10 @@ static int32_t no_os_cb_operation(struct no_os_circular_buffer *desc,
 		if (ret == -NO_OS_EOVERRUN)
 			sticky_overrun = true;
 
+		/* If no data is available return error */
+		if(!available_size)
+			return -1;
```
buha, line 283: let's try to be creative with these return codes (use matching ones from errno.h), they can be really helpful when debugging with printfs

rbolboac:
- implemented review findings

drivers/api/no_os_timer.c
rbolboac: done

drivers/platform/stm32/stm32_timer.c
rbolboac: Now -EIO is being used. Unfortunately STM32 HAL in this case returns either HAL_OK or HAL_ERROR so there is no useful information to be returned in case of an error (just the fact that an error occurred).

drivers/platform/stm32/stm32_timer.c
rbolboac: Now -EIO is being used. Unfortunately STM32 HAL in this case returns either HAL_OK or HAL_ERROR so there is no useful information to be returned in case of an error (just the fact that an error occurred).

drivers/platform/stm32/stm32_timer.c
rbolboac: Now -EIO is being used. Unfortunately STM32 HAL in this case returns either HAL_OK or HAL_ERROR so there is no useful information to be returned in case of an error (just the fact that an error occurred).

util/no_os_circular_buffer.c
rbolboac: done

rbolboac:
- added no_os_timer_get_elapsed_time_nsec in stm32_timer.c to avoid undefined behavior in case someone wants to use the function. However, at the moment it is not implemented because it is not needed. Currently the timers are used only for triggers in IIO application.

CiprianRegus (approved):
Looks good.

adrimbarean (approved):
Looks good to me.

PR #1388, opened 2022-05-24
tools:scripts:build_projects.py: Fixes bsp regeneration for stm32

rbolboac:
If platform is stm32, always perform make reset.
Fixes: d2018ba ("tools:scripts:build_projects.py: Always regenerate bsp for stm32")


amiclaus (approved):

amiclaus (changes requested):
these needs more testing, it seems that regenerating each time the BSP for stm32 increases considerably the projects build duration, resulting eventually in a timeout/hang.

buha:
i don't see how else we can do this, each stm32 project has different .ioc (think HARDWARE, like .hdf) files describing the hardware and reusing such an .ioc for multiple projects doesn't make sense like for xilinx

amiclaus (approved):

PR #1393, opened 2022-05-26
Staging/adxl313 iio driver

danmois:
The implementation of the ADXL313 IIO driver and the addition of the IIO example for the ADXL312Z, ADXL313Z, and ADXL314Z development boards.

### DIFF
--- a/drivers/accel/adxl313/iio_adxl313.c
+++ b/drivers/accel/adxl313/iio_adxl313.c
@@ -314,13 +314,13 @@ static int adxl313_iio_read_raw(void *dev, char *buf, uint32_t len,
 
 		switch (channel->address) {
 		case CHAN_X:
-			ret = no_os_sign_extend32(accel_x, MIN_SHIFT + adxl313->resolution);
+			ret = no_os_sign_extend16(accel_x, MIN_SHIFT + adxl313->resolution);
 			break;
 		case CHAN_Y:
-			ret = no_os_sign_extend32(accel_y, MIN_SHIFT + adxl313->resolution);
+			ret = no_os_sign_extend16(accel_y, MIN_SHIFT + adxl313->resolution);
 			break;
 		case CHAN_Z:
-			ret = no_os_sign_extend32(accel_z, MIN_SHIFT + adxl313->resolution);
+			ret = no_os_sign_extend16(accel_z, MIN_SHIFT + adxl313->resolution);
 			break;
 		default:
 			return -EINVAL;
@@ -1009,15 +1009,15 @@ static int adxl313_iio_read_samples(void* dev, int16_t* buff, uint32_t samples)
 			return ret;
 
 		if (iio_adxl313->active_channels & NO_OS_BIT(0)) {
-			buff[i] = no_os_sign_extend32(data_x, MIN_SHIFT + adxl313->resolution);
+			buff[i] = no_os_sign_extend16(data_x, MIN_SHIFT + adxl313->resolution);
 			i++;
 		}
 		if (iio_adxl313->active_channels & NO_OS_BIT(1)) {
-			buff[i] = no_os_sign_extend32(data_y, MIN_SHIFT + adxl313->resolution);
+			buff[i] = no_os_sign_extend16(data_y, MIN_SHIFT + adxl313->resolution);
 			i++;
 		}
 		if (iio_adxl313->active_channels & NO_OS_BIT(2)) {
-			buff[i] = no_os_sign_extend32(data_z, MIN_SHIFT + adxl313->resolution);
+			buff[i] = no_os_sign_extend16(data_z, MIN_SHIFT + adxl313->resolution);
 			i++;
 		}
 	}
--- a/drivers/accel/adxl313/iio_adxl313.c
+++ b/drivers/accel/adxl313/iio_adxl313.c
@@ -1,0 +1,1135 @@
+/*******************************************************************************
+ *   @file   iio_adxl313.c
+ *   @brief  Implementation of IIO ADXL313 Driver.
+ *   @author GMois (george.mois@analog.com)
+********************************************************************************
+ * Copyright 2022(c) Analog Devices, Inc.
+ *
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *  - Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ *  - Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in
+ *    the documentation and/or other materials provided with the
+ *    distribution.
+ *  - Neither the name of Analog Devices, Inc. nor the names of its
+ *    contributors may be used to endorse or promote products derived
+ *    from this software without specific prior written permission.
+ *  - The use of this software may or may not infringe the patent rights
+ *    of one or more patent holders.  This license does not release you
+ *    from the requirement that you obtain separate licenses from these
+ *    patent holders to use this software.
+ *  - Use of the software either in source or binary form, must be run
+ *    on or directly connected to an Analog Devices Inc. component.
+ *
+ * THIS SOFTWARE IS PROVIDED BY ANALOG DEVICES "AS IS" AND ANY EXPRESS OR
+ * IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, NON-INFRINGEMENT,
+ * MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED.
+ * IN NO EVENT SHALL ANALOG DEVICES BE LIABLE FOR ANY DIRECT, INDIRECT,
+ * INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, INTELLECTUAL PROPERTY RIGHTS, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+*******************************************************************************/
+
+/******************************************************************************/
+/***************************** Include Files **********************************/
+/******************************************************************************/
+#include <stdlib.h>
+#include <stdio.h>
+#include "no_os_error.h"
+#include "no_os_util.h"
+#include "iio_adxl313.h"
+#include "adxl313.h"
+
+/******************************************************************************/
+/********************** Macros and Constants Definitions **********************/
+/******************************************************************************/
+#define SCALE_FACTOR 1000000000UL
+#define MIN_SHIFT    9
+
+static const int adxl313_iio_odr_table[10][2] = {
+	{6, 250000},
+	{12, 500000},
+	{25, 0},
+	{50, 0},
+	{100, 0},
+	{200, 0},
+	{400, 0},
+	{800, 0},
+	{1600, 0},
+	{3200, 0}
+};
+
+static const int adxl312_iio_range_table[4][2] = {
+	{1, 500000},
+	{3, 0},
+	{6, 0},
+	{12, 0},
+};
+
+static const int adxl313_iio_range_table[4][2] = {
+	{0, 500000},
+	{1, 0},
+	{2, 0},
+	{4, 0}
+};
+
+static const int adxl314_iio_range = 200;
+
+static const int adxl312_iio_scale_table[4] = { 28439285, 56878570, 113757140, 227514280 };
+
+static const int adxl313_iio_scale_table[4] = { 9576806, 19153613, 38307226, 76614453 };
+
+static const int adxl314_iio_scale = 478858719;
+
+/******************************************************************************/
+/*************************** Types Declarations *******************************/
+/******************************************************************************/
+enum adxl313_iio_accel_chans {
+	CHAN_X = 0,
+	CHAN_Y = 1,
+	CHAN_Z = 2
+};
+
+/******************************************************************************/
+/************************ Functions Declarations ******************************/
+/******************************************************************************/
+extern int adxl313_read(struct adxl313_dev *dev,
+			uint8_t base_address, uint16_t size, uint8_t *read_data);
+extern int adxl313_write(struct adxl313_dev *dev,
+			 uint8_t base_address, uint16_t size, uint8_t *write_data);
+static int adxl313_iio_find_2d_row(const int (*freq_tbl)[2], const int n,
+				   const int val1, const int val2);
+static int adxl313_iio_read_reg(struct adxl313_iio_dev *dev, uint32_t reg,
+				uint32_t *readval);
+static int adxl313_iio_write_reg(struct adxl313_iio_dev *dev, uint32_t reg,
+				 uint32_t writeval);
+static int adxl313_iio_read_raw(void *dev, char *buf, uint32_t len,
+				const struct iio_ch_info *channel, intptr_t priv);
+static int adxl313_iio_read_scale(void *dev, char *buf, uint32_t len,
+				  const struct iio_ch_info *channel, intptr_t priv);
+static int adxl313_iio_write_scale(void *dev, char *buf,
+				   uint32_t len, const struct iio_ch_info *channel, intptr_t priv);
+static int adxl313_iio_read_calibbias(void *dev, char *buf, uint32_t len,
+				      const struct iio_ch_info *channel, intptr_t priv);
+static int adxl313_iio_write_calibbias(void *dev, char *buf, uint32_t len,
+				       const struct iio_ch_info *channel, intptr_t priv);
+static int adxl313_iio_read_sampling_freq(void *dev, char *buf, uint32_t len,
+		const struct iio_ch_info *channel, intptr_t priv);
+static int adxl313_iio_write_sampling_freq(void *dev, char *buf,
+		uint32_t len, const struct iio_ch_info *channel, intptr_t priv);
+static int adxl313_iio_read_range(void *dev, char *buf, uint32_t len,
+				  const struct iio_ch_info *channel, intptr_t priv);
+static int adxl313_iio_write_range(void *dev, char *buf,
+				   uint32_t len, const struct iio_ch_info *channel, intptr_t priv);
+static int adxl313_iio_read_samp_freq_avail(void *dev, char *buf,
+		uint32_t len, const struct iio_ch_info *channel, intptr_t priv);
+static int adxl313_iio_read_range_avail(void *dev, char *buf,
+					uint32_t len, const struct iio_ch_info *channel, intptr_t priv);
+static int adxl313_iio_read_scale_avail(void *dev, char *buf,
+					uint32_t len, const struct iio_ch_info *channel, intptr_t priv);
+static int adxl313_iio_read_samples(void* dev, int16_t* buff, uint32_t samples);
+static int adxl313_iio_update_channels(void* dev, uint32_t mask);
+/******************************************************************************/
+/************************ Variable Declarations *******************************/
+/******************************************************************************/
+
+static struct iio_attribute adxl313_iio_accel_attrs[] = {
+	{
+		.name  = "calibbias",
+		.show  = adxl313_iio_read_calibbias,
+		.store = adxl313_iio_write_calibbias,
+	},
+	{
+		.name = "raw",
+		.show = adxl313_iio_read_raw,
+	},
+	{
+		.name   = "sampling_frequency",
+		.shared = IIO_SHARED_BY_TYPE,
+		.show   = adxl313_iio_read_sampling_freq,
+		.store  = adxl313_iio_write_sampling_freq,
+	},
+	{
+		.name   = "sampling_frequency_available",
+		.shared = IIO_SHARED_BY_TYPE,
+		.show   = adxl313_iio_read_samp_freq_avail,
+	},
+	{
+		.name   = "range",
+		.shared = IIO_SHARED_BY_TYPE,
+		.show   = adxl313_iio_read_range,
+		.store  = adxl313_iio_write_range,
+	},
+	{
+		.name   = "range_available",
+		.shared = IIO_SHARED_BY_TYPE,
+		.show   = adxl313_iio_read_range_avail,
+	},
+	{
+		.name   = "scale",
+		.shared = IIO_SHARED_BY_TYPE,
+		.show   = adxl313_iio_read_scale,
+		.store  = adxl313_iio_write_scale,
+	},
+	{
+		.name   = "scale_available",
+		.shared = IIO_SHARED_BY_TYPE,
+		.show   = adxl313_iio_read_scale_avail,
+	},
+	END_ATTRIBUTES_ARRAY
+};
+
+static struct scan_type adxl313_iio_accel_scan_type = {
+	.sign = 's',
+	.realbits = 13,
+	.storagebits = 16,
+	.shift = 0,
+	.is_big_endian = false
+};
+
+#define ADXL313_ACCEL_CHANNEL(index, reg, axis) { \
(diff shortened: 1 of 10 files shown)


drivers/accel/adxl313/iio_adxl313.c
```diff
@@ -0,0 +1,1200 @@
...
+			vals[1] = 0; //fractional part
+			break;
+		default:
+			return -EINVAL;
+		}
+
+		return iio_format_value(buf, len, IIO_VAL_INT_PLUS_MICRO, 1, vals);
+	} else {
+		return -EINVAL;
+	}
```
rbolboac, line 772: you can remove the else and simply leave return -EINVAL; check the whole file.

drivers/accel/adxl313/iio_adxl313.c
```diff
@@ -0,0 +1,1200 @@
...
+
+	/* Enables full resolution for ADXL312 and ADXL313 */
+	if (desc->adxl313_dev->dev_type != ID_ADXL314) {
+		ret = adxl313_enable_full_res(desc->adxl313_dev);
+		if (ret)
+			return ret;
+	}
+
+	// Set operation mode
```
rbolboac, line 1164: nitpick: use the same comment format for line comments. check through the file.
suggestion: use /* */ for line comments (format used for Linux development)

drivers/accel/adxl313/iio_adxl313.c
```diff
@@ -0,0 +1,1200 @@
...
+			if (ret)
+				goto error_config;
+		}
+	}
+
+	/* Enables full resolution for ADXL312 and ADXL313 */
+	if (desc->adxl313_dev->dev_type != ID_ADXL314) {
+		ret = adxl313_enable_full_res(desc->adxl313_dev);
+		if (ret)
+			return ret;
```
rbolboac, line 1161: memory leak: use goto error_config

drivers/accel/adxl313/iio_adxl313.c
```diff
@@ -0,0 +1,1200 @@
...
+		return -EINVAL;
+
+	iio_adxl313 = (struct adxl313_iio_dev *)dev_data->dev;
+
+	if (!iio_adxl313->adxl313_dev)
+		return -EINVAL;
+
+	adxl313 = iio_adxl313->adxl313_dev;
+
+	adxl313_get_raw_xyz(adxl313, &x, &y, &z);
```
rbolboac, line 1100: use return value

drivers/accel/adxl313/iio_adxl313.c
```diff
@@ -0,0 +1,1200 @@
...
+
+	iio_adxl313 = (struct adxl313_iio_dev *)dev;
+
+	if (!iio_adxl313->adxl313_dev)
+		return -EINVAL;
+
+	adxl313 = iio_adxl313->adxl313_dev;
+
+	for(uint32_t i = 0; i < samples*iio_adxl313->no_of_active_channels;) {
+		adxl313_get_raw_xyz(adxl313, &data_x, &data_y, &data_z);
```
rbolboac, line 1019: use return value

drivers/accel/adxl313/iio_adxl313.c
```diff
@@ -0,0 +1,1200 @@
...
+	iio_adxl313->active_channels = mask;
+
+	if (mask & NO_OS_BIT(0))
+		counter ++;
+	if (mask & NO_OS_BIT(1))
+		counter ++;
+	if (mask & NO_OS_BIT(2))
+		counter ++;
+	if (mask & NO_OS_BIT(3))
+		counter ++;
```
rbolboac, line 1066: there are only 3 channels used, the mask will never have b3 set. you may remove these lines.

rbolboac (commented):
just a few remarks on my side

drivers/accel/adxl313/iio_adxl313.c
```diff
@@ -0,0 +1,1200 @@
...
+#include "iio_adxl313.h"
+#include "adxl313.h"
+
+/******************************************************************************/
+/********************** Macros and Constants Definitions **********************/
+/******************************************************************************/
+#define ACCEL_AXIS_X (uint32_t) 0
+#define ACCEL_AXIS_Y (uint32_t) 1
+#define ACCEL_AXIS_Z (uint32_t) 2
+#define MEGA 1000000UL
```
amiclaus, line 56: just a thought: these kind for definitions might fit into a generic header file. similar to what it is done on linux:
https://github.com/torvalds/linux/blob/master/include/linux/units.h

drivers/accel/adxl313/iio_adxl313.c
```diff
@@ -0,0 +1,1200 @@
...
+		uint32_t len, const struct iio_ch_info *channel, intptr_t priv);
+static int adxl313_iio_read_range_avail(void *dev, char *buf,
+					uint32_t len, const struct iio_ch_info *channel, intptr_t priv);
+static int adxl313_iio_read_scale_avail(void *dev, char *buf,
+					uint32_t len, const struct iio_ch_info *channel, intptr_t priv);
+static int adxl313_iio_read_samples(void* dev, int* buff, uint32_t samples);
+static int adxl313_iio_update_channels(void* dev, uint32_t mask);
+static int32_t adxl313_trigger_handler(struct iio_device_data *dev_data);
+// /******************************************************************************/
+// /************************ Variable Declarations ******************************/
```
amiclaus, line 145: nit: remove `//`

drivers/accel/adxl313/iio_adxl313.c
```diff
@@ -0,0 +1,1200 @@
...
+			else
+				vals[0] = adxl313_iio_scale_table[adxl313->range];
+			return iio_format_value(buf, len, IIO_VAL_FRACTIONAL, 2, vals);
+		case ID_ADXL314:
+			vals[0] = adxl314_iio_scale;
+			return iio_format_value(buf, len, IIO_VAL_FRACTIONAL, 2, vals);
+		default:
+			return -EINVAL;
+		}
+	} else {
```
amiclaus, line 403: no need to use `else` here.  just return directly. (same as you did for `adxl313_iio_write_scale`)

drivers/accel/adxl313/iio_adxl313.c
```diff
@@ -0,0 +1,1200 @@
...
+			val = adxl313->y_offset_raw * multiplier;
+			return iio_format_value(buf, len, IIO_VAL_INT, 1, &val);
+		case ACCEL_AXIS_Z:
+			val = adxl313->z_offset_raw * multiplier;
+			return iio_format_value(buf, len, IIO_VAL_INT, 1, &val);
+
+		default:
+			return -EINVAL;
+		}
+
```
amiclaus, line 524: nit: remove line

drivers/accel/adxl313/iio_adxl313.c
```diff
@@ -0,0 +1,1200 @@
...
+		case ACCEL_AXIS_Z:
+			ret = adxl313_set_offset(adxl313, val, ADXL313_Z_AXIS);
+			break;
+		default:
+			ret = -EINVAL;
+			break;
+		}
+
+		if (ret) {
+			adxl313_set_op_mode(adxl313, op_mode);
```
amiclaus, line 591: is this function call supposed to be here?

drivers/accel/adxl313/iio_adxl313.c
```diff
@@ -0,0 +1,1200 @@
...
+	if (!iio_adxl313->adxl313_dev)
+		return -EINVAL;
+
+	adxl313 = iio_adxl313->adxl313_dev;
+
+	if (channel->type == IIO_ACCEL) {
+		vals[0] = adxl313_iio_odr_table[adxl313->odr][0];
+		vals[1] = adxl313_iio_odr_table[adxl313->odr][1];
+		return iio_format_value(buf, len, IIO_VAL_INT_PLUS_MICRO, 1, vals);
+	} else {
```
amiclaus, line 634: ditto: no need for else. return directly

drivers/accel/adxl313/iio_adxl313.c
```diff
@@ -0,0 +1,1200 @@
...
+		case ID_ADXL314:
+			vals[0] = adxl314_iio_range;
+			vals[1] = 0; //fractional part
+			break;
+		default:
+			return -EINVAL;
+		}
+
+		return iio_format_value(buf, len, IIO_VAL_INT_PLUS_MICRO, 1, vals);
+	} else {
```
amiclaus, line 770: ditto: remove `else`

drivers/accel/adxl313/iio_adxl313.c
```diff
@@ -0,0 +1,1200 @@
...
+			vals[0] = adxl314_iio_range;
+			vals[1] = 0; //fractional part
+			iio_format_value(buffer, sizeof(buffer), IIO_VAL_INT_PLUS_MICRO, 2, vals);
+			strcat(buf, buffer);
+			break;
+		default:
+			return -EINVAL;
+		}
+		return strlen(buf);
+	} else {
```
amiclaus, line 906: dito: remove else

drivers/accel/adxl313/iio_adxl313.c
```diff
@@ -0,0 +1,1200 @@
...
+		case ID_ADXL314:
+			vals[0] = adxl314_iio_scale;
+			iio_format_value(buffer, sizeof(buffer), IIO_VAL_FRACTIONAL, 2, vals);
+			strcat(buf, buffer);
+			break;
+		default:
+			return -EINVAL;
+		}
+		return strlen(buf);
+	} else {
```
amiclaus, line 985: ditto: remove else

amiclaus (commented):
lgtm overall, just some minor comments

drivers/accel/adxl313/iio_adxl313.c
```diff
@@ -0,0 +1,1200 @@
...
+#include <stdio.h>
+#include "no_os_error.h"
+#include "no_os_util.h"
+#include "iio_adxl313.h"
+#include "adxl313.h"
+
+/******************************************************************************/
+/********************** Macros and Constants Definitions **********************/
+/******************************************************************************/
+#define ACCEL_AXIS_X (uint32_t) 0
```
amiclaus, line 53: nit: shouldn't these defines be in `.h` ?

drivers/accel/adxl313/iio_adxl313.c
amiclaus: with https://github.com/analogdevicesinc/no-OS/pull/1394 you can use the generic header file for these specific units

drivers/accel/adxl313/iio_adxl313.c
```diff
@@ -0,0 +1,1200 @@
...
+	ADXL313_ACCEL_CHANNEL(1, chan_y, Y),
+	ADXL313_ACCEL_CHANNEL(2, chan_z, Z),
+};
+
+static struct iio_device adxl313_iio_dev = {
+	.num_ch = NO_OS_ARRAY_SIZE(adxl313_channels),
+	.channels = adxl313_channels,
+	.pre_enable = (int32_t (*)())adxl313_iio_update_channels,
+	.trigger_handler = (int32_t (*)())adxl313_trigger_handler,
+	.read_dev = (int32_t (*)())adxl313_iio_read_samples,
```
rbolboac, line 225: I'm curious if you have tested this functionality. To me it does not look like it would work seeing how you have not defined on application level a read buffer which would store the data.
You can check if this is working by using the plot functionality in IIO Osc.

drivers/accel/adxl313/iio_adxl313.c
```diff
@@ -0,0 +1,1200 @@
...
+	ADXL313_ACCEL_CHANNEL(0, chan_x, X),
+	ADXL313_ACCEL_CHANNEL(1, chan_y, Y),
+	ADXL313_ACCEL_CHANNEL(2, chan_z, Z),
+};
+
+static struct iio_device adxl313_iio_dev = {
+	.num_ch = NO_OS_ARRAY_SIZE(adxl313_channels),
+	.channels = adxl313_channels,
+	.pre_enable = (int32_t (*)())adxl313_iio_update_channels,
+	.trigger_handler = (int32_t (*)())adxl313_trigger_handler,
```
rbolboac, line 224: Seeing how there isn't a trigger example available of this one I would remove it until the example is available and the functionality is successfully being tested.

drivers/accel/adxl313/iio_adxl313.c
danmois: Thank you! I went through all the file and corrected the format.

drivers/accel/adxl313/iio_adxl313.c
danmois: Modified in all occurrences of the function.

drivers/accel/adxl313/iio_adxl313.c
danmois: Thank you for the observation! However, I removed the units defines since they are not used in the file.

drivers/accel/adxl313/iio_adxl313.c
danmois: Thanks for pointing to this problem. I replaced the switch with an if here. In my opinion, it is more simple to understand this way.

drivers/accel/adxl313/iio_adxl313.c
danmois: If adxl313_set_offset fails, the adxl313_iio_write_calibbias function should return the error code generated by adxl313_set_offset, and not by the adxl313_set_op_mode(which could be executed sucessfully). However, the chip was set to standby mode for performing the setting, even if the setting action had failed. This function call tries to set the device to the state previous to standby (could have been MEASURE).

drivers/accel/adxl313/iio_adxl313.c
danmois: The functionality has been tested. However, the variant that includes the buffer is the right one, thank you for pointing that out.

drivers/accel/adxl313/iio_adxl313.c
danmois: They are used in static functions in the .c file. It seemed more appropriate to place them here, where they are used. The header contains only the definitions for the init and remove functions, that do not deal with the accelerometer's axes.

danmois:
V1

- Formatted comments
- Removed redundant code for return values

danmois:
V2

- removed the trigger handler

drivers/accel/adxl313/iio_adxl313.c
danmois: You are right, this should be removed.

danmois:
V3

- Addition of read data buffer to the example

drivers/accel/adxl313/iio_adxl313.h
```diff
@@ -0,0 +1,74 @@
...
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+*******************************************************************************/
+#ifndef IIO_ADXL313_H
+#define IIO_ADXL313_H
+
+/******************************************************************************/
+/***************************** Include Files **********************************/
+/******************************************************************************/
+#include "iio.h"
+#include "no_os_irq.h"
```
rbolboac, line 46: nitpick: I don't think this include is needed at the moment (needed only for hardware trigger)

projects/eval-adxl313z/src/platform/stm32/main.c
```diff
@@ -62,6 +66,12 @@ int main()
 
 	stm32_init();
 
+#ifdef IIO_EXAMPLE
+	ret = iio_example_main();
+	if (ret < 0)
```
rbolboac, line 71: nitpick: if (ret) syntax is being used on Linux, would be nice to use the same syntax

projects/eval-adxl313z/src/platform/stm32/main.c
```diff
@@ -75,6 +85,12 @@ int main()
 		goto error;
 #endif
 
+#if (IIO_EXAMPLE + DUMMY_EXAMPLE > 1)
```
rbolboac, line 88: this won't work as expected seeing how BASIC_EXAMPLE is used instead of DUMMY_EXAMPLE and DUMMY_EXAMPLE is never defined.
replace DUMMY_EXAMPLE with BASIC_EXAMPLE.

danmois:
V4

- modified storagebits to 16 (maximum realbits for all devices is 13 bits)

drivers/accel/adxl313/iio_adxl313.c
```diff
@@ -0,0 +1,1200 @@
...
+		.shared = IIO_SHARED_BY_TYPE,
+		.show   = adxl313_iio_read_scale_avail,
+	},
+	END_ATTRIBUTES_ARRAY
+};
+
+static struct scan_type adxl313_iio_accel_scan_type = {
+	.sign = 's',
+	.realbits = 20,
+	.storagebits = 32,
```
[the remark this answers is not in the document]
danmois: Thank you for the observation, the values were correctly set now.

danmois:
V5

- removed unused include (no_os_irq.h)
- adopted Linux-like syntax for checking for errors and corrected project options in main.c (platform/stm32)

drivers/accel/adxl313/iio_adxl313.c
```diff
@@ -0,0 +1,1146 @@
...
+	iio_adxl313 = (struct adxl313_iio_dev *)dev;
+
+	if (!iio_adxl313->adxl313_dev)
+		return -EINVAL;
+
+	adxl313 = iio_adxl313->adxl313_dev;
+
+	vals[1] = SCALE_FACTOR;
+
+	if (channel->type == IIO_ACCEL) {
```
CiprianRegus, line 370: Is this `if` necessary, since you only have `IIO_ACCEL` channels?

drivers/accel/adxl313/iio_adxl313.c
```diff
@@ -0,0 +1,1146 @@
...
+				return adxl313_enable_full_res(adxl313);
+			else
+				return adxl313_disable_full_res(adxl313);
+		case ID_ADXL313:
+			iio_parse_value(buf, IIO_VAL_FRACTIONAL, &vals[0], &vals[1]);
+			/* Enable/disable FULL_RES depending on selection */
+			if (vals[1] == adxl313_iio_scale_table[0])
+				return adxl313_enable_full_res(adxl313);
+			else
+				return adxl313_disable_full_res(adxl313);
```
CiprianRegus, line 446: Redundant elses.

drivers/accel/adxl313/iio_adxl313.c
```diff
@@ -0,0 +1,1146 @@
...
+		switch (channel->address) {
+		case ACCEL_AXIS_X:
+			val = adxl313->x_offset_raw * multiplier;
+			return iio_format_value(buf, len, IIO_VAL_INT, 1, &val);
+		case ACCEL_AXIS_Y:
+			val = adxl313->y_offset_raw * multiplier;
+			return iio_format_value(buf, len, IIO_VAL_INT, 1, &val);
+		case ACCEL_AXIS_Z:
+			val = adxl313->z_offset_raw * multiplier;
+			return iio_format_value(buf, len, IIO_VAL_INT, 1, &val);
```
CiprianRegus, line 511: You can have only 1 return after the `switch`, which will simplify the code a bit.

drivers/accel/adxl313/iio_adxl313.c
```diff
@@ -0,0 +1,1146 @@
...
+
+	if (!iio_adxl313->adxl313_dev)
+		return -EINVAL;
+
+	adxl313 = iio_adxl313->adxl313_dev;
+
+	if (channel->type == IIO_ACCEL) {
+		vals[0] = adxl313_iio_odr_table[adxl313->odr][0];
+		vals[1] = adxl313_iio_odr_table[adxl313->odr][1];
+		return iio_format_value(buf, len, IIO_VAL_INT_PLUS_MICRO, 1, vals);
```
CiprianRegus, line 624: Shouldn't this  `len` be 2?

drivers/accel/adxl313/iio_adxl313.c
```diff
@@ -0,0 +1,1146 @@
...
+	iio_adxl313 = (struct adxl313_iio_dev *)dev;
+
+	iio_adxl313->active_channels = mask;
+
+	if (mask & NO_OS_BIT(0))
+		counter ++;
+	if (mask & NO_OS_BIT(1))
+		counter ++;
+	if (mask & NO_OS_BIT(2))
+		counter ++;
```
CiprianRegus, line 1057: You can use `no_os_hweight8` instead.

drivers/accel/adxl313/iio_adxl313.c
danmois: Yest, this is the correct version, although it works this way (work with array).

drivers/accel/adxl313/iio_adxl313.c
danmois: Thank you for the suggestion!

danmois:
V6

- removed redundant else for disabling full resolution
- use of no_os_hweight8() for updating channels
- corrected formatting value for read frequency
- simplified switch statement regarding reading offset values

drivers/accel/adxl313/iio_adxl313.c
```diff
@@ -0,0 +1,1140 @@
...
+#include "no_os_error.h"
+#include "no_os_util.h"
+#include "iio_adxl313.h"
+#include "adxl313.h"
+
+/******************************************************************************/
+/********************** Macros and Constants Definitions **********************/
+/******************************************************************************/
+#define ACCEL_AXIS_X (uint32_t) 0
+#define ACCEL_AXIS_Y (uint32_t) 1
```
amiclaus, line 54: by the way, why using cast?
also, these defines might fit in an enum.
Actually, you can use `adxl313_iio_accel_chans` inline

drivers/accel/adxl313/iio_adxl313.c
```diff
@@ -0,0 +1,1140 @@
...
+ *
+ * @return ret     - Table index which contains the searched data
+*******************************************************************************/
+static int adxl313_iio_find_2d_row(const int (*freq_tbl)[2], const int size,
+				   const int val1, const int val2)
+{
+	int i;
+
+	for (i = 0; i < size; i++) {
+		if (freq_tbl[i][0] == val1 && freq_tbl[i][1] == val2)
```
amiclaus, line 243: hmm, i think brackets should be used here:

`((freq_tbl[i][0] == val1) && (freq_tbl[i][1] == val2))`

drivers/accel/adxl313/iio_adxl313.c
```diff
@@ -0,0 +1,1140 @@
...
+				return adxl313_enable_full_res(adxl313);
+			break;
+		case ID_ADXL313:
+			iio_parse_value(buf, IIO_VAL_FRACTIONAL, &vals[0], &vals[1]);
+			/* Enable/disable FULL_RES depending on selection */
+			if (vals[1] == adxl313_iio_scale_table[0])
+				return adxl313_enable_full_res(adxl313);
+			break;
+		case ID_ADXL314:
+			return ret;
```
amiclaus, line 447: drop `ret` variable completely and just return 0.

drivers/accel/adxl313/iio_adxl313.c
```diff
@@ -0,0 +1,1140 @@
...
+			range_idx = adxl313_iio_find_2d_row(adxl313_iio_range_table,
+							    NO_OS_ARRAY_SIZE(adxl313_iio_range_table),
+							    vals[0], vals[1]);
+			if (range_idx < 0)
+				return range_idx;
+
+			/* Set the ODR value */
+			return adxl313_set_range(adxl313, range_idx);
+		case ID_ADXL314:
+			return ret;
```
amiclaus, line 822: drop ret and return 0 directly.

projects/eval-adxl313z/src/platform/stm32/main.c
```diff
@@ -75,6 +85,12 @@ int main()
 		goto error;
 #endif
 
+#if (IIO_EXAMPLE + BASIC_EXAMPLE > 1)
+#error Selected example projects cannot be enabled at the same time. \
+Please enable only one example and re-build the project.
+#endif
+
+
```
amiclaus, line 93: nit: extra line

amiclaus (commented):
some comments on my side

drivers/accel/adxl313/iio_adxl313.c
danmois: Thank you for the suggestion, I have removed these defines and used the enum.

danmois:
V7

- removed ret variable where it was not needed
- placed parentheses in if for removing ambiguity
- used enum members instead of defines for identifying channels

danmois:
V8

- changed adxl313_iio_accel_chans enum member names to uppercase

danmois:
V9

- modified ADXL314 scale with the value provided by the datasheet

amiclaus (approved):

PR #1394, opened 2022-05-30
 include: add no_os_units.h

amiclaus:
Similar to linux version:
https://github.com/torvalds/linux/blob/master/include/linux/units.h

Add units header file containing standard units definition.

This is a minimal implementation, subject to being expanded.

The purpose of this file is to avoid redefinition of these kind for
macros in drivers that require them.

For now, it can be used in conjunction with https://github.com/analogdevicesinc/no-OS/pull/1393 which redefines `TERA` and `MEGA`

### DIFF
--- a/include/no_os_fifo.h
+++ b/include/no_os_fifo.h
@@ -1,6 +1,6 @@
 /***************************************************************************//**
  *   @file   no_os_fifo.h
- *   @brief  Implementation of fifo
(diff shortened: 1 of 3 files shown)


rbolboac (approved):
lgtm

CiprianRegus (approved):

PR #1395, opened 2022-05-31
daq2/daq3: fix PN sequence mismatch

amiclaus:
In order to properly perform the PN sequence tests, the LINK_MODE_CTRL3
register from ad9680 needs to be written with the appropriate value for
the PN test sequence.

According to the datasheet: a transition from test modes (Register
0x0573 ≠ 0x00) to normal mode (Register 0x0573 = 0x00) requires an SPI
soft reset.

Therefore the recommended ADC JESD204 link startup sequence needs to be
performed again so that the capture works properly.

Fixes:
"ad9680 - PN9 sequence mismatch!"
"ad9680 - PN23 sequence mismatch!"

Signed-off-by: Paul Pop <paul.pop@analog.com>


amiclaus:
Closing this PR workaround, the issues should be solved in hdl.

PR #1399, opened 2022-06-03
MAX32666 platform drivers

CiprianRegus:
Added platform drivers support for MAX32666.


drivers/platform/maxim/maxim_spi.c
```diff
@@ -56,6 +57,121 @@
...
+
+	/** This function will be a no-op for other chips, because this step is
+	 * done in the SDK in those cases.
+	*/
+#if TARGET_NUM != 32650 && TARGET_NUM != 32665
+	return 0;
+#endif
+
+	switch(id) {
+	case 0:
```
amiclaus, line 78: maybe we can use an enum with appropriate namings for these ids.

drivers/platform/maxim/maxim_uart.c
```diff
@@ -265,30 +265,30 @@ int32_t no_os_uart_init(struct no_os_uart_desc **desc,
 		parity = MXC_UART_PARITY_DISABLE;
 		break;
 	case NO_OS_UART_PAR_MARK:
-#if TARGET_NUM == 32660 || TARGET_NUM == 32650
+#if TARGET_NUM == 32660 || TARGET_NUM == 32650 || TARGET_NUM == 32665
```
amiclaus, line 268: i'm wondering if there is an alternative to these ifdefs (we keep expanding them, and it starts to look a bit ugly :) )

amiclaus (commented):
some small comments on my side.

drivers/platform/maxim/maxim_spi.c
CiprianRegus: In this case, the `id` specifies the index of the SPI port (the `device_id` of a SPI descriptor), so I don't think that there is a point in creating a define for it.

adrimbarean (approved):
Looks good to me.

PR #1402, opened 2022-06-06
drivers:dac:ltc268x: Minor Fix.

pmallick-adi:
* Moved the channel array inside the source file so as to avoid linking issues with different compilers.

### DIFF
--- a/drivers/dac/ltc268x/ltc268x.h
+++ b/drivers/dac/ltc268x/ltc268x.h
@@ -90,9 +90,6 @@
 #define LTC268X_PWDN(x)				(1 << ((x) & 0xF))
 #define LTC268X_DITH_EN(x)			(1 << ((x) & 0xF))
 
-/* Number of channels corresponding to device */
-const uint8_t ltc268x_channels[] = { 8, 16};
-
 /******************************************************************************/
 /*************************** Types Declarations *******************************/
 /******************************************************************************/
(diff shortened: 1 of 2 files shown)


drivers/dac/ltc268x/ltc268x.h
```diff
@@ -91,7 +91,7 @@
 #define LTC268X_DITH_EN(x)			(1 << ((x) & 0xF))
 
 /* Number of channels corresponding to device */
-const uint8_t ltc268x_channels[] = { 8, 16};
+extern const uint8_t ltc268x_channels[];
```
rbolboac, line 94: is this array going to be accessed outside of the driver? otherwise I would remove the external declaration and simply define it statically in ltc268x.c.

pmallick-adi: I think this would be useful for the application layer to get number of channels for the desired part number.

rbolboac: Then I would suggest adding a new field uint8_t num_channels in struct ltc268x_dev, which is populated at init time and is accessible to the application layer, and then use that field instead of the array in the driver.

pmallick-adi:
Change log:

* Addressed review comments and incorporated the "num_channels" parameter to the descriptor.

rbolboac (approved):
Looks good to me. Just two minor remarks.
It looks like the commit message contains a typo. I think you intended the message to be "Minor Updates." instead of "Minor Updated."

amiclaus (approved):

PR #1405, opened 2022-06-07
adc_demo, dac_demo: use submit function instead of read_dev, write_dev functions for buffer handling

rbolboac:
- Use submit function instead of read_dev for buffer readings.
- Move adc_submit_samples and adc_demo_trigger_handler implementations to iio_adc_demo.c.
- Use submit function instead of write_dev for buffer writings.
- Move dac_submit_samples and dac_demo_trigger_handler implementations to iio_dac_demo.c.

### DIFF
--- a/drivers/adc/adc_demo/iio_adc_demo.c
+++ b/drivers/adc/adc_demo/iio_adc_demo.c
@@ -43,6 +43,32 @@
 #include "no_os_error.h"
 #include "no_os_util.h"
 #include "iio_adc_demo.h"
+#include "iio.h"
+
+/**
+ * @brief utility function for computing next upcoming channel
+ * @param ch_mask - active channels .
+ * @param last_idx -  previous index.
+ * @param new_idx - upcoming channel index, return param.
+ * @return 1 if there are more channels, 0 if done.
+ */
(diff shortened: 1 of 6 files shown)


drivers/adc/adc_demo/iio_adc_demo.c
```diff
@@ -43,6 +43,32 @@
...
+#include "iio.h"
+
+/**
+ * @brief utility function for computing next upcoming channel
+ * @param ch_mask - active channels .
+ * @param last_idx -  previous index.
+ * @param new_idx - upcoming channel index, return param.
+ * @return 1 if there are more channels, 0 if done.
+ */
+static bool get_next_ch_idx(uint32_t ch_mask, uint32_t last_idx,
```
adrimbarean, line 55: Should we do this in `no_os_util.c`?

adrimbarean (approved):
Looks good to me. One question, but not too vital.

amiclaus (approved):

drivers/adc/adc_demo/iio_adc_demo.c
rbolboac: Yes, we can move it seeing how it is already used in 3 different drivers. Thanks!
However, I would prefer doing that in another PR, if that's OK.

PR #1408, opened 2022-06-08
stm32 build hang workaround

buha:
Due to a yet elusive problem, stm32cubemx sometimes hangs while building
on azure.

This commit adds a workaround to not block the other builds by
time bounding the make all execution.

It seems reverting stm32cubemx to 6.3.0 would fix the problem, but 6.3.0
has another issue that was the reason we currently  use 6.5.0.

So this workaround should be removed once 6.6.0 arrives and fixes the
code generation hanging problem.

### DIFF
--- a/tools/scripts/build_projects.py
+++ b/tools/scripts/build_projects.py
@@ -214,7 +214,7 @@ class BuildConfig:
 			err = run_cmd(cmd + ' reset')
 			if err != 0:
 				return err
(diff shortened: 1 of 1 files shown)


amiclaus (approved):

rbolboac (approved):

buha:
you can see in the CI build log that the build for one stm32 project failed with timeout and the build process continued building other projects

dbogdan (approved):

PR #1412, opened 2022-06-10
drivers:afe:ad4110:Corrections in AD4110 driver

jansunil:
1) Replaced AD4110_ENABLE with the correct data to be written
2) Set I_EXEC_SEL to 100uA in case of RTD 2-wire as that is the only permissible option for 2-wire mode

### DIFF
--- a/drivers/afe/ad4110/ad4110.c
+++ b/drivers/afe/ad4110/ad4110.c
@@ -45,6 +45,7 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include "ad4110.h"
+#include "no_os_util.h"
 #include "no_os_error.h"
 #include "no_os_irq.h"
 #include "no_os_print_log.h"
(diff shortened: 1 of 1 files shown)


drivers/afe/ad4110/ad4110.c
```diff
@@ -308,14 +308,14 @@ int32_t ad4110_set_op_mode(struct ad4110_dev *dev, enum ad4110_op_mode mode)
 		return ad4110_spi_int_reg_write_msk(dev,
 						    A4110_AFE,
 						    AD4110_REG_AFE_CNTRL2,
-						    AD4110_ENABLE,
+						    AD4110_REG_AFE_CNTRL2_IMODE_MSK,
```
amiclaus, line 311: we shouldn't pass masks as data parameters.
Please use `no_os_field_prep` utility function inline:
`no_os_field_prep(AD4110_REG_AFE_CNTRL2_IMODE_MSK, AD4110_ENABLE)`.

Or create similar defines to: https://github.com/analogdevicesinc/no-OS/blob/4dc973f8be0ccd623e4b1e5e2a11b5bc33d0b944/drivers/afe/ad4110/ad4110.h#L123-L124

jansunil:
V2 changes: Used no_os_field_prep as suggested in the review comments

drivers/afe/ad4110/ad4110.c
jansunil: The change has been made- no_os_field_prep() has been used

jansunil:
@amiclaus , Do these changes look good?

amiclaus (approved):

adrimbarean (approved):

PR #1415, opened 2022-06-14
stm32 irq rework

rbolboac:
Create gpio irq controller and make necessary adaptations:
- remove gpio irq handling from stm32_irq
- add gpio irq handling to stm32_gpio_irq
- add check for irq controller instance
- adapt eval-adxl355-pmdz project for new gpio irq handling

Set irq priority for triggers in the projects which are using hardware triggers.

### DIFF
--- a/projects/eval-adxl355-pmdz/src/examples/iio_trigger_example/iio_trigger_example.c
+++ b/projects/eval-adxl355-pmdz/src/examples/iio_trigger_example/iio_trigger_example.c
@@ -86,6 +86,11 @@ int iio_trigger_example_main()
 	ret = no_os_irq_ctrl_init(&adxl355_irq_desc, &adxl355_gpio_irq_ip);
 	if (ret)
 		return ret;
+
+	ret = no_os_irq_set_priority(adxl355_irq_desc, adxl355_gpio_trig_ip.irq_id, 1);
+	if (ret)
+		return ret;
+
 	adxl355_gpio_trig_ip.irq_ctrl = adxl355_irq_desc;
 
 	/* Initialize hardware trigger */
--- a/projects/eval-adxl355-pmdz/src/platform/maxim/main.c
+++ b/projects/eval-adxl355-pmdz/src/platform/maxim/main.c
@@ -95,10 +95,6 @@ int main()
 	if (ret)
 		return ret;
 
-	ret = no_os_irq_set_priority(nvic_desc, NVIC_GPIO_IRQ, 1);
-	if (ret)
-		return ret;
-
 	ret = no_os_irq_enable(nvic_desc, NVIC_GPIO_IRQ);
 	if (ret)
 		return ret;


drivers/platform/stm32/stm32_gpio_irq.c
```diff
@@ -0,0 +1,512 @@
...
+
+	while (0 == no_os_list_get_first(actions, (void **)&discard))
+		free(discard);
+
+	no_os_list_remove(actions);
+	free(desc->extra);
+	free(desc);
+
+	return 0;
+}
```
CiprianRegus, line 250: I think this function needs to set an `initialized` entry as false (the same would apply to the remove function from the NVIC controller).

drivers/platform/stm32/stm32_gpio_irq.c
```diff
@@ -0,0 +1,512 @@
...
+		gpio_irq_ctrl_id = 6;
+
+	if (!initialized[gpio_irq_ctrl_id]) {
+
+		gpio_irq_desc = calloc(1, sizeof(*gpio_irq_desc));
+		if (!gpio_irq_desc)
+			return -ENOMEM;
+
+		sdesc = (struct stm32_gpio_irq_desc*)calloc(1,
+				sizeof(struct stm32_gpio_irq_desc));
```
CiprianRegus, line 199: nitpick: `sizeof(*sdesc)`

drivers/platform/stm32/stm32_gpio_irq.c
```diff
@@ -0,0 +1,485 @@
...
+	case 7:
+	case 8:
+	case 9:
+		*irq_id = EXTI9_5_IRQn;
+		break;
+	case 10:
+	case 11:
+	case 12:
+	case 13:
+	case 14:
```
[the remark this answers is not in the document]
rbolboac: yes it should be, thanks!

drivers/platform/stm32/stm32_gpio_irq.c
rbolboac: You are right, good catch! Thank you!

rbolboac:
- implemented review changes

CiprianRegus (approved):
lgtm

drivers/platform/stm32/stm32_gpio_irq.c
```diff
@@ -0,0 +1,497 @@
...
+	/* Check port number */
+	if (!IS_EXTI_GPIO_PORT(gpio_irq_ip->port_nb))
+		return -EINVAL;
+
+	/* Check pin number */
+	if (!IS_EXTI_GPIO_PIN(param->irq_ctrl_id))
+		return -EINVAL;
+
+	/* Obtain gpio irq controller id */
+	if (param->irq_ctrl_id <= 4)
```
buha, line 187: same comment here, i think this would work well for the families mentioned above, but this mapping is not true for other families

i think there are families with just 3 interrupt events instead of 6 and they map them differently:
- L0 and possibly F0 for example have EXTI0_1_IRQHandler, EXTI2_3_IRQHandler, EXTI4_15_IRQHandler

drivers/platform/stm32/stm32_gpio_irq.h
```diff
@@ -0,0 +1,70 @@
...
+ */
+struct stm32_gpio_irq_init_param {
+	uint8_t port_nb;
+};
+
+/**
+ * @struct stm32_gpio_irq_desc
+ * @brief stm32 platform specific GPIO IRQ descriptor
+ */
+struct stm32_gpio_irq_desc {
```
buha, line 58: this should probably also have a pointer to its parent, no_os_irq_desc

drivers/platform/stm32/stm32_gpio_irq.c
```diff
@@ -0,0 +1,497 @@
...
+
+/**
+ * @brief Enable all GPIO interrupts.
+ * @param desc - the GPIO irq descriptor.
+ * @return 0
+ */
+static int32_t stm32_gpio_irq_global_enable(struct no_os_irq_ctrl_desc *desc)
+{
+	/* Enable all available EXTI line interrupts */
+	HAL_NVIC_EnableIRQ(EXTI0_IRQn);
```
buha, line 406: and if you implement the desc->parent part, you may simply do no_os_gpio_irq_global_enable(desc->parent)

drivers/platform/stm32/stm32_gpio_irq.c
```diff
@@ -0,0 +1,497 @@
...
+/******************************************************************************/
+/************************ Functions Definitions *******************************/
+/******************************************************************************/
+/**
+ * @brief Maps pin_nb to NVIC IRQ id
+ * @param pin_nb - pin number to be mapped
+ * @param irq_id - NVIC IRQ id corresponding to given pin_nb
+ * @return 0 in case of success, errno error codes otherwise.
+ */
+static int stm32_pin_get_irq(uint8_t pin_nb, IRQn_Type *irq_id)
```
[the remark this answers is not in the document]
rbolboac: I have made some new changes which no longer use specifically EXTI_n_IRQn in the driver. please take a look and let me know your thoughts!

drivers/platform/stm32/stm32_gpio_irq.c
rbolboac: I have made some new changes which no longer use specifically EXTI_n_IRQn in the driver. please take a look and let me know your thoughts!

drivers/platform/stm32/stm32_gpio_irq.h
rbolboac: seeing how we always use the HAL APIs to handle gpio irq controller, I don't really see any usage for a parent reference.

drivers/platform/stm32/stm32_gpio_irq.c
rbolboac: In this case stm32 gpio controller does not allow to globally disable or enable all gpio interrupts so I decided to simply return -ENOSYS. However, for example on maxim platform, the gpio irq controller allows to disable/enable all gpio interrupts. So the general idea is that gpio_irq_global_enable/disable should be used to enable/disable all gpio specifc interrupts if possible, otherwise it should return -ENOSYS. On the other hand if a user wants to enable/disable global interrupts he should use the api from NVIC controller directly.

rbolboac:
- added script to generate get_irq_id_from_pin based on the available EXTI_x_IRQn values

buha (approved):
nice work !
now the user doesn't need to input anyting, the exti_script.py translates pin number to irq_id

PR #1416, opened 2022-06-16
drivers:adc:ad469x: Minor Updates.

pmallick-adi:
* Added macros for configuring reference control register.
* Updated the ad469x_std_pin_pairing() function so that only the desired
  bits of the register are updated.


amiclaus (approved):
lgtm, but please split into two commits since the two changes are unrelated

PR #1421, opened 2022-06-22
No os cn0561 wifi update

adrimbarean:
Update Wifi for the new interrupts and update iio_add1080 project for wifi purposes.


buha:
closing this for as it will be partially integrated when #1516 gets merged

PR #1435, opened 2022-07-11
drivers:adc:ad738x Added standard SPI support

mphalke:
Change 1:
Added standard SPI support for ad738x drivers to work with baremetal systems.
- The SPI engine functionality has been enabled by default. If macro 'USE_STANDARD_SPI' is added,
- the SPI engine code is replaced with standard no-os platform drivers SPI code.
- This change is similar to what has been done on ad4696 drivers.

Change 2:
Fixed the register data read function
As per ad7380 datasheet (Rev A, Page 27, section-ADDRESSING REGISTERS), to read the register data, the command needs to be send twice.
First word (2-bytes) of data read request, holds the command to inform register address to ADC.
Second word (2-bytes) of data request, is dummy read (NOP) to read the contents of that register.
Behavior is tested on SDP-K1 board using AD7380FMCZ board.

### DIFF
--- a/drivers/adc/ad738x/ad738x.c
+++ b/drivers/adc/ad738x/ad738x.c
@@ -44,7 +44,9 @@
 #include "stdio.h"
 #include "stdlib.h"
 #include "stdbool.h"
+#if !defined(USE_STANDARD_SPI)
 #include "spi_engine.h"
+#endif
 #include "ad738x.h"
 #include "no_os_delay.h"
 #include "no_os_error.h"
@@ -141,8 +143,8 @@ int32_t ad738x_spi_single_conversion(struct ad738x_dev *dev,
 	ret = no_os_spi_write_and_read(dev->spi_desc, buf, rx_buf_len);
 
 	/*
-	 *  Conversion data is 16 bits long in 1-wire mode and
-	 *  32 bits in 2-wire mode
+	 *  Conversion data is 32 bits long in 1-wire mode and
+	 *  16 bits in 2-wire mode
 	 */
 	adc_data[0] =  (buf[0] << 8) | buf[1];
 	adc_data[1] =  (buf[2] << 8) | buf[3];
@@ -280,6 +282,8 @@ int32_t ad738x_read_data(struct ad738x_dev *dev,
 			 uint16_t samples)
 {
 	int32_t ret;
+
+#if !defined(USE_STANDARD_SPI)
 	uint32_t commands_data[2] = {0, 0};
 	struct spi_engine_offload_message msg;
 	uint32_t spi_eng_msg_cmds[3] = {
@@ -303,10 +307,14 @@ int32_t ad738x_read_data(struct ad738x_dev *dev,
 
 	if (dev->dcache_invalidate_range)
 		dev->dcache_invalidate_range(msg.rx_addr, samples * 2);
+#else
+	ret = no_os_spi_write_and_read(dev->spi_desc, buf, samples);
+	if (ret != 0)
+		return ret;
+#endif
 
 	return ret;
 }
-
 
 /**
  * Initialize the device.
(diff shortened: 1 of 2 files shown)


drivers/adc/ad738x/ad738x.c
```diff
@@ -303,11 +307,15 @@ int32_t ad738x_read_data(struct ad738x_dev *dev,
 
 	if (dev->dcache_invalidate_range)
 		dev->dcache_invalidate_range(msg.rx_addr, samples * 2);
+#else
+	ret = no_os_spi_write_and_read(dev->spi_desc, buf, samples);
+	if (ret != 0)
```
amiclaus, line 318: nit: `if (ret)`

drivers/adc/ad738x/ad738x.c
```diff
@@ -325,15 +333,17 @@ int32_t ad738x_init(struct ad738x_dev **device,
...
 	dev->dcache_invalidate_range = init_param->dcache_invalidate_range;
+#endif
 	dev->conv_mode = init_param->conv_mode;
 	dev->ref_sel = init_param->ref_sel;
 
 	ret = no_os_spi_init(&dev->spi_desc, init_param->spi_param);
-
 	ret |= ad738x_reset(dev, HARD_RESET);
 	no_os_mdelay(1000);
+
```
amiclaus, line 352: ditto

drivers/adc/ad738x/ad738x.c
```diff
@@ -347,6 +357,7 @@ int32_t ad738x_init(struct ad738x_dev **device,
 
 	return ret;
 }
+
```
amiclaus, line 366: ditto

drivers/adc/ad738x/ad738x.c
```diff
@@ -68,10 +68,16 @@ int32_t ad738x_spi_reg_read(struct ad738x_dev *dev,
 	int32_t ret;
 	uint8_t buf[2];
 
+	/* Load the register (read) address */
 	buf[0] = AD738X_REG_READ(reg_addr);
 	buf[1] = 0x00;
+	ret = no_os_spi_write_and_read(dev->spi_desc, buf, 2);
```
amiclaus, line 74: perform `ret` check here.

drivers/adc/ad738x/ad738x.c
```diff
@@ -68,10 +68,16 @@ int32_t ad738x_spi_reg_read(struct ad738x_dev *dev,
...
 
+	/* Read the register content */
+	buf[0] = 0x00;
+	buf[1] = 0x00;
 	ret = no_os_spi_write_and_read(dev->spi_desc, buf, 2);
```
amiclaus, line 79: perform `ret` check here and return 0 at the end of the function.

drivers/adc/ad738x/ad738x.c
```diff
@@ -44,7 +44,9 @@
 #include "stdio.h"
 #include "stdlib.h"
 #include "stdbool.h"
+#if !defined(USE_STANDARD_SPI)
```
amiclaus, line 47: I think standard SPI should be the default code usage of the driver.
I'd do something like:
`if defined(USE_SPI_ENGINE)` insead.

mphalke: This make sense to me as well, however the original drivers were created to use with SPI engine, so retained that default functionality. The same thing was done for ad4696. I am okay with modifying this code to use standard SPI as default mode, so just let me know your inputs and I can modify it accordingly.

mphalke:
V1.0:
Review comment fixes:
- Reverted back the unrelated changes.
- Corrected/fixed the return checks for functions

amiclaus (approved):

PR #1440, opened 2022-07-19
Staging/adxcvr dev - Synchronization with Linux driver

danmois:
Synchronization of ADXCVR no-OS driver with the Linux variant.


drivers/axi_core/jesd204/axi_adxcvr.c
```diff
@@ -232,8 +233,8 @@ int adxcvr_clk_set_rate(struct adxcvr *xcvr,
 	uint32_t i;
 	int ret;
 
-//	dev_dbg(st->dev, "%s: Rate %lu Hz Parent Rate %lu Hz",
-//			__func__, rate, parent_rate);
+	pr_debug("%s: Rate %lu Hz Parent Rate %lu Hz\n",
```
amiclaus, line 236: it'd do this directly in the first commit, instead of commenting code and changing it afterwards. check the entire commit for similar changes.

amiclaus (commented):
looks good overall. you also might want to fit the commit message in the 50/72 rule.

amiclaus (approved):

drivers/axi_core/jesd204/axi_adxcvr.c
danmois: This was done so that specific modifications are separated in different commits. Squashing the commits would turn the PR into a large commit with a large amount of different modifications.

PR #1444, opened 2022-07-22
Removal of  warnings generated during make for the fmcdaq2 project

danmois:
Fixes for solving the problems that generate warnings during the make process for the fmcdaq2 project. 

These will decrease the number of warnings for other projects using the modified driver files.

### DIFF
--- a/drivers/dac/ad9144/ad9144.c
+++ b/drivers/dac/ad9144/ad9144.c
@@ -687,7 +687,6 @@ int32_t ad9144_short_pattern_test(struct ad9144_dev *dev,
 
 	uint32_t dac = 0;
 	uint32_t sample = 0;
-	uint8_t status = 0;
 	int32_t ret = 0;
 
 	for (dac = 0; dac < dev->num_converters; dac++) {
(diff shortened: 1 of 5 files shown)


drivers/frequency/ad9523/ad9523.c
```diff
@@ -235,7 +235,7 @@ int32_t ad9523_calibrate(struct ad9523_dev *dev)
 			AD9523_READBACK_1,
 			&reg_data);
 	if ((reg_data & 0x1) != 0x0) {
-		printf("AD9523: VCO calibration failed (%x)!\n", reg_data);
+		printf("AD9523: VCO calibration failed (0x%04lx)!\n", reg_data);
```
buha, line 238: just a suggestion,

instead of explicitly putting the 0x in front, you could add the # specifier after the % like:
%#04lx
which will add the 0x to the output for you.

drivers/dac/ad9144/ad9144.c
```diff
@@ -687,7 +687,6 @@ int32_t ad9144_short_pattern_test(struct ad9144_dev *dev,
 
 	uint32_t dac = 0;
 	uint32_t sample = 0;
-	uint8_t status = 0;
```
buha, line 690: please put in the whole word in the title "remove" instead of "rem"

no need to write the whole path drivers:dac:ad9144:ad9144.c
ad9144.c is enough, because this file is unique and resides only in one place in no-OS anyway

drivers/frequency/ad9523/ad9523.c
danmois: Thank you for the suggestion, this is very useful. The "0x" prefix is added, but it seems that this way, the padding does not work as expected.

danmois:
V2

- implemented review suggested modifications, where possible

danmois:
V3

- formatting of strings in printf calls was updated as suggested by review

buha (approved):

PR #1450, opened 2022-08-01
fmcjesdadc1 fixes

buha:
minor fixes and improvements in the ad9517 driver, used by the fmcjesdadc1 project

### DIFF
--- a/drivers/frequency/ad9517/ad9517.c
+++ b/drivers/frequency/ad9517/ad9517.c
@@ -287,6 +287,7 @@ int32_t ad9517_read(struct ad9517_dev *dev,
 	int32_t ret = 0;
 	uint8_t tx_buffer[3] = {0, 0, 0};
 	uint8_t i = 0;
+	*reg_value = 0;
 
 	reg_address = AD9517_READ + AD9517_ADDR(reg_addr);
 	for(i = 0; i < AD9517_TRANSF_LEN(reg_addr); i++) {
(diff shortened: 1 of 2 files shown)


drivers/frequency/ad9517/ad9517.c
```diff
@@ -78,6 +79,14 @@ int32_t ad9517_setup(struct ad9517_dev **device,
 
 	/* Initializes the SPI peripheral */
 	ret = no_os_spi_init(&dev->spi_desc, &init_param.spi_init);
+	if(ret < 0)
```
amiclaus, line 82: nit: why not `if (ret)`?

drivers/frequency/ad9517/ad9517.c
```diff
@@ -287,6 +287,7 @@ int32_t ad9517_read(struct ad9517_dev *dev,
 	int32_t ret = 0;
 	uint8_t tx_buffer[3] = {0, 0, 0};
 	uint8_t i = 0;
+	*reg_value = 0;
```
amiclaus, line 290: is this really necessary? 
it looks like `*reg_value <<= 8;` does something with the previous reg_value. it might be wrong though, but i'd rather change the implementation.

amiclaus (commented):
some minor comments on my side

drivers/frequency/ad9517/ad9517.c
buha: *reg_value <<= 8; was there for multi-byte registers.
And yes, this is really necessary, the reg_value contains junk otherwise in the 3 uppermost bytes (when reading 1 byte registers).

drivers/frequency/ad9517/ad9517.c
buha: true, will change

amiclaus (approved):

PR #1452, opened 2022-08-03
drivers:platform:maxim:max32670 Initial implementation of UART platform drivers

mphalke:
The baseline code for this is taken from UART implementation of MAX32660.
Below minor changes done w.r.t. to implementation:
1. Removed the switch case for UART 'mark' and 'space' parity as MAX32670 does not
support these parity modes. For odd and even parity, MAX32670 provides two modes-
mode0 and mode1. As no_os_uart_parity enum provides only odd and even parity settings,
mode0 of odd and even parity is used for max32670. This is done in uart init function.

2. MAX32670 does not support low and high flow control. It supports only enable/disable of flow control.
So, for both low and high flow enum switch case, MXC_UART_FLOW_EN value is set. This is done in uart init function.

3. 3rd argument to MXC_UART_Init() function is set as MXC_UART_APB_CLK. This is the max UART clock frequency
possible for this device (system_clock / 2 = 100Mhz/2 = 50Mhz).

4. Removed the FIFO implementation for uart write and read functions and replaced it wih non-fifo blocking
uart write/read. While transmitting data to IIO client over UART, it has been observed that FIFO implementation
is quite slower and take more execution time. While capturing ADC data in the background using interrupt and
periodically transmitting data over UART (based on request from IIO client), the FIFO implementation
results into loss of UART communication after running data capture for minute or so. With non-fifo
implementation, the UART communication and continuous data capturing works for longer time without any failure.
Both FIFO and non-fifo are of blocking type, so assumption is made that we have choice to use one of it.

### DIFF
--- a/drivers/platform/maxim/max32670/maxim_uart.c
+++ b/drivers/platform/maxim/max32670/maxim_uart.c
@@ -1,0 +1,401 @@
+/***************************************************************************//**
+ *   @file   maxim_uart.c
+ *   @brief  Source file of UART driver for STM32
+ *   @author Ciprian Regus (ciprian.regus@analog.com)
+********************************************************************************
+ * Copyright 2022(c) Analog Devices, Inc.
+ *
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *  - Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ *  - Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in
+ *    the documentation and/or other materials provided with the
+ *    distribution.
+ *  - Neither the name of Analog Devices, Inc. nor the names of its
+ *    contributors may be used to endorse or promote products derived
+ *    from this software without specific prior written permission.
+ *  - The use of this software may or may not infringe the patent rights
+ *    of one or more patent holders.  This license does not release you
+ *    from the requirement that you obtain separate licenses from these
+ *    patent holders to use this software.
+ *  - Use of the software either in source or binary form, must be run
+ *    on or directly connected to an Analog Devices Inc. component.
+ *
+ * THIS SOFTWARE IS PROVIDED BY ANALOG DEVICES "AS IS" AND ANY EXPRESS OR
+ * IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, NON-INFRINGEMENT,
+ * MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED.
+ * IN NO EVENT SHALL ANALOG DEVICES BE LIABLE FOR ANY DIRECT, INDIRECT,
+ * INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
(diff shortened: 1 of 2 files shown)


CiprianRegus:
Looks good! I only have one question. Have you tested the IIO client communication with a baud rate of 115200 or higher? From my experience, a few of these MCUs have an issue with the UART baud rate accuracy when the APB is used as a clock source. The error is ~2%, but it's enough to mess up the data transmission by 1 bit. If everything works well in your case using the APB is acceptable, but otherwise you may want to set the clock source to a dedicated baud rate generator one (it's usually 7.37MHz).

mphalke:
> Looks good! I only have one question. Have you tested the IIO client communication with a baud rate of 115200 or higher? From my experience, a few of these MCUs have an issue with the UART baud rate accuracy when the APB is used as a clock source. The error is ~2%, but it's enough to mess up the data transmission by 1 bit. If everything works well, using the APB is acceptable, but otherwise you may want to set the clock source to a dedicated baud rate generator one (it's usually 7.37MHz).

@CiprianRegus I tested at 115200 baud rate and IIO client communication works perfectly fine. Above 115200 it does not work. Even I was wondering why this is happening and assumed it's an issue with physical communication link. Thanks for clarifying this.
Could you please be more descriptive about this dedicated baud rate generator option? Is there any reference code from other Maxim MCUs?

CiprianRegus:
Hi @mphalke, the UART generator clock can be selected by using the `MXC_UART_IBRO_CLK` option instead of `MXC_UART_APB_CLK` in the `MXC_UART_Init` function call (although you may need to enable it first, I'm not sure). Currently, I haven't used it in other maxim drivers (because at the time, the MAX32650 which would require it had other baud rate computation bugs in the SDK, thus using the IBRO clock wouldn't have made a difference). Anyway, the idea is that 7.3728 MHz can be divided by a power of 2 to get a baud rate value (57600, 115200, etc) without any rounding. A short description of this clock source is given at page 40 in the MAX32670 user guide.

mphalke:
Thanks for the inputs @CiprianRegus. I will look into this and see if I can able to make it work above 115200.

drivers/platform/maxim/max32670/maxim_uart.c
```diff
@@ -0,0 +1,402 @@
...
+			if (ret)
+				return i ? (int32_t)i : -EAGAIN;
+		}
+		return i;
+	}
+
+	ret = MXC_UART_Read(MXC_UART_GET_UART(desc->device_id), data,
+			    (int *)&bytes_number);
+	if (ret != E_SUCCESS)
+		return ret;
```
CiprianRegus, line 109: This function should return the number of bytes actually read (which is a positive number). I see that unlike the FIFO implementation, this is not possible here, so returning `bytes_number` is fine by me. In case you're returning an error code, it must be a negative value defined in the `errno.h` header (which could be `-EIO` in this case) as opposed to a SDK defined err code.

mphalke: Make sense, will return -EIO from errno.h file

mphalke:
V1.0:
Review comment fixes:
- Changed the default UART clock frequency to IRBO (7.3728Mhz) for avoiding error in baud rate calculations.
- Returning -EIO error code from uart read/write functions in case of failure.

CiprianRegus (approved):

mphalke:
V2.0:

- Added 'MAX_' prefix to max_uart_flow_ctrl enum members.

amiclaus (approved):

PR #1454, opened 2022-08-03
drivers:adc:ad400x: Added macro USE_STANDARD_SPI

naveengs-adi:
SPI engine used in this driver is FGPA specific. Hence added a
macro to switch between using standard SPI for specifically with
MCU.

Signed-off-by: Naveen Godekeresiddaramaiah <Naveen.Godekeresiddaramaiah@analog.com>

### DIFF
--- a/drivers/adc/ad400x/ad400x.h
+++ b/drivers/adc/ad400x/ad400x.h
@@ -40,7 +40,13 @@
 #ifndef SRC_AD400X_H_
 #define SRC_AD400X_H_
 
+#include <stdbool.h>
+
+#if !defined(USE_STANDARD_SPI)
 #include "spi_engine.h"
+#else
+#include "no_os_spi.h"
+#endif
 
 /******************************************************************************/
 /********************** Macros and Constants Definitions **********************/
--- a/drivers/adc/ad400x/ad400x.c
+++ b/drivers/adc/ad400x/ad400x.c
@@ -44,7 +44,9 @@
 #include "stdlib.h"
 #include "stdbool.h"
 #include "ad400x.h"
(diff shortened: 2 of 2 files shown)


drivers/adc/ad400x/ad400x.h
```diff
@@ -40,7 +40,13 @@
 #ifndef SRC_AD400X_H_
 #define SRC_AD400X_H_
 
+#include <stdbool.h>
```
amiclaus, line 43: is this include related to the commit description/subject?

naveengs-adi: Yes, It is required.
* In ad400x.h, the ad400x_init_param struct contains few members of type bool.
* When spi_engine.h is included in ad400x.h, it internally includes spi_engine_private.h and it contains <stdbool.h>
* But, When we add compile time switch USE_STANDARD_SPI, it is required to include <stdbool.h> in header file ad400x.h.

buha (approved):
Looks good to me, although i think it'd be more intuitive to not use negation every time:
```
#if ! USE_STANDARD_SPI
    spi_engine...()
#endif
```

It could as well be:
```
#if USE_SPI_ENGINE
    spi_engine...()
#endif
```

I'm approving as-is, it's up to you to choose one or the other.

naveengs-adi:
> Looks good to me, although i think it'd be more intuitive to not use negation every time:
@buha Yes I agree with the statement.

I think we are good to go as is, because EVAL-AD400xFMCZ should support SPI Engine by default and the macro USE_STANDARD_SPI is the compile time switch that should be validated to check if we want to support these drivers usage with MCUs.

amiclaus (approved):

PR #1461, opened 2022-08-05
drivers:platform:aducm3029: UART init param changes and file renaming

rbolboac:
- Move UART initialization parameters (size, stop bits and parity) from extra to no_os_init_param for aducm3029 platform.
- Perform driver files renaming to aducm3029_drive format.


amiclaus (approved):
wondering if we can squash some of these commits so we can align to 5-6 commits per PR.

rbolboac:
> wondering if we can squash some of these commits so we can align to 5-6 commits per PR.

I would prefer to leave the commits as they are (each commit is for a driver).

CiprianRegus (approved):

PR #1467, opened 2022-08-12
drivers:platform:stm32:stm32_gpio.c Fix build for L5 family

rbolboac:
GPIO_MODE is defined in stm32l5xx_hal_gpio.c instead of stm32l5xx_hal_gpio.h so it has to be redefined in stm32_gpio.c before it is being used.
MODE_INPUT and MODE_OUTPUT are not defined in STM32 L5 hal, instead GPIO_MODE_INPUT and GPIO_MODE_OUTPUT_PP have to be used to obtain the same functionality.

### DIFF
--- a/drivers/platform/stm32/stm32_gpio.c
+++ b/drivers/platform/stm32/stm32_gpio.c
@@ -42,6 +42,10 @@
 #include "no_os_gpio.h"
 #include "stm32_gpio.h"
 
+#ifndef GPIO_MODE && defined(STM32L5)
+#define GPIO_MODE 0x00000003U
+#endif
+
 /******************************************************************************/
 /************************ Functions Definitions *******************************/
 /******************************************************************************/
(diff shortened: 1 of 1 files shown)


drivers/platform/stm32/stm32_gpio.c
```diff
@@ -42,6 +42,10 @@
 #include "no_os_gpio.h"
 #include "stm32_gpio.h"
 
+#ifndef GPIO_MODE && defined(STM32L5)
+#define GPIO_MODE (0x00000003U)
```
amiclaus, line 46: nit: do we need brackets here?

amiclaus (approved):
lgtm, one minor nit.

drivers/platform/stm32/stm32_gpio.c
rbolboac: I left the brackets there because I copied the definition directly from STM HAL.

amiclaus: I guess it depends on the family: (see line 142) https://www.disca.upv.es/aperles/arm_cortex_m3/llibre/st/STM32F439xx_User_Manual/stm32f4xx__hal__gpio_8c_source.html

rbolboac: I removed the brackets now. Let's wait for @buha 's approval before merging.

PR #1475, opened 2022-08-22
Dev/stm32 exti irq l5 support

rbolboac:
Add handling for HAL_GPIO_EXTI_Rising_Callback and HAL_GPIO_EXTI_Rising_Callback EXTI callbacks.
HAL_GPIO_EXTI_Callback is not defined in L5 family, instead HAL_GPIO_EXTI_Rising_Callback and HAL_GPIO_EXTI_Rising_Callback are used.

### DIFF
--- a/drivers/platform/stm32/stm32_gpio_irq.c
+++ b/drivers/platform/stm32/stm32_gpio_irq.c
@@ -83,10 +83,10 @@ static bool initialized[STM32_IRQ_CTRL_NB] =  {false};
 /************************ Functions Definitions *******************************/
 /******************************************************************************/
 /**
- * @brief EXTI GPIO Interrupt handler callback
+ * @brief Generic Interrupt handler callback
  * @param pin pin number on which the interrupt occurred (GPIO_PIN_pin)
  */
-void HAL_GPIO_EXTI_Callback(uint16_t pin)
+static inline void stm32_handle_generic_callback(uint16_t pin)
 {
 	int ret;
 	struct irq_action *action;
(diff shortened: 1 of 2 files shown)


drivers/platform/stm32/stm32_gpio_irq.c
```diff
@@ -83,28 +83,50 @@ static bool initialized[STM32_IRQ_CTRL_NB] =  {false};
 /************************ Functions Definitions *******************************/
 /******************************************************************************/
 /**
- * @brief EXTI GPIO Interrupt handler callback
+ * @brief Generic Interrupt handler callback
  * @param pin pin number on which the interrupt occurred (GPIO_PIN_pin)
  */
-void HAL_GPIO_EXTI_Callback(uint16_t pin)
+static void stm32_handle_generic_callback(uint16_t pin)
```
buha, line 89: I would also declare it `inline` so that there is 1 less entry in the execution call stack.

drivers/platform/stm32/stm32_gpio_irq.c
```diff
@@ -83,28 +83,50 @@ static bool initialized[STM32_IRQ_CTRL_NB] =  {false};
...
 {
 	int ret;
 	struct irq_action *action;
 	struct irq_action key;
 
 	key.irq_id = no_os_find_first_set_bit(pin);
 
-	if (__HAL_GPIO_EXTI_GET_IT(pin)) {
```
buha, line 97: is this removed by mistake ?

rbolboac: no, it is not removed by mistake. until now HAL_GPIO_EXTI_Callback was called directly from EXTI IRQ handler (due to the rename in stm32.mk script). Now, HAL_GPIO_EXTI_IRQHandler is called instead which is taking care of checking and clearing the interrupt.

rbolboac:
- made stm32_handle_generic_callback function inline

CiprianRegus (approved):
Tested with AD74413R.

amiclaus (approved):

PR #1481, opened 2022-08-25
stm32 spi: configure peripheral before transfer

buha:
Previously, the stm32_spi_write_and_read function did not
configure the spi peripheral with the settings stored in the
spi device descriptor. This means that it was impossible to use
the driver to drive multiple slaves with different spi settings (
bus speed, spi mode, bit order) on the same spi bus.

This commit changes the behavior of spi_write_and_read on stm32
platform to reconfigure the spi peripheral before each transfer if
needed.

The condition to assess the need for reconfiguration is based on
 - spi instance
 - chip select

If any of the above changes between consecutive transfers, the
reconfiguration takes place. If they stay the same, reconfiguration
is not performed (optimization).

Tested on hardware with adxrs290 project, used 4 MHz and 1 Mhz as maximum SPI speeds (init param) for 2 adxrs290 slaves.
With 24MHz AHB1 clock, nearest spi clock frequencies should be 3MHz and 750kHz which is exactly what was observed using a logic analyzer.

### DIFF
--- a/drivers/platform/stm32/stm32_spi.c
+++ b/drivers/platform/stm32/stm32_spi.c
@@ -44,65 +44,21 @@
 #include "no_os_spi.h"
 #include "stm32_spi.h"
 
-/**
- * @brief Initialize the SPI communication peripheral.
- * @param desc - The SPI descriptor.
- * @param param - The structure that contains the SPI parameters.
- * @return 0 in case of success, -1 otherwise.
- */
-int32_t stm32_spi_init(struct no_os_spi_desc **desc,
-		       const struct no_os_spi_init_param *param)
+static int stm32_spi_config(struct no_os_spi_desc *desc)
 {
-	int32_t ret;
-	uint32_t input_clock;
+	int ret;
+
 	const uint32_t prescaler_default = SPI_BAUDRATEPRESCALER_64;
 	const uint32_t prescaler_min = SPI_BAUDRATEPRESCALER_2;
 	const uint32_t prescaler_max = SPI_BAUDRATEPRESCALER_256;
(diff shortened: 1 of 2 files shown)


drivers/platform/stm32/stm32_spi.c
```diff
@@ -250,6 +270,19 @@ int32_t stm32_spi_write_and_read(struct no_os_spi_desc *desc,
...
+	// no need to reconfigure SPI. Otherwise, reconfigure it.
+	slave_id = ((uint64_t)(uintptr_t)sdesc->hspi.Instance << 32) |
+		   sdesc->chip_select->number;
+	if (slave_id != last_slave_id) {
+		ret = stm32_spi_config(desc);
+		if (ret)
+			return ret;
+	}
+
+	last_slave_id = slave_id;
```
amiclaus, line 284: this can go inside `if (slave_id != last_slave_id)`

amiclaus (commented):
looks good overall, might be a good idea to have a more generic approach in the future, not only for the stm32 platform

drivers/platform/stm32/stm32_spi.c
buha: yeah true, i will move it inside

buha:
@amiclaus i also think it would be good to move this stuff in the SPI thin-layer but then we'd need a no_os_spi_configure() function that is public

i think it's ok for now if we just implement [platform]_spi_configure() function only for the platforms that don't have this feature; it has less impact

buha:
V2:
- moved last_slave_id in the if condition, as per suggestion above

amiclaus (approved):

PR #1482, opened 2022-08-26
max11205 driver + iio support

rbolboac:
- max11205 driver
- max11205 basic example on maxim platform
- iio driver for max11205
- iio example for max11205 on maixm platform

### DIFF
--- a/drivers/adc/max11205/iio_max11205.c
+++ b/drivers/adc/max11205/iio_max11205.c
@@ -1,0 +1,302 @@
+/***************************************************************************//**
+ *   @file   iio_max11205.c
+ *   @brief  Implementation of iio max11205 driver
+ *   @author RBolboac (ramona.bolboaca@analog.com)
+ *******************************************************************************
+ * Copyright 2022(c) Analog Devices, Inc.
+ *
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *  - Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ *  - Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in
+ *    the documentation and/or other materials provided with the
+ *    distribution.
+ *  - Neither the name of Analog Devices, Inc. nor the names of its
+ *    contributors may be used to endorse or promote products derived
+ *    from this software without specific prior written permission.
+ *  - The use of this software may or may not infringe the patent rights
+ *    of one or more patent holders.  This license does not release you
+ *    from the requirement that you obtain separate licenses from these
+ *    patent holders to use this software.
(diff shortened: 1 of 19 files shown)


amiclaus (commented):
some comments on my side.

drivers/adc/max11205/max11205.c
```diff
@@ -0,0 +1,239 @@
...
+ * @return 0 in case of success, negative error code otherwise.
+ */
+int max11205_get_data_mv(struct max11205_dev *dev, int16_t raw_data,
+			 int32_t *data_mv)
+{
+	if(!dev || !data_mv)
+		return -EINVAL;
+
+	*data_mv = NO_OS_DIV_ROUND_CLOSEST(raw_data*dev->vref_mv,
+					   NO_OS_BIT(15)-1);
```
amiclaus, line 203: also, i think we can use NO_OS_GENMASK here.

rbolboac:
- implemented review findings

amiclaus (approved):
lgtm, some minor comments

drivers/adc/max11205/iio_max11205.c
```diff
@@ -0,0 +1,261 @@
...
+ */
+static int max11205_iio_read_raw(void *dev, char *buf, uint32_t len,
+				 const struct iio_ch_info *channel, intptr_t priv)
+{
+	int ret;
+	int32_t val;
+	int16_t data_raw;
+	bool new_data_avail = false;
+	struct max11205_iio_dev *iio_max11205;
+	uint32_t timeout = NO_OS_GENMASK(31,0);
```
CiprianRegus, line 74: Is there a point in using such high of a value for this timeout? Wouldn't `no_os_mdelay()` fit better?

rbolboac: The timeout is used as an exit condition of the while loop such that an infinite loop is avoided in case something happens with the device and it does not work anymore (no data ready interrupts are generated anymore). The timeout is a relatively large value to make sure platform with high CPU frequencies also allow for a minimum data conversion time (for max11205b can be up to 73ms). Taking for example a 120MHz processor, it will take about 36 seconds*number of instructions needed to execute the loop to reach the timeout. Of course, it can be lowered but the main reason for using the timeout was to avoid an infinite while loop.
no_os_mdelay would not fit in this scenario because the idea is to read the data as soon as new data is available, without any further delays.

CiprianRegus (approved):

drivers/adc/max11205/iio_max11205.c
rbolboac: The timeout value was lowered. The timeout will now take 100msec * number of instructions needed to execute the loop instructions.

rbolboac:
- implemented review findings
- changed iio device name from max11205 to max11205a

rbolboac:
- added sampling_frequency attribute based on the selected device id

buha (approved):

PR #1492, opened 2022-09-09
iio: Add cyclic buffer support

rbolboac:
Add cyclic buffer support when cyclic flag is given in OPEN command. The cyclic flag is kept until a CLOSE command is received.
When a device is opened in cyclic mode, the first buffer of samples pushed by the client, will be pushed continuously to the iio driver. submit function will be called continuously with the given samples.

### DIFF
--- a/iio/iio.c
+++ b/iio/iio.c
@@ -1068,6 +1068,9 @@ static int iio_open_dev(struct iiod_ctx *ctx, const char *device,
 	if (!mask)
 		return -ENOENT;
 
+	dev->buffer.public.cyclic_info.is_cyclic = cyclic;
+	dev->buffer.public.cyclic_info.buff_index = 0;
+
 	dev->buffer.public.active_mask = mask;
 	dev->buffer.public.bytes_per_scan =
 		bytes_per_scan(dev->dev_descriptor->channels, mask);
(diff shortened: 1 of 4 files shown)


amiclaus (approved):

iio/iio.c
```diff
@@ -1172,6 +1175,8 @@ static int iio_call_submit(struct iiod_ctx *ctx, const char *device,
 		return -EINVAL;
 
 	dev->buffer.public.dir = dir;
+	dev->buffer.public.cyclic_info.buff_index = 0;
```
pmallick-adi, line 1178: Hi @rbolboac,
Thanks for this PR.

I have one comment here. 
Should this line be inside the `if (dev->dev_descriptor->submit && dev->trig_idx==NO_TRIGGER)` ?
Doing this will ensure that the buff_index is being reset only when the submit function is defined and there are no triggers.

If the user makes use of trigger to push data to DAC, then resetting the cyclic buffer index to zero should be take care by the application layer or should be take care in the iio_buffer_pop scan function?

buha (approved):

iio/iio.c
rbolboac: Hi @pmallick-adi . I did not take into consideration the usage of triggers. You are right. Some adaptations have to be done for these scenario. I am currently working on implementing the needed changes.

pmallick-adi: Thanks for considering this.

rbolboac:
- added support for cyclic buffers with triggers

PR #1494, opened 2022-09-15
IIOD pointer fix

rbolboac:
- Reset the trigger id assigned to a device when the close command is received.
- Use array of chars instead of pointers for device id, trigger id, channel id and attribute name. With current implementation the pointers point to parser_buf which can change over the execution of the program, leading to unwanted changes for device id, trigger id, channel id or attribute name. This can happen when iiod_read_line reads a few characters, modifying parser_buf, but then returns an error (because not all data has been yet received). In this scenario iiod_parse_line is not called and device id, trigger id, channel id or attribute name will have garbage values until iiod_parse_line is called again.

### DIFF
--- a/iio/iiod.h
+++ b/iio/iiod.h
@@ -48,6 +48,11 @@
 #define IIOD_VERSION		"1.1.0000000"
 #define IIOD_VERSION_LEN	(sizeof(IIOD_VERSION) - 1)
 
+#define MAX_DEV_ID		64
+#define MAX_TRIG_ID		64
+#define MAX_CHN_ID		64
+#define MAX_ATTR_NAME		256
+
 enum iio_attr_type {
 	IIO_ATTR_TYPE_DEBUG,
 	IIO_ATTR_TYPE_BUFFER,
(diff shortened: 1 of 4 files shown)


amiclaus (approved):
lgtm

CiprianRegus (approved):

iio/iiod.h
```diff
@@ -48,6 +48,11 @@
 #define IIOD_VERSION		"1.1.0000000"
 #define IIOD_VERSION_LEN	(sizeof(IIOD_VERSION) - 1)
 
+#define MAX_DEV_ID		21
```
buha, line 51: I'd say that rather do some bug hunting in the future, perhaps it's better to just increase 21 and 22 to 32 or 64. It's still better than 256 (linux) but large enough so that we won't ever be increasing it again.

rbolboac:
- changed MAX_TRIG_ID and MAX_DEV_ID to 64.

iio/iiod.h
rbolboac: done. thanks!

PR #1495, opened 2022-09-23
projects: ad7616 updates

amiclaus:
- use pr_info 
- update buf to array 
-  print samples


amiclaus (commented):
v2:
- print each 16-bit sample independently.

buha (approved):

PR #1500, opened 2022-09-28
iio: Updates and issue fixes related to IIO context attributes

mphalke:
In previous implementation of IIO context attributes, the attributes were tied-up with IIO device. However, the context attributes must be global and independent of devices/triggers. In case if none of the device is present (or not initialized), context attributes must be populated into IIO client to convey IIO/firmware global information.
Updated code to remove device dependency for context attributes and tested with IIO oscilloscope client.

### DIFF
--- a/iio/iio.c
+++ b/iio/iio.c
@@ -145,6 +145,15 @@ struct iio_buffer_priv {
 };
 
 /**
+ * @struct iio_cntx_attr_priv
+ * @brief Context attributes private structure
+ */
+struct iio_cntx_attr_priv {
+	/** Pointer to list of context attributes (name and value) */
+	struct iio_context_attribute *attributes;
+};
+
+/**
  * @struct iio_dev_priv
  * @brief Links a physical device instance "void *dev_instance"
  * with a "iio_device *iio" that describes capabilities of the device.
@@ -192,6 +201,8 @@ struct iio_desc {
 	void			*phy_desc;
 	char			*xml_desc;
 	uint32_t		xml_size;
+	struct iio_cntx_attr_priv *cntx_attributes;
+	uint32_t nb_cntx_attr;
 	struct iio_dev_priv	*devs;
 	uint32_t		nb_devs;
 	struct iio_trig_priv	*trigs;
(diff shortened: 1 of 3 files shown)


rbolboac:
Hi @mphalke,

I have some remarks.
Indeed the current implementation links the context to the device. However, that part of the code is never used in our repo.
The context is visible with and without the changes you provided, because it is actually added in the XML header in iio.c as follows:
```
"<context name=\"xml\" description=\"no-OS analog 1.1.0-g0000000 #1 Tue Nov 26 09:52:32 IST 2019 armv7l\" >"
"<context-attribute name=\"no-OS\" value=\"1.1.0-g0000000\" />";
```

I am curious about the use-case, do you plan to add context attributes from application level?
Are the context attributes specific to the running application, or rather specific to the IIO daemon? 
In your opinion, where should new context attributes be added?

With the implementation you provided, cntx_attrs structure in iio_init_param is never populated. 
If the purpose is to populate the attributes from application level, then further changes have to be made.
The application will always have to call either iio_app_run_with_trigs if triggers are needed or iio_app_run if triggers are not needed. With the current implementation, the parameters of iio_app_run_with_trigs or iio_app_run do not contain any context attributes, and thus iio_init_param will not contain any context attributes.

mphalke:
@rbolboac 
I was the one who implemented previous version of context attributes. However, I discovered that it is not correct way as if none of the device is initialized in the firmware, then it should display at-least context attributes on IIO client.

Context attributes can be added from application layer. I think they are specific to running application and not specific to any device.

This is the way someone can create context attributes:

```
void main()
{
     struct iio_cntx_attr_init iio_cntx_attr_init_params;

     get_iio_context_attributes(&iio_cntx_attr_init_params,
			&iio_init_params.nb_cntx_attrs);

     iio_init_params.cntx_attrs = &iio_cntx_attr_init_params;

     iio_init(&p_ad7689_iio_desc, &iio_init_params);
}

void get_iio_context_attributes(struct iio_cntx_attr_init *params,
		uint32_t *attrs_cnt)
{
         struct iio_context_attribute *context_attributes;
	context_attributes = (struct iio_context_attribute *)calloc(1,
			     sizeof(struct iio_context_attribute) * num_of_context_attributes);
	if (!context_attributes) {
		return -ENOMEM;
	}
        
       // Add context attributes into context_attributes  structure

        params->descriptor = context_attributes;
	*attrs_cnt = num_of_context_attributes;
}
```

I have added condition in iio.c file that ensures context attributes are not stored into xml if 'iio_cntx_attr_init_params' structure is NULL (not defined) or parameters of iio_cntx_attr_init_params are NULL

```
(iio.c, Line 1846)
if (init_param->cntx_attrs && init_param->cntx_attrs->descriptor) {
		ret = iio_init_contxt_attrs(ldesc, init_param->cntx_attrs,
					    init_param->nb_cntx_attrs);
		if (NO_OS_IS_ERR_VALUE(ret))
			goto free_devs;
	}
```

rbolboac:
@mphalke ,
Thank you for explaining the use-case you had in mind.

At the moment, no project in no-os calls specifically iio_init function, instead, iio_app_run_with_trigs or iio_app_run are called.
I think these are the functions the application is supposed to call because they are doing more than calling iio_init. They are also configuring uart/networking and irq capabilities for iio.
With this in mind, I am wondering if we should also add the context attributes as parameter for iio_app_run_with_trigs and iio_app_run APIs.
Otherwise, if the users choose to use iio_app_run_with_trigs or iio_app_run, they won't be able set context attributes.
If they choose to use iio_init, following the example you provided, they will have to take care of uart/network and irq setup and things get complicated.

Let me know what you think.

mphalke:
Hi @rbolboac 
In terms of using additional argument to 'iio_app_run_with_trigs ' or 'iio_app_run' functions to pass context attribute also make sense (but totally optional), particularly for no-os projects. I agree some additional changes needed to make this work in these 2 functions. So, I think it depends upon what you and your team think about it.

mphalke:
Hi, Any thoughts/inputs on above comment and PR changes?

rbolboac:
Hi @mphalke ,
In my opinion these changes can be merged now, and then in the future we can add support for context attributes also in  iio_app_run_with_trigs and iio_app_run.
I will also talk with my colleagues tomorrow in our sync team about this and add it to our todo list.

rbolboac (approved):

mphalke:
> Hi @mphalke , In my opinion these changes can be merged now, and then in the future we can add support for context attributes also in iio_app_run_with_trigs and iio_app_run. I will also talk with my colleagues tomorrow in our sync team about this and add it to our todo list.

Thanks Ramona.

buha (approved):
separating context from device is a good thing

amiclaus (approved):

PR #1502, opened 2022-09-29
drivers:platform:pico:pico_spi: Add support for multiple slaves

rbolboac:
Add support for spi driver on pico platform such that communication is supported with multiple spi slaves, with different bus configuration.

PR #1509, opened 2022-10-03
Maxim SPI reconfiguration

CiprianRegus:
- implement SPI reconfiguration before transfers.
- replace `spi_write_and_read()` with a `spi_transfer()` call.

### DIFF
--- a/drivers/platform/maxim/max32650/maxim_spi.c
+++ b/drivers/platform/maxim/max32650/maxim_spi.c
@@ -142,6 +142,57 @@ static int32_t _max_spi_enable_ss(uint32_t id, uint32_t chip_select)
 }
 
 /**
+ * @brief Configure a SPI peripheral
+ * @param spi - The SPI descriptor which stores the configuration parameters.
+ * @return 0 in case of success, a negative error code otherwise
+ */
+static int _max_spi_config(struct no_os_spi_desc *desc)
+{
+	struct max_spi_init_param *eparam;
+	int32_t ret;
+
+	eparam = desc->extra;
+
+	ret = MXC_SPI_Init(MXC_SPI_GET_SPI(desc->device_id), SPI_MASTER_MODE,
+			   SPI_SINGLE_MODE,
+			   eparam->numSlaves, eparam->polarity, desc->max_speed_hz);
+	if (ret) {
+		ret = -EINVAL;
+		goto err_init;
+	}
+
+	ret = _max_spi_enable_ss(desc->device_id, desc->chip_select);
+	if (ret) {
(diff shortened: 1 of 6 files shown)


drivers/platform/maxim/max32650/maxim_spi.c
```diff
@@ -169,6 +169,7 @@ int32_t max_spi_init(struct no_os_spi_desc **desc,
 	descriptor->mode = param->mode;
 	descriptor->bit_order = param->bit_order;
 	descriptor->platform_ops = &max_spi_ops;
+	descriptor->extra = eparam;
```
amiclaus, line 223: i think all the changes in this commit should be distributed to each of the next commits, corresponding to each MAX part.

CiprianRegus:
- Implemented review change request.
- Removed unused variable.

drivers/platform/maxim/max32670/maxim_spi.c
```diff
@@ -199,12 +175,23 @@ int32_t max_spi_transfer(struct no_os_spi_desc *desc,
...
 {
+	static uint32_t last_slave_id[MXC_SPI_INSTANCES];
 	mxc_spi_req_t req;
+	uint32_t slave_id;
 	int32_t ret;
 
 	if (!desc || !msgs)
 		return -EINVAL;
 
+	slave_id = desc->chip_select;
```
buha, line 186: slave_id should be a combination of device_id and chip_select

buha: like here: https://github.com/analogdevicesinc/no-OS/pull/1502/commits/3743b17170792b61507e2ffcfe5e3543b2c1a5d2#diff-9072e7b9877324e833f75a13a501eb0af2c8a032d8280e44e9f583bd3c8e5661R230

CiprianRegus: `last_slave_id` is now an array, so the behavior is the same, since we save the chip_select number for each SPI interface individually. I think this is better than the current implementation, since there could be multiple SPI peripherals, and the following transfer sequence may occur:

[SPI0 | CS0] ... [SPI1 | CS0] ... [SPI0 | CS0]

In this case, SPI0 will be reconfigured before the last transfer, even though it's not needed, since we're talking with the same device as in the first transfer.

buha: true, you're right, i guess we need to change the other implementations to follow this logic, they are now too aggressive with changing the settings :D

buha (approved):

PR #1520, opened 2022-10-06
drivers:platform:mbed Added I2C frequency set function call

mphalke:
Added mbed API call to set the I2C frequency set in init parameters.


amiclaus (approved):
ltgm, please use path / partial path in the commit title, something like:
`platform: mbed: Add I2C frequency set function call`

CiprianRegus (approved):

PR #1535, opened 2022-10-12
Set explicit CS on Maxim platform

CiprianRegus:
- Set the CS line before a SPI transfer.


amiclaus (approved):

CiprianRegus:
- Did the change for the MAX32665 too.

buha (approved):

PR #1539, opened 2022-10-17
ADXL359

rbolboac:
- add support for adxl359 device (driver, iio driver and project example)

### DIFF
--- a/drivers/accel/adxl355/adxl355.c
+++ b/drivers/accel/adxl355/adxl355.c
@@ -49,6 +49,10 @@
 /******************************************************************************/
 static uint8_t shadow_reg_val[5] = {0, 0, 0, 0, 0};
 static const uint8_t adxl355_scale_mul[4] = {0, 1, 2, 4};
+static const uint8_t adxl355_part_id[2] = {
+	[ID_ADXL355] = GET_ADXL355_RESET_VAL(ADXL355_PARTID),
+	[ID_ADXL359] = GET_ADXL355_RESET_VAL(ADXL359_PARTID),
+};
 
 /******************************************************************************/
 /************************ Functions Declarations ******************************/
@@ -56,7 +60,7 @@ static const uint8_t adxl355_scale_mul[4] = {0, 1, 2, 4};
 static uint32_t adxl355_accel_array_conv(struct adxl355_dev *dev,
 		uint8_t *raw_array);
 static int64_t adxl355_accel_conv(struct adxl355_dev *dev, uint32_t raw_accel);
-static int64_t adxl355_temp_conv(uint16_t raw_temp);
+static int64_t adxl355_temp_conv(struct adxl355_dev *dev, uint16_t raw_temp);
 
 /******************************************************************************/
 /************************ Functions Definitions *******************************/
@@ -140,6 +144,9 @@ int adxl355_init(struct adxl355_dev **device,
 	int ret;
 	uint8_t reg_value;
 
+	if(init_param.dev_type != ID_ADXL355 && init_param.dev_type != ID_ADXL359)
+		return -EINVAL;
+
 	dev = (struct adxl355_dev *)calloc(1, sizeof(*dev));
 
 	if (!dev)
@@ -157,6 +164,8 @@ int adxl355_init(struct adxl355_dev **device,
 			goto error_dev;
 	}
 
+	dev->dev_type = init_param.dev_type;
+
 	ret = adxl355_read_device_data(dev, ADXL355_ADDR(ADXL355_DEVID_AD),
 				       GET_ADXL355_TRANSF_LEN(ADXL355_DEVID_AD), &reg_value);
 	if (ret || (reg_value != GET_ADXL355_RESET_VAL(ADXL355_DEVID_AD)))
@@ -169,7 +178,8 @@ int adxl355_init(struct adxl355_dev **device,
 
 	ret = adxl355_read_device_data(dev, ADXL355_ADDR(ADXL355_PARTID),
 				       GET_ADXL355_TRANSF_LEN(ADXL355_PARTID),&reg_value);
-	if (ret || (reg_value != GET_ADXL355_RESET_VAL(ADXL355_PARTID)))
+
+	if (ret || reg_value != adxl355_part_id[dev->dev_type])
 		goto error_com;
 
 	// Get shadow register values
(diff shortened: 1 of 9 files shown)


amiclaus (commented):
some minor comments on my side

drivers/accel/adxl355/adxl355.c
```diff
@@ -182,6 +195,7 @@ int adxl355_init(struct adxl355_dev **device,
 
 	// Default range is set
 	dev->range = GET_ADXL355_RESET_VAL(ADXL355_RANGE) & ADXL355_RANGE_FIELD_MSK;
+	adxl355_set_range(dev, 2);
```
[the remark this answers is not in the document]
rbolboac: this was a line added for testing, it has already been removed from the code

rbolboac:
- implemented review changes (using switch cases instead of if else and removed enum explicit initialization)

drivers/accel/adxl355/adxl355.c
```diff
@@ -169,8 +174,21 @@ int adxl355_init(struct adxl355_dev **device,
...
+
+	switch(dev->dev_type) {
+	case ID_ADXL355:
+		if (reg_value != GET_ADXL355_RESET_VAL(ADXL355_PARTID))
+			goto error_com;
+		break;
+	case ID_ADXL359:
+		if (reg_value != GET_ADXL355_RESET_VAL(ADXL355_PARTID))
+			goto error_com;
+		break;
```
CiprianRegus, line 188: These cases look the same.

drivers/accel/adxl355/adxl355.c
```diff
@@ -924,8 +962,16 @@ static int64_t adxl355_accel_conv(struct adxl355_dev *dev,
...
-		(int64_t)ADXL355_TEMP_SCALE_FACTOR);
+	switch(dev->dev_type) {
+	case ID_ADXL355:
+		return ((raw_temp*ADXL355_TEMP_OFFSET_DIV +  ADXL355_TEMP_OFFSET) *
+			(int64_t)ADXL355_TEMP_SCALE_FACTOR);
+	case ID_ADXL359:
+		return ((raw_temp*ADXL359_TEMP_OFFSET_DIV +  ADXL359_TEMP_OFFSET) *
+			(int64_t)ADXL359_TEMP_SCALE_FACTOR);
+	default:
+		return 0;
```
CiprianRegus, line 975: `-EINVAL`?

drivers/accel/adxl355/iio_adxl355.c
```diff
@@ -414,11 +414,33 @@ static int adxl355_iio_read_offset(void *dev, char *buf, uint32_t len,
 				   const struct iio_ch_info *channel, intptr_t priv)
 {
 	int32_t vals[2];
+	struct adxl355_iio_dev *iio_adxl355;
+	struct adxl355_dev *adxl355;
+
+	if (!dev)
+		return -EINVAL;
+
+	iio_adxl355 = (struct adxl355_iio_dev *)dev;
```
CiprianRegus, line 423: You don't need to cast.

drivers/accel/adxl355/iio_adxl355.c
```diff
@@ -441,11 +463,33 @@ static int adxl355_iio_read_scale(void *dev, char *buf, uint32_t len,
 				  const struct iio_ch_info *channel, intptr_t priv)
 {
 	int32_t vals[2];
+	struct adxl355_iio_dev *iio_adxl355;
+	struct adxl355_dev *adxl355;
+
+	if (!dev)
+		return -EINVAL;
+
+	iio_adxl355 = (struct adxl355_iio_dev *)dev;
```
CiprianRegus, line 472: Ditto.

projects/eval-adxl355-pmdz/src/platform/maxim/main.c
```diff
@@ -100,6 +98,8 @@ int main()
 		return ret;
 
 	ret = iio_trigger_example_main();
+#else
+#error IIO_TRIGGER_EXAMPLE is currently supported only on max32655 and max78000 targets.
```
CiprianRegus, line 102: Isn't this check already done in the `parameters.h` file?

drivers/accel/adxl355/adxl355.c
```diff
@@ -912,8 +942,16 @@ static int64_t adxl355_accel_conv(struct adxl355_dev *dev,
...
-			  adxl355_scale_mul[dev->range]));
+	switch (dev->dev_type) {
+	case ID_ADXL355:
+		return ((int64_t)(accel_data * ADXL355_ACC_SCALE_FACTOR_MUL *
+				  adxl355_scale_mul[dev->range]));
+	case ID_ADXL359:
+		return ((int64_t)(accel_data * ADXL359_ACC_SCALE_FACTOR_MUL *
+				  adxl355_scale_mul[dev->range]));
+	default:
+		return 0;
```
CiprianRegus, line 953: `-EINVAL` ?

drivers/accel/adxl355/adxl355.c
rbolboac: in this case the returned value can also be negative. the default case will never be reached. 0 return value in case of an error is suitable in this case.

drivers/accel/adxl355/adxl355.c
rbolboac: in this case the returned value can also be negative. the default case will never be reached. 0 return value in case of an error is suitable in this case.

drivers/accel/adxl355/adxl355.c
rbolboac: Thanks!

projects/eval-adxl355-pmdz/src/platform/maxim/main.c
rbolboac: yes, it is already done, I removed it from the main file now.

rbolboac:
- implemented review findings (added correct part id checking, removed not needed cast, removed not needed target check)

drivers/accel/adxl355/adxl355.c
```diff
@@ -612,10 +630,22 @@ int adxl355_get_temp(struct adxl355_dev *dev, struct adxl355_frac_repr *temp)
 	int ret;
 
 	ret = adxl355_get_raw_temp(dev, &raw_temp);
-	if (!ret)
-		temp->integer = no_os_div_s64_rem(adxl355_temp_conv(raw_temp),
-						  ADXL355_TEMP_OFFSET_DIV*ADXL355_TEMP_SCALE_FACTOR_DIV,
-						  &(temp->fractional));
+	if (!ret) {
```
buha, line 633: If we can simplify code by reducing the amount of nested statements we should do so. Here it's better to:
```
if (ret)
    return ret;
// continue here
```
instead of:
```
if (!ret) {
    // continue here
}
```

drivers/accel/adxl355/adxl355.c
```diff
@@ -612,10 +630,22 @@ int adxl355_get_temp(struct adxl355_dev *dev, struct adxl355_frac_repr *temp)
...
+		switch(dev->dev_type) {
```
buha, line 634: This switch has almost identical code in its branches. I think duplicating code like that results in higher code size in .text and we should avoid it. One idea is to create a variable for that one parameter that is different, and call the function after the switch.

drivers/accel/adxl355/adxl355.c
```diff
@@ -169,8 +174,21 @@ int adxl355_init(struct adxl355_dev **device,
 
 	ret = adxl355_read_device_data(dev, ADXL355_ADDR(ADXL355_PARTID),
 				       GET_ADXL355_TRANSF_LEN(ADXL355_PARTID),&reg_value);
-	if (ret || (reg_value != GET_ADXL355_RESET_VAL(ADXL355_PARTID)))
+	if (ret)
+		goto error_com;
+
+	switch(dev->dev_type) {
```
buha, line 180: you could get rid of this switch with a simple array where index is ID_ADXL... which contains PARTID's.

rbolboac:
- simplified code as per coded review findings

buha (approved):

amiclaus (approved):

CiprianRegus (approved):

PR #1540, opened 2022-10-18
drivers:eeprom:24xx32a setting i2c stop bit during eeprom address write

mphalke:
As per datasheet, the stop condition is not expected after sending address bytes. Instead, repeated start condition needs to be generated (which would be done by subsequent i2c_read function call). Therefore, set stop bit to 0.


amiclaus (approved):
please comply to the 50/72 commit message rule..

mphalke:
> please comply to the 50/72 commit message rule..

Sure, will keep in mind. I have updated the commit message.

mphalke:
V1.0:

- Updated the commit message (no functionality changes done)

buha (approved):

PR #1541, opened 2022-10-18
iio: Fixed bug in storing context attrs into xml string

mphalke:
Corrected the for loop condition while copying context attr values into IIO xml string.


buha (approved):
looks good, i see that desc->nb_cntx_attr is initialized in iio_init_contxt_attrs

PR #1543, opened 2022-10-25
MAX32665: change UART pins mapping

CiprianRegus:
Set the UART mapping, so that the RX/TX signals will be routed to the JTAG header on the MAX32665 FTHR board. Even though this is board specific, I think it would be useful to have these changes applied, as it's the only MAX32665/6 board we're using right now.

The MAP_X sets the GPIOs that will be configured as UART.
MAP_A: pins 20, 21.
MAP_B: pins 12, 13.


amiclaus (commented):
can't we make this configurable by the user?

amiclaus (approved):

PR #1544, opened 2022-10-25
JTAG_CABLE_ID specifier for maxim platform

buha:
To find the serial number of the CMSIS-DAP interface:
	lsusb -v
	lsusb -v | grep iSerial

Then pass the serial number to the run command:
	make run JTAG_CABLE_ID=0123456789abcdef

This allows to target one particular CMSIS-DAP programmer if several of them are connected to the PC.


buha:
we should propagate this to all platforms that use openocd
tested both on windows and linux, merging now

PR #1545, opened 2022-10-25
JTAG_CABLE_ID for aducm3029 platform

buha:
Move part of the logic to generic.mk (previously in maxim.mk). Use the generated .id file to specify the JTAG device serial no for both aducm3029 and maxim platforms.


CiprianRegus (approved):
Tested on Windows.

PR #1550, opened 2022-11-11
drivers:platform:mbed Update i2c slave address dynamically

mphalke:
Added function to update/store i2c slave address at run time, as there are few cases where data needs to be read from different slaves without reinitializing i2c drivers

### DIFF
--- a/drivers/platform/mbed/mbed_i2c.cpp
+++ b/drivers/platform/mbed/mbed_i2c.cpp
@@ -97,6 +97,7 @@ int32_t mbed_i2c_init(struct no_os_i2c_desc **desc,
 	i2c->frequency(param->max_speed_hz);
 
 	mbed_i2c_desc->i2c_port = i2c;
+	mbed_i2c_desc->i2c_slave_addr = i2c_desc->slave_address;
 	i2c_desc->extra = mbed_i2c_desc;
 
 	*desc = i2c_desc;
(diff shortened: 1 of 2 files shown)


amiclaus (approved):

mphalke:
Hi, are we good to merge these changes to master branch?

buha:
So @mphalke ,
Every platform implementation of the no-OS I2C API uses desc->slave_address. Examples:
https://github.com/analogdevicesinc/no-OS/blob/master/drivers/platform/mbed/mbed_i2c.cpp#L163
https://github.com/analogdevicesinc/no-OS/blob/master/drivers/platform/maxim/max32650/maxim_i2c.c

What you could do for now is to simply change that in application code prior to i2c_write()/i2c_read() by explictly assigning device.slave_address = 0x34;
i2c_write(device...)

This is not elegant but it's still better than duplicating that field.

I plan on doing something similar on I2C as we've done it for SPI so that you have a device descriptor per slave and the write/read functions would take care that the peripheral gets reinitialized (or not) as needed.
https://github.com/analogdevicesinc/no-OS/blob/master/drivers/platform/maxim/max32650/maxim_spi.c#L280-L287

mphalke:
Hi @buha 
Thanks for the inputs. I think this is simpler solution and make sense and will be consistent across platforms.
We can delete this PR now.

PR #1552, opened 2022-11-14
adc: ad463x: Implement ad463x_spi_reg_read_masked()

dbogdan:
The function was present in the header file but its implementation was missing from the source file.

### DIFF
--- a/drivers/adc/ad463x/ad463x.c
+++ b/drivers/adc/ad463x/ad463x.c
@@ -101,6 +101,31 @@ int32_t ad463x_spi_reg_write(struct ad463x_dev *dev,
 }
 
 /**
+ * @brief SPI read device register using a mask.
+ * @param dev - The device structure.
+ * @param reg_addr - The register address.
+ * @param mask - The mask.
+ * @param data - The data read from the register.
+ * @return 0 in case of success, negative error code otherwise.
+ */
+int32_t ad463x_spi_reg_read_masked(struct ad463x_dev *dev,
+				   uint16_t reg_addr,
+				   uint8_t mask,
+				   uint8_t *data)
+{
+	uint8_t reg_data;
+	int32_t ret;
+
+	ret = ad463x_spi_reg_read(dev, reg_addr, &reg_data);
+	if (ret != 0)
+		return ret;
+
+	*data = no_os_field_get(mask, reg_data);
+
+	return ret;
+}
+
+/**
  * @brief SPI write device register using a mask.
  * @param dev - The device structure.
  * @param reg_addr - The register address.


drivers/adc/ad463x/ad463x.c
```diff
@@ -100,6 +100,31 @@ int32_t ad463x_spi_reg_write(struct ad463x_dev *dev,
...
+ * @param mask - The mask.
+ * @param data - The data read from the register.
+ * @return 0 in case of success, negative error code otherwise.
+ */
+int32_t ad463x_spi_reg_read_masked(struct ad463x_dev *dev,
+				    uint16_t reg_addr,
+				    uint8_t mask,
+				    uint8_t *data)
+{
+	uint8_t reg_data;
```
amiclaus, line 116: nit: i think we can get rid of this extra variable and use `data` inline.

amiclaus (commented):
one nit + seems like astyle build is complaining.

drivers/adc/ad463x/ad463x.c
```diff
@@ -100,6 +100,31 @@ int32_t ad463x_spi_reg_write(struct ad463x_dev *dev,
...
+	int32_t ret;
+
+	ret = ad463x_spi_reg_read(dev, reg_addr, &reg_data);
+	if (ret != 0)
```
danmois, line 120: Could we omit here the comparison with 0?

amiclaus: I think is due to the way the driver was initially written.

danmois: Understood.

buha (approved):

drivers/adc/ad463x/ad463x.c
```diff
@@ -100,6 +100,31 @@ int32_t ad463x_spi_reg_write(struct ad463x_dev *dev,
...
+ */
+int32_t ad463x_spi_reg_read_masked(struct ad463x_dev *dev,
+				    uint16_t reg_addr,
+				    uint8_t mask,
+				    uint8_t *data)
+{
+	uint8_t reg_data;
+	int32_t ret;
+
+	ret = ad463x_spi_reg_read(dev, reg_addr, &reg_data);
```
amiclaus, line 119: shouldn't we pass `reg_data` together with the mask to the `data` pointer? or as I mentioned before, we can use only `data` pointer for both reading from the register and applying mask.

amiclaus (changes requested):

drivers/adc/ad463x/ad463x.c
```diff
@@ -100,6 +100,31 @@ int32_t ad463x_spi_reg_write(struct ad463x_dev *dev,
...
+	if (ret != 0)
+		return ret;
+
+	*data &= mask;
```
amiclaus, line 123: also, we can use here directly `no_os_field_get` so we perform also the shifting.

amiclaus (commented):
v2:
- properly handle `read_data`
- use `no_os_field_get`

CiprianRegus (approved):

amiclaus (approved):

PR #1553, opened 2022-11-15
iio:iio.c: Reset trigger id of the device optional macro.

pmallick-adi:
Suggest to add this macro as the firmware based on iio code is used across multiple ecosystems ranging from pyadi-iio to Matlab and not all of these platform support setting trigger id using impulse generator. Also the impulse generator in IIO oscilloscope doesn't work for windows version.

### DIFF
--- a/iio/iio.c
+++ b/iio/iio.c
@@ -1171,7 +1171,9 @@ static int iio_close_dev(struct iiod_ctx *ctx, const char *device)
 		trig = &desc->trigs[dev->trig_idx];
 		if (trig->descriptor->disable)
 			ret = trig->descriptor->disable(trig->instance);
+#ifndef IIO_IGNORE_SET_TRIGGER
 		dev->trig_idx = NO_TRIGGER;
+#endif
 	}
 
 	return ret;


rbolboac (commented):
Can you please explain a little bit the following:
"not all of these platform support setting trigger id using impulse generator"
So do you actually mean that these clients do not ever call SETTRIG and GETTRIG commands?
In that case, how do you select the trigger? Because by default there is no trigger assigned to the device when using iio_app. The SETTRIG command still has to be called to assign a trigger to a device.
Do you assign the trigger directly in the software? Are you using iio_app_run_with_trigs, or do you have another application running on top of iio?

Also, please adjust the commit message to no exceed 72 characters per line.

pmallick-adi:
Hi @rbolboac,

I would give you a background on how our firmware works. We're not using the iio_app files. We have our own application code that makes use of the iio_init_param and iio_device_init descriptors to initialize IIO. We set the trigger_id to **"trigger0**" in the iio_device_init structure. This is then passed as a parameter of the iio_init_param descriptor in to the iio_init() function.  Below a picture for your reference.

image

This is way we have been initializing the trigger for our application. The iio trigger comes up as another device on the clients which we never use. The trigger on the firmware is attached to callback function which collects push data into the iio buffer after each conversion. This is the way we have using triggers to facilitate data capture in our application code. Nor sure if this is the intended way to use IIO trigger support.

Coming on to your questions:
1) There are set trigger wrappers for python. However, I don't find a lot devices using it. https://github.com/analogdevicesinc/pyadi-iio/search?q=_set_trigger

2) We don't use SETTRIG Command to set the trigger ID. As mentioned, above we have our own application running on top of IIO which sets the trigger ID.

Will keep comments within 72 characters per line from here on.

rbolboac:
Changes done with https://github.com/analogdevicesinc/no-OS/pull/1557
I am closing this PR.

PR #1564, opened 2022-11-28
drivers: jesd204 :jesd204-core: copy_link_params() sysref param fix

danmois:
jesd204_copy_link_params() copy the SYSREF settings form source to destination.


amiclaus (approved):
I think this might fit as PR also for 2021_R2 branch. similar to what you did on the linux tree. Also a fixes tag would be nice.

danmois:
V2

- added a **Fixes** tag to the commit

PR #1566, opened 2022-11-28
iio: fix the issue with buffer size allocation

SaikiranGudla:
Corrected the size of the circular buffer allocated in the iio_open_dev function.

Signed-off-by: Gudla <Saikiran.Gudla@analog.com>

### DIFF
--- a/iio/iio.c
+++ b/iio/iio.c
@@ -1066,6 +1066,7 @@ static int iio_open_dev(struct iiod_ctx *ctx, const char *device,
 	uint32_t ch_mask;
 	int32_t ret;
 	int8_t *buf;
+	uint32_t buf_size;
 
 	dev = get_iio_device(ctx->instance, device);
 	if (!dev)
@@ -1091,6 +1092,7 @@ static int iio_open_dev(struct iiod_ctx *ctx, const char *device,
 			/* Need a bigger buffer or to allocate */
 			return -ENOMEM;
 
+		buf_size = dev->buffer.raw_buf_len;
 		buf = dev->buffer.raw_buf;
 	} else {
 		if (dev->buffer.allocated) {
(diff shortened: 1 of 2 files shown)


iio/iio.c
```diff
@@ -1091,20 +1092,22 @@ static int iio_open_dev(struct iiod_ctx *ctx, const char *device,
...
 		}
+		buf_size = dev->buffer.public.size;
 		buf = (int8_t *)calloc(dev->buffer.public.size, sizeof(*buf));
 		if (!buf)
 			return -ENOMEM;
 		dev->buffer.allocated = 1;
 	}
 
-	ret = no_os_cb_cfg(&dev->buffer.cb, buf, dev->buffer.public.size);
+	ret = no_os_cb_cfg(&dev->buffer.cb, buf, buf_size);
```
rbolboac, line 1110: Hi @SaikiranGudla ,

Could you please explain what is the issue with the current implementation in details? What is the observed bug and in which scenario?

Taking a look at the code I see the following:
- we need a number of bytes available based on the number of samples and bytes per scan: 
dev->buffer.public.size = dev->buffer.public.bytes_per_scan * samples;
- now, if we are using the raw_buf we have to make sure that enough memory in allocated, that is why we check the following:
if (dev->buffer.raw_buf && dev->buffer.raw_buf_len) {
		if (dev->buffer.raw_buf_len < dev->buffer.public.size)
- if not enough memory is allocated, we return -ENOMEM. 
- However, if the raw_buf has more memory than needed, we will still use only dev->buffer.public.size bytes, because that is the size for the requested data.
- finally, no_os_cb_cfg is called, but the size should be equal to the requested size, and not the total available memory in raw_buf. because the IIO client expects to read/write dev->buffer.public.size bytes, and not the size of raw_buf

Why would you want to call no_os_cb_cfg with dev->buffer.raw_buf_len and not with dev->buffer.public.size?
Even though you allocate a bigger buffer, still only the first dev->buffer.public.size bytes are needed.

rbolboac (approved):

PR #1569, opened 2022-11-30
include:no_os_clk.h util:no_os_clk.c clk ops

danmois:
Add clk ops similar to spi no-OS implementation.


include/no_os_clk.h
```diff
@@ -47,6 +47,17 @@
...
+	const char	*name;
+	/** Channel number */
+	uint8_t		hw_ch_num;
+	/** CLK function pointers */
+	const struct no_os_clk_platform_ops *platform_ops;
+	/**  CLK hardware device descriptor */
+	void		*dev_desc;
+};
+
 struct no_os_clk_hw {
```
buha, line 61: I think that following this change, struct no_os_clk_hw might not be needed.

danmois: This was kept so that current implementation does not suffer.

danmois:
V2

- check init and remove ops existence before call

util/no_os_clk.c
```diff
@@ -46,83 +46,111 @@
...
+ * @param param - The structure that contains CLK parameters.
+ * @return 0 in case of success, -1 otherwise.
+ */
+int32_t no_os_clk_init(struct no_os_clk_desc **desc,
+		       const struct no_os_clk_init_param *param)
+{
+	if (!param)
+		return -1;
+
+	if (desc->platform_ops->init)
```
buha, line 61: I was thinking more like:
```
if (!param || !param->platform_ops->init)
    return -EINVAL;

... the rest of the function
```

Same for all other functions.

danmois:
V3

- fix return value if ops not declared
- check init before call

danmois:
V4

- implement checks for every function existence before call

util/no_os_clk.c
```diff
@@ -46,83 +46,132 @@
...
+ * @param desc - CLK descriptor.
+ * @param param - The structure that contains CLK parameters.
+ * @return 0 in case of success, -1 otherwise.
+ */
+int32_t no_os_clk_init(struct no_os_clk_desc **desc,
+		       const struct no_os_clk_init_param *param)
+{
+	int ret;
+
+	if (!param || !param->platform_ops)
```
buha, line 60: Check also `desc` here

buha (approved):

PR #1573, opened 2022-12-02
cppcheck misc fixes 2

amiclaus:
Multiple fixes related to the updated version of cppcheck.

### DIFF
--- a/drivers/axi_core/jesd204/axi_adxcvr.c
+++ b/drivers/axi_core/jesd204/axi_adxcvr.c
@@ -380,7 +380,7 @@ int32_t adxcvr_status_error(struct adxcvr *xcvr)
  */
 static int adxcvr_reset(struct adxcvr *xcvr)
 {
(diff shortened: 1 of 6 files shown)


drivers/axi_core/jesd204/axi_adxcvr.c
```diff
@@ -380,7 +380,7 @@ int32_t adxcvr_status_error(struct adxcvr *xcvr)
  */
 static int adxcvr_reset(struct adxcvr *xcvr)
 {
-	int ret, retry = 1;
+	int ret = 0, retry = 1;
```
danmois, line 383: Wouldn't it be better to initialize with error value? It would be like an additional check that the routine is executed correctly.

amiclaus (commented):
v2:
- initialized ret with the error value returned by the `adxcvr_status_error`

rbolboac (approved):

PR #1581, opened 2022-12-16
Fix the API pointer checks

CiprianRegus:
- Return the actual error code instead of -1.
- Add pointer checks where they were missing.

### DIFF
--- a/drivers/api/no_os_eeprom.c
+++ b/drivers/api/no_os_eeprom.c
@@ -51,11 +51,17 @@
 int32_t no_os_eeprom_init(struct no_os_eeprom_desc **desc,
 			  const struct no_os_eeprom_init_param *param)
 {
-	if (!param)
+	int32_t ret;
+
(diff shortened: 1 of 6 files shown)


drivers/api/no_os_eeprom.c
```diff
@@ -51,11 +51,17 @@
...
 			  const struct no_os_eeprom_init_param *param)
 {
-	if (!param)
-		return -EINVAL;
+	int32_t ret;
 
-	if ((param->platform_ops->init(desc, param)))
-		return -EINVAL;
+	if (!param || !param->platform_ops)
+		return -ENXIO;
```
buha, line 57: I think ENXIO is used for situations where you're searching for a device on a bus and cna't find it. EINVAL is more suitable here.

CiprianRegus:
Replaced `ENXIO` error code with `EINVAL`

amiclaus (approved):

PR #1561, opened 2022-11-22
SPI delays for maxim platform

CiprianRegus:

PR #1592, opened 2022-12-23
Maxim peripherals VDDIO configuration

CiprianRegus:
This has to be merged first: https://github.com/analogdevicesinc/no-OS/pull/1561

Draft until properly tested on multiple MCUs

### DIFF
--- a/drivers/platform/maxim/max32650/maxim_spi.c
+++ b/drivers/platform/maxim/max32650/maxim_spi.c
@@ -64,18 +64,21 @@
 /******************************************************************************/
 
 /**
- * @brief Enable the chip select gpio on MAX32665 or MAX32650
- * @param spi - The SPI port registers
- * @param chip_select - cs index
+ * @brief Configure the VDDIO level for a SPI interface
+ * @param desc - the SPI descriptor
  * @return 0 in case of success, -EINVAL otherwise
  */
-static int32_t _max_spi_enable_ss(uint32_t id, uint32_t chip_select)
+static int32_t _max_spi_config_pins(struct no_os_spi_desc *desc)
 {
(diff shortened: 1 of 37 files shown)


drivers/platform/maxim/max32660/maxim_spi.c
```diff
@@ -151,6 +157,9 @@ static int _max_spi_config(struct no_os_spi_desc *desc)
 		goto err_init;
 	}
 
+	spi_pins.vssel = eparam->vssel;
```
buha, line 160: Just a remark, for the sake of completeness, if vssel is not initialized by driver API user, its value will default to 0 which means VDDIO (not VDDIOH). I think this is good behavior.

buha:
let's not forget about documentation, perhaps you could edit these pages:
https://wiki.analog.com/resources/no-os/drivers/i2c
https://wiki.analog.com/resources/no-os/drivers/spi
https://wiki.analog.com/resources/no-os/drivers/uart

And add a collapsible area by platform where you show how exactly to fill the xuip/slave_xip on maxim.

drivers/platform/maxim/max32650/maxim_spi.c
```diff
@@ -142,6 +151,8 @@ static int32_t _max_spi_enable_ss(uint32_t id, uint32_t chip_select)
 		return -EINVAL;
 	}
 
+	spi_pins.vssel = st->init_param->vssel;
```
amiclaus, line 154: i think we should do also `cs.vssel = st->init_param->vssel`. Or is the vssel for the chip select treated somewhere else?

CiprianRegus:
Set the vssel value to CS pins.

amiclaus (approved):

PR #1593, opened 2023-01-03
Staging/ad9361 Modify ad9361 project defines

danmois:
Remove duplicate code and rename DMA IRQ define.

### DIFF
--- a/projects/ad9361/src/main.c
+++ b/projects/ad9361/src/main.c
@@ -714,7 +714,6 @@ int main(void)
 		return status;
 #endif
 
-#ifdef FMCOMMS5
 	struct axi_dma_transfer transfer = {
 		// Number of bytes to write/read
 		.size = sizeof(sine_lut_iq),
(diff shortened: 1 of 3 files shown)


projects/ad9361/src/main.c
```diff
@@ -160,7 +160,7 @@ struct axi_dac_init tx_dac_init = {
 struct axi_dmac_init rx_dmac_init = {
 	"rx_dmac",
 	CF_AD9361_RX_DMA_BASEADDR,
-#ifdef ADC_DMA_IRQ_EXAMPLE
```
amiclaus, line 163: any specific reason for the rename? the previous versions looks more explicit for the user. Also, if you do this change please make sure that the .json file is updated accordingly.

amiclaus (commented):
looks good, one comment.

projects/ad9361/src/main.c
danmois: In the current implementation, the define is valid for both ADC and DAC, and not only for the first one. Therefore, removing ADC from the define makes the example selection clearer.

danmois:
V2

- update builds.json with the new define for DMA IRQ

amiclaus (commented):
you might want to squash the last 2 commits since they are interdependent.

projects/ad9361/src/main.c
amiclaus: got it, thanks.

danmois:
V3

- add modifications in builds.json to last commit

danmois:
V4

- update DMA IRQ define in app_config.h

buha (approved):

amiclaus (approved):

PR #1599, opened 2023-01-11
Staging/time work

danmois:
Handle timekeeping requests for pr_info (pr_*) calls for the cases in which the code runs on Microblaze (system time not available).

### DIFF
--- a/drivers/platform/xilinx/xilinx_delay.c
+++ b/drivers/platform/xilinx/xilinx_delay.c
@@ -43,7 +43,9 @@
 
 #include "no_os_delay.h"
 #include <sleep.h>
-#include <xtime_l.h>
+#ifdef _XPARAMETERS_PS_H_
+#include "xtime_l.h"
+#endif
 
 /******************************************************************************/
 /************************ Functions Definitions *******************************/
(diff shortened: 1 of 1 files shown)


danmois:
Should we use a preprocessor warning or an error if the user is trying to display timestamps (PRINT_TIME defined) for Microblaze-based systems, where time-related information is not available using the basic setup?

include/no_os_print_log.h
```diff
@@ -42,6 +42,12 @@
 
 #include <stdio.h>
 
+#ifdef XILINX_PLATFORM
+#if defined(PRINT_TIME) && !defined(_XPARAMETERS_PS_H_)
+#error The processor time is not available. PRINT_TIME cannot be used.
```
buha, line 47: On a second thought, I think there are 2 better options:
- drop this commit entirely
- or put a warning in xilinx_delay.c something like below
```
#ifdef _XPARAMETERS_PS_H_
#include "xtime_l.h"
#elif defined(PRINT_TIME)
#warning Time information is not available on Microblaze.
#endif
```

I think I prefer no1 point above, silently returning a time of 0.000000.

danmois:
V2

- drop commit containing modifications that generate error if no time available and PRINT_TIME enabled

buha (approved):

PR #1600, opened 2023-01-12
platform:maxim:maxim_delay.c Add time function

danmois:
Add Maxim implementation for function getting current time value.

### DIFF
--- a/drivers/platform/maxim/max32650/maxim_delay.c
+++ b/drivers/platform/maxim/max32650/maxim_delay.c
@@ -1,4 +1,4 @@
-/***************************************************************************//**
+/*******************************************************************************
  *   @file   maxim_delay.c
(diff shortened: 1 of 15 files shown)


buha:
I don't think using RTC is a good idea.
RTC should be available to the user, no-os shouldnt monopolize it.
See how delay is done on stm32, some sort of a debugger timer is used which is available on these Cortex-M devices. That one could be used for timing info while debugging.

danmois:
> I don't think using RTC is a good idea. RTC should be available to the user, no-os shouldnt monopolize it. See how delay is done on stm32, some sort of a debugger timer is used which is available on these Cortex-M devices. That one could be used for timing info while debugging.

You are right. We will have to decide the best way to implement timekeeping.

buha (approved):
looks good

PR #1602, opened 2023-01-12
xilinx:xilinx_delay.c Remove use of float

danmois:
Remove use of float in microsecond calculation.

### DIFF
--- a/drivers/platform/xilinx/xilinx_delay.c
+++ b/drivers/platform/xilinx/xilinx_delay.c
@@ -44,6 +44,7 @@
 #include "no_os_delay.h"
 #include <sleep.h>
 #ifdef _XPARAMETERS_PS_H_
+#include "no_os_util.h"
 #include "xtime_l.h"
 #endif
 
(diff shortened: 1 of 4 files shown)


drivers/platform/xilinx/xilinx_delay.c
```diff
@@ -42,6 +42,7 @@
 /******************************************************************************/
 
 #include "no_os_delay.h"
+#include "no_os_util.h"
```
buha, line 45: You're adding a new file, which might break some projects which don't include this file in src.mk.

Let's wait for merging #1599 which fixes unrelated build errors and then we'll see if this commit breaks anything.

danmois:
V2

- rebase
- reorganize instructions in no_os_get_time()

CiprianRegus (approved):

amiclaus (approved):

danmois:
V3

- add no_os_util files to INCS and SRCS where needed

PR #1624, opened 2023-01-31
iio_app update

danmois:
Update the iio_app API for limiting use of defines for initializations.


iio/iio_app/iio_app.h
```diff
@@ -71,21 +74,54 @@ struct iio_app_device {
...
+struct iio_app_init_param {
+	struct iio_ctx_attr *ctx_attrs;
+	uint32_t nb_ctx_attr;
+	struct iio_app_device *devices;
+	uint32_t nb_devices;
+	struct iio_trigger_init *trigs;
+	int32_t nb_trigs;
+	struct no_os_uart_init_param user_uart_init_params;
+	void *irq_desc;
+	struct iio_desc *iio_desc;
```
buha, line 115: iio_desc is not to be initialized by user, it's created by iio_app, so remove it from init

iio/iio_app/iio_app.h
```diff
@@ -71,21 +74,54 @@ struct iio_app_device {
...
+	uint32_t nb_ctx_attr;
+	struct iio_app_device *devices;
+	uint32_t nb_devices;
+	struct no_os_uart_init_param *user_uart_params;
+	/** IIO triggers to be used */
+	struct iio_trigger_init *trigs;
+	/** Number of triggers to be used */
+	int32_t nb_trigs;
+	/** IRQ descriptor to be used */
+	void *irq_desc;
```
buha, line 93: I am not 100% sure but I think all members of this struct up until irq_desc can be discarded, there is no reason to keep them in desc.

iio/iio_app/iio_app.h
```diff
@@ -71,21 +74,54 @@ struct iio_app_device {
...
+	int (*app_callback)(void *arg);
+	void *arg;
+};
+
+/** Register devices for an iio application */
+int32_t iio_app_init(struct iio_app_desc **app,
+	struct iio_app_init_param app_init_param);
+
+/** Start an iio application */
+int32_t iio_app_run(struct iio_app_desc *app);
```
buha, line 125: I'm wondering if we should add a remove() function too. It doesn't make much sense yet it looks weird to allocate memory without a function which deallocates it.

iio/iio_app/iio_app.c
```diff
@@ -122,6 +111,7 @@ static inline uint32_t _calc_uart_xfer_time(uint32_t len, uint32_t baudrate)
 #if !defined(LINUX_PLATFORM) && !defined(NO_OS_NETWORKING)
 static int32_t iio_print_uart_info_message(struct no_os_uart_desc **uart_desc,
 		struct no_os_uart_init_param *uart_init_par,
+		struct no_os_uart_init_param *user_uart_params,
```
buha, line 114: it doesn't make sense to have 2 uart_init_params!

this function only needs 1, the one from the user

the other one (with the defaults) you can create internally with hardcoded values (115200,8n1n)

iio/iio_app/iio_app.c
```diff
@@ -413,33 +396,90 @@ int32_t iio_app_run_with_trigs(struct iio_ctx_attr *ctx_attrs,
...
+int32_t iio_app_run(struct iio_app_desc *app)
+{
+	int32_t	status;
+
+	status = iio_app_prepare(app);
+	if (status < 0)
+		return status;
+
+	do {
+		if (app->app_callback)
```
buha, line 476: I would rename app_callback to something more suggestive like `post_step` and move it such that it executes after iio_step.

buha, line 476: Let's execute the callback after step.
Also, not sure why the step status is not being verified, let's fix that too.

iio/iio_app/iio_app.c
```diff
@@ -322,24 +316,13 @@ static int32_t irq_setup(struct no_os_irq_ctrl_desc **irq_desc)
...
+ *
+ * @param app - the iio application descriptor
+ * @return 0 on success, negative value otherwise
  */
-int32_t iio_app_run_with_trigs(struct iio_ctx_attr *ctx_attrs,
-			       uint32_t nb_ctx_attr,
-			       struct iio_app_device *devices, uint32_t nb_devs,
-			       struct iio_trigger_init *trigs, int32_t nb_trigs,
-			       void *irq_desc, struct iio_desc **iio_desc)
+static int iio_app_prepare(struct iio_app_desc *app)
```
buha, line 325: move the content of iio_app_prepare into iio_app_init

iio/iio_app/iio_app.h
danmois: Removed

iio/iio_app/iio_app.h
danmois: Removed.

iio/iio_app/iio_app.h
danmois: Yes, we can.

iio/iio_app/iio_app.c
danmois: Only one uart_init_params used.

iio/iio_app/iio_app.c
danmois: Done.

iio/iio_app/iio_app.c
danmois: Moved.

iio/iio_app/iio_app.h
buha: It could be that uart_desc is needed, you're right

iio/iio_app/iio_app.c
```diff
@@ -413,33 +373,80 @@ int32_t iio_app_run_with_trigs(struct iio_ctx_attr *ctx_attrs,
...
-	} while (true);
+	*app = application;
+
+	return 0;
 error:
-	status = print_uart_error_message(&uart_desc, uart_init_par, status);
+	free(application);
+
+	status = print_uart_error_message(&uart_desc,
+					  &app_init_param.user_uart_init_params, status);
```
buha, line 395: i think uart_remove is also needed here, at the end of `error` label

iio/iio_app/iio_app.c
```diff
@@ -413,33 +373,80 @@ int32_t iio_app_run_with_trigs(struct iio_ctx_attr *ctx_attrs,
...
+{
+	int ret;
+
+#if defined(ADUCM_PLATFORM) || (defined(XILINX_PLATFORM) && !defined(PLATFORM_MB)) || defined(STM32_PLATFORM)
+	ret = no_os_irq_ctrl_remove(app->irq_desc);
+	if (ret)
+		return ret;
 #endif
+
+	// ret = no_os_uart_remove(app->iio_desc->uart_desc);
```
buha, line 439: why is this commented out ?

iio/iio_app/iio_app.c
```diff
@@ -37,24 +37,17 @@
  * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 *******************************************************************************/
 
-#ifdef IIO_SUPPORT
+// #ifdef IIO_SUPPORT
```
buha, line 40: also this, shouldn't be commented out I guess

projects/eval-adxl355-pmdz/src/examples/iio_trigger_example/iio_trigger_example.c
```diff
@@ -75,6 +75,19 @@ int iio_trigger_example_main()
 	struct iio_hw_trig *adxl355_trig_desc;
 	struct no_os_irq_ctrl_desc *adxl355_irq_desc;
 	struct iio_desc *iio_desc;
+	struct iio_app_desc *app;
+	struct iio_app_init_param app_init_param = {
```
buha, line 79: just say `struct iio_app_init_param app_init_param = {0};`

(... see continuation below)

projects/eval-adxl355-pmdz/src/examples/iio_trigger_example/iio_trigger_example.c
```diff
@@ -115,7 +128,15 @@ int iio_trigger_example_main()
...
 
-	return iio_app_run_with_trigs(NULL, 0, iio_devices,
-				      NO_OS_ARRAY_SIZE(iio_devices),
-				      trigs, NO_OS_ARRAY_SIZE(trigs), adxl355_irq_desc, &iio_desc);
+	app_init_param.devices = iio_devices;
+	app_init_param.nb_devices = NO_OS_ARRAY_SIZE(iio_devices);
+	app_init_param.trigs = trigs;
+	app_init_param.nb_trigs = NO_OS_ARRAY_SIZE(trigs);
+	app_init_param.irq_desc = adxl355_irq_desc;
+
```
buha, line 136: and add `app_init_param.user_uart_init_params = adxl355_uart_ip;`

iio/iio_app/iio_app.h
```diff
@@ -71,21 +74,55 @@ struct iio_app_device {
...
+	/** Array of IIO devices */
+	struct iio_app_device *devices;
+	/** Number of devices */
+	uint32_t nb_devices;
+	/** IIO triggers to be used */
+	struct iio_trigger_init *trigs;
+	/** Number of triggers to be used */
+	int32_t nb_trigs;
+	/** UART init params */
+	struct no_os_uart_init_param user_uart_init_params;
```
buha, line 109: `uart_init_param` should be enough

iio/iio_app/iio_app.c
danmois: Just for development time.

buha:
I see INTC_DEVICE_ID is still in the code - has to be removed

buha:
Also, all aducm_uart_ops, linux_uart_ops etc. should disappear

buha:
```
#if defined(STM32_PLATFORM) || defined(MAXIM_PLATFORM) || defined(ADUCM_PLATFORM) || defined(PICO_PLATFORM)
		.irq_id = UART_IRQ_ID,
		.asynchronous_rx = true,
#endif
```

This part also depends on UART_IRQ_ID. All this platform dependant code should disappear. 
IIO APP should only get a uart_init_param as input and not care about such details.

iio/iio_app/iio_app.h
buha: rename user_uart_init_params ---> uart_init_param or uart_params or whatever, but shorter

iio/iio_app/iio_app.c
danmois: A separate member of iio_app_desc was needed.

buha:
Looks good now!

There is still some platform dependant code but it's unrelated to this code (it's related to differing IRQ implementations)

Also those #includes could be simplified with #include $(PLATFORM)_uart.c etc. since we now have the same naming in all platforms.

iio/iio_app/iio_app.c
```diff
@@ -406,33 +352,85 @@ int32_t iio_app_run_with_trigs(struct iio_ctx_attr *ctx_attrs,
...
 
-	return iio_app_run_with_trigs(ctx_attrs, nb_ctx_attr, devices, len, NULL, 0,
-				      irq_desc, &iio_desc);
+	do {
+		status = iio_step(app->iio_desc);
+		if (status && status != -EAGAIN)
+			return status;
+		if (app->post_step_callback) {
+			status = app->post_step_callback(app->arg);
+			if (status)
```
buha, line 400: `if (status && status != -EAGAIN)` here as well

projects/ad713x_fmcz/src/ad713x_fmc.c
```diff
@@ -291,6 +291,29 @@ int main()
 		.dcache_invalidate_range = (void (*)(uint32_t, uint32_t))Xil_DCacheInvalidateRange,
 	};
 
+	struct xil_uart_init_param platform_uart_init_par = {
+#ifdef XPAR_XUARTLITE_NUM_INSTANCES
```
buha, line 295: Use `PLATFORM_MB` instead

buha (approved):
I'm approving ~~with the mention that we should probably use PLATFORM_MB for the xilinx uart pl/ps distinction instead of the whatever is currently being used. AFter that it can be merged in my opinion.~~

projects/ad713x_fmcz/src/ad713x_fmc.c
buha: taking this back, the current condition is the correct one

danmois: You could use a UARTLite IP even if you have a PS (it depends on the HDL design and not on the processor used).

danmois:
V2
-reordered first commits

amiclaus (commented):
Ideally, no commit introduced should break the builds. Therefore, i'd suggest squashing most of the commits.

buha:
yeah, squash at least the project commits to be able to merge

danmois:
V3

- squash project updates commits

danmois:
V4

- correct commit messages

buha (approved):

PR #1625, opened 2023-01-31
drivers: meter: ade9430: initial commit

amiclaus:
The ADE9430 is a highly accurate, fully integrated, polyphase energy and power quality monitoring device.

This a basic driver implementation capable of reading IRMS, VRMS, Power and Temperature.


drivers/meter/ade9430/ade9430.c
```diff
@@ -0,0 +1,324 @@
...
+	*device = dev;
+
+	return 0;
+
+error_spi:
+	no_os_spi_remove(dev->spi_desc);
+error_dev:
+	free(dev);
+
+	return 0;
```
danmois, line 215: Return ret?

amiclaus (commented):
v2:
- return ret instead of 0 on the initialization error path

drivers/meter/ade9430/ade9430.c
```diff
@@ -0,0 +1,324 @@
...
+	ret = ade9430_read(dev, ADE9430_REG_CONFIG5, &chip_id);
+	if (ret)
+		goto error_spi;
+
+	ret = ade9430_write(dev, ADE9430_REG_RUN, 0);
+	if (ret)
+		goto error_spi;
+
+	if (chip_id != ADE9430_CHIP_ID)
+		return -EINVAL;
```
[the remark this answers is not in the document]
amiclaus: it should jump to the error handling step, indeed :)

amiclaus (commented):
v3:
- remove extra comment line
- replace return with conditional jump

drivers/meter/ade9430/ade9430.c
```diff
@@ -0,0 +1,323 @@
...
+{
+	int ret;
+	uint32_t data;
+
+	ret = ade9430_read(dev, reg_addr, &data);
+	if (ret)
+		return ret;
+
+	data &= ~mask;
+	data |= reg_data;
```
CiprianRegus, line 125: `reg_data & mask` to prevent modifying unwanted bits.

drivers/meter/ade9430/ade9430.c
```diff
@@ -0,0 +1,323 @@
...
+	buff[0] = reg_addr >> 4;
+	buff[1] = reg_addr << 4;
+
+	if (reg_addr >= ADE9430_REG_RUN && reg_addr <= ADE9430_REG_VERSION) {
+		no_os_put_unaligned_be16(reg_data, &buff[2]);
+		return no_os_spi_write_and_read(dev->spi_desc, buff, 4);
+	} else {
+		no_os_put_unaligned_be32(reg_data, &buff[2]);
+		return no_os_spi_write_and_read(dev->spi_desc, buff, 6);
+	}
```
CiprianRegus, line 102: return early and remove the `else` branch.

drivers/meter/ade9430/ade9430.h
```diff
@@ -0,0 +1,918 @@
...
+#define ADE9430_IA_DISADC		NO_OS_BIT(0)
+
+/* ADE9430_VAR_DIS Bit Definition */
+#define ADE9430_VARDIS			NO_OS_BIT(0)
+
+/* Miscellaneous Definitions */
+#define ADE9430_CHIP_ID			0x63
+#define ADE9430_I_RES_DEFAULT		4018ULL / 1000000000
+#define ADE9430_V_RES_DEFAULT		13357ULL / 1000000000
+#define ADE9430_W_RES_DEFAULT		7203ULL / 1000000
```
CiprianRegus, line 851: Can we use a subunit and avoid the division?

amiclaus (commented):
v4:
- use `reg_data & mask` to avoid setting unwanted bits
- remove redundant else
- use `no_os_units.h` for conversions

drivers/meter/ade9430/ade9430.c
```diff
@@ -0,0 +1,325 @@
...
+	/* Enable Temperature */
+	ret = ade9430_update_bits(dev, ADE9430_REG_TEMP_CFG, ADE9430_TEMP_EN,
+				  no_os_field_prep(ADE9430_TEMP_EN, 1));
+	if (ret)
+		goto error_spi;
+
+	ret = ade9430_write(dev, ADE9430_REG_RUN, 1);
+	if (ret)
+		goto error_spi;
+
```
buha, line 206: So this whole initialization sequence above seems tailored for a specific project because the drive hardcodes those values.

Device driver should be "project-agnostic" and allow for configuration of these things through init params to be able to fit any project.

amiclaus: Indeed, in this driver there is only one implementation of the three User Energy Register Use Models that are available at:
https://wiki.analog.com/resources/eval/user-guides/ade9430

I can add support for the other 2 also.

amiclaus (commented):
v5:
- create separate function and add support for all three user energy use models.

amiclaus (commented):
v6:
- fix redefined macros

CiprianRegus (approved):

drivers/meter/ade9430/ade9430.c
```diff
@@ -0,0 +1,360 @@
...
+	/** Use a valid register with default value different from 0*/
+	ret = ade9430_read(dev, ADE9430_REG_CONFIG5, &chip_id);
+	if (ret)
+		goto error_spi;
+
+	if (chip_id != ADE9430_CHIP_ID)
+		goto error_spi;
+
+	/* Enable Temperature Sensor */
+	ret = ade9430_update_bits(dev, ADE9430_REG_TEMP_CFG, ADE9430_TEMP_EN,
```
buha, line 326: This is an optional change:

I can see why you always want to enable temperature (you have an API for reading temperature that assumes temperature sensor is enabled), but it'd probably be better to specify this through init_param.

amiclaus: Sure, will do.

amiclaus (commented):
v8: 
- add init param for enbaling/disabling the temperature sensor

PR #1631, opened 2023-02-03
scripts:generic_variables.mk Fix NO_OS_VERSION

danmois:
Rewrite the NO_OS_VERSION compiler define and modify iio.c so that it is correctly included in the xml header.

Fixes: fdfaea1c8a6e

### DIFF
--- a/iio/iio.c
+++ b/iio/iio.c
@@ -73,6 +73,9 @@
 #define IIOD_CONN_BUFFER_SIZE	0x1000
 #define NO_TRIGGER				(uint32_t)-1
 
+#define NO_OS_STRINGIFY(x) #x
+#define NO_OS_TOSTRING(x) NO_OS_STRINGIFY(x)
+
 /******************************************************************************/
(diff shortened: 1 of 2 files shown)


iio/iio.c
```diff
@@ -73,6 +73,9 @@
 #define IIOD_CONN_BUFFER_SIZE	0x1000
 #define NO_TRIGGER				(uint32_t)-1
 
+#define STRINGIFY(x) #x
```
buha, line 76: elegant solution but please move these 2 macros in no_os_util.h, they will be useful everywhere

buha: and obviously prefix them

dbogdan (approved):

danmois:
V2
- place defines in no_os_util

buha:
cppcheck, astyle fail

danmois:
V3

- run astyle

danmois:
V4
- remove defines from no_os_util.h (cppcheck fail)

PR #1639, opened 2023-02-08
hmc630x iio device

buha:
Make some changes to the hmc630x device driver (new features, mainly getter functions and more init params).

Add the hmc630x iio device implementation that exposes the device driver to iio clients through iiod.

### DIFF
--- a/drivers/rf-transceiver/hmc630x/hmc630x.c
+++ b/drivers/rf-transceiver/hmc630x/hmc630x.c
@@ -1,5 +1,5 @@
 /***************************************************************************//**
- *   @file   hmc630x.h
+ *   @file   hmc630x.c
  *   @brief  hmc6300 and hmc6301 device driver implementation.
  *   @author Darius Berghe (darius.berghe@analog.com)
 ********************************************************************************
@@ -196,6 +196,45 @@ int hmc630x_init(struct hmc630x_dev **dev, struct hmc630x_init_param *init)
 	if (ret)
 		goto error;
 
+	ret = hmc630x_set_enable(d, init->enabled);
+	if (ret)
+		goto error;
+
+	ret = hmc630x_set_temp_en(d, init->temp_en);
+	if (ret)
+		goto error;
+
+	ret = hmc630x_set_if_attn(d, init->if_attn);
+	if (ret)
+		goto error;
+
+	if (init->type == HMC6300) {
+		ret = hmc6300_set_rf_attn(d, init->tx.rf_attn);
+		if (ret)
+			goto error;
+	} else {
+		ret = hmc6301_set_lna_gain(d, init->rx.lna_gain);
+		if (ret)
+			goto error;
+
+		ret = hmc6301_set_bb_attn(d, init->rx.bb_attn1, init->rx.bb_attn2);
+		if (ret)
+			goto error;
+
+		ret = hmc6301_set_bb_attn_fine(d, init->rx.bb_attni_fine,
+					       init->rx.bb_attnq_fine);
+		if (ret)
+			goto error;
+
+		ret = hmc6301_set_bb_lpc_hpc(d, init->rx.bb_lpc, init->rx.bb_hpc);
+		if (ret)
+			goto error;
+	}
+
+	ret = hmc630x_set_vco(d, init->vco);
+	if (ret)
+		goto error;
+
 	*dev = d;
 	return 0;
 error:
@@ -244,6 +283,14 @@ int hmc630x_remove(struct hmc630x_dev *dev)
 	dev = NULL;
 
 	return 0;
+}
+
+enum hmc630x_type hmc630x_type(struct hmc630x_dev *dev)
+{
+	if (dev)
+		return dev->type;
+	else
+		return -EINVAL;
 }
 
 /* Write a device row using GPIO bit-banging. */
@@ -343,7 +390,6 @@ int hmc630x_write_regmap(struct hmc630x_dev *dev, const uint8_t *regmap)
 		skip2 = 15;
 		break;
 	default:
-	case HMC6300:
 		r = 1;
 		skip1 = 13;
 		skip2 = 15;
@@ -436,7 +482,7 @@ int hmc630x_read_regmap(struct hmc630x_dev *dev, uint8_t *regmap)
 }
 
 /* Enable temperature sensor. */
-int hmc630x_enable_temp(struct hmc630x_dev *dev, bool enable)
+int hmc630x_set_temp_en(struct hmc630x_dev *dev, bool enable)
 {
 	int ret;
 	uint16_t tempflash, tempsensor;
@@ -460,6 +506,38 @@ int hmc630x_enable_temp(struct hmc630x_dev *dev, bool enable)
 	return hmc630x_write(dev, tempsensor, !enable);
 }
 
+/* Is temperature sensor enabled? */
+int hmc630x_get_temp_en(struct hmc630x_dev *dev, bool *enable)
+{
+	int ret;
+	uint16_t tempflash, tempsensor;
+	uint8_t ena1, ena2;
+
+	switch (dev->type) {
+	case HMC6301:
+		tempflash = HMC6301_EN_TEMPFLASH;
+		tempsensor = HMC6301_ENBAR_TEMPS;
+		break;
+	default:
+	case HMC6300:
+		tempflash = HMC6300_EN_TEMPFLASH;
+		tempsensor = HMC6300_TEMPSENSOR_PWRDN;
+		break;
+	};
+
+	ret = hmc630x_read(dev, tempflash, &ena1);
+	if (ret)
+		return ret;
+
+	ret = hmc630x_read(dev, tempsensor, &ena2);
+	if (ret)
+		return ret;
+
+	*enable = ena1 && !ena2;
+
+	return 0;
+}
+
 /* Get a temperature reading. */
 int hmc630x_get_temp(struct hmc630x_dev *dev, uint8_t *temp)
 {
@@ -467,7 +545,7 @@ int hmc630x_get_temp(struct hmc630x_dev *dev, uint8_t *temp)
 }
 
 /* Enable FSK/MSK modulation inputs. */
-int hmc6300_enable_fm(struct hmc630x_dev *dev, bool enable)
+int hmc6300_set_fm_en(struct hmc630x_dev *dev, bool enable)
 {
 	if (!dev)
 		return -EINVAL;
@@ -479,8 +557,21 @@ int hmc6300_enable_fm(struct hmc630x_dev *dev, bool enable)
 	return hmc630x_write(dev, HMC6300_ENABLE_FM, (uint8_t)enable);
 }
 
+/* Are FSK/MSK modulation inputs enabled. */
+int hmc6300_get_fm_en(struct hmc630x_dev *dev, bool *enable)
+{
+	if (!dev)
+		return -EINVAL;
+
+	// this is a transmitter specific feature
+	if (dev->type != HMC6300)
+		return -ENOSYS;
+
+	return hmc630x_read(dev, HMC6300_ENABLE_FM, (uint8_t *)enable);
+}
+
 /* Power On/Off the chip. */
-int hmc630x_enable(struct hmc630x_dev *dev, bool enable)
+int hmc630x_set_enable(struct hmc630x_dev *dev, bool enable)
 {
 	int ret = -EFAULT;
 	uint8_t row;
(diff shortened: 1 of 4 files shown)


drivers/rf-transceiver/hmc630x/hmc630x.c
```diff
@@ -522,6 +614,54 @@ int hmc630x_enable(struct hmc630x_dev *dev, bool enable)
...
+			return ret;
+		ena1 = (bool)row;
+
+		ret = hmc630x_read(dev, HMC630X_EN_SYNTH_LDO, &row);
+		if (ret)
+			return ret;
+		ena2 = (bool)row;
+	} else {
+		/* row0 */
+		mask = no_os_field_prep(HMC630X_MASK(HMC6301_IFVGA_PWRDN), 1);
```
amiclaus, line 637: nit: i'd initialized only once the mask and avoid using `|=`

drivers/rf-transceiver/hmc630x/hmc630x.c
```diff
@@ -522,6 +614,54 @@ int hmc630x_enable(struct hmc630x_dev *dev, bool enable)
...
+		mask |= no_os_field_prep(HMC630X_MASK(HMC6301_LNA_PWRDWN), 1);
+		ret = hmc630x_read_row(dev, HMC630X_ROW(HMC6301_LNA_PWRDWN), &row);
+		if (ret)
+			return ret;
+
+		ena1 = (bool)(row & mask);
+
+		/* row1 */
+		mask = no_os_field_prep(HMC630X_MASK(HMC6301_ASK_PWRDN), 1);
+		mask |= no_os_field_prep(HMC630X_MASK(HMC6301_IF_BGMUX_PWRDN), 1);
```
amiclaus, line 650: ditto

drivers/rf-transceiver/hmc630x/hmc630x.c
```diff
@@ -710,3 +990,23 @@ int hmc6301_set_bb_lpc_hpc(struct hmc630x_dev *dev, enum hmc6301_bb_lpc lpc,
...
+	// this is a receiver specific feature
+	if (dev->type != HMC6301)
+		return -ENOSYS;
+
+	ret = hmc630x_read(dev, HMC6301_BBAMP_SELBW, (uint8_t *)lpc);
+	if (ret)
+		return ret;
+
+	return hmc630x_read(dev, HMC6301_BBAMP_SELFASTREC, (uint8_t *)hpc);
+}
```
amiclaus, line 1012: nit: new line

drivers/rf-transceiver/hmc630x/iio_hmc630x.c
```diff
@@ -0,0 +1,378 @@
...
+#include "hmc630x.h"
+#include <string.h>
+
+static int32_t _hmc630x_read_register2(struct hmc630x_iio_dev *iiodev,
+				       uint32_t reg,
+				       uint32_t *readval)
+{
+	int ret;
+	uint8_t val;
+	ret = hmc630x_read_row(iiodev->dev, reg, &val);
```
CiprianRegus, line 19: `readval` may be directly passed to this function.

drivers/rf-transceiver/hmc630x/iio_hmc630x.c
```diff
@@ -0,0 +1,378 @@
...
+	*readval = val;
+	return 0;
+}
+
+static int32_t _hmc630x_write_register2(struct hmc630x_iio_dev *iiodev,
+					uint32_t reg,
+					uint32_t writeval)
+{
+	uint8_t val = writeval;
+	return hmc630x_write_row(iiodev->dev, reg, val);
```
CiprianRegus, line 32: `hmc630x_write_row(iiodev->dev, reg, writeval);`

drivers/rf-transceiver/hmc630x/iio_hmc630x.c
```diff
@@ -0,0 +1,378 @@
...
+				 intptr_t priv)
+{
+	int ret;
+	struct hmc630x_iio_dev *iiodev = (struct hmc630x_iio_dev *)device;
+	struct hmc630x_dev *d = (struct hmc630x_dev *)iiodev->dev;
+	int32_t val;
+
+	switch (priv) {
+	case HMC630X_IIO_ATTR_ENABLED:
+		bool enable;
```
CiprianRegus, line 46: This is a compilation error (same for all the other variable declarations in this `switch`).

drivers/rf-transceiver/hmc630x/iio_hmc630x.c
```diff
@@ -0,0 +1,378 @@
...
+}
+
+static int hmc630x_iio_write_attr(void *device, char *buf,
+				  uint32_t len, const struct iio_ch_info *channel,
+				  intptr_t priv)
+{
+	int ret = 0;
+	int32_t val;
+	struct hmc630x_iio_dev *iiodev = (struct hmc630x_iio_dev *)device;
+	struct hmc630x_dev *d = (struct hmc630x_dev *)iiodev->dev;
```
CiprianRegus, line 119: The cast is not needed.

drivers/rf-transceiver/hmc630x/iio_hmc630x.c
```diff
@@ -0,0 +1,378 @@
...
+		return -EINVAL;
+
+	ret = hmc630x_get_avail_vco(d, &freqs, &nfreqs);
+	if (ret)
+		return ret;
+
+	buf[0] = '\0';
+	for (n = 0; n < nfreqs; n++) {
+		sprintf(fstr, "%lu ", freqs[n]/1000); // divide because too large for int
+		strcat(buf, fstr);
```
CiprianRegus, line 196: You can write to `buf` directly: 

```
size_t len = 0;
len = sprintf(buf + len, "%llu ", freqs[n]/1000);
```
Also, the `%lu` modifier will only print 32 bits.

drivers/rf-transceiver/hmc630x/iio_hmc630x.c
```diff
@@ -0,0 +1,378 @@
...
+	}
+	ret = strlen(buf);
+
+	return ret;
```
CiprianRegus, line 200: `return strlen(buf)`

drivers/rf-transceiver/hmc630x/iio_hmc630x.c
```diff
@@ -0,0 +1,378 @@
...
+	d2 = (struct iio_device *)calloc(1, sizeof(*d2));
+	if (!d2)
+		goto end_0;
+
+	*d2 = hmc630x_iio_device_template;
+
+	d->iio_dev = d2;
+
+	ret = hmc630x_init(&d->dev, init_param->ip);
+	if (ret != 0)
```
CiprianRegus, line 344: nitpick: `if (ret)`

drivers/rf-transceiver/hmc630x/iio_hmc630x.c
```diff
@@ -0,0 +1,378 @@
...
+	if (!d)
+		return -ENOMEM;
+
+	d2 = (struct iio_device *)calloc(1, sizeof(*d2));
+	if (!d2)
+		goto end_0;
+
+	*d2 = hmc630x_iio_device_template;
+
+	d->iio_dev = d2;
```
CiprianRegus, line 341: `d->iio_dev = &hmc630x_iio_device_template` and then you don't need to allocate `d2` anymore.

drivers/rf-transceiver/hmc630x/iio_hmc630x.c
buha: i don't think so, behavior is different based on endianess

drivers/rf-transceiver/hmc630x/iio_hmc630x.c
buha: in this case, implicit casting would work, but it's good to explicitly cast (it avoids bugs)

probable the better version would be `return hmc630x_write_row(iiodev->dev, reg, (uint8_t)val);`

buha: will fix in v2

drivers/rf-transceiver/hmc630x/iio_hmc630x.c
buha: true, will fix in v2

drivers/rf-transceiver/hmc630x/iio_hmc630x.c
buha: will do in v2

drivers/rf-transceiver/hmc630x/iio_hmc630x.c
buha: will do in v2

drivers/rf-transceiver/hmc630x/iio_hmc630x.c
buha: will do in v2

drivers/rf-transceiver/hmc630x/hmc630x.c
buha: will fix in v2

drivers/rf-transceiver/hmc630x/hmc630x.c
buha: i'm not sure i understand

drivers/rf-transceiver/hmc630x/iio_hmc630x.c
buha: And what if you have multiple devices ? You're going the end up with the same d->iio_dev for all of them, which is not ok.

buha:
V2:
- fixed problems found in review

buha:
V3:
- fix attempt for cppcheck

drivers/rf-transceiver/hmc630x/hmc630x.c
amiclaus: Something like this:
```
mask = no_os_field_prep(HMC630X_MASK(HMC6301_IFVGA_PWRDN), 1) |
            no_os_field_prep(HMC630X_MASK(HMC6301_MIXER_PWRDN), 1) |
	    no_os_field_prep(HMC630X_MASK(HMC6301_BBAMP_PWRDN_I), 1) |
	    no_os_field_prep(HMC630X_MASK(HMC6301_TRIPLER_PWRDN), 1) |
	    no_os_field_prep(HMC630X_MASK(HMC6301_LNA_PWRDWN), 1);
```

buha:
V4:
- added bb_hpc and bb_lpc attributes for rx

drivers/rf-transceiver/hmc630x/hmc630x.c
```diff
@@ -460,14 +507,46 @@ int hmc630x_enable_temp(struct hmc630x_dev *dev, bool enable)
...
+	uint16_t tempflash, tempsensor;
+	uint8_t ena1, ena2;
+
+	switch (dev->type) {
+	case HMC6301:
+		tempflash = HMC6301_EN_TEMPFLASH;
+		tempsensor = HMC6301_ENBAR_TEMPS;
+		break;
+	default:
+	case HMC6300:
```
CiprianRegus, line 523: nitpick: this `case` is not needed.

drivers/rf-transceiver/hmc630x/hmc630x.c
```diff
@@ -246,6 +285,14 @@ int hmc630x_remove(struct hmc630x_dev *dev)
...
 }
 
+enum hmc630x_type hmc630x_type(struct hmc630x_dev *dev)
+{
+	if (dev)
+		return dev->type;
+	else
+		return -EINVAL;
+}
+
```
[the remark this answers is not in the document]
buha: Well, yes, but accessing it would also give users write access and i don't want this for this driver.
As a matter of fact, i think this should be the case for all drivers... 
Or at least have a public and a private structure, but i'd rather not go that way for 1 single member.

drivers/rf-transceiver/hmc630x/hmc630x.c
buha: true, will fix

drivers/rf-transceiver/hmc630x/iio_hmc630x.c
```diff
@@ -0,0 +1,425 @@
...
+	int ret;
+
+	if (!iiodev)
+		return 0;
+
+	ret = hmc630x_remove(iiodev->dev);
+	if (ret != 0)
+		return ret;
+
+	free(iiodev->dev);
```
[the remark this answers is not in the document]
buha: good catch !!

drivers/rf-transceiver/hmc630x/hmc630x.c
buha: ah i see,
how mandatory is this ? :D

amiclaus: it's just a matter of preference I think :D

buha:
V6:
- fixed some code commets

CiprianRegus (approved):

amiclaus (approved):

PR #1646, opened 2023-02-10
iio: iio.c: Handle return error if no triggers are present

rbolboac:
With the current implementation -EINVAL error is returned when searching for triggers. However, libiio exptects -ENOENT error if no triggers are present.
This commit returns the correct error if no triggers are found.

Signed-off-by: Ramona Bolboaca <ramona.bolboaca@analog.com>

### DIFF
--- a/iio/iio.c
+++ b/iio/iio.c
@@ -926,7 +926,16 @@ static int iio_get_trigger(struct iiod_ctx *ctx, const char *device,
 			   char *trigger, uint32_t len)
 {
 	struct iio_dev_priv *dev;
+	struct iio_trig_priv *trig;
 	struct iio_desc *desc = ctx->instance;
+
+	if (!desc->nb_trigs)
+		return -ENOENT;
+
+	trig = get_iio_trig_device(desc, device);
+	/* Device is a trigger and triggers cannot have other triggers */
+	if (trig)
+		return -ENOENT;
 
 	dev = get_iio_device(desc, device);
 	if (!dev)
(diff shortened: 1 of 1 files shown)


iio/iio.c
```diff
@@ -926,8 +926,17 @@ static int iio_get_trigger(struct iiod_ctx *ctx, const char *device,
 			   char *trigger, uint32_t len)
 {
 	struct iio_dev_priv *dev;
+	struct iio_trig_priv *trig;
 	struct iio_desc *desc = ctx->instance;
 
+	if (!desc->nb_trigs)
+		return -ENOENT;
+
+	trig = get_iio_trig_device(desc,device);
```
buha, line 935: `, `

iio/iio.c
```diff
@@ -954,9 +963,19 @@ static int iio_set_trigger(struct iiod_ctx *ctx, const char *device,
 			   const char *trigger, uint32_t len)
 {
 	struct iio_dev_priv	*dev;
+	struct iio_trig_priv	*trig;
 	uint32_t i;
+	struct iio_desc *desc = ctx->instance;
```
buha, line 968: maybe `ctx` could be NULL (not sure, double-check)

rbolboac: the ctx cannot be null with the current implementation...it comes from iiod: struct iiod_ctx ctx = IIOD_CTX(desc, conn);
also, it is never checked in iio.c

buha: ok, fair enough, that's why i asked

buha (approved):

amiclaus (approved):

CiprianRegus (approved):

PR #1649, opened 2023-02-16
drivers:adc:ad7779:Update initialization sequence

jansunil:
1) Added a delay after enabling clocks to the enabled channels. This is done to ensure that the next operation happens after the clocks are successfully enabled for every channel. 
2) Set and cleared AD7779_REG_SRC_UPDATE bit to reflect on the updated output data rate

### DIFF
--- a/drivers/adc/ad7779/ad7779.c
+++ b/drivers/adc/ad7779/ad7779.c
@@ -1554,6 +1554,8 @@ int32_t ad7779_init(ad7779_dev **device,
 			ret |= ad7779_set_state(dev, (ad7779_ch)i, dev->state[i]);
 	}
 
+	no_os_mdelay(50);
+
 	for (i = AD7779_CH0; i <= AD7779_CH7; i++) {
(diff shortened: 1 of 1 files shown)


amiclaus (approved):

jansunil:
Hi Team- Any review comments on this?

drivers/adc/ad7779/ad7779.c
```diff
@@ -1554,6 +1554,8 @@ int32_t ad7779_init(ad7779_dev **device,
 			ret |= ad7779_set_state(dev, (ad7779_ch)i, dev->state[i]);
 	}
 
+	no_os_mdelay(50);
```
buha, line 1557: is this delay value a trial-and-error value or does it come from the datasheet ?

jansunil: I do not see any value specific to this in the datasheet. It is a trial and error value that has been tested with two STM32 boards interfaced with the AD7770 chip

buha (approved):

PR #1650, opened 2023-02-16
drivers:platform:stm32:Add API for TDM DMA read.

jansunil:
1) Added no_OS driver API for TDM DMA read
2) Added no_OS driver API for Stop TDM DMA read
3) Added TDM DMA Read in STM32 platform driver
4) Added function for Stop TDM DMA read in STM32 platform driver.

### DIFF
--- a/drivers/platform/stm32/stm32_tdm.c
+++ b/drivers/platform/stm32/stm32_tdm.c
@@ -44,6 +44,9 @@
 #include "stm32_tdm.h"
 #include "no_os_error.h"
 #include "no_os_alloc.h"
+#include "no_os_irq.h"
+#include "stm32_irq.h"
+#include "no_os_lf256fifo.h"
 
 /**
  * @brief stm32 platform specific TDM platform ops structure
@@ -51,6 +54,9 @@
 const struct no_os_tdm_platform_ops stm32_tdm_platform_ops = {
 	.tdm_ops_init = &stm32_tdm_init,
 	.tdm_ops_read = &stm32_tdm_read,
+	.tdm_ops_stop = &stm32_stop_tdm_transfer,
+	.tdm_ops_pause = &stm32_pause_tdm_transfer,
+	.tdm_ops_resume = &stm32_resume_tdm_transfer,
 	.tdm_ops_remove = &stm32_tdm_remove
 };
 
@@ -130,6 +136,7 @@ int32_t stm32_tdm_init(struct no_os_tdm_desc **desc,
 		goto error;
 		break;
 	};
+	tdm_desc->irq_id = param->irq_id;
 	tdesc->hsai.Init.DataSize = tmp;
 	tdesc->hsai.Init.FirstBit = param->data_lsb_first ? SAI_FIRSTBIT_LSB :
 				    SAI_FIRSTBIT_MSB;
@@ -159,6 +166,60 @@ int32_t stm32_tdm_init(struct no_os_tdm_desc **desc,
 		goto error;
 	}
 
+	if(param->rx_complete_callback) {
+		ret = lf256fifo_init(&tdm_desc->rx_fifo);
+		if (ret < 0)
+			goto error;
+
+		struct no_os_irq_init_param nvic_rx_cplt = {
+			.platform_ops = &stm32_irq_ops
+		};
+
+		ret = no_os_irq_ctrl_init(&tdesc->nvic_rxcplt, &nvic_rx_cplt);
+		if (ret < 0)
+			goto error;
+
+		tdesc->rx_callback.callback = param->rx_complete_callback;
+		tdesc->rx_callback.ctx = tdm_desc;
+		tdesc->rx_callback.event = NO_OS_EVT_DMA_RX_COMPLETE;
+		tdesc->rx_callback.peripheral = NO_OS_DMA_IRQ;
+		tdesc->rx_callback.handle = &tdesc->hsai;
+
+		ret = no_os_irq_register_callback(tdesc->nvic_rxcplt, param->irq_id,
+						  &tdesc->rx_callback);
+		if (ret < 0)
+			goto error;
+
+		ret = no_os_irq_enable(tdesc->nvic_rxcplt, tdm_desc->irq_id);
+		if (ret < 0)
+			goto error;
+	}
+
+	if(param->rx_half_complete_callback) {
+		struct no_os_irq_init_param nvic_rx_half_cplt = {
+			.platform_ops = &stm32_irq_ops
+		};
+
+		ret = no_os_irq_ctrl_init(&tdesc->nvic_rx_halfcplt, &nvic_rx_half_cplt);
+		if (ret < 0)
+			goto error;
+
+		tdesc->rx_half_callback.callback = param->rx_complete_callback;;
+		tdesc->rx_half_callback.ctx = tdm_desc;
+		tdesc->rx_half_callback.event = NO_OS_EVT_DMA_RX_HALF_COMPLETE;
+		tdesc->rx_half_callback.peripheral = NO_OS_DMA_IRQ;
+		tdesc->rx_half_callback.handle = &tdesc->hsai;
+
+
+		ret = no_os_irq_register_callback(tdesc->nvic_rx_halfcplt, param->irq_id,
+						  &tdesc->rx_half_callback);
+		if (ret < 0)
+			goto error;
+
+		ret = no_os_irq_enable(tdesc->nvic_rx_halfcplt, tdm_desc->irq_id);
+		if (ret < 0)
+			goto error;
+	}
 	*desc = tdm_desc;
 
 	return 0;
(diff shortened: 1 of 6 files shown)


jansunil:
Hi Team- Any review comments on this?

include/no_os_tdm.h
```diff
@@ -107,8 +107,12 @@ struct no_os_tdm_platform_ops {
 				const struct no_os_tdm_init_param *);
 	/** TDM read operation function pointer */
 	int32_t (*tdm_ops_read)(struct no_os_tdm_desc *, void *, uint16_t);
+	/** TDM DMA read operation function pointer */
+	int32_t (*tdm_ops_read_dma)(struct no_os_tdm_desc *, void *, uint16_t);
```
buha, line 111: So these read_dma/write_dma ops are direct mappings of STM32 HAL in this no-OS top level include file. I think we shouldn't do this. In this file we need to be as generic as possible to cover all possible TDM peripherals and what we can do with them.

A better option would be to keep the read/write functions as they are but:
- add ops for stop()
- add ops for pause()
- add ops for resume()
- in no_os_tdm_init_param add callback for void tx_complete_callback(void *arg) and a void *tx_arg
- in no_os_tdm_init_param add callback for void rx_complete_callback(void *arg) and a void *rx_arg

All these new members can be left uninitialized and read()/write() will have normal blocking behavior.
However, if a callback is initialized, read()/write() should have non-blocking behavior.

Pause/Stop/Resume have no meaning in blocking mode so they can either silently return 0 or -EBADSTATE.

drivers/platform/stm32/stm32_tdm.c
```diff
@@ -219,3 +221,54 @@ int32_t stm32_tdm_read(struct no_os_tdm_desc *desc,
...
 }
+
+/**
+ * @brief Read data using SAI TDM DMA mode.
+ * @param desc - The TDM descriptor.
+ * @param data - The buffer to fill with the received data.
+ * @param nb_samples - Number of samples to read.
+ * @return 0 in case of success, negative error code otherwise.
+ */
+int32_t stm32_tdm_read_dma(struct no_os_tdm_desc *desc, void *data,
```
buha, line 232: ...continuation of my other comment:

This way, you will not implement new ops here, but rather modify existing read/write ops.
In them you will be looking of callbacks are defined and behave differently based on that.

```
if (desc->ops->tx_complete_callback)
    HAL_SAI_Receive_DMA();
else
    HAL_SAI_Receive();
```

jansunil: Got your point here.

include/no_os_tdm.h
jansunil: Do you suggest something similar to the [UART implementation](https://github.com/analogdevicesinc/no-OS/blob/master/drivers/platform/stm32/stm32_uart.c#L130) here?

buha: Exactly !

jansunil:
V2: Code changes as per review comments:
1) Removed the stm32_tdm_read_dma() function and added a condition check in the stm32_tdm_read() with the parameter ``rx_complete_callback``

2) Added a ``rx_complete_callback`` and a ``tx_complete_callback`` in the no_os_tdm_init_param.

3) Added the callback parameter in the no_os_tdm_desc so that the parameter check can be made in the stm32_tdm_read() function

include/no_os_tdm.h
jansunil: Added ops for stop/pause and resume explicitly.
Added a structure member for the callback functions as well.

drivers/platform/stm32/stm32_tdm.c
jansunil: The callback defined check has been made with ``desc->rx_complete_callback``. Would it be relevant to add the callback and make the check with the no_os_tdm_desc?

jansunil:
Hi @buha - are there any discrepancies between the proposed structure and the latest update?

include/no_os_tdm.h
```diff
@@ -109,6 +117,12 @@ struct no_os_tdm_platform_ops {
 	int32_t (*tdm_ops_read)(struct no_os_tdm_desc *, void *, uint16_t);
 	/** TDM write operation function pointer */
 	int32_t (*tdm_ops_write)(struct no_os_tdm_desc *, void *, uint16_t);
+	/** Pause TDM DMA transfer */
+	int32_t (*tdm_ops_pause_dma)(struct no_os_tdm_desc *);
```
buha, line 121: I'd drop the `_dma` suffix, it is unnecessary. Technically the implementation of the non-blocking TDM read/write could be done with just interrupts and memcpy's instead of DMA and this generic API needs to cover that case too.

drivers/platform/stm32/stm32_tdm.c
```diff
@@ -219,3 +228,46 @@ int32_t stm32_tdm_read(struct no_os_tdm_desc *desc,
...
+ * @return 0 in case of success, negative error code otherwise.
+ */
+int32_t stm32_stop_tdm_dma_transfer(struct no_os_tdm_desc *desc)
+{
+	int32_t	ret;
+	struct stm32_tdm_desc *tdesc;
+
+	if (!desc)
+		return -EINVAL;
+
```
buha, line 244: I think an extra check is needed here:

```
if (blocking)
    return -ENOSYS;
```

buha: it is up to you how you detect if user configured the API in blocking or non-blocking

drivers/platform/stm32/stm32_tdm.c
```diff
@@ -219,3 +228,46 @@ int32_t stm32_tdm_read(struct no_os_tdm_desc *desc,
...
+}
+
+/**
+ * @brief Pause SAI DMA transfer
+ * @param desc - The TDM descriptor.
+ * @return 0 in case of success, negative error code otherwise.
+ */
+int32_t stm32_pause_tdm_dma_transfer(struct no_os_tdm_desc *desc)
+{
+	return 0;
```
buha, line 262: resume/pause need to be implemented just like the stop

buha (commented):
looks good, we're almost there in my opinion

drivers/platform/stm32/stm32_tdm.c
buha: (and this comment applies to the implementations of pause and resume too)

include/no_os_tdm.h
jansunil: Noted. The "_dma" suffix has been removed everywhere

drivers/platform/stm32/stm32_tdm.c
jansunil: The blocking check has been made with respect to the definition of rx_callback.

drivers/platform/stm32/stm32_tdm.c
jansunil: Resume and Pause are also implemented now.

drivers/platform/stm32/stm32_tdm.c
```diff
@@ -129,6 +132,10 @@ int32_t stm32_tdm_init(struct no_os_tdm_desc **desc,
 		goto error;
 		break;
 	};
+	if (param->rx_complete_callback) {
```
buha, line 135: you don't need to check, you can simply copy it, if it's NUL you will anyway detect it later when checking `desc->rx_complete_callback`

jansunil: Removed the check here

buha:
One other thing is that the tx and rx callbacks are not used. Also, their argument should be passed by the user in init_param. The callbacks are provided by the user and they should be called by your implementation (in stm32_tdm.c) upon transfer completion.

But I'm still cooking the details of the implementation, I need to consult with the team.

buha:
So @jansunil ,
I think this is how it should be implemented.
As you already know, UART implements something similar, we can use that as a template.

Initialize the DMA Tx Complete interrupts as it is done here:
https://github.com/analogdevicesinc/no-OS/blob/master/drivers/platform/stm32/stm32_uart.c#L135-L156

Of course, for this to work you need to insert two new entries here, NO_OS_EVT_TDM_TX_COMPLETE, NO_OS_EVT_TDM_RX_COMPLETE and make the correlation with HAL_SAI_TX_COMPLETE_CB_ID, HAL_SAI_RX_COMPLETE_CB_ID
https://github.com/analogdevicesinc/no-OS/blob/8708a859e2c7a8d69c57db0905f8638d83132aa9/drivers/platform/stm32/stm32_irq.c#L68

And of course, a new entry is required here, for the TDM:
https://github.com/analogdevicesinc/no-OS/blob/7fc98116d819e9c0b0bec4cff8986f96f553d79e/include/no_os_irq.h#L87

buha:
All this, so that the user-passed tx/rx complete callbacks will be called when the events occur.

jansunil:
Thanks for this @buha . Let me work on this and get back.

buha:
But really, you should test this on your side, I have no setup that could test this, so make sure the interrupts work and that the callbacks get called ;)

buha:
@jansunil any updates on this ?

jansunil:
Hi @buha - May I take some more time to get back on this?

buha:
sure, of course

jansunil:
V3: Implemented SAI DMA callbacks
Added register callback for SAI DMA similar to UART interrupt mode.

jansunil:
V4: Fixed build errors

jansunil:
Hi @buha ,

Apologies for getting back late on this activity. The PR is ready for review now.

buha (approved):
Looks good, I think this will work.

I still have a question about what happens if we use DMA in a project for both (for example) SPI, UART and SAI - how will we differentiate between them ?

buha:
I'm merging this anyway, we can deal with that separately.

PR #1655, opened 2023-02-20
IMU support

rbolboac:
- added generic support for IMUs
- added driver for adis16505
- added examples for adis16505


drivers/imu/iio_adis.c
```diff
@@ -0,0 +1,1414 @@
...
+	if (!iio_adis->adis_dev)
+		return -EINVAL;
+
+	adis = iio_adis->adis_dev;
+
+	switch (channel->ch_num) {
+	case ADIS_GYRO_X:
+		ret = adis_read_xg_bias(adis, &res);
+		if (ret)
+			return ret;
```
CiprianRegus, line 158: You can move this if outside the `switch`.

drivers/imu/iio_adis.c
```diff
@@ -0,0 +1,1414 @@
...
+int iio_adis_read_rang_mdl(struct adis_dev* adis, char *buf)
+{
+	int ret;
+	unsigned int rang_mdl;
+	size_t len;
+	ret = adis_read_rang_mdl(adis, &rang_mdl);
+	if (ret)
+		return ret;
+
+	len = snprintf(buf, MAX_TXT_SIZE, rang_mdl_txt[rang_mdl]);
```
CiprianRegus, line 570: return directly

drivers/imu/adis.c
```diff
@@ -0,0 +1,2070 @@
...
+		msgs_to_send[send_idx] = msgs[2];
+		msgs_to_send[send_idx + 1] = msgs[3];
+		send_idx += 2;
+		break;
+	default:
+		return -EINVAL;
+	}
+
+	ret = no_os_spi_transfer(adis->spi_desc, msgs_to_send, send_idx);
+	if (ret < 0)
```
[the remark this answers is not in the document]
rbolboac: this is actually related to linux platform.... linux_spi_transfer will actually return a positive value upon success, that is why if (ret) would not work in this case

drivers/imu/adis.c
```diff
@@ -0,0 +1,2070 @@
...
+		break;
+	default:
+		return -EINVAL;
+	}
+
+	for (i = 1; i <= size; i++, send_idx ++)
+		msgs_to_send[send_idx] = msgs[i];
+
+	ret = no_os_spi_transfer(adis->spi_desc, msgs_to_send, send_idx);
+	if (ret < 0)
```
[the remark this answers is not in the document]
rbolboac: same as above

drivers/imu/adis.c
```diff
@@ -0,0 +1,2075 @@
...
+
+	if(data->has_paging)
+		dev->current_page = -1;
+	else
+		dev->current_page = 0;
+
+	ret = no_os_gpio_get_optional(&dev->gpio_reset, data->adis_ip->gpio_reset);
+	if (ret) {
+		pr_warning("No reset pin found \n");
+		goto error_gpio;
```
CiprianRegus, line 120: Should this be an error?

rbolboac: no, it should be just a warning and should continue with initialization, will modify, thanks!

buha (approved):

rbolboac:
- fixed maxim build
- changed scan mode to big endian false

PR #1663, opened 2023-02-21
platform:mbed Added initial lvgl interface support

mphalke:
1. Added mbed drivers for initial lvgl interface functionality
2. Added lvgl API and include files

### DIFF
--- a/drivers/platform/mbed/mbed_lvgl.h
+++ b/drivers/platform/mbed/mbed_lvgl.h
@@ -1,0 +1,87 @@
+/***************************************************************************//*
+ * @file   mbed_lvgl.h
+ * @brief  Header containing extra types for lvgl library interface
+******************************************************************************
+ * Copyright (c) 2023 Analog Devices, Inc.
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *  - Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ *  - Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in
+ *    the documentation and/or other materials provided with the
+ *    distribution.
+ *  - Neither the name of Analog Devices, Inc. nor the names of its
+ *    contributors may be used to endorse or promote products derived
+ *    from this software without specific prior written permission.
+ *  - The use of this software may or may not infringe the patent rights
+ *    of one or more patent holders.  This license does not release you
+ *    from the requirement that you obtain separate licenses from these
+ *    patent holders to use this software.
+ *  - Use of the software either in source or binary form, must be run
+ *    on or directly connected to an Analog Devices Inc. component.
+ *
+ * THIS SOFTWARE IS PROVIDED BY ANALOG DEVICES "AS IS" AND ANY EXPRESS OR
+ * IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, NON-INFRINGEMENT,
+ * MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED.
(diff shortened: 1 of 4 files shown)


drivers/platform/mbed/mbed_lvgl.h
```diff
@@ -0,0 +1,87 @@
...
+/******************************************************************************/
+/********************** Variables and User defined data types *****************/
+/******************************************************************************/
+
+/**
+* @struct mbed_lvgl_init_param
+* @brief Structure holding the mbed specific lvgl init parameters
+*/
+struct mbed_lvgl_init_param {
+	uint32_t tick_period_us;	// lvgl tick period in usec
```
buha, line 62: i'm not familiar with lvgl, but is this tick period lvgl-specific or it's a setting that is only possible when using mbed ?

mphalke: lvgl is a graphics library for portable display based embedded applications. lvgl task handler need one periodic timer with period in msec range. In case of mbed platform, we have used the mbed ticket module (derived from systick timer) to generate ticks for lvgl task handler.
For STM32 platform with lvgl, the system tick handler callback function takes care of tick count for lvgl task handler. The implementation is here: https://github.com/lvgl/lv_port_stm32f769_disco/blob/master/src/stm32f7xx_it.c#L39

So to answer, tick period is lvgl specific but in my opinion it is better to keep tick period handling separate for mbed and other platforms. Each platform may implement timer in different ways.

buha: Then I really don't understand the purpose of this PR.
It seems you only need a ticker at a specific frequency that calls a callback where you have to perform `lv_tick_inc(1)` .

In that case either deal with it in application code (calling HAL stuff directly) or, if you care about your project running on multiple platforms, extend the mbed_timer.cpp to be able to configure a timer that fires an interrupt which calls your desired callback (in this example it's heartbeat()):
https://github.com/analogdevicesinc/no-OS/blob/microwave/projects/microwave/src/examples/iio_admv96x1/main.c#L31

The example above uses RTC, you could use the RTC implementation on Maxim to get inspiration on how to do that for Mbed Timer.

mphalke: There could be other lvgl platform specific functions as well. I am still exploring them, so please hold off the review until I figure our those. If not, as suggested, ticker can be implemented through timer.

mphalke: @buha I have added few more supporting functions to init lvgl and related h/w.

So you need lvgl_init() function which is part of lvgl library and is purely abstracted, so no h/w dependency.
You also need touchpad and tft/display init functions, which must be vendor specific (e.g. stm displays- https://github.com/lvgl/lv_port_stm32f769_disco/tree/master/hal_stm_lvgl). These 2 functions can be part of application.

Let me know if this makes sense to you.

dbogdan:
@mphalke, I don't think that this is the right approach: similar to the rest of the external libraries, lvgl should go to no-OS/libraries and it shouldn't require any specific drivers/platform layer. If you need a lvgl_init(), this will be platform agnostic (and it will be still located to no-OS/libraries) - the application will pass the platform dependent information similar to a standard device driver.

mphalke:
> 

Hi @dbogdan It make sense. I agree that lvgl_init() function can be part of application and lvgl library can go into libraries folder. But the other 2 functions I added for tft and touchpad init are platform/vendor dependent and I have added them as a callback functions to init param structure. So basically lvgl library make these functions mandatory. Can those be part of lvgl no-os init param structure then?

mphalke:
Closing this PR as I need to look this further and see better way of adding platform dependency for lvgl implementation.
For now, lvgl time tick can be derived by directly calling platform driver APIs (ticker irq in case of Mbed), so this may not be needed as a part of lvgl platform drivers layer.
The other thing that I need to look is, are there any APIs in lvgl library that needs to be initialized through platform drivers layer. I highlighted tft_init and touchpad_init, but if they have naming consistency across multiple vendors for the lvgl library, then these functions could also be invoked through application/library level.

PR #1664, opened 2023-02-21
add PID control utility to no-OS

buha:
This PID supports:
- setting Kp, Ki, Kd gains
- clipping output to a certain range to match the range of whatever PID is controlling (for example, for an 8-bit PWM, the algorithm would clip the output to a range between 0 and 255).
- hysteresis
- clipping internal integrator to a certain range to avoid overflow

The no_os_pid_control() must be caled in a loop and user is responsible for detecting when the control should stop by not calling it anymore.

This can be done either by characterizing the algorithm for a particular set of configuration parameters and finding a minimum number of iterations that make the output stable or by looking at the output and detecting when it sits within a satisfactory tolerance from the set point.

### DIFF
--- a/include/no_os_pid.h
+++ b/include/no_os_pid.h
@@ -1,0 +1,83 @@
+/***************************************************************************//**
+ *   @file   no_os_pid.h
+ *   @brief  Header file for PID control utility.
+ *   @author Darius Berghe (darius.berghe@analog.com)
+********************************************************************************
+ * Copyright 2023(c) Analog Devices, Inc.
+ *
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *  - Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ *  - Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in
+ *    the documentation and/or other materials provided with the
+ *    distribution.
+ *  - Neither the name of Analog Devices, Inc. nor the names of its
+ *    contributors may be used to endorse or promote products derived
+ *    from this software without specific prior written permission.
+ *  - The use of this software may or may not infringe the patent rights
+ *    of one or more patent holders.  This license does not release you
+ *    from the requirement that you obtain separate licenses from these
+ *    patent holders to use this software.
+ *  - Use of the software either in source or binary form, must be run
+ *    on or directly connected to an Analog Devices Inc. component.
+ *
+ * THIS SOFTWARE IS PROVIDED BY ANALOG DEVICES "AS IS" AND ANY EXPRESS OR
+ * IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, NON-INFRINGEMENT,
+ * MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED.
+ * IN NO EVENT SHALL ANALOG DEVICES BE LIABLE FOR ANY DIRECT, INDIRECT,
+ * INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, INTELLECTUAL PROPERTY RIGHTS, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+*******************************************************************************/
+#ifndef _NO_OS_PID_H
+#define _NO_OS_PID_H
+#include <stdlib.h>
+#include <stdio.h>
+#include <stdint.h>
+
+/**
+ * @struct no_os_pid_range
+ * @brief Range definition for limiting PID control output or internal integrator accumulator
+ */
+struct no_os_pid_range {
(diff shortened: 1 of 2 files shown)


include/no_os_pid.h
```diff
@@ -0,0 +1,66 @@
...
+ * MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED.
+ * IN NO EVENT SHALL ANALOG DEVICES BE LIABLE FOR ANY DIRECT, INDIRECT,
+ * INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, INTELLECTUAL PROPERTY RIGHTS, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+*******************************************************************************/
+#include <stdlib.h>
```
buha, line 39: forgot to add header guard, will fix

util/no_os_pid.c
```diff
@@ -0,0 +1,145 @@
...
+	return pid->output;
+}
+
+/* De-initialize a PID controller by freeing the allocated memory. */
+int no_os_pid_remove(struct no_os_pid *pid)
+{
+	if (!pid)
+		return -EINVAL;
+
+	free(pid);
```
CiprianRegus, line 142: nitpick: set pid to `NULL`

util/no_os_pid.c
```diff
@@ -0,0 +1,145 @@
...
+	(*pid)->output = config.initial;
+	(*pid)->iacc = 0;
+	(*pid)->dacc = 0;
+
+	return 0;
+}
+
+/* Perform PID control given a controller, a set-point and a process variable.
+ * This has to be called several times, until output stabilizes.
+ * The PID output is returned by the function. */
```
CiprianRegus, line 73: I guess these functions need doxygen style comment headers.

util/no_os_pid.c
```diff
@@ -0,0 +1,145 @@
...
+int no_os_pid_control(struct no_os_pid *pid, int SP, int PV)
+{
+	int ERR; // error
+	int64_t P; // proportional component
+	int64_t I; // integral component
+	int64_t D; // derivative component
+
+	if (!pid)
+		return 0;
```
amiclaus, line 82: do we want to return a success code here?

amiclaus (commented):
some comments on my side.

include/no_os_pid.h
```diff
@@ -0,0 +1,66 @@
...
+
+/* Range definition for limiting PID control output or internal integrator accumulator. */
+struct no_os_pid_range {
+	int high; // high limit of the range
+	int low; // low limit of the range
+};
+
+/* Configuration of the PID. */
+struct no_os_pid_config {
+	unsigned int Kp; // proportional gain (micro-units)
```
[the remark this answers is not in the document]
buha: I agree in general but everywhere in PID theory I found these parameters written with capital K.
Whomever is maintains this code finds it easier to read Kp compared to p_gain or k_p or whatever because it's exactly as in theory...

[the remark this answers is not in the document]
buha: same goes for SP and PV

util/no_os_pid.c
```diff
@@ -0,0 +1,145 @@
...
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+*******************************************************************************/
+#include <errno.h>
+#include "no_os_pid.h"
+#include "no_os_print_log.h"
+
+struct no_os_pid {
+	int iacc; // integral accumulator
```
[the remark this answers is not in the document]
buha: because user should not have access to the members, this is internal/private stuff

util/no_os_pid.c
buha: So basically what this does, if user passes pid = 0 (the only wrong value we can detect, there can be countless others), the control function returns a constant output.

This is acceptable because it's the kind of error that you catch in development.

Although actually it should output not 0 but min(output_clip.low, output_clip.high).

buha: I guess I will change the function to return a status and the output will be a pointer argument...

util/no_os_pid.c
```diff
@@ -0,0 +1,145 @@
...
+	return 0;
+}
+
+/* Perform PID control given a controller, a set-point and a process variable.
+ * This has to be called several times, until output stabilizes.
+ * The PID output is returned by the function. */
+int no_os_pid_control(struct no_os_pid *pid, int SP, int PV)
+{
+	int ERR; // error
+	int64_t P; // proportional component
```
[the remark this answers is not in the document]
buha: I've changed ERR, I, P, D  to lowercase in v2

buha:
v2:
- added doxygen
- changed some variables to lowercase
- control function outputs status rather than the output (which is now output via a pointer passed as argument)

include/no_os_pid.h
```diff
@@ -0,0 +1,66 @@
...
+#include <stdio.h>
+#include <stdint.h>
+
+/* Range definition for limiting PID control output or internal integrator accumulator. */
+struct no_os_pid_range {
+	int high; // high limit of the range
+	int low; // low limit of the range
+};
+
+/* Configuration of the PID. */
```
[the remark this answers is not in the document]
buha: done in v2

amiclaus (approved):

buha:
V3:
- fixed problem regarding pid control output being returned instead of written to output argument
- moved declaration of variables at the beginning of pid control function

CiprianRegus (approved):

PR #1666, opened 2023-02-22
libraries: add azure-sdk-for-c support

amiclaus:
Integrate azure-sdk-for-c into the no-os framework.

### DIFF
--- a/tools/scripts/azure_sdk_for_c.mk
+++ b/tools/scripts/azure_sdk_for_c.mk
@@ -1,0 +1,5 @@
+AZURE_DIR = $(NO-OS)/libraries/azure-sdk-for-c
+AZURE_DIR_BUILD = $(AZURE_DIR)/build
+AZURE_DIR_BUILD_LIBS = $(AZURE_DIR_BUILD)/sdk/src/azure
+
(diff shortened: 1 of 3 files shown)


tools/scripts/azure_sdk_for_c.mk
```diff
@@ -0,0 +1,5 @@
+AZURE_DIR = $(NO-OS)/libraries/azure-sdk-for-c
+AZURE_DIR_BUILD = $(AZURE_DIR)/build/
+AZURE_DIR_BUILD_LIBS = $(AZURE_DIR_BUILD)sdk/src/azure/
```
buha, line 3:
```
$(AZURE_DIR_BUILD)/sdk/src/azure/
..................^
```

This slash is needed in case AZURE_DIR_BUILD doesn't contain a trailing slash

amiclaus (commented):
v2:
- rework trailing slashes when appending variables in the makefiles

tools/scripts/azure_sdk_for_c.mk
buha: LOL I thought it comes from the user.

PR #1668, opened 2023-02-24
mbedtls_config: update to default values

amiclaus:
[network: mbedtls_config: use default buffer length](https://github.com/analogdevicesinc/no-OS/commit/422cf57ab2d7cea79a02a3b754409d0b2206c32d)

### DIFF
--- a/network/noos_mbedtls_config.h
+++ b/network/noos_mbedtls_config.h
@@ -90,7 +90,7 @@
  * 2000 is a eoungh to do the tls handshake and is no to much
  * platforms with memory constrains like ADuCM3029
  */
-#define MAX_CONTENT_LEN 2500
+//#define MAX_CONTENT_LEN 2500
 
 /*
  * ENABLE_MEMORY_OPTIMIZATIONS should be defined in the case memory


amiclaus (commented):
v2:
- do not remove comments related to specific platform since they might be useful for the user when configuring own application based on a specific platform

amiclaus (commented):
v3:
- remove PEM cert related commit since it is no longer relevant after https://github.com/analogdevicesinc/no-OS/pull/1670 is merged.

buha:
I don't understand this PR.
If this MAX_CONTENT_LEN is not supposed to be used, there are some occurences of it left which should be removed:
```
$ ag MAX_CONTENT_LEN -w --cc
network/tcp_socket.c
68:#ifdef MAX_CONTENT_LEN
69:#define DEFAULT_CONNECTION_BUFFER_SIZE MAX_CONTENT_LEN
72:#endif /* MAX_CONTENT_LEN */

network/noos_mbedtls_config.h
93:#define MAX_CONTENT_LEN 2500
129:#ifdef MAX_CONTENT_LEN
130:#define MBEDTLS_SSL_MAX_CONTENT_LEN	MAX_CONTENT_LEN
```

If it is supposed to be left in the code and configured by the user (by editing network/noos_mbedtls_config.h) than we should really think of a better way to define it by project, not globally.

amiclaus (commented):
This PR is not quite mandatory after https://github.com/analogdevicesinc/no-OS/pull/1670 was introduced. And indeed, the config file is application specific, not framework specific. But I'd keep this configuration file in `network/noos_mbedtls_config.h` as a reference for users who will define their own config file.

The idea of this PR was to make the generic noos config file as default as possible. And the MAX_CONTENT_LEN define was specific to aducm3029.

amiclaus:
Since the PR is not mandatory after #1670 was opened, will close it.

PR #1670, opened 2023-02-27
tools: scripts: allow custom mbedtls config

amiclaus:
mbedtls library can be build based on custom configuration file defined by the user. Although there is already a "generic" configuration file provided for the no-os framework (noos_mbedtls_config.h), allow the user to create its own custom config file since these are applications specific rather than framework specific.


rbolboac (approved):

buha:
Just saw this PR after seeing the other one. Can you put them together ?

amiclaus (commented):
Together with which of the PRs?

PR #1677, opened 2023-03-02
spi_transfer null buffers fix

rbolboac:
Use rx_buff and tx_buff in spi_transfer only if they are not null.
Fixes for stm32 and pico platform.

### DIFF
--- a/drivers/platform/stm32/stm32_spi.c
+++ b/drivers/platform/stm32/stm32_spi.c
@@ -259,6 +259,8 @@ int32_t stm32_spi_transfer(struct no_os_spi_desc *desc,
 	uint64_t slave_id;
 	static uint64_t last_slave_id;
 	int ret;
+	uint32_t tx_cnt = 0;
+	uint32_t rx_cnt = 0;
 
 	if (!desc || !desc->extra || !msgs)
 		return -EINVAL;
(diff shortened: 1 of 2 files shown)


amiclaus (approved):

buha:
I don't quite understand the context here. Are the HAL functions not handling NULL pointers correctly or why do we need to do this ?

drivers/platform/stm32/stm32_spi.c
```diff
@@ -288,14 +292,24 @@ int32_t stm32_spi_transfer(struct no_os_spi_desc *desc,
...
 		__HAL_SPI_ENABLE(&sdesc->hspi);
-		while(msgs[i].bytes_number--) {
+
+		while ((msgs[i].rx_buff && rx_cnt < msgs[i].bytes_number) ||
+		       (msgs[i].tx_buff && tx_cnt < msgs[i].bytes_number)) {
 			while(!(SPIx->SR & SPI_SR_TXE))
 				;
-			*(volatile uint8_t *)&SPIx->DR = *msgs[i].tx_buff++;
+			if (msgs[i].tx_buff)
+				*(volatile uint8_t *)&SPIx->DR = msgs[i].tx_buff[tx_cnt++];
```
buha, line 301: ok, for stm32 I understand what you did, HAL is not used for transfer so we need to take care of the pointers, but the question still remains for pico platform

rbolboac: the same goes for pico platform, please see the code from: https://github.com/raspberrypi/pico-sdk/blob/master/src/rp2_common/hardware_spi/spi.c#L76
dst and src are not checked if they are ```NULL```

PR #1693, opened 2023-03-13
drivers: adc: ad7779: fix reference set

amiclaus:
According to the datasheet (page 78), the ADC reference is set using the ADC_MUX_CONFIG register, REF_MUX_CTRL bits.

Fixes: 2b8c1f7 ("ad7779: Initial Preliminary Version")

NOTE: the implementation was done according to the driver previous implementation (mask defines, bit defines, etc)


amiclaus (commented):
v2:
- fix also the getter function

dbogdan (approved):

buha (approved):

PR #1694, opened 2023-03-13
Fix Maxim drivers warnings

CiprianRegus:


drivers/platform/maxim/max32650/maxim_i2c.c
```diff
@@ -40,6 +40,7 @@
 
 #include <stdint.h>
 #include <stdlib.h>
+#include <string.h>
```
amiclaus, line 43: this change is unrelated to the commit. is this include required? same for all the other files changed in this commit.

buha: i think it is, probably for strcpy() usage

buha: or memcpy()

buha (approved):

drivers/platform/maxim/max32650/maxim_i2c.c
CiprianRegus: Indeed, it's for memcpy(), but I'll change the commit message, since it's totally unrelated.

PR #1696, opened 2023-03-14
drivers:platform:mbed:mbed_gpio_irq.cpp:Add Set Priority

jansunil:
Added `mbed_gpio_irq_set_priority()` function. In order to do this, the interrupt number EXTI_IRQn (n=0,1,..16) needs to be retrieved. There is a child class derived from the `InterruptIn` class to inherit the protected member `gpio_irq`

### DIFF
--- a/drivers/platform/mbed/mbed_gpio_irq.cpp
+++ b/drivers/platform/mbed/mbed_gpio_irq.cpp
@@ -41,6 +41,7 @@
 
 #include <stdio.h>
 #include <mbed.h>
+#include <InterruptIn.h>
 
 // Platform support needs to be C-compatible to work with other drivers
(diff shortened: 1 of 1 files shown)


CiprianRegus (approved):
Looks good.

jansunil:
Deleting this PR as the code change seems to be target specific (specific to STM family). The [https://github.com/analogdevicesinc/no-OS/pull/1696/commits/51fe4d9dda05e0aea26bfc3dcb52bf0333156935#diff-84911571ffb76aac1634b7eca526f8f765a671d48ef2b780bc0f89149fe72a2fR87](irq_n) structure member is seen only in case of STM targets and not in the others. 
This change would make the mbed driver not generic. Hence deleting this.

PR #1702, opened 2023-03-20
ADT7420-pmdz fixes

CiprianRegus:
A few fixes for the ADT7420 driver and project.


projects/adt7420-pmdz/src/examples/dummy/dummy_example.c
```diff
@@ -85,6 +85,8 @@ int dummy_example_main()
 	if (ret)
 		goto error_adt7420;
 
+	no_os_mdelay(240);
```
amiclaus, line 88: nit: usually for these specific delay values, if they are datasheet related, we add a comment in the code.

amiclaus (approved):
one nit

CiprianRegus:
- Changed the I2C address to 0x48, which is the default for the ADT7420 PMOD board.
- Added a comment which explains the delay.

PR #1715, opened 2023-03-23
Dev/adis support

rbolboac:
eval-adis project related changes:
- Remove IIO example and linux platform support
- Remove read_samples API
- Make adis trigger synchronous
- Add lost samples attribute
- Add some extra checks for register writes
- Change sync mode update implementation with adis init and adis frequency setting updates
- Rework burst data reading implementation and adapt trigger handler accordingly


drivers/imu/iio_adis.c
```diff
@@ -948,38 +957,93 @@ int adis_iio_trigger_handler(struct iio_device_data *dev_data)
...
-	if (dev_data->buffer->active_mask & NO_OS_BIT(ADIS_DELTA_VEL_Y))
-		buff[i++] = adis->burst_sel ? burst_data.y_accl : 0;
-	if (dev_data->buffer->active_mask & NO_OS_BIT(ADIS_DELTA_VEL_Z))
-		buff[i++] = adis->burst_sel ? burst_data.z_accl : 0;
-
-	return iio_buffer_push_scan(dev_data->buffer, &buff[0]);
+	uint8_t temp_offset = iio_adis->burst_size ? 13 : 7;
+	uint8_t data_cntr_offset = iio_adis->burst_size ? 14 : 8;
+
+	uint16_t current_data_cntr = no_os_get_unaligned_be16((uint8_t*)
```
CiprianRegus, line 967: Move these in the variable declaration block, there are several other places where this is the case. Also, using `no_os_bswap_constant_16()` would be cleaner.

drivers/imu/iio_adis.c
```diff
@@ -948,38 +957,93 @@ int adis_iio_trigger_handler(struct iio_device_data *dev_data)
...
+				if(res1 > res2) {
+					iio_adis->samples_lost += res1/res2;
+					if(res1%res2 < res2/2)
+						iio_adis->samples_lost--;
+				}
+			}
+
+		} else /* data counter overflowed occurred */
+			if (iio_adis->sync_mode != ADIS_SYNC_SCALED)
+				iio_adis->samples_lost += 0xFFFFU - iio_adis->data_cntr + current_data_cntr;
```
CiprianRegus, line 987: `NO_OS_U16_MAX`

drivers/imu/iio_adis.c
```diff
@@ -948,38 +957,93 @@ int adis_iio_trigger_handler(struct iio_device_data *dev_data)
...
+				uint32_t res2 = NO_OS_DIV_ROUND_CLOSEST(1000000, iio_adis->sampling_frequency);
+
+				if(res1 > res2) {
+					iio_adis->samples_lost += res1/res2;
+					if(res1%res2 < res2/2)
+						iio_adis->samples_lost--;
+				}
+			}
+
+		} else /* data counter overflowed occurred */
```
CiprianRegus, line 985: overflowed -> overflow

drivers/imu/iio_adis.h
```diff
@@ -65,13 +65,21 @@ enum adis16505_chan_type {
...
 	struct iio_device *iio_dev;
 	uint32_t active_channels;
 	uint8_t no_of_active_channels;
+	uint16_t samples_lost;
+	uint16_t data_cntr;
+	uint32_t sampling_frequency;
+	unsigned int burst_size;
+	unsigned int burst_sel;
+	unsigned int sync_mode;
+	uint16_t data[26];
```
CiprianRegus, line 82: Wouldn't it be simpler to use `uint8_t` instead?

rbolboac: Not really... The burst data is either on 16 bits or on 32 so it is easier to add it in the buffer

drivers/imu/adis.c
```diff
@@ -494,22 +471,43 @@ static void adis_update_diag_flags(struct adis_dev *adis, uint16_t diag_stat)
...
 /**
  * @brief Read burst data.
  * @param adis       - The adis device.
  * @param burst_data - Array filled with read data.
+ * @param burst_size - Burst size encoded value.
  * @return 0 in case of success, error code otherwise.
  */
-int adis_read_burst_data(struct adis_dev *adis,
-			 struct adis_burst_data *burst_data)
+int adis_read_burst_data(struct adis_dev *adis, uint16_t *burst_data,
```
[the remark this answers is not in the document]
rbolboac: So I think there was a confusion here so I changed the prototype of the function...
Now we have burst_data_size which should be the size of burst_data and I renamed burst_size to burst_size_selection to make it clear that this related to the device configuration and not the size of the burst_data.

drivers/imu/adis.c
```diff
@@ -527,53 +525,12 @@ int adis_read_burst_data(struct adis_dev *adis,
...
-		burst_data->y_accl = no_os_get_unaligned_be32(
-					     &buffer[ADIS_YACCL_IDX_32_BIT_BURST + ADIS_READ_BURST_DATA_CMD_SIZE]);
-		burst_data->z_accl = no_os_get_unaligned_be32(
-					     &buffer[ADIS_ZACCL_IDX_32_BIT_BURST + ADIS_READ_BURST_DATA_CMD_SIZE]);
-		burst_data->temp_out = no_os_get_unaligned_be16(
-					       &buffer[ADIS_TEMP_IDX_32_BIT_BURST + ADIS_READ_BURST_DATA_CMD_SIZE]);
-		burst_data->data_cntr = no_os_get_unaligned_be16(
-						&buffer[ADIS_CNT_IDX_32_BIT_BURST + ADIS_READ_BURST_DATA_CMD_SIZE]);
-	}
+	/* All to buffer all data, except checksum data */
```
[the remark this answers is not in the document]
rbolboac: Made a few adjustments to the comment as well

CiprianRegus (approved):

PR #1724, opened 2023-04-03
iio: iio.c: Fix buffer maximum length based on received request

rbolboac:
When using the statically allocated buffer with circular buffer, make sure the used size is a multiple of the total requested number of bytes from the iio client.
This has to be done because the circular buffer cannot handle memory wrapping.


danmois (approved):
lgtm

amiclaus (approved):

PR #1739, opened 2023-04-17
drivers/platform/chibios: bugfix

rbudai98:
The calloc replacements needed extra var plus missing .h in memory alloc

### DIFF
--- a/drivers/platform/chibios/chibios_alloc.c
+++ b/drivers/platform/chibios/chibios_alloc.c
@@ -39,6 +39,7 @@
 
 #include "no_os_alloc.h"
 #include "hal.h"
+#include <string.h>
 
 /**
  * @brief Allocate memory and return a pointer to it.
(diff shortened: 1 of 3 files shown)


drivers/platform/chibios/chibios_alloc.c
```diff
@@ -39,6 +39,7 @@
 
 #include "no_os_alloc.h"
 #include "hal.h"
+#include <string.h>
```
rbolboac, line 42: I would add this line to a different commit and a Fixes tag would be nice for both commits

rbudai98:
v2:
- split into different commits

rbolboac (commented):
for platform/chbios/chibios_spi.c: param fix commit message
it looks like you have a wrong commit hash in Fixes, I think it should be 8487f1b

amiclaus (approved):

PR #1746, opened 2023-04-25
CI:configure new agent for projects build

ccraluca:
New Self hosted agent was configured for no-OS projects on Azure Pipelines.

### DIFF
--- a/build_projects.yml
+++ b/build_projects.yml
@@ -12,8 +12,11 @@ variables:
   value: $[eq(variables['Build.SourceBranch'], 'refs/heads/master')]
 
 jobs:
-- job: Projects
-  pool: LocalPool
+- job: Xilinx
+  pool: 
+    name: Default
+    demands:
+      - agent.name -equals no-OS_xilinx
   steps:
   - checkout: self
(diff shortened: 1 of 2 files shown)


build_projects.yml
```diff
@@ -13,17 +14,19 @@ variables:
...
+  pool: 
+    name: Default
+    demands:
+      - agent.name -equals no-OS
   steps:
   - checkout: self
     fetchDepth: 50
     submodules: true
     clean: true
-    persistCredentials: true
```
amiclaus, line 22: this might be redundant.

build_projects.yml
```diff
@@ -13,17 +14,19 @@ variables:
...
   - script: 'python3 ./tools/scripts/build_projects.py . 
                -export_dir $(RELEASE_DIR)
                -log_dir $(LOG_DIR)
-               -builds_dir $(BUILDS_DIR)'
+               -builds_dir /no-OS_builds/builds'
```
amiclaus, line 29: we should modify the BUILDS_DIR value in the azure variables group instead of hardcoding the path.

PR #1748, opened 2023-04-27
drivers: imu: adis: Refactor adis code

rbolboac:
Refactor adis driver related code:
- add missing documentation
- rename structures and APIs
- reorder APIs in files
- remove data which is not used


amiclaus (commented):
there seem to be some valid doxygen issues.

amiclaus (approved):

buha (approved):
i like the field idea, fits well in this scenario

PR #1756, opened 2023-05-04
drivers:io-expander: added ltc4306 support

jsanbuen:
Added driver implementation for ltc4306

More information: https://www.analog.com/media/en/technical-documentation/data-sheets/4306.pdf

### DIFF
--- a/drivers/io-expander/ltc4306/ltc4306.c
+++ b/drivers/io-expander/ltc4306/ltc4306.c
@@ -1,0 +1,817 @@
+/***************************************************************************//**
+ *   @file   ltc4306.c
+ *   @brief  Implementation of ltc4306 driver.
+ *   @author Jose Ramon San Buenaventura (jose.sanbuenaventura@analog.com)
+********************************************************************************
+ * Copyright 2023(c) Analog Devices, Inc.
+ *
+ * All rights reserved.
+ *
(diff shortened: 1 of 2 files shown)


drivers/io-expander/ltc4306/ltc4306.c
```diff
@@ -0,0 +1,817 @@
...
+ * @param dev - The device structure.
+ * @param upstream_en - Set true to enable upstream rise time accelerator, false otherwise.
+ *
+ * @return 0 in case of success, negative error code otherwise.
+*******************************************************************************/
+int ltc4306_set_upstream_accel(struct ltc4306_dev *dev, bool upstream_en)
+{
+	dev->upstream = upstream_en;
+
+	if (upstream_en)
```
amiclaus, line 392: wondering if you can't just pass `upstream_en` directly to `ltc4306_reg_update`. this will reduce the number of function calls to 1 in this function.

jsanbuen: similar structure was used from previous submissions i.e., MAX9611 and MAX31865 as recommended by one of the reviewers. I can change and test the suggested revision for this submission but I might also need to do the same with the other previous submissions for uniformity.

amiclaus (approved):

PR #1758, opened 2023-05-05
drivers:adc:ad4858 Added ad4858 driver support

mphalke:
Initial implementation of the ad4858 no-os drivers

### DIFF
--- a/drivers/adc/ad4858/ad4858.c
+++ b/drivers/adc/ad4858/ad4858.c
@@ -1,0 +1,998 @@
+/***************************************************************************//**
+ *   @file   ad4858.c
+ *   @brief  Source file for the ad4858 drivers
+********************************************************************************
+ * Copyright 2023(c) Analog Devices, Inc.
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *  - Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ *  - Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in
+ *    the documentation and/or other materials provided with the
+ *    distribution.
+ *  - Neither the name of Analog Devices, Inc. nor the names of its
+ *    contributors may be used to endorse or promote products derived
+ *    from this software without specific prior written permission.
+ *  - The use of this software may or may not infringe the patent rights
+ *    of one or more patent holders.  This license does not release you
+ *    from the requirement that you obtain separate licenses from these
+ *    patent holders to use this software.
+ *  - Use of the software either in source or binary form, must be run
+ *    on or directly connected to an Analog Devices Inc. component.
+ *
+ * THIS SOFTWARE IS PROVIDED BY ANALOG DEVICES "AS IS" AND ANY EXPRESS OR
+ * IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, NON-INFRINGEMENT,
+ * MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED.
+ * IN NO EVENT SHALL ANALOG DEVICES BE LIABLE FOR ANY DIRECT, INDIRECT,
+ * INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, INTELLECTUAL PROPERTY RIGHTS, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+*******************************************************************************/
+#include <stdio.h>
+#include <stdlib.h>
+#include <stdbool.h>
+#include <string.h>
+#include "ad4858.h"
+#include "no_os_delay.h"
+#include "no_os_print_log.h"
+
+/**
+ * @brief Write device register.
+ * @param dev- The device structure.
+ * @param reg_addr - The register address.
+ * @param reg_val - The data to be written.
+ * @return 0 in case of success, negative error code otherwise.
+ * @note Multibyte write with streaming mode is supported by default.
+ */
+int ad4858_reg_write(struct ad4858_dev *dev, uint32_t reg_addr,
+		     uint32_t reg_val)
+{
+	int32_t ret;
+	uint32_t addr;
+	uint8_t cnt = 0;
+	uint8_t reg_size = AD4858_LEN(reg_addr);
+	uint8_t buf[6] = {0};
+
+	if (!dev)
+		return -EINVAL;
+
+	if (!dev->addr_ascension_enable)
+		addr = AD4858_ADDR(reg_addr) + reg_size - 1;
+	else
+		addr = AD4858_ADDR(reg_addr);
+
+	buf[cnt++] = (uint8_t)(addr >> 8);
+	buf[cnt++] = (uint8_t)addr;
+
+	if (dev->big_endian) {
+		if (dev->addr_ascension_enable)
+			no_os_memswap64((uint8_t *)(&reg_val + 1) - reg_size, reg_size, reg_size);
+		memcpy(&buf[cnt], (uint8_t *)(&reg_val + 1) - reg_size, reg_size);
+	} else {
+		if (!dev->addr_ascension_enable)
+			no_os_memswap64(&reg_val, reg_size, reg_size);
+		memcpy(&buf[cnt], &reg_val, reg_size);
+	}
+
+	/* Add the register data size to calculate an actual length of SPI frame */
+	cnt += reg_size;
+
+	ret = no_os_spi_write_and_read(dev->spi_desc, buf, cnt);
+	if (ret)
+		return ret;
+
+	return 0;
+}
+
+/**
+ * @brief Read device register.
+ * @param dev - The device structure.
+ * @param reg_addr - The register address.
+ * @param reg_val - The data read from the register.
+ * @return 0 in case of success, negative error code otherwise.
+ * @note Multibyte read with streaming mode is supported by default.
+ */
+int ad4858_reg_read(struct ad4858_dev *dev, uint32_t reg_addr,
+		    uint32_t *reg_val)
+{
+	int32_t ret;
+	uint32_t addr;
+	uint8_t cnt = 0;
+	uint8_t dindx = 0;
+	uint8_t reg_size = AD4858_LEN(reg_addr);
+	uint8_t buf[6] = {0};
+
+	if (!dev || !reg_val)
+		return -EINVAL;
+
+	if (!dev->addr_ascension_enable)
+		addr = AD4858_ADDR(reg_addr) + reg_size - 1;
+	else
+		addr = AD4858_ADDR(reg_addr);
+
+	/* Store the address (16-bit instruction phase with R/W=1) */
+	buf[cnt++] = (uint8_t)(AD4858_REG_RD_BIT_MSK | (addr >> 8));
+	buf[cnt++] = (uint8_t)addr;
+
+	/* Save the data phase start index */
+	dindx = cnt;
+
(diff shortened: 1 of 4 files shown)


drivers/adc/ad4858/ad4858.c
```diff
@@ -0,0 +1,1130 @@
...
+#include "ad4858.h"
+#include "ad4858_regs.h"
+#include "no_os_delay.h"
+#include "no_os_print_log.h"
+
+/**
+ * @brief Check big endianess of the host processor.
+ * @return Big endianess status (true/false)
+ */
+static bool _is_big_endian(void)
```
amiclaus, line 50: check please if you can't use already some functions available in https://github.com/analogdevicesinc/no-OS/blob/master/util/no_os_util.c or https://github.com/analogdevicesinc/no-OS/blob/master/util/

Otherwise add these generic functions (_is_big_endian and _memswap64) there in a separate commit prior to the driver addition.

amiclaus (commented):
- one inline comment
- some build checks are failing
- i'd suggest dropping extra `_regs`  header/source files and integrate the registers in the main driver header file as macro definitions.

mphalke:
> * one inline comment
> * some build checks are failing
> * i'd suggest dropping extra `_regs`  header/source files and integrate the registers in the main driver header file as macro definitions.

-Will fix the build issues
-On 3rd point, this was just added to print the register contents in debug mode. If you see, already these register macros are part of the drivers main header file. What I have done is added them into an array and I am using that array to print the register addresses along with data (done in the init function). This is mainly needed for our application, so if you okay, I can remove these two files completely and remove the debug code added to print  the register contents.

drivers/adc/ad4858/ad4858.c
mphalke: I didn't find the function in util.c that does what I intended- basically wanted to swap the memory content at byte level and also depending upon the step level (e.g. how many bytes I want to swap). Also no function to check the endianness as well. Just to clarify, I have taken these functions from existing ad4170 drivers (developed by Darius) which is present there on our internal bitbucket repo.
Just to be clear, do you suggest to add two separate commits for this? first commit with only these two functions and 2nd with other driver functions?

mphalke:
V1.0: Review comments fixes

1. Created two commits for the changes. First commit contains additional memswap and endianess check function. 2nd commit contains entire drivers implementation
2. Removed the ad4858_regs.c/.h file as they were added for register debug porpose.
3. Fixed doxygen build errors

drivers/adc/ad4858/ad4858.c
amiclaus: sorry i wasn't explicit enough. the functions for memswap and is_big_endian should be placed in the no_os_util.c file prior to the driver addition. these are generic functions that should be implemented only once and should not be dependent on a specific part driver.

amiclaus (commented):
please add memswap and is_big_endian in no_os_util source files.

drivers/adc/ad4858/ad4858.h
```diff
@@ -0,0 +1,407 @@
...
+#define AD4858_R3B				(3ul << 16)
+#define AD4858_R4B				(4ul << 16)
+#define AD4858_LEN(x)			((x) >> 16)
+#define AD4858_ADDR(x)			((x) & 0xFFFF)
+
+/** Register definitions */
+#define AD4858_REG_INTERFACE_CONFIG_A		(AD4858_R1B | 0x00)
+#define AD4858_REG_INTERFACE_CONFIG_B		(AD4858_R1B | 0x01)
+#define AD4858_REG_DEVICE_CONFIG			(AD4858_R1B | 0x02)
+#define AD4858_REG_CHIP_TYPE				(AD4858_R1B | 0x03)
```
amiclaus, line 58: the indentation is messed up. check the raw file: https://raw.githubusercontent.com/analogdevicesinc/no-OS/82587278b4871ab99907fcb3f088bfa8c1552b6d/drivers/adc/ad4858/ad4858.h

drivers/adc/ad4858/ad4858.h
```diff
@@ -0,0 +1,407 @@
...
+#define AD4858_DEF_CHN_OR			0x7ffff0
+#define AD4858_DEF_CHN_UR			0x800000
+
+/**
+ * @enum ad4858_operating_mode
+ * @brief Operating modes
+ */
+enum ad4858_operating_mode {
+	AD4858_NORMAL_OP_MODE,
+	AD4858_LOW_POWER_OP_MODE=0x3,
```
amiclaus, line 149: if one member of the enum is explicitly initialized, then all of them should be.

drivers/adc/ad4858/ad4858.c
```diff
@@ -0,0 +1,1069 @@
...
+ * @note Multibyte write with streaming mode is supported by default.
+ */
+int ad4858_reg_write(struct ad4858_dev *dev, uint32_t reg_addr,
+		     uint32_t reg_val)
+{
+	int32_t ret;
+	uint32_t addr;
+	uint8_t cnt = 0;
+	uint8_t reg_size = AD4858_LEN(reg_addr);
+	uint8_t buf[6] = { 0 };
```
amiclaus, line 60: nit: `{0}`

drivers/adc/ad4858/ad4858.c
```diff
@@ -0,0 +1,1069 @@
...
+
+	if (!dev)
+		return -EINVAL;
+
+	if (!dev->addr_ascension_enable) {
```
amiclaus, line 65: drop the brackets.

drivers/adc/ad4858/ad4858.c
```diff
@@ -0,0 +1,1069 @@
...
+		addr = AD4858_ADDR(reg_addr) + reg_size - 1;
+	} else {
+		addr = AD4858_ADDR(reg_addr);
+	}
+
+	buf[cnt++] = (uint8_t)(addr >> 8);
```
amiclaus, line 71: use no_os_field_get

drivers/adc/ad4858/ad4858.c
```diff
@@ -0,0 +1,1069 @@
...
+	buf[cnt++] = (uint8_t)addr;
```
amiclaus, line 72: ditto.

drivers/adc/ad4858/ad4858.c
```diff
@@ -0,0 +1,1069 @@
...
+	uint32_t addr;
+	uint8_t cnt = 0;
+	uint8_t dindx = 0;
+	uint8_t reg_size = AD4858_LEN(reg_addr);
+	uint8_t buf[6] = { 0 };
+
+	if (!dev || !reg_val)
+		return -EINVAL;
+
+	if (!dev->addr_ascension_enable) {
```
amiclaus, line 115: drop brackets.

drivers/adc/ad4858/ad4858.c
```diff
@@ -0,0 +1,1069 @@
...
+		addr = AD4858_ADDR(reg_addr) + reg_size - 1;
+	} else {
+		addr = AD4858_ADDR(reg_addr);
+	}
+
+	/* Store the address (16-bit instruction phase with R/W=1) */
+	buf[cnt++] = (uint8_t)(AD4858_REG_RD_BIT_MSK | (addr >> 8));
```
amiclaus, line 122: no_os_field_get

drivers/adc/ad4858/ad4858.c
```diff
@@ -0,0 +1,1069 @@
...
+	dindx = cnt;
+
+	/* Add the register data size to calculate an actual length of SPI frame */
+	cnt += reg_size;
+
+	ret = no_os_spi_write_and_read(dev->spi_desc, buf, cnt);
+	if (ret)
+		return ret;
+
+	*reg_val = 0;
```
amiclaus, line 135: is this explicit initialization required?

drivers/adc/ad4858/ad4858.c
```diff
@@ -0,0 +1,1069 @@
...
+		if (gpio_val == NO_OS_GPIO_LOW)
+			break;
+		timeout--;
+	} while (timeout);
+
+	if (!timeout)
+		return -ETIMEDOUT;
+
+	/* Read SPI data */
+	ret = no_os_spi_write_and_read(dev->spi_desc, buff, nb_bytes);
```
amiclaus, line 611: return directly.

drivers/adc/ad4858/ad4858.c
```diff
@@ -0,0 +1,1069 @@
...
+
+	ret = no_os_gpio_remove(dev->gpio_cmos_lvds);
+	if (ret)
+		return ret;
+
+	ret = no_os_gpio_remove(dev->gpio_cnv);
+	if (ret)
+		return ret;
+
+	ret = no_os_gpio_remove(dev->gpio_busy);
```
amiclaus, line 937: return directly.

drivers/adc/ad4858/ad4858.c
mphalke: do you mean no_os_put_unaligned_be16? This is the only relevant function I see.
However, the problem with this function is, it's name indicates "be- big endian" but my processor is little endian (as most ARM processors are LE). So this might create some confusion.

drivers/adc/ad4858/ad4858.c
mphalke: do you mean no_os_put_unaligned_be16?

drivers/adc/ad4858/ad4858.c
mphalke: No I guess, will remove this

drivers/adc/ad4858/ad4858.c
```diff
@@ -0,0 +1,1069 @@
...
+ * @param dev - Pointer to the device structure.
+ * @param chn - ADC input channel.
+ * @param raw - Pointer to uint32 variable where to store the data.
+ * @return 0 in case of success, negative error code otherwise.
+ */
+int ad4858_read_chn_data(struct ad4858_dev *dev, uint8_t chn, uint32_t *raw)
+{
+	int ret;
+	uint8_t indx;
+	uint8_t buff[32] = {};
```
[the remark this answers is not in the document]
mphalke: This is just done to send 0 over MOSI line in the SPI data read command as few devices expects 0 over line for data read. However, verified with ad4858 and it doesn't need 0 over MOSI line, so will remove this explicit initialization

drivers/adc/ad4858/ad4858.c
```diff
@@ -0,0 +1,1069 @@
...
+			return ret;
+	}
+
+	/* BUSY GPIO */
+	ret = no_os_gpio_get_optional(&dev->gpio_busy, init_param->gpio_busy);
+	if (ret)
+		return ret;
+
+	if (dev->gpio_busy) {
+		ret = no_os_gpio_direction_input(dev->gpio_busy);
```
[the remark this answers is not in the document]
mphalke: I think we should not return as this is optional gpio and if not initialized in init structure, then this if condition won't execute. In that case, we must come to else part and return 0. Correct me if I am wrong.

drivers/adc/ad4858/ad4858.h
mphalke: I am using VS and I see alignment properly. Is there a way to check the alignments in raw format on github and adjust accordingly?

image

drivers/adc/ad4858/ad4858.c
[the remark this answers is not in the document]
amiclaus: then `uint8_t buff[32] = {0};` would be more explicit.

drivers/adc/ad4858/ad4858.c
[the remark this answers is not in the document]
amiclaus: by return directly I did not mean removing the `return 0;` at the end of the function. Just remove
```
if (ret)
	return ret;
```
and do here:
`return no_os_gpio_direction_input(dev->gpio_busy)`

drivers/adc/ad4858/ad4858.h
amiclaus: check all the defines. for some of them I can see that there is a mix of spaces and tabs. for the example you provided here the indentation is correct.

mphalke:
V2.0: Review comments fixes and updated to data capture interface functions

1. Fixed the review comments
2. Fixed issues with data interface logic

drivers/adc/ad4858/ad4858.h
```diff
@@ -0,0 +1,419 @@
...
+	AD4858_RANGE_NEG_20_0V_TO_POS_20_0V,
+	AD4858_RANGE_0V_TO_25_0V,
+	AD4858_RANGE_NEG_25_0V_TO_POS_25_0V,
+	AD4858_RANGE_0V_TO_40_0V,
+	AD4858_RANGE_NEG_40_0V_TO_POS_40_0V,
+	AD4858_NUM_OF_SOFTSPAN
+};
+
+/**
+ * @enum ad4858_conv_data
```
amiclaus, line 233: struct

mphalke:
V3.0; Fixed typo
-Fixed typo in doxygen header for ad4858_conv_data structure

amiclaus (approved):

PR #1761, opened 2023-05-10
Xilinx/timer init bug fix

gatsby0203:
Since dev->platform_ops is not assigned no_os_timer_count_clk_set and no_os_timer_counter_set will always return -EINVAL. The timer will never be initialized.

### DIFF
--- a/drivers/platform/xilinx/xilinx_timer.c
+++ b/drivers/platform/xilinx/xilinx_timer.c
@@ -90,6 +90,7 @@ int32_t xilinx_timer_init(struct no_os_timer_desc **desc,
 	dev->id = param->id;
 	dev->freq_hz = param->freq_hz;
 	dev->ticks_count = param->ticks_count;
+	dev->platform_ops = param->platform_ops;
 	dev->extra = xdesc;
 	xdesc->active_tmr = xinit->active_tmr;
 	xdesc->type = xinit->type;


amiclaus (commented):
Please sign the CLA

amiclaus (changes requested):
The assignment of the `platform_ops` is done in the generic timer api.
https://github.com/analogdevicesinc/no-OS/blob/f018f149e118d537c1dbf475f68446c33b94b7d9/drivers/api/no_os_timer.c#L71

Therefore you should pass the platform specific ops (in this case `xil_timer_ops`) to the `no_os_timer_init_param` structure inside your specific application.
See an example here:
https://github.com/analogdevicesinc/no-OS/blob/f018f149e118d537c1dbf475f68446c33b94b7d9/projects/adv7511/src/main.c#L371

danmois (approved):

gatsby0203:
> The assignment of the `platform_ops` is done in the generic timer api.
> 
> https://github.com/analogdevicesinc/no-OS/blob/f018f149e118d537c1dbf475f68446c33b94b7d9/drivers/api/no_os_timer.c#L71
> 
> Therefore you should pass the platform specific ops (in this case `xil_timer_ops`) to the `no_os_timer_init_param` structure inside your specific application. See an example here:
> 
> https://github.com/analogdevicesinc/no-OS/blob/f018f149e118d537c1dbf475f68446c33b94b7d9/projects/adv7511/src/main.c#L371

I know `platform_ops` will be assigned in the generic timer api. But in initialize function `xilinx_timer_init` line 110 and 114.

https://github.com/analogdevicesinc/no-OS/blob/9264d8983c4e434affd171c7e6b394826089fdb0/drivers/platform/xilinx/xilinx_timer.c#L110-L116

it uses `dev` to initialize the timer clock and timer value, which always returns `-EINVAL`, since dev->platform_ops is not assigned.

And I know it kind of breaks the coherence of code, but I think it's ok to take it as a special case.

gatsby0203:
> > > The assignment of the `platform_ops` is done in the generic timer api.
> > > https://github.com/analogdevicesinc/no-OS/blob/f018f149e118d537c1dbf475f68446c33b94b7d9/drivers/api/no_os_timer.c#L71
> > > 
> > > Therefore you should pass the platform specific ops (in this case `xil_timer_ops`) to the `no_os_timer_init_param` structure inside your specific application. See an example here:
> > > https://github.com/analogdevicesinc/no-OS/blob/f018f149e118d537c1dbf475f68446c33b94b7d9/projects/adv7511/src/main.c#L371
> > 
> > 
> > I know `platform_ops` will be assigned in the generic timer api. But in initialize function `xilinx_timer_init` line 110 and 114.
> > https://github.com/analogdevicesinc/no-OS/blob/9264d8983c4e434affd171c7e6b394826089fdb0/drivers/platform/xilinx/xilinx_timer.c#L110-L116
> > 
> > it uses `dev` to initialize the timer clock and timer value, which always returns `-EINVAL`, since dev->platform_ops is not assigned.
> > And I know it kind of breaks the coherence of code, but I think it's ok to take it as a special case.
> 
> This was a bug on our side, thanks for pointing that out! I created a patch to fix the issue #1764 . Using generic api functions inside platform specific init functions is not the correct approach taking into consideration the way we developed the apis for all our peripherals.

I thought so too at first.

but `xilinx_timer_count_clk_set` and  `xilinx_timer_counter_set`  are called before definitions, this is not valid in C99 and later.

amiclaus (commented):
> > > > The assignment of the `platform_ops` is done in the generic timer api.
> > > > https://github.com/analogdevicesinc/no-OS/blob/f018f149e118d537c1dbf475f68446c33b94b7d9/drivers/api/no_os_timer.c#L71
> > > > 
> > > > Therefore you should pass the platform specific ops (in this case `xil_timer_ops`) to the `no_os_timer_init_param` structure inside your specific application. See an example here:
> > > > https://github.com/analogdevicesinc/no-OS/blob/f018f149e118d537c1dbf475f68446c33b94b7d9/projects/adv7511/src/main.c#L371
> > > 
> > > 
> > > I know `platform_ops` will be assigned in the generic timer api. But in initialize function `xilinx_timer_init` line 110 and 114.
> > > https://github.com/analogdevicesinc/no-OS/blob/9264d8983c4e434affd171c7e6b394826089fdb0/drivers/platform/xilinx/xilinx_timer.c#L110-L116
> > > 
> > > it uses `dev` to initialize the timer clock and timer value, which always returns `-EINVAL`, since dev->platform_ops is not assigned.
> > > And I know it kind of breaks the coherence of code, but I think it's ok to take it as a special case.
> > 
> > 
> > This was a bug on our side, thanks for pointing that out! I created a patch to fix the issue #1764 . Using generic api functions inside platform specific init functions is not the correct approach taking into consideration the way we developed the apis for all our peripherals.
> 
> I thought so too at first.
> 
> but `xilinx_timer_count_clk_set` and `xilinx_timer_counter_set` are called before definitions, this is not valid in C99 and later.

You are right, thank you. I added an extra patch in the same PR that should fix this.

PR #1764, opened 2023-05-10
platform:xilinx:timer: fix init function

amiclaus:
Since the platform ops are populated in the generic timer api after the platform specific timer initialization, the `no_os_timer_count_clk_set` won't work properly when called inside the platform specific init function. Instead use directly the `xilinx_timer_count_clk_set` function.


rbolboac (approved):
lgtm

buha (approved):

PR #1769, opened 2023-05-15
drivers: io-link: max22516 support

amiclaus:
Initial commit for max22516 driver.

### DIFF
--- a/drivers/io-link/max22516/max22516.c
+++ b/drivers/io-link/max22516/max22516.c
@@ -1,0 +1,885 @@
+/***************************************************************************//**
+ *   @file   max22516.c
+ *   @brief  Implementation of MAX22516 Driver.
+ *   @author Antoniu Miclaus (antoniu.miclaus@analog.com)
+********************************************************************************
+ * Copyright 2023(c) Analog Devices, Inc.
+ *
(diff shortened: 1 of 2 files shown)


amiclaus (commented):
Subject for initial review.

amiclaus (commented):
Driver not tested yet with hardware.

drivers/io-link/max22516/max22516.c
```diff
@@ -0,0 +1,892 @@
...
+/**
+ * @brief Writes data to max22516 over SPI.
+ * @param dev - The device structure.
+ * @param reg_addr - The register address.
+ * @param data - Data value to write.
+ * @return Returns 0 in case of success or negative error code otherwise.
+ */
+int max22516_write(struct max22516_dev *dev, uint8_t reg_addr, uint8_t data)
+{
+	uint8_t buff[MAX22516_BUFF_SIZE_BYTES];
```
buha, line 62: We typically place such buffers in device descriptor rather than on the stack.
In your case one can get away with 2 bytes, but still it's good practice to avoid putting buffers on the stack.

buha (approved):

PR #1770, opened 2023-05-15
drivers:adc-dac:ad74416h: initial commit

amiclaus:
Add support for ad74416h.

### DIFF
--- a/drivers/adc-dac/ad74416h/ad74416h.c
+++ b/drivers/adc-dac/ad74416h/ad74416h.c
@@ -1,0 +1,798 @@
+/***************************************************************************//**
+ *   @file   ad74416h.c
+ *   @brief  Source file of AD74416H Driver.
+ *   @author Antoniu Miclaus (antoniu.miclaus@analog.com)
+********************************************************************************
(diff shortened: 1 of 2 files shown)


drivers/adc-dac/ad74416h/ad74416h.c
```diff
@@ -0,0 +1,802 @@
...
+/************************ Functions Definitions *******************************/
+/******************************************************************************/
+
+/**
+ * @brief Converts a millivolt value in the corresponding DAC 16 bit code.
+ * @param mvolts - The millivolts value.
+ * @param code - The resulting DAC code.
+ * @return 0 in case of success, -EINVAL otherwise
+ */
+int ad74416h_dac_voltage_to_code(uint32_t mvolts, uint32_t *code)
```
buha, line 82: I'd return the code and get rid of `uint32_t *code` argument.

It could be done by clipping the output to the max/min possible code instead of returning an error.

In my opinion it's best for the driver API user to not handle so many error cases, but you can leave it as-is if you prefer.

buha (approved):

PR #1771, opened 2023-05-16
drivers/platform/mbed: typo fix.

pmallick-adi:


amiclaus (commented):
as usually, please provide a commit body expressing what you did in your patch. Also, if there is a fix patch then `Fixes:` tag should be attached to the commit.

pmallick-adi:
> Fixes:

Have update the comment.

amiclaus (approved):

PR #1772, opened 2023-05-16
projects: add eval-ltc4306-pmdz example project

jsanbuen:
add example project for eval-ltc4306-pmdz eval board.


projects/eval-ltc4306/src/examples/dummy/dummy_example.c
```diff
@@ -0,0 +1,173 @@
...
+		if (ret)
+			goto error_ltc4306;
+
+		no_os_udelay(50000);
+
+		ret = ltc4306_set_gpio_output_state(ltc4306, 1, false);
+		if (ret)
+			goto error_ltc4306;
+
+		no_os_udelay(50000);
```
amiclaus, line 109: why not `no_os_mdelay`? same for the other delay function calls.

amiclaus (commented):
lgtm, one comment.

buha:
looks good, one remark though which is not optional, please rename `dummy` to `basic` or something like that, dummy is not really a fit word for these examples, we should change all occurrences

buha (approved):

amiclaus (approved):

PR #1773, opened 2023-05-16
drivers: adc: ad4080: add support

amiclaus:
AD4080 is a high speed, low noise, low distorsion, 20-bit Easy Driver, succesive approxmiation register (SAR) analog-to-digital converter ADC.

### DIFF
--- a/drivers/adc/ad4080/ad4080.c
+++ b/drivers/adc/ad4080/ad4080.c
@@ -1,0 +1,1046 @@
+/***************************************************************************//**
+ *   @file   ad4080.c
+ *   @brief  Implementation of AD4080 Driver.
+ *   @author Antoniu Miclaus (antoniu.miclaus@analog.com)
+********************************************************************************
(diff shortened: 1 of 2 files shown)


buha (approved):

drivers/adc/ad4080/ad4080.c
```diff
@@ -0,0 +1,882 @@
...
+}
+
+/**
+ * @brief Get Address Ascension
+ * @param dev - The device structure.
+ * @return Address ascension value
+ */
+enum ad4080_addr_asc ad4080_get_addr_asc(struct ad4080_dev *dev)
+{
+	return dev->addr_asc;
```
CiprianRegus, line 181: Wouldn't it be safer to read the actual register value, instead of the cached one? This approach seems wrong if any of these fields are written using `ad4080_write`.

amiclaus (commented):
v2:
- update getter functions to read from hardware instead of cached values.

PR #1777, opened 2023-05-21
drivers: adc-dac: ad74416h: update regmap

amiclaus:
Update regmap for ad74416h and add all bitfields for all registers.


rbolboac (approved):
lgtm

PR #1619, opened 2023-01-30
ADIN1110 driver

CiprianRegus:
Supports both the ADIN1110 and ADIN2111.
    
The ADIN1110 is an ultra low power, single port, 10BASE-T1L
transceiver design for industrial Ethernet applications and is com-
pliant with the IEEE 802.3cg-2019 Ethernet standard for long
reach, 10 Mbps single pair Ethernet (SPE). Featuring an integrated
media access control (MAC) interface, the ADIN1110 enables direct
connectivity with a variety of host controllers via a serial peripheral
interface (SPI).

PR #1778, opened 2023-05-22
IIOD networking using lwip

CiprianRegus:
https://github.com/analogdevicesinc/no-OS/pull/1619 has to be merged first.


dbogdan (approved):

CiprianRegus:
- Added CI build support for the new lwip example.
- The lwip example runs on the ADIN1110EBZ, and some changes regarding the SPI and UART interface numbers had to be made. There changes should not affect the already existing examples (indented for SDP-K1).

libraries/lwip/configs/lwipopts.h
```diff
@@ -0,0 +1,179 @@
+/*
```
buha, line 1: For now lwipcfg and lwipopts headers are probably ok here, but i think a project will pop-up which requires different configuration and will need these 2 files in the project folder to override this default configuration.

I guess we can override this default configuration with -I directives order but we can think about the details later on.

tools/scripts/libraries.mk
```diff
@@ -108,6 +108,13 @@ CFLAGS += -I$(AZURE_DIR)/sdk/inc
 
 endif
 
+ifneq ($(if $(findstring lwip, $(LIBRARIES)), 1),)
+include $(NO-OS)/tools/scripts/lwip.mk
+INCS += $(NO-OS)/libraries/lwip/arch
+INCS += $(NO-OS)/libraries/lwip/configs/lwipcfg.h
```
buha, line 114: as i said, this will need to be configurable in the future, but leave it as-is for now

buha (approved):

PR #1785, opened 2023-05-24
drivers:adc:ad400x: Add support for ADAQ4003

machschmitt:
Include ADAQ4003 in the list of devices supported by the ad400x driver. ADAQ4003 and AD400x devices share the same register read/write protocol, configuration register layout, and many other characteristics such as timing specifications, number of channels, resolution bits, and sampling rate.


amiclaus (commented):
is this ready for review?
Please sign the CLA :)

machschmitt:
The AD400x and ADAQ400x devices are very similar, so I think updating de list of supported devices may be enough to support the newer chip. If supporting another device requires more than updating the supported device list, then some guidance on what additional changes to do would be of great help.

These changes should not break anything so pulling them should cause no regressions. Despite that, this has yet to be tested with an actual ADAQ4003 device and so we may find out additional changes to be required for supporting that design. Thus, set this as draft to distinguish from more mature set of updates.

buha (approved):

PR #1795, opened 2023-05-31
tools:scripts: cflags issue fix

CMinajigi:
Adding condition on adding more backslashs for flags. 

We need add extra backslashes when the macro contains the quotes. like if macro is CFLAGS=\"flag\" then we need to add extra backslashes in linux machine so that it become CFLAGS=\\\"flag\\\". when we put this macro in text file it finally becomes CFLAGS=\"flag\" in text file. This only occur in linux machines, in windows we don't need to add extra backslashes.

### DIFF
--- a/tools/scripts/generic.mk
+++ b/tools/scripts/generic.mk
@@ -81,11 +81,14 @@ set_one_time_rule = echo Target file. Do not delete > $1
 # Append text to file
 APPEND_TEXT_TO_FILE = echo $1 >> $2
 
-# Add blank line to a file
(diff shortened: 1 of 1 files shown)


amiclaus (commented):
what does this commit fix? please be more explicit in the commit body. This will help the review process.

amiclaus (commented):
you did not change the commit as requested, just the PR description which does not remain in the commit history for better tracking of the changes.

buha (approved):

PR #1797, opened 2023-06-07
drivers:adc:ad713x:Added a delay after GPIO operations

jansunil:
Added a delay after the GPIO operations as the device needs a 10 ms delay to come out of power down state. The power down operation has been added as a part of the ad713x_init_gpio().

### DIFF
--- a/drivers/adc/ad713x/ad713x.c
+++ b/drivers/adc/ad713x/ad713x.c
@@ -453,6 +453,8 @@ static int32_t ad713x_init_gpio(struct ad713x_dev *dev,
 		no_os_mdelay(100);
 	}
 
+	no_os_mdelay(10);
+
 	return 0;
 }
 


drivers/adc/ad713x/ad713x.c
```diff
@@ -549,6 +549,8 @@ int32_t ad713x_init(struct ad713x_dev **device,
 	if(NO_OS_IS_ERR_VALUE(ret))
 		goto error_gpio;
 
+	no_os_mdelay(10);
```
amiclaus, line 552: just curious, is this delay related to the datasheet specifications? 

Also i'd move it in the gpio_init function.

jansunil: Yes, this delay has been mentioned in the datashseet for the PDN functionality.

jansunil:
V1: Addressed review comments

Moved the delay to the ad713x_init_gpio() function

amiclaus (approved):

PR #1801, opened 2023-06-13
projects: demo_esp: rework project

amiclaus:
Restructure demo_esp8266 project and rename it to demo_esp with the scope of adding support for multiple platforms / esp versions.


amiclaus (commented):
v2:
- update builds.json and use BASIC_EXAMPLE flag

danmois (approved):

PR #1805, opened 2023-06-20
test: build_projects: check stm32 plaform build

amiclaus:
Test the time required by the azure agent to perform builds only for stm32 platform.


amiclaus:
closing this since it was only for testing purposes.

PR #1811, opened 2023-06-22
drivers:adc:ad4858 Data interface mode switch API

mphalke:
Added API to change to ADC data interface mode

### DIFF
--- a/drivers/adc/ad4858/ad4858.c
+++ b/drivers/adc/ad4858/ad4858.c
@@ -261,6 +261,22 @@ int ad4858_set_config_interface_mode(struct ad4858_dev *dev)
 }
 
 /**
+ * @brief Set device data interface mode.
+ * @param dev - Pointer to the device structure.
+ * @return 0 in case of success, negative error code otherwise.
+ */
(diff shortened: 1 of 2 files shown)


amiclaus (commented):
is this the standard approach to enter in conversion mode recommended in the datasheet?

mphalke:
> is this the standard approach to enter in conversion mode recommended in the datasheet?

Nop. The problem with this part is, it has 2 interfaces- CMOS Config and CMOS Data and ideally we would need two 4/3-wire SPI ports to interface them separately. However, there is a feasibility of using a single SPI port for both the interfaces but they won't work simultaneously (after discussing with design engineer). With single SPI port, few SPI pins are shared between data and config interface (e.g. SCK and SDO0).
This function has been added based on my observations while working with the part using STM32 MCU platform. In order to enter into data interface mode, first you need to perform dummy SPI transaction to read previous conversion results. This way, I get correct data and properly able to switch between data and config interface modes.
I might want to have another word with design engineers and see if there is any better way to do that, but for now this solution works for me. If I see any issues in the future, will rectify them. This part is bit tricky to use with single SPI port instance.

amiclaus (approved):

PR #1813, opened 2023-06-26
API: Add i2c and spi bus instance

rbudai98:
Adding SPI and I2C bus descriptors to the api:
* changed STM32 initialization
* added extra bus_desc struct to header files

### DIFF
--- a/drivers/api/no_os_i2c.c
+++ b/drivers/api/no_os_i2c.c
@@ -41,6 +41,13 @@
 #include "no_os_i2c.h"
 #include <stdlib.h>
 #include "no_os_error.h"
+#include "no_os_mutex.h"
+#include "no_os_alloc.h"
+
+/**
+ * @brief i2c_table contains the pointers towards the i2c buses
+*/
+static void *i2c_table[I2C_MAX_BUS_NUMBER + 1];
 
 /**
  * @brief Initialize the I2C communication peripheral.
@@ -58,15 +65,54 @@ int32_t no_os_i2c_init(struct no_os_i2c_desc **desc,
 
 	if (!param->platform_ops->i2c_ops_init)
 		return -ENOSYS;
-
+	if ( param->device_id > I2C_MAX_BUS_NUMBER)
+		return -EINVAL;
+	// Initializing BUS descriptor
+	if (i2c_table[param->device_id] == NULL) {
+		ret = no_os_i2cbus_init(param);
+		if (ret)
+			return ret;
+	}
+	// Initilize I2C descriptor
 	ret = param->platform_ops->i2c_ops_init(desc, param);
 	if (ret)
 		return ret;
-
+	(*desc)->bus = i2c_table[param->device_id];
 	(*desc)->platform_ops = param->platform_ops;
 
 	return 0;
 }
+
+/**
+ * @brief Initialize the i2c bus communication peripheral.
+ * @param param - The structure that containes the i2c bus parameters
+ * @return 0 in case of success, error code otherwise
+*/
+int32_t no_os_i2cbus_init(const struct no_os_i2c_init_param *param)
+{
+	struct no_os_i2cbus_desc *bus = (struct no_os_i2cbus_desc *)no_os_calloc(1,
+					sizeof(struct no_os_i2cbus_desc));
+
+	if (!bus)
+		return -ENOMEM;
+
+	no_os_mutex_init(bus->mutex);
+
+	if (bus->mutex == NULL) {
+		no_os_free(bus);
+		return -ENOMEM;
+	}
+
+	bus->device_id = param->device_id;
+	bus->max_speed_hz = param->max_speed_hz;
+	bus->platform_ops = param->platform_ops;
+	bus->extra = param->extra;
+
+	i2c_table[param->device_id] = bus;
+
+	return 0;
+}
+
 
 /**
  * @brief Free the resources allocated by no_os_i2c_init().
@@ -75,6 +121,9 @@ int32_t no_os_i2c_init(struct no_os_i2c_desc **desc,
  */
 int32_t no_os_i2c_remove(struct no_os_i2c_desc *desc)
 {
+	// Remove I2C bus
+	no_os_i2cbus_remove(desc->device_id);
+
 	if (!desc || !desc->platform_ops)
 		return -EINVAL;
 
@@ -82,6 +131,25 @@ int32_t no_os_i2c_remove(struct no_os_i2c_desc *desc)
 		return -ENOSYS;
 
 	return desc->platform_ops->i2c_ops_remove(desc);
+}
+
+/**
+ * @brief Removes i2c bus instance
+ * @param bus_number - i2c bus number
+*/
+void no_os_i2cbus_remove(uint32_t bus_number)
+{
+	struct no_os_i2cbus_desc *i2c_bus_desriptor = (struct no_os_i2cbus_desc *)
+			i2c_table[bus_number];
+
+	if (i2c_bus_desriptor->mutex != NULL) {
+		no_os_mutex_remove(i2c_bus_desriptor->mutex);
+		i2c_bus_desriptor->mutex = NULL;
+	}
+	if (i2c_bus_desriptor != NULL) {
+		no_os_free(i2c_bus_desriptor);
+		i2c_bus_desriptor = NULL;
+	}
 }
 
 /**
--- a/drivers/api/no_os_spi.c
+++ b/drivers/api/no_os_spi.c
@@ -41,6 +41,13 @@
 #include "no_os_spi.h"
 #include <stdlib.h>
 #include "no_os_error.h"
+#include "no_os_mutex.h"
+#include "no_os_alloc.h"
+
+/**
+ * @brief spi_table contains the pointers towards the SPI buses
+*/
+static void *spi_table[SPI_MAX_BUS_NUMBER + 1];
 
 /**
  * @brief Initialize the SPI communication peripheral.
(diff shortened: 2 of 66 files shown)


drivers/platform/stm32/stm32_i2c.c
```diff
@@ -43,6 +43,52 @@
...
+/**
+ * @brief i2c_table contains the pointers towards the I2C buses
+*/
+static void *i2c_table[4];
+
+/**
+ * @brief STM32 I2C bus initializer
+*/
+int32_t stm32_i2c_initialize_bus(uint32_t bus_number,
+				 struct no_os_i2c_init_param *param)
```
rbudai98, line 55: Here should I initialize a different spibus_init_param? For users this part is not exposed and allocating extra variable for this just makes the runtime allocation slower

include/no_os_spi.h
```diff
@@ -114,6 +114,25 @@ struct no_os_spi_msg {
  */
 struct no_os_spi_platform_ops ;
 
+/**
+ * @struct no_os_spibus_init_param
+ * @brief Structure for SPI bus initialization parameter
+*/
+struct no_os_spibus_init_param {
+	/** SPI bus chip ID */
```
buha, line 122: change the comment here to something like "SPI bus number (0 for SPI0, 1 for SPI1 etc.)"

include/no_os_spi.h
```diff
@@ -136,11 +155,33 @@ struct no_os_spi_init_param {
...
 };
 
+/**
+ * @struct no_os_spibus_desc
+ * @brief SPI bus descriptor
+*/
+struct no_os_spibus_desc {
+	/** SPI bus mutex (lock) */
+	void 		*mutex;				// created internally in no_os_spibus_init() at every call
+	/** SPI bus device id */
```
buha, line 165: same here

include/no_os_spi.h
```diff
@@ -136,11 +155,33 @@ struct no_os_spi_init_param {
 	struct no_os_spi_desc *parent;
 };
 
+/**
+ * @struct no_os_spibus_desc
+ * @brief SPI bus descriptor
+*/
+struct no_os_spibus_desc {
+	/** SPI bus mutex (lock) */
+	void 		*mutex;				// created internally in no_os_spibus_init() at every call
```
buha, line 164: i'd remove that "creted internally..." comment

include/no_os_i2c.h
```diff
@@ -57,6 +57,23 @@
...
+ * @struct no_os_i2cbus_init_param
+ * @brief Structure holding I2C bus descriptor
+*/
+struct no_os_i2cbus_init_param {
+	/** I2C bus Device ID */
+	uint32_t	device_id;
+	/** I2C bus maximum transfer speed supported */
+	uint32_t	max_speed_hz;
+	/** I2C bus slave address */
+	uint8_t		slave_address;
```
buha, line 70: remove slave_address, this is slave specific info (like Chip select for SPI)

include/no_os_i2c.h
```diff
@@ -74,11 +91,33 @@ struct no_os_i2c_init_param {
...
+ */
+struct no_os_i2cbus_desc {
+	/** I2C bus mutex(lock)*/
+	void* mutex;
+	/** I2C bus Device ID */
+	uint32_t	device_id;
+	/** I2C bus maximum transfer speed supported */
+	uint32_t	max_speed_hz;
+	/** I2C bus Slave address */
+	uint8_t		slave_address;
```
buha, line 106: same here

drivers/platform/stm32/stm32_i2c.c
buha: this is hidden from the user so you can simply use no_os_i2c_init_param, you're right

drivers/platform/stm32/stm32_spi.c
```diff
@@ -46,6 +46,52 @@
...
 
+/**
+ * @brief spi_table contains the pointers towards the SPI buses
+*/
+static void *spi_table[8];
+
+/**
+ * @brief STM32 SPI bus memory initializer
+*/
+int stm32_spibus_init(uint32_t bus_number, struct no_os_spi_init_param *param)
```
buha, line 57: One important remark here.
It seems your implementation doesn't have anything to do with stm32.
I think this could somehow be moved to no_os_spi.c, could you please investigate that ?
That way there wouldn't be a need to implement it in each [platform]_spi.c

include/no_os_spi.h
```diff
@@ -136,11 +155,33 @@ struct no_os_spi_init_param {
 	struct no_os_spi_desc *parent;
 };
 
+/**
+ * @struct no_os_spibus_desc
+ * @brief SPI bus descriptor
+*/
+struct no_os_spibus_desc {
```
buha, line 162: And if you do the no_os_spi.c modification that I proposed in another commit, you could then proceed to this one:

To make sure this desc is not modified by the user, you could declare here in the header:

```
struct no_os_spibus_desc;
```

And in the no_os_spi.c you could fully define the structure:
```
struct no_os_spibus_desc {
    ...
}

This way, you can only access the structure members by name in the no_os_spi.c

File: drivers/platform/stm32/stm32_spi.c (hunk 7 above)
Comment (rbudai98) [reply drivers/platform/stm32/stm32_spi.c:57]: Yes, it would make sense.

File: drivers/platform/stm32/stm32_i2c.c [hunk 9]
```diff
@@ -43,6 +43,52 @@
...
+{
+	i2c_table[bus_number] = (struct no_os_i2cbus_desc *)no_os_calloc(1,
+				sizeof(*no_os_i2cbus_desc));
+
+	if (!i2c_table[bus_number])
+		return -ENOMEM;
+
+	no_os_mutex_init(i2c_table[bus_number].mutex);
+
+	if (i2c_table[bus_number].mutex == NULL)
```
amiclaus, line 65: i'd keep the code consistent: here you to `if (x == NULL)` and right above you do `if (!x)`

drivers/platform/stm32/stm32_i2c.c
```diff
@@ -114,9 +160,22 @@ int32_t stm32_i2c_init(struct no_os_i2c_desc **desc,
 	descriptor->device_id = param->device_id;
 	descriptor->max_speed_hz = param->max_speed_hz;
 	descriptor->slave_address = param->slave_address;
+
+	//Memory allocation for I2C bus descritor
+	if (i1c_table[i2c_desc->device_id] == NULL) {
+		ret = stm32_i2c_initialize_bus(param->device_id, param);
+		if (ret)
+			goto error;
```
amiclaus, line 168: i don't see any label named `error`  in this function.

include/no_os_i2c.h
```diff
@@ -74,11 +91,33 @@ struct no_os_i2c_init_param {
...
+	const struct no_os_i2c_platform_ops *platform_ops;
+	/** I2C bus extra parameters (device specific parameters) */
+	void		*extra;
+};
+
+
 /**
  * @struct no_os_i2c_desc
- * @brief Structure holding I2C descriptor
+ * @brief Structure holding I2C address descriptor
```
amiclaus, line 116: is this change required?

include/no_os_spi.h
```diff
@@ -136,11 +155,33 @@ struct no_os_spi_init_param {
...
+	/** SPI bus extra */
+	void		*extra;
+};
+
 /**
  * @struct no_os_spi_desc
  * @brief Structure holding SPI descriptor.
  */
 struct no_os_spi_desc {
+	struct no_os_spibus_desc	*bus;
```
amiclaus, line 184: if you added comment for the i2cbus_desc, then add here also.

amiclaus (commented):
please split into to commits: one for spi, one for i2c

include/no_os_i2c.h
rbudai98: Do you have a better naming version? It should differentiate the bus descriptor from the i2c descriptor that has address associated to it

drivers/api/no_os_i2c.c
```diff
@@ -58,23 +64,62 @@ int32_t no_os_i2c_init(struct no_os_i2c_desc **desc,
...
 }
 
+/**
+ * @brief Initialize the i2c bus communication peripheral.
+ * @param param - The structure that containes the i2c bus parameters
+ * @return 0 in case of success, error code otherwise
+*/
+int32_t no_os_i2cbus_init(const struct no_os_i2c_init_param *param)
+{
+	i2c_table[param->device_id] = (struct no_os_i2cbus_desc *)no_os_calloc(1,
```
CiprianRegus, line 92: `param->device_id` should be verified first. And maybe a macro for the `i2c_table`'s size instead of 8 would simplify thing a bit.

drivers/api/no_os_i2c.c
```diff
@@ -84,6 +129,25 @@ int32_t no_os_i2c_remove(struct no_os_i2c_desc *desc)
...
+/**
+ * @brief Removes i2c bus instance
+ * @param bus_number - i2c bus number
+*/
+void no_os_i2cbus_remove(uint32_t bus_number)
+{
+	struct no_os_i2cbus_desc i2c_bus_desriptor = (struct no_os_i2cbus_desc)
+			i2c_table[bus_number];
+
+	if (i2c_bus_desriptor.mutex != NULL) {
```
CiprianRegus, line 143: descriptor instead of desriptor

rbudai98:
v2:
* commits separated
* addressed PR comments
* moved BUS initialization from platform to api level
* memory leak solved in bus init

drivers/api/no_os_spi.c
```diff
@@ -58,33 +65,94 @@ int32_t no_os_spi_init(struct no_os_spi_desc **desc,
 
 	if (!param->platform_ops->init)
 		return -ENOSYS;
-
+	if (param->device_id > SPI_MAX_BUS_NUMBER)
+		return -EINVAL;
+	//Initializing BUS descriptor
```
buha, line 70: This PR looks good to me now.
1 comment though about aesthetics, you seem to have a lot of these `//Text` comments whereas we always leave a blank after `// Text`, please insert those whitespaces, otherwise no-OS code will look very heterogenous :D

rbudai98:
v3:
* spacing added
* added commit 34ebcab443e42c3fb0f8c3036fd480b96ba25942 to include mutex files in all projects

buha:
Looks good,
But please put the 3rd commit (the one with changes in projects) to be the 1st commit because the others depend on it

rbudai98:
v4:
* reorder commits

buha (approved):
i'm merging this, i see the commit order changed

PR #1816, opened 2023-06-28
Free the IIO xml desc

CiprianRegus:
Currently the xml decriptor is not freed when iio_remove() is called, which causes a memory leak. Free desc->xml_desc to fix the issue.


rbolboac (approved):
lgtm, can be merged after build is successful

amiclaus (approved):

PR #1818, opened 2023-07-03
Dev/eval adis memory leak fix

rbolboac:
- Remove memory leaks from eval-adis project in iio_trigger_example
- Add missing include of "no_os_print_log.h", needed for pr_info function call in iio_trigger_example.c for eval-adis project.
- Remove memory leaks from eval-adis project in dummy_example.


amiclaus (approved):
lgtm

PR #1820, opened 2023-07-04
tools:scripts: Fixed echo command printing into text file issue

mphalke:
pre_build rule used in generic.mk file copies the dependencies such as c/cpp flags, object/include files paths into the text files, which are used later during compilation and linking stage. On windows-os, during the copying of dependencies into text file using echo inbuilt command, the echo command string itself is getting printed into the text files. Compiler is therefore unable to interpret this echo string during compile and link stage. This seems to be the limitation of windows-os as this issue has not been detected on the linux-os yet.
This patch is created as workaround for this issue. A batch file has been added, which searches and replaces any echo string present in the text files during pre-build stage. This is done if make build is running only on the windows-os.

### DIFF
--- a/tools/scripts/echo_remove.bat
+++ b/tools/scripts/echo_remove.bat
@@ -1,0 +1,16 @@
+:: Search and replace 'echo' string in input text file with single space
+@echo off 
+setlocal enabledelayedexpansion
+setlocal enableextensions
+
+set "search=echo"
+set "replace="
+
+:: Loop through all input text files
+for %%x in (%*) do (
+    echo %%x
+    for /f "delims=" %%i in ('type "%%x" ^& break ^> "%%x" ') do (
+        set "line=%%i"
+        >>"%%x" echo(!line:%search%=%replace%!)
+    )
+)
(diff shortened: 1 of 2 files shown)


mphalke:
Hi,
Any review comments on this PR?

buha:
I think echo can be handled more elegantly with "@echo off" or something like that, no need to post process those "echo" occurences from the text files if you can make the system avoid generating them...

mphalke:
> I think echo can be handled more elegantly with "@echo off" or something like that, no need to post process those "echo" occurences from the text files if you can make the system avoid generating them...

This has already been tried out but not working well.

CiprianRegus:
I just tried a build on Maxim platform (on windows), and I wasn't able to reproduce this issue. There is no "echo" string in any of the generated files containing flags. This is true even without what @buha suggested. Can you provide a bit more details about what you mean by "not working well"?

mphalke:
> I just tried a build on Maxim platform (on windows), and I wasn't able to reproduce this issue. There is no "echo" string in any of the generated files containing flags. This is true even without what @buha suggested. Can you provide a bit more details about what you mean by "not working well"?

The issue is very random, I will share the build logs with you on separate thread. When I say not working well, I meant this is working sometime and sometimes not. We observed the issue on local machines as well as on our Jenkins CI builder machine when we try to build multiple projects/matrixes.

mphalke:
Hi, any decision made regards to this PR?

buha:
@mphalke , we were able to reproduce this on one of our machines, we will investigate the root cause

mphalke:
> @mphalke , we were able to reproduce this on one of our machines, we will investigate the root cause

Thanks for the info @buha

mphalke:
@buha Any updates on this issue? Just wanted to check if any work is being done on this

buha:
@mphalke I'm currently trying to convert the whole Windows build to use Git Bash and cut off pretty much anything Windows-native. See #1867

buha:
closing due to #1867

PR #1823, opened 2023-07-06
drivers:dac:ad3552r: Update ad3552r driver

CMinajigi:
Updated existing driver to latest changes in dac.
Added support to ad354xr and ad355xr devices.


CiprianRegus:
The commit should be split, since there are 2 different changes (update the existing driver and adding support for the new parts).

CMinajigi:
It's not two different changes, i have updated the driver to so that to support updates in hardware features.

drivers/dac/ad3552r/ad3552r.h
```diff
@@ -171,13 +170,20 @@
 #define AD3552R_STORAGE_BITS_FAST_MODE			16
 #define AD3552R_MAX_OFFSET				511
 #define AD3552R_LDAC_PULSE_US				1
+#define AD3552R_BOTH_CH_SELECT			(NO_OS_BIT(0) | NO_OS_BIT(1))
+#define AD3552R_BOTH_CH_DESELECT		0x0
+
+/* Maximum number of channels in this family of devices */
+#define AD3552R_MAX_NUM_CH		2
```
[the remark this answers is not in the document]
CMinajigi: yes. Earlier this driver support 2 channel devices now it support device with 1 channel also.

drivers/dac/ad3552r/ad3552r.h
```diff
@@ -472,19 +494,19 @@ int32_t ad3552r_set_ch_value(struct ad3552r_desc *desc,
 			     enum ad3552r_ch_attributes attr,
 			     uint8_t ch,
 			     uint16_t val);
-
```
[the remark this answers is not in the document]
CMinajigi: i will revert.

drivers/dac/ad3552r/ad3552r.c
```diff
@@ -1209,16 +1253,23 @@ int32_t ad3552r_reset(struct ad3552r_desc *desc)
...
+		if (is_fast)
+			return ad3552r_write_reg(desc,
+						 AD3552R_REG_ADDR_SW_LDAC_16B,
+						 mask);
+		else
+			return ad3552r_write_reg(desc,
+						 AD3552R_REG_ADDR_SW_LDAC_24B,
+						 mask);
+	} else {
+		no_os_gpio_set_value(desc->ldac, NO_OS_GPIO_LOW);
```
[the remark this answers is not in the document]
CMinajigi: i will add.

CMinajigi:
Any other review comments?

amiclaus (approved):

drivers/dac/ad3552r/ad3552r.c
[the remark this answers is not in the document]
amiclaus: this comment is still valid from my side

drivers/dac/ad3552r/ad3552r.c
```diff
@@ -1209,16 +1253,30 @@ int32_t ad3552r_reset(struct ad3552r_desc *desc)
...
+		if (NO_OS_IS_ERR_VALUE(err))
+			return err;
 
+		no_os_udelay(AD3552R_LDAC_PULSE_US);
+		no_os_gpio_set_value(desc->ldac, NO_OS_GPIO_HIGH);
```
amiclaus, line 1276: ditto: error check and handling.

amiclaus (changes requested):

drivers/dac/ad3552r/ad3552r.c
amiclaus: Where do you assign the return value of `no_os_gpio_set_value(desc->ldac, NO_OS_GPIO_HIGH);` to `err` variable?

drivers/dac/ad3552r/ad3552r.h
```diff
@@ -457,6 +457,8 @@ struct ad3552r_init_param {
 
 uint8_t ad3552r_reg_len(uint8_t addr);
 
+uint8_t get_code_reg_addr(uint8_t ch, uint8_t is_dac, uint8_t is_fast);
```
buha, line 460: why should this function be public ? can you provide a scenario where it's needed ?

also, if it's public, it should have the ad3552r prefix

CMinajigi: we can use this function to get address of dac or channel register based on mode so that we can directly write to register address.

drivers/dac/ad3552r/ad3552r.c
```diff
@@ -607,16 +605,44 @@ static int32_t _ad3552r_set_code_value(struct ad3552r_desc *desc,
...
+	addr = ad3552r_get_code_reg_addr(ch, 1, desc->ch_data[ch].fast_en);
+	return ad3552r_write_reg(desc, addr, val);
+}
+
+int32_t ad3552r_simulatneous_update_enable(struct ad3552r_desc *desc)
+{
+	if (desc->ch_data[0].fast_en != desc->ch_data[1].fast_en)
+		return -EINVAL;
+	if (desc->is_simultaneous) {
+		if (desc->ch_data[0].fast_en)
```
amiclaus, line 619: some more code optimization that can be done in your patch: if you have the following scenario:
```
if (statement)
   return x;
else
   return y;
```
This can be reduced to:
```
if (statement)
   return x;
return y;
```
There are a lot of occurrences like this in the code that can be optimized.

amiclaus (commented):
one more comment regarding code optimization on my side.

drivers/dac/ad3552r/ad3552r.c
CMinajigi: Addressed this comment.

PR #1806, opened 2023-06-20
Freertos initial

rbudai98:
* FreeRTOS generic makefile added
* Current build commands need: `FREERTOS=y` for enabling the library and `FREERTOS_KERNEL=/path/to/Kernel` for specifying the path towards freeRTOS Kernel (https://github.com/FreeRTOS/FreeRTOS-Kernel)
* FreeRTOS.config file is linked from project folder
* Current implementation has the ARM4F CMSIS file linked for max devices, in future it will need generalizations
* (example project in different PR)

PR #1825, opened 2023-07-07
projects/iio_demo_freeRTOS: initial commit

rbudai98:
Currently only maxim platform beeing supported,
Demo contains two tasks started in the same time,
one is a blinking task and another one is the iio_demo task.

This project is intended as example project for further development and act as a base line solution

Example to build and run the project:
(After downloading: https://github.com/FreeRTOS/FreeRTOS-Kernel)
```
export FREERTOS=y
export MAXIM_LIBRARIES=/home/robi/Desktop/MaximSDK/MaximSDK/Libraries
export FREERTOS_KERNEL=/home/robi/Workspace_no-os/FreeRTOS-Kernel
make PLATFORM=maxim TARGET=max78000 run
```

### DIFF
--- a/drivers/platform/freeRTOS/freertos_mutex.c
+++ b/drivers/platform/freeRTOS/freertos_mutex.c
@@ -40,6 +40,7 @@
 #include <FreeRTOS.h>
 #include "no_os_mutex.h"
 #include "semphr.h"
+#include "queue.h"
 
 /**
  * @brief Initialize mutex.
@@ -48,9 +49,12 @@
  */
 __attribute__((weak)) inline void no_os_mutex_init(void **mutex)
 {
-	if ((*mutex) == NULL) {
-		(*mutex) = (SemaphoreHandle_t *)no_os_malloc(sizeof(SemaphoreHandle_t));
-		(*mutex) = xSemaphoreCreateMutex();
+	if (*mutex == NULL) {
+		SemaphoreHandle_t* mutexTmp = (SemaphoreHandle_t *)no_os_calloc(1,
+					      sizeof(SemaphoreHandle_t));
+
+		*mutex = xSemaphoreCreateBinary();
+		xSemaphoreGive(*mutex);
 	}
 }
 
@@ -62,10 +66,8 @@ __attribute__((weak)) inline void no_os_mutex_init(void **mutex)
 __attribute__((weak)) inline void no_os_mutex_lock(void *mutex)
 {
 	if (mutex != NULL)
-		xSemaphoreTake(mutex,
-			       freeRTOS_WAIT_TIME); //replace with define from freertos config file
+		xSemaphoreTake((SemaphoreHandle_t)mutex, portMAX_DELAY);
 }
-
 /**
  * @brief Unlock mutex.
  * mutex - Pointer toward the mutex.
@@ -74,7 +76,7 @@ __attribute__((weak)) inline void no_os_mutex_lock(void *mutex)
 __attribute((weak)) inline void no_os_mutex_unlock(void *mutex)
 {
 	if (mutex != NULL)
-		xSemaphoreGive(mutex);
+		xSemaphoreGive((SemaphoreHandle_t)mutex);
 }
 
 /**
@@ -84,7 +86,8 @@ __attribute((weak)) inline void no_os_mutex_unlock(void *mutex)
  */
 __attribute__((weak)) inline void no_os_mutex_remove(void *mutex)
 {
-	if (mutex != NULL)
-		no_os_free(mutex);
+	if (mutex != NULL) {
+		vSemaphoreDelete((SemaphoreHandle_t)mutex);
+	}
 }
 
--- a/projects/iio_demo_freeRTOS/src/common/common_data.c
+++ b/projects/iio_demo_freeRTOS/src/common/common_data.c
@@ -1,0 +1,93 @@
+/***************************************************************************//**
+ *   @file   common_data.c
+ *   @brief  Defines common data to be used by iio_demo_freeRTOS examples.
+ *   @author Robert Budai (robert.budai@analog.com)
+********************************************************************************
+ * Copyright 2023(c) Analog Devices, Inc.
+ *
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *  - Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ *  - Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in
+ *    the documentation and/or other materials provided with the
+ *    distribution.
+ *  - Neither the name of Analog Devices, Inc. nor the names of its
+ *    contributors may be used to endorse or promote products derived
+ *    from this software without specific prior written permission.
+ *  - The use of this software may or may not infringe the patent rights
+ *    of one or more patent holders.  This license does not release you
+ *    from the requirement that you obtain separate licenses from these
+ *    patent holders to use this software.
+ *  - Use of the software either in source or binary form, must be run
+ *    on or directly connected to an Analog Devices Inc. component.
+ *
+ * THIS SOFTWARE IS PROVIDED BY ANALOG DEVICES "AS IS" AND ANY EXPRESS OR
+ * IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, NON-INFRINGEMENT,
+ * MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED.
+ * IN NO EVENT SHALL ANALOG DEVICES BE LIABLE FOR ANY DIRECT, INDIRECT,
+ * INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, INTELLECTUAL PROPERTY RIGHTS, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+*******************************************************************************/
+
+/******************************************************************************/
+/***************************** Include Files **********************************/
+/******************************************************************************/
+#include "common_data.h"
+
+/******************************************************************************/
+/********************** Macros and Constants Definitions **********************/
+/******************************************************************************/
+#ifdef ENABLE_LOOPBACK
+static uint16_t loopback_buffs[DEMO_CHANNELS][SAMPLES_PER_CHANNEL];
+#endif
+
+struct no_os_gpio_init_param gpio_ip = {
+	.port = GPIO_PORT,
+	.number = 0,
+	.pull = NO_OS_PULL_NONE,
+	.platform_ops = GPIO_OPS,
+	.extra = GPIO_EXTRA,
+};
+
+struct no_os_uart_init_param iio_demo_uart_ip = {
+	.device_id = UART_DEVICE_ID,
+	.irq_id = UART_IRQ_ID,
+	.asynchronous_rx = true,
+	.baud_rate = UART_BAUDRATE,
+	.size = NO_OS_UART_CS_8,
+	.parity = NO_OS_UART_PAR_NO,
+	.stop = NO_OS_UART_STOP_1_BIT,
+	.extra = UART_EXTRA,
+	.platform_ops = UART_OPS,
+};
+
+struct adc_demo_init_param adc_init_par = {
+	.ext_buff_len = SAMPLES_PER_CHANNEL,
+	.ext_buff = (uint16_t **)loopback_buffs,
+	.dev_global_attr = 3333,
+	.dev_ch_attr = {
+		1111, 1112, 1113, 1114, 1115, 1116, 1117, 1118,
+		1119, 1120, 1121, 1122, 1123, 1124, 1125, 1126
+	}
+};
+
+struct dac_demo_init_param dac_init_par = {
+	.loopback_buffer_len = SAMPLES_PER_CHANNEL,
+	.loopback_buffers = (uint16_t **)loopback_buffs,
+	.dev_global_attr = 4444,
+	.dev_ch_attr = {
+		1111, 1112, 1113, 1114, 1115, 1116, 1117, 1118,
+		1119, 1120, 1121, 1122, 1123, 1124, 1125, 1126
+	}
+};
+
+uint8_t in_buff[MAX_SIZE_BASE_ADDR] = {0};
+uint8_t out_buff[MAX_SIZE_BASE_ADDR] = {0};
--- a/projects/iio_demo_freeRTOS/src/common/common_data.h
+++ b/projects/iio_demo_freeRTOS/src/common/common_data.h
@@ -1,0 +1,92 @@
+/***************************************************************************//**
+ *   @file   iio_demo/src/common/common_data.h
+ *   @brief  Defines common data to be used by iio_demo_freeRTOS examples.
+ *   @author Robert Budai (robert.budai@analog.com)
+********************************************************************************
+ * Copyright 2023(c) Analog Devices, Inc.
+ *
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *  - Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ *  - Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in
+ *    the documentation and/or other materials provided with the
+ *    distribution.
+ *  - Neither the name of Analog Devices, Inc. nor the names of its
+ *    contributors may be used to endorse or promote products derived
+ *    from this software without specific prior written permission.
+ *  - The use of this software may or may not infringe the patent rights
+ *    of one or more patent holders.  This license does not release you
+ *    from the requirement that you obtain separate licenses from these
+ *    patent holders to use this software.
+ *  - Use of the software either in source or binary form, must be run
+ *    on or directly connected to an Analog Devices Inc. component.
+ *
+ * THIS SOFTWARE IS PROVIDED BY ANALOG DEVICES "AS IS" AND ANY EXPRESS OR
+ * IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, NON-INFRINGEMENT,
+ * MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED.
+ * IN NO EVENT SHALL ANALOG DEVICES BE LIABLE FOR ANY DIRECT, INDIRECT,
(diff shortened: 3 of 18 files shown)


rbudai98:
Requires: https://github.com/analogdevicesinc/no-OS/pull/1806

drivers/platform/freeRTOS/freertos_mutex.c
```diff
@@ -40,6 +40,9 @@
 #include <FreeRTOS.h>
 #include "no_os_mutex.h"
 #include "semphr.h"
+#include "queue.h"
+
+#define freeRTOS_WAIT_TIME 0
```
buha, line 45: I think this should be portMAX_DELAY and you could use it directly at line 73
https://www.freertos.org/a00122.html

projects/iio_demo_freeRTOS/src/platform/maxim/main.c
```diff
@@ -0,0 +1,184 @@
...
+#error "Please update ISR macro for UART CONSOLE_UART"
+#endif
+
+/**
+ * @brief iiod example task
+*/
+
+void iiodTask(void *pvParameters)
+{
+	iio_example_main();
```
buha, line 76: I think for this example it's ok to have one blinking LED task and one IIO task. Two iiod's don't really make sense as they can't really work together with one UART.

drivers/platform/freeRTOS/freertos_mutex.c
buha: Apparently 0 really means 0 timeout and it used for "semaphore polling" and portMAX_DELAY can either mean the maximum wait time (0xFFFFFFFF on some platforms) or it can mean infinite wait time

projects/iio_demo_freeRTOS/src/platform/maxim/main.c
```diff
@@ -0,0 +1,184 @@
...
+	while(1) {
+		no_os_mdelay(10);
+	}
+}
+
+void iiodTask2(void *pvParameters)
```
buha, line 82: you should remove iiodTask2

projects/iio_demo_freeRTOS/src/platform/maxim/main.c
```diff
@@ -0,0 +1,184 @@
...
+{
+	//Initial wait time
+	for (int i = 0; i < 0x00FFFF; i++) {}
+
+	/* Configure task */
+	if ((xTaskCreate(iiodTask, (const char *)"iiodTask", configIIO_APP_STACK_SIZE,
+			 NULL,
+			 tskIDLE_PRIORITY + 1, NULL) != pdPASS)) {
+		printf("xTaskCreate() failed to create a task.\n");
+        }else if ((xTaskCreate(iiodTask2, (const char *)"iiodTask2", configIIO_APP_STACK_SIZE,
```
buha, line 130: remove this whole Task2 section

projects/iio_demo_freeRTOS/src/platform/maxim/main.c
rbudai98: removed

projects/iio_demo_freeRTOS/src/platform/maxim/main.c
rbudai98: removed

projects/iio_demo_freeRTOS/src/platform/maxim/main.c
```diff
@@ -0,0 +1,157 @@
...
+	} else if ((xTaskCreate(blinkingTask, (const char *)"blinkingTask",
+				configMINIMAL_STACK_SIZE,
+				NULL,
+				tskIDLE_PRIORITY + 1, NULL) != pdPASS)) {
+		printf("xTaskCreate() failed to create a task.\n");
+	} else {
+		/* Start scheduler */
+		printf("Starting scheduler.\n");
+		vTaskStartScheduler();
+	}
```
rbolboac, line 127: I am not sure this is the best form for the API calls...also, do we want to start the scheduler if a task creation failed?
I would propose something like:
```
	ret = xTaskCreate(iiodTask, (const char *)"iiodTask", configIIO_APP_STACK_SIZE,
			  NULL,
			  tskIDLE_PRIORITY + 1, NULL);
	if (ret != pdPASS) {
		printf("xTaskCreate() failed to create iiodTask task.\n");
		return ret
	}
	ret = xTaskCreate(blinkingTask, (const char *)"blinkingTask",
			  configMINIMAL_STACK_SIZE,
			  NULL,
			  tskIDLE_PRIORITY + 1, NULL);
	if (ret != pdPASS) {
		printf("xTaskCreate() failed to create blinkingTask task.\n");
		return ret;
	}
```

rbolboac (commented):
a first review from my side with things which caught my eye...
also you have some commits which are missing commit description. I think it is mandatory to have a commit description in No-Os

projects/iio_demo_freeRTOS/src/platform/maxim/main.c
```diff
@@ -0,0 +1,148 @@
...
+int create_tasks(void)
+{
+	int ret;
+	for (int i = 0; i < 0x00FFFF; i++) {}
+
+	ret = xTaskCreate(iiodTask, (const char *)"iiodTask", configIIO_APP_STACK_SIZE,
+			  NULL, tskIDLE_PRIORITY + 1, NULL);
+	if (ret != pdPASS) {
+		printf("xTaskCreate() failed to create iiodTask task.\n");
+		return ret
```
buha, line 109: missing `;`

buha: not even sure how this can compile

projects/iio_demo_freeRTOS/src/platform/maxim/main.c
```diff
@@ -0,0 +1,148 @@
...
+			  configMINIMAL_STACK_SIZE, NULL, tskIDLE_PRIORITY + 1, NULL);
+	if (ret != pdPASS) {
+		printf("xTaskCreate() failed to create blinkingTask task.\n");
+		return ret;
+	}
+
+	printf("Starting scheduler.\n");
+	vTaskStartScheduler();
+
+	while (1) {
```
buha, line 121: I think this loop is unnecessary, the control only gets here if there is insufficient memory for the scheduler. Otherwise the scheduler has it's own loop.
Also, when exiting main, there is an infinite loop typically implemented in assembly (depends on toolchain).
So you can safely remove this.

projects/iio_demo_freeRTOS/src/platform/maxim/main.c
rbudai98: So far I just corrected the comments, I still have to test it and will write a versioning log after

projects/iio_demo_freeRTOS/src/platform/maxim/main.c
rbudai98: Will remove it than

buha (approved):

projects/iio_demo_freeRTOS/src/common/app_config.h
```diff
@@ -0,0 +1,47 @@
+/***************************************************************************//**
+ *   @file   app_config.h
+ *   @brief  Config file of IIO Demo FreeRTOS project.
+ *   @author Robert Budai (robert.budai@analog.com)
+********************************************************************************
+ * Copyright 2015(c) Analog Devices, Inc.
```
rbolboac, line 6: *2023

projects/iio_demo_freeRTOS/src/common/app_config.h
```diff
@@ -0,0 +1,47 @@
...
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+*******************************************************************************/
+#ifndef CONFIG_H_
+#define CONFIG_H_
+
+#define ENABLE_LOOPBACK
+
+#define WIFI_SSID	"RouterSSID"
+#define WIFI_PWD	"******"
```
rbolboac, line 45: don't think these are used

projects/iio_demo_freeRTOS/src/common/common_data.c
```diff
@@ -0,0 +1,93 @@
+/***************************************************************************//**
+ *   @file   common_data.c
+ *   @brief  Defines common data to be used by iio_demo_freeRTOS examples.
+ *   @author Robert Budai (robert.budai@analog.com)
+********************************************************************************
+ * Copyright 2022(c) Analog Devices, Inc.
```
rbolboac, line 6: *2023

projects/iio_demo_freeRTOS/src/common/common_data.c
```diff
@@ -0,0 +1,93 @@
...
+	.platform_ops = UART_OPS,
+};
+
+struct adc_demo_init_param adc_init_par = {
+	.ext_buff_len = SAMPLES_PER_CHANNEL,
+	.ext_buff = (uint16_t **)loopback_buffs,
+	.dev_global_attr = 3333,
+	.dev_ch_attr = {
+		1111, 1112, 1113, 1114, 1115, 1116, 1117, 1118,
+		1119, 1120,	1121, 1122, 1123, 1124, 1125, 1126
```
rbolboac, line 78: please fix alignment

projects/iio_demo_freeRTOS/src/common/common_data.h
```diff
@@ -0,0 +1,78 @@
+/***************************************************************************//**
+ *   @file   iio_demo/src/common/common_data.h
+ *   @brief  Defines common data to be used by iio_demo_freeRTOS examples.
+ *   @author Robert Budai (robert.budai@analog.com)
+********************************************************************************
+ * Copyright 2022(c) Analog Devices, Inc.
```
rbolboac, line 6: *2023

projects/iio_demo_freeRTOS/src/examples/iio_example/iio_example.c
```diff
@@ -0,0 +1,108 @@
+/***************************************************************************//**
+ *   @file   iio_example.c
+ *   @brief  Implementation of IIO example for iio_demo_freeRTOS project.
+ *   @author Robert Budai (robert.budai@analog.com)
+********************************************************************************
+ * Copyright 2022(c) Analog Devices, Inc.
```
rbolboac, line 6: *2023

projects/iio_demo_freeRTOS/src/examples/iio_example/iio_example.h
```diff
@@ -0,0 +1,51 @@
+/***************************************************************************//**
+ *   @file   iio_example.h
+ *   @brief  IIO example header for iio_demo_freeRTOS project
+ *   @author Robert Budai (robert.budai@analog.com)
+********************************************************************************
+ * Copyright 2022(c) Analog Devices, Inc.
```
rbolboac, line 6: *2023

rbolboac (commented):
just some nits from my side

rbudai98:
v3:
* removed not needed while loops
* added returns for task creations
* styling/file format fixed

buha (approved):
lgtm

projects/iio_demo_freeRTOS/src/platform/maxim/main.c
```diff
@@ -0,0 +1,143 @@
...
+
+/**
+ * @brief Creating iio example with blinking led task
+ * @return  Returns error code
+ *
+*/
+int create_tasks(void)
+{
+	int ret;
+	for (int i = 0; i < 0x00FFFF; i++) {}
```
rbolboac, line 103: it's not all that clear why this for is needed. maybe a comment would be useful. also, if this is intended to be a delay, maybe it would be better to use no_os_delay

projects/iio_demo_freeRTOS/src/platform/maxim/parameters.h
```diff
@@ -0,0 +1,92 @@
...
+
+#define INTC_DEVICE_ID	0
+#define UART_IRQ_ID    	UART0_IRQn
+#define UART_DEVICE_ID	0
+#define UART_BAUDRATE	57600
+#define UART_EXTRA      &iio_demo_uart_extra_ip
+#define UART_OPS        &max_uart_ops
+#define GPIO_EXTRA      &gpio_extra_ip
+#define GPIO_OPS        &max_gpio_ops
+#define GPIO_PORT		2
```
rbolboac, line 87: please use the same alignment

projects/iio_demo_freeRTOS/src/examples/iio_example/iio_example.c
```diff
@@ -0,0 +1,108 @@
...
+		.buff = (void *)DAC_DDR_BASEADDR,
+		.size = MAX_SIZE_BASE_ADDR
+	};
+
+	status = adc_demo_init(&adc_desc, &adc_init_par);
+	if (status)
+		return status;
+
+	status = dac_demo_init(&dac_desc, &dac_init_par);
+	if (status)
```
rbolboac, line 89: this is something that is missing from other projects.. but in case of an error adc_demo_remove should be called

projects/iio_demo_freeRTOS/src/examples/iio_example/iio_example.c
```diff
@@ -0,0 +1,108 @@
...
+		IIO_APP_DEVICE("dac_demo", dac_desc,
+			       &dac_demo_iio_descriptor,NULL, &dac_buff, NULL)
+	};
+
+	app_init_param.devices = devices;
+	app_init_param.nb_devices = NO_OS_ARRAY_SIZE(devices);
+	app_init_param.uart_init_params = iio_demo_uart_ip;
+
+	status = iio_app_init(&app, app_init_param);
+	if (status)
```
rbolboac, line 104: this is something that is missing from other projects.. but in case of an error adc_demo_remove and dac_demo_remove should be called

projects/iio_demo_freeRTOS/src/examples/iio_example/iio_example.c
```diff
@@ -0,0 +1,108 @@
...
+		return status;
+
+	return iio_app_run(app);
```
rbolboac, line 107: this is something that is missing from other projects.. but in case of an error adc_demo_remove, dac_demo_remove and iio_app_remove should be called before returning to avoid memory leaks

projects/iio_demo_freeRTOS/src/platform/maxim/main.c
```diff
@@ -0,0 +1,143 @@
...
+int blinkingTask()
+{
+	int ret = 0;
+	struct no_os_gpio_desc *led_pin;
+	ret = no_os_gpio_get(&led_pin, &gpio_ip);
+	if (ret)
+		return ret;
+
+	ret = no_os_gpio_direction_output(led_pin, NO_OS_GPIO_LOW);
+	if (ret)
```
rbolboac, line 72: no_os_gpio_free should be called in case of an error before returning to avoid memory leaks, see also the cases below

projects/iio_demo_freeRTOS/src/platform/maxim/main.c
```diff
@@ -0,0 +1,143 @@
...
+			  NULL, tskIDLE_PRIORITY + 1, NULL);
+	if (ret != pdPASS) {
+		printf("xTaskCreate() failed to create iiodTask task.\n");
+		return ret;
+	}
+	ret = xTaskCreate(blinkingTask, (const char *)"blinkingTask",
+			  configMINIMAL_STACK_SIZE, NULL, tskIDLE_PRIORITY + 1, NULL);
+	if (ret != pdPASS) {
+		printf("xTaskCreate() failed to create blinkingTask task.\n");
+		return ret;
```
rbolboac, line 115: is there any dynamic allocation happening when xTaskCreate is called? if so, before returning make sure that memory is freed

rbolboac (commented):
some other comments from my side

projects/iio_demo_freeRTOS/src/platform/maxim/main.c
rbudai98: In case the task creation is not successful, there should not be any allocated memory. There is a xTaskDelete function for deleting tasks, I have added that as a safety step.

projects/iio_demo_freeRTOS/src/platform/maxim/main.c
rbudai98: added

projects/iio_demo_freeRTOS/src/examples/iio_example/iio_example.c
rbudai98: added

projects/iio_demo_freeRTOS/src/examples/iio_example/iio_example.c
rbudai98: added

projects/iio_demo_freeRTOS/src/examples/iio_example/iio_example.c
rbudai98: added

projects/iio_demo_freeRTOS/src/platform/maxim/main.c
rbudai98: removed, not needed

rbudai98:
v4:
* added memory free for led, adc, dac buffers and task delete

projects/iio_demo_freeRTOS/src/examples/iio_example/iio_example.c
```diff
@@ -0,0 +1,116 @@
...
+	};
+
+	struct iio_data_buffer dac_buff = {
+		.buff = (void *)DAC_DDR_BASEADDR,
+		.size = MAX_SIZE_BASE_ADDR
+	};
+
+	status = adc_demo_init(&adc_desc, &adc_init_par);
+	if (status)
+		goto error_adc_buffer;
```
rbolboac, line 86: this does not look like it is handled properly...see all the cases in which you applied the same logic.
so basically if adc_demo_init is unsuccessful no memory allocation is performed and there is no need to call adc_demo_remove. adc_demo_remove should be called only if dac_demo_init is not successful.

projects/iio_demo_freeRTOS/src/platform/maxim/main.c
```diff
@@ -0,0 +1,151 @@
...
+			vTaskDelete(iio_app_handle);
+		return ret;
+	}
+	ret = xTaskCreate(blinkingTask, (const char *)"blinkingTask",
+			  configMINIMAL_STACK_SIZE, NULL, tskIDLE_PRIORITY + 1, &led_task_handle);
+	if (ret != pdPASS) {
+		printf("xTaskCreate() failed to create blinkingTask task.\n");
+		if (led_task_handle != NULL)
+			vTaskDelete(led_task_handle);
+		return ret;
```
rbolboac, line 123: in case iiodTask is created successfully, shouldn't we also delete that task before returning?

rbudai98: In this case I think it is more up to the developer. Maybe one task fails, but you might still want another driver to publish some data. (but in this case I still have to remove the returns :) )

rbudai98: Also in case someone links the FreeRTOS_CLI for terminal access, you can't stop the entire rtos, just because one task is failing.

rbudai98: As discussed will delete all the tasks in case of an error.

projects/iio_demo_freeRTOS/src/examples/iio_example/iio_example.c
```diff
@@ -0,0 +1,114 @@
...
+
+	app_init_param.devices = devices;
+	app_init_param.nb_devices = NO_OS_ARRAY_SIZE(devices);
+	app_init_param.uart_init_params = iio_demo_uart_ip;
+
+	status = iio_app_init(&app, app_init_param);
+	if (status)
+		goto error_iio_app;
+
+	status = iio_app_run(app);
```
rbolboac, line 107: one last thing: in case status is not 0, iio_app_remove should be called to avoid memory leaks

rbolboac (approved):

PR #1828, opened 2023-07-12
drivers: imu: iio_adis: Add direct register access APIs

rbolboac:


buha (approved):

projects/eval-adis/src/examples/examples_src.mk
```diff
@@ -7,6 +7,7 @@ endif
 ifeq (y,$(strip $(IIO_TRIGGER_EXAMPLE)))
 TINYIIOD=y
 CFLAGS += -DIIO_TRIGGER_EXAMPLE=1
+CFLAGS += -O2
```
CiprianRegus, line 10: I think it would be better to add a `RELEASE` make flag, so this flag will be set in generic.mk.

rbolboac:
- implemented review findings

amiclaus (approved):

CiprianRegus (approved):

PR #1835, opened 2023-07-19
Print binary size for Maxim

CiprianRegus:
Print the executable size (for .hex and .elf) after the compilation for Maxim platform.


amiclaus (approved):

rbolboac (approved):

buha:
Let's not rush it and implement it in generic.mk for all platforms, all toolchains should have `size` tool. It's important to have the same build/make experience regardless of platform and printing the program size shouldn't be an exception.

CiprianRegus:
Updated to run on every platform.

amiclaus (approved):

buha:
There were some remaining questions about whether Microblaze or Altera have the `size` tool in the toolchain, I've checked this and they do so I think we can safely merge.

PR #1845, opened 2023-08-01
add maxim programming script (openocd based)

buha:
This commit adds a standalone script that can program a Maxim micro using CMSIS-DAP and swd:

maxfla.sh file.elf

One can also specify the swd programmer id from dmesg:

maxfla.sh file.elf 1234567896d9088800000000000000000000000097969906

Please try it out and on your system and let me know if you have any issues.

### DIFF
--- a/tools/scripts/platform/maxim/maxfla.sh
+++ b/tools/scripts/platform/maxim/maxfla.sh
@@ -1,0 +1,109 @@
+#!/bin/bash
+# Copyright 2023(c) Analog Devices, Inc.
+# 
+# All rights reserved.
+# 
+# Redistribution and use in source and binary forms, with or without modification,
+# are permitted provided that the following conditions are met:
+#     - Redistributions of source code must retain the above copyright
+#       notice, this list of conditions and the following disclaimer.
+#     - Redistributions in binary form must reproduce the above copyright
+#       notice, this list of conditions and the following disclaimer in
+#       the documentation and/or other materials provided with the
+#       distribution.
+#     - Neither the name of Analog Devices, Inc. nor the names of its
+#       contributors may be used to endorse or promote products derived
+#       from this software without specific prior written permission.
+#     - The use of this software may or may not infringe the patent rights
+#       of one or more patent holders.  This license does not release you
+#       from the requirement that you obtain separate licenses from these
+#       patent holders to use this software.
+#     - Use of the software either in source or binary form, must be run
+#       on or directly connected to an Analog Devices Inc. component.
+#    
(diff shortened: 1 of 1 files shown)


tools/maxfla.sh
```diff
@@ -0,0 +1,106 @@
...
+Usage:
+	maxfla.sh path/to/binary.elf [adapter_serial_number]
+
+Author:
+	darius.berghe@analog.com
+END
+
+)
+
+function print_run
```
amiclaus, line 51: similar to this function, i'd do one for the prerequisites (openocd/hidapi) and call it prior to `
print_run ~/.maxflash/openocd/src/openocd -s ~/.maxflash/openocd/tcl -c "adapter driver cmsis-dap; transport select swd; $DAP_SERIAL" -f "target/$TARGET.cfg"  -c "program $1 verify reset exit"`
In this manner we keep this modularized :)

tools/maxfla.sh
```diff
@@ -0,0 +1,106 @@
...
+
+	"$@"
+}
+
+if [ $# -eq 0 ]; then
+	echo "$USAGE"
+	exit 0
+fi
+
+if [ ! -d ~/.maxflash/hidapi ]; then
```
amiclaus, line 65: just a thought: maybe restrict the prerequisites paths to the current no-OS path ($(pwd) ), even though the script is not necessarily dependent on it. See:
https://github.com/analogdevicesinc/no-OS/blob/464d752890670e64ed5aca4db51596144d7d6686/ci/run_build.sh#L7

amiclaus (commented):
tested on my machine and works fine. Some minor comments/thoughts.

buha:
V2:
- rebased on master
- fixed Debian dependencies
- added ldconfig step after installing hidapi

amiclaus (commented):
lgtm, tested last version and works. some thoughts:
 - maybe we should find a more suitable location for this script (maybe `tools/scripts/flash/maxfla.sh`?)
 - i'd still make a separate functions: one for the prerequisites procedure and one for the input arguments handling.

tools/maxfla.sh
buha: This script is supposed to be run completely standalone (separate from no-OS) even though I publish it in no-OS.

So somebody could (and will!) wget download this script alone and flash a binary without going through no-OS Maxim prerequisite section at all. So I can't really do what you suggest here.

amiclaus: Makes sense.

buha:
V4:
- moved script to `tools/scripts/platform/maxim`

CiprianRegus (approved):

danmois (approved):

PR #1850, opened 2023-08-07
Dma example support versal

danmois:
Add support for Versal in the script that downloads memory contents from the device.


amiclaus (approved):
are these patches valid for our release branch too?

danmois:
> are these patches valid for our release branch too?

Yes.

PR #1856, opened 2023-08-08
network:lwip_raw_socket: Print the IP after config

CiprianRegus:
Print the IP address, network mask and gateway's IP if the configuration is successful (meaning that the network interface got an IP either by DHCP or AutoIP).


amiclaus (approved):
lgtm, maybe we can use "print_log.h"

PR #1857, opened 2023-08-08
parameters.h  Update defines to match the latest HDL

danmois:
Update AXI SPI related & other defines in no-OS to match the latest HDL.

HDL introduced a new SPI hierarchy which is reflected in the macros generated in xparameters.h from .xsa.

For example, what used to be called before XPAR_CN0561_SPI_AXI_BASEADDR is now called XPAR_SPI_CN0561_SPI_CN0561_AXI_REGMAP_BASEADDR.

Git blame in hdl:
https://github.com/analogdevicesinc/hdl/commit/553774319a12254bfd5792b9869234d36345afd9#diff-bc564d9ee1a3935befd0bb57534492fed38a6c78dd280fb07219510fc886d980R73


amiclaus (approved):
looks good, let's wait for the build to finish successfully for the Xilinx platform.

amiclaus (changes requested):
re-triggered the Xilinx build since there are still some issues.

projects/ad9361/src/parameters.h
```diff
@@ -122,9 +122,14 @@
 #define ADC_DDR_BASEADDR			XPAR_DDR3_SDRAM_S_AXI_BASEADDR + 0x800000
 #define DAC_DDR_BASEADDR			XPAR_DDR3_SDRAM_S_AXI_BASEADDR + 0xA000000
 #else
+#if defined(XPAR_AXI_DDR_CNTRL_BASEADDR)
```
amiclaus, line 125: this looks odd, and not very deductive. If you need to differentiate between boards you should go for the approach:
https://github.com/analogdevicesinc/no-OS/blob/6d75bc376eb7f1a4fd992cac75a7e33c0c89e37e/projects/ad9361/builds.json#L60

projects/ad7768-evb/src/ad7768_evb.c
```diff
@@ -256,15 +255,6 @@ int main(void)
 	if (ret != 0)
 		return -1;
 
-	axi_gpio_init.type = GPIO_PL;
```
amiclaus, line 259: I wouldn't remove this code yet. Rather check inline for the definition of `GPIO_STATUS_DEVICE_ID` and only if available perfrom these status checks.

dbogdan (approved):

PR #1860, opened 2023-08-11
Xilinx scripts

buha:
- move xilinx_capture script
- add script to plot the output of xilinx_capture
Screenshot from 2023-08-09 17-52-19


buha:
when and if this is merged we need to update this page:

https://wiki.analog.com/resources/no-os/dac_dma_example

danmois (approved):

PR #1861, opened 2023-08-15
tools: scripts: maxfla: update script

amiclaus:
Clearly define the steps performend in the script (parse_input, build_prerequisites and program device) by adding suggestive functions.

No functional changes.


amiclaus (commented):
v2:
- make HALF_CORES as local variable and move to the build_prerequisites function since it is used only there.

buha (approved):
looks good but somebody should test this (i don't have hardware on me)

amiclaus (commented):
> looks good but somebody should test this (i don't have hardware on me)

I've tested it with hardware and works as expected. But a double check would be nice :)

CiprianRegus (approved):

CiprianRegus:
Also tested on my side (Debian + max32650). Works as expected.

PR #1867, opened 2023-08-21
Build on Windows using Git Bash only

buha:
To test this, you'll need to have Git Bash available on Windows and execute in it the `tools/scripts/git-bash.sh`.

The rest works as usual, just bear in mind that the paths should look like `/c/Path/to/YOUR/file` and they should be case sensitive. (`/c` stands for Windows `C:`, `/d` stands for Windows `D:` etc.):
- xilinx: `export PATH=export PATH=/c/Xilinx/Vitis/2022.2/bin:/c/Xilinx/Vitis/2022.2/gnu/aarch32/nt/gcc-arm-none-eabi/bin/:$PATH` instead of the usual 'source settings64.sh`
- maxim: `export MAXIM_LIBRARIES=/c/MaximSDK/Libraries`
- aducm3029: `export CCES_HOME=/c/ADI/cces`, a path without spaces is required, the default installation path of CCES contains spaces.

Tested (all, project, update, build, reset, clean, run, sdkopen) and works on Windows for the following platforms:
- [x] xilinx
- [x] maxim
- [x] aducm3029
- [ ] altera

sdkbuild, sdkclean and debug are untested but they are not critical, they are more intended for us, no-OS maintainers


rbolboac (approved):
lgtm, as long as it does the trick and works also on Windows

buha:
@dbogdan i need to update wiki when this gets merged

amiclaus (approved):

PR #1869, opened 2023-08-23
include: no_os_uart: Increase the value of UART_MAX_NUMBER

rbolboac:
For stm32 UART instance 5 is used. With the current implementation this would result in an out of bounds array access.


CiprianRegus (commented):
It would be more future proof to increase it a bit more. For example, the STM32F413 has 10 UART interfaces (from what I can find this is the highest number of any stm32 mcu). And also a bounds check before accessing the `uart_mutex_table` array would be nice.

CiprianRegus (approved):

PR #1877, opened 2023-09-01
AD74413R updates

CiprianRegus:
## Pull Request Description

Fixes/ new features related to the AD74413R driver. Required for the SWIOT1L project.

## PR Type

## PR Checklist


drivers/adc-dac/ad74413r/ad74413r.c
```diff
@@ -1085,6 +1085,24 @@ int ad74413r_remove(struct ad74413r_desc *desc)
 	if (!desc)
 		return -EINVAL;
 
+	/* Perform a reset to bring the device in the default state */
+	if (desc->reset_gpio) {
```
amiclaus, line 1105: this looks like duplicated code. A separate function for hardware reset might fit. And then call it both in init function and remove function.

drivers/adc-dac/ad74413r/ad74413r.c
```diff
@@ -156,7 +156,7 @@ int ad74413r_dac_voltage_to_code(uint32_t mvolts, uint32_t *code)
  * @param val - Register value.
  * @param buff - The communication buffer.
  */
-static void ad74413r_format_reg_write(uint8_t reg, uint16_t val, uint8_t *buff)
```
amiclaus, line 159: the data width for this device is exactly 16-bit, according to datasheet, therefore we already reduce the risk considerably of sending random data that might be passed with 32-bit wide values when not properly initialized. Moreover, is it a good idea to adapt a driver based on the upper layers? It should be vice-versa from my point of view.

drivers/adc-dac/ad74413r/ad74413r.c
```diff
@@ -312,8 +311,8 @@ int ad74413r_set_info(struct ad74413r_desc *desc, uint16_t mode)
 static int ad74413r_scratch_test(struct ad74413r_desc *desc)
 {
 	int ret;
-	uint16_t val;
-	uint16_t test_val = 0x1234;
+	uint32_t val = 0;
```
amiclaus, line 314: is this explicit initialization required?

CiprianRegus: it was added by mistake, will remove.

drivers/adc-dac/ad74413r/ad74413r.c
CiprianRegus: for this exact cast there is no risk of sending random data, since the `no_os_put_unaligned_be16()` takes a `uint16_t` as a parameter, thus it will cast it to 16 bit (meaning that the upper 16 bits will just be ignored). However I agree that since this function doesn't receive a pointer to the `val` parameter there is no good reason for this change and I can revert it. The behavior is the same in both cases.

What I meant by "requiring extra handling in the upper layers" is the following case: `ad74413r_get_adc_range(struct ad74413r_desc *, uint32_t, uint16_t *)` -  this function has to receive a pointer to a 16 bit wide integer. Passing a pointer to `uint32_t` on a little endian cpu will result in writing at the wrong address. I think it would be a better API to just use pointers to 32 bit integers for all the cases where a value has to be read.

buha: I think this 16 -> 32 bits change may introduce bugs

For example line 162 here, i think it might not work on any endianess because it's now picking 2 bytes out of 4 and rearranges them. With val being 16 bits, it would work as intended, picking 2 bytes out of 2 (all good!) and rearranging them... 
https://github.com/analogdevicesinc/no-OS/blob/a78abcb369f233b9086c0ccf8ae0287e761f4648/drivers/adc-dac/ad74413r/ad74413r.c#L159-L164

drivers/adc-dac/ad74413r/iio_ad74413r_trig.c
```diff
@@ -47,7 +47,7 @@
 /************************ Variable Declarations *******************************/
 /******************************************************************************/
 struct iio_trigger ad74413r_iio_trig_desc = {
-	.is_synchronous = false,
+	.is_synchronous = true,
```
buha, line 50: This seems like project specific behavior and makes me wonder why iio_ad74413r_trig.c file is where it is.
It seems the ad74413r_iio_trig_desc definition should be in application code and not under drivers/.

drivers/adc-dac/ad74413r/ad74413r.c
```diff
@@ -364,6 +374,17 @@ int ad74413r_set_channel_function(struct ad74413r_desc *desc,
...
+	if (ret)
+		return ret;
+
+	ret = ad74413r_reg_update(desc, AD74413R_CH_FUNC_SETUP(ch),
+				  AD74413R_CH_FUNC_SETUP_MASK, AD74413R_HIGH_Z);
+	if (ret)
+		return ret;
+
+	/* Each function should be selected for at least 130 us. */
+	no_os_udelay(150);
```
amiclaus, line 387: nit: it should be 130 here.

drivers/adc-dac/ad74413r/ad74413r.c
```diff
@@ -338,13 +337,24 @@ int ad74413r_reset(struct ad74413r_desc *desc)
...
+						  NO_OS_GPIO_LOW);
+		if (ret)
+			return ret;
 
-	ret = ad74413r_reg_write(desc, AD74413R_CMD_KEY, AD74413R_CMD_KEY_RESET_2);
-	if (ret)
-		return ret;
+		/* Minimum RESET signal pulse duration */
+		no_os_udelay(50);
+		ret = no_os_gpio_set_value(desc->reset_gpio, NO_OS_GPIO_HIGH);
```
amiclaus, line 348: missing error check. It will return 0 even though gpio_set_value fails.

drivers/adc-dac/ad74413r/ad74413r.c
```diff
@@ -338,13 +337,24 @@ int ad74413r_reset(struct ad74413r_desc *desc)
 {
 	int ret;
 
-	ret = ad74413r_reg_write(desc, AD74413R_CMD_KEY, AD74413R_CMD_KEY_RESET_1);
-	if (ret)
-		return ret;
+	if (desc->reset_gpio) {
```
amiclaus, line 340: please update the function documentation since now it does not do only a "soft reset".

CiprianRegus:
- Removed the commit which changed the function parameters to be `uint32_t`, instead fixed incorrect parameter passing in some places.
- Removed the `iio_ad74413r_trig.c` file, since the trigger type should be specific to each project.
- Addressed all the other review comments

drivers/adc-dac/ad74413r/iio_ad74413r.c
```diff
@@ -309,6 +310,8 @@ static int ad74413r_iio_read_offset(void *dev, char *buf, uint32_t len,
 		} else {
 			ad74413r_get_adc_range(iio_desc->ad74413r_desc,
 					       channel->address, &reg_val);
+			if (ret)
+				return ret;
```
rbolboac: am I missing something here? I don't see where ret value is set

CiprianRegus: `ret` has to be set, will change.

buha (approved):
minus that remark from Ramona, looks really good

amiclaus (approved):

rbolboac (approved):

PR #1885, opened 2023-09-06
drivers:adc:ad4080: Fix driver init

ribdp:
## Pull Request Description

Fix ad4080 driver init method
1. Prevent premature assignment of device structure instance before reg configuration is successful
2. Return error if ID does not match

## PR Type

## PR Checklist

### DIFF
--- a/drivers/adc/ad4080/ad4080.c
+++ b/drivers/adc/ad4080/ad4080.c
@@ -978,8 +978,10 @@ int ad4080_init(struct ad4080_dev **device,
 	if (ret)
 		goto error_spi;
 
-	if (data != AD4080_CHIP_ID)
+	if (data != AD4080_CHIP_ID) {
+		ret = -EINVAL;
 		goto error_spi;
+	}
 
 	/* Software Reset */
 	ret = ad4080_soft_reset(dev);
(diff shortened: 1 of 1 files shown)


drivers/adc/ad4080/ad4080.c
```diff
@@ -979,7 +979,7 @@ int ad4080_init(struct ad4080_dev **device,
 		goto error_spi;
 
 	if (data != AD4080_CHIP_ID)
-		goto error_spi;
```
amiclaus, line 982: here you should assign ret=-EFAULT and keep the conditional jump.

buha (approved):
now it looks correct, i'm merging

drivers/adc/ad4080/ad4080.c
```diff
@@ -978,8 +978,10 @@ int ad4080_init(struct ad4080_dev **device,
 	if (ret)
 		goto error_spi;
 
-	if (data != AD4080_CHIP_ID)
+	if (data != AD4080_CHIP_ID) {
+		ret = -EFAULT;
```
CiprianRegus, line 982: `EFAULT` means that a function argument points to an invalid memory address. Doesn't really fit here.

ribdp: Oh, I referred to a recently created driver, for handling this case - https://github.com/analogdevicesinc/no-OS/blob/master/drivers/adc/ad4858/ad4858.c#L969
Would you recommend using EBADMSG? Or a different error code?

CiprianRegus: I think `EINVAL` would be ok.

ribdp: Fixed this

CiprianRegus (approved):

PR #1892, opened 2023-09-07
Fix the IIO attribute negative representation

CiprianRegus:
## PR Type

## PR Checklist

### DIFF
--- a/iio/iio.c
+++ b/iio/iio.c
@@ -668,7 +668,7 @@ int32_t iio_parse_value(char *buf, enum iio_val fmt, int32_t *val,
 int iio_format_value(char *buf, uint32_t len, enum iio_val fmt,
 		     int32_t size, int32_t *vals)
 {
-	uint64_t tmp;
+	int64_t tmp;
 	int32_t integer, fractional;
 	bool dB = false;
 	int32_t i = 0;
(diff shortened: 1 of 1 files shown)


amiclaus (approved):

buha:
ok for the 2nd commit but with first commit i have some issues:
1. It's too similar to the linux implementation
2. There are 2 possible modulo formulas which depend on the way rounding is done
```
Type 1, truncation:  rem = a − b * (a / b)
Type 2, flooring:    rem = a - b * floor(a / b)
```

I put together this code to show the difference:
```
#include <stdio.h>
#include <stdint.h>
#include <math.h>

int64_t div_s64_rem(int64_t dividend, int32_t divisor, int32_t *remainder)
{
	*remainder = dividend % divisor;
	return dividend / divisor;
}

int64_t div_s64_rem_floor(int64_t dividend, int32_t divisor, int32_t *remainder)
{
	*remainder = dividend - divisor * (int64_t)floor((double)dividend / divisor);
	return dividend / divisor;
}

void compute(int64_t a, int32_t b)
{
    int64_t div;
    int32_t rem;
    div = div_s64_rem(a, b, &rem);
    printf("a=%ld, b=%d\n", a, b);
    printf("\ttrunc a/b=%ld rem=%d\n", div, rem);
    div_s64_rem_floor(a, b, &rem);
    printf("\tfloor a/b=%ld rem=%d\n", div, rem);
}

int main()
{
    compute(-1, 10);
    compute(-2, 9);
    compute(-7, 5);
    compute(-5, 3);

    return 0;
}
```

In my opinion we should both these points by:
1. implement it in a different way than linux due to licensing issues
2. do not delete the previous implementation which is fine, provide alternative flooring implementation (or make the flooring implementation as default but keep the other one)

CiprianRegus:
- Removed the `no_os_div_s64_rem()` changes.

rbolboac (approved):

amiclaus (approved):

PR #1894, opened 2023-09-07
drivers: api: set desc to NULL on remove 

amiclaus:
## Pull Request Description
[drivers: api: set desc to NULL on remove](https://github.com/analogdevicesinc/no-OS/commit/85d4f66f04464fa98ec7b2306dc09e86066ebcad)

[drivers: remove peripheral descriptor NULL set](https://github.com/analogdevicesinc/no-OS/commit/379135263424a8e776397efcbccf9980deffd380)

## PR Type

## PR Checklist


buha:
This wont work, you can check for yourself:
```
#include <stdio.h>

void function(int *p)
{
    p = NULL;
}

int main()
{
    int a;
    
    printf("before: %x\n", &a);
    function(&a);
    printf("after: %x\n", &a);

    return 0;
}
```

PR #1900, opened 2023-09-13
drivers: remove NULL set in remove function

amiclaus:
## Pull Request Description

The current approach of the `_remove` functions prototype does not allow proper set of the descriptor to NULL.

The function arguments are passed by value, therefore changing the local pointer won't affect the original pointer in the function being called.

## PR Type

## PR Checklist


amiclaus (commented):
perform rebase.

buha:
So as discussed in the meeting, this should be the step 1 of our attempt to properly handle deallocation of resources. My question here is how do we know these are all the places where we do this badly ?

(Somewhat unrelated) Step 2 should probably be double pointers for all remove() functions because without it, we get into very nasty issues like:
* dangling pointers
* double free

amiclaus (commented):
> So as discussed in the meeting, this should be the step 1 of our attempt to properly handle deallocation of resources. My question here is how do we know these are all the places where we do this badly ? I couldn't find an easy efficient way to recursively search for where we do this useless assignment to NULL so maybe you have a script/regex, whatever...
> 
> (Somewhat unrelated) Step 2 should probably be double pointers for all remove() functions because without it, we get into very nasty issues like:
> 
>     * dangling pointers
> 
>     * double free

Regarding step 1, i used search for all NULL assignments in the `drivers` and `projects` path, since there is where we encounter the `_remove` functions. And removed the ones that were done incorrectly.

buha (approved):

PR #1901, opened 2023-09-13
doc: sphinx: add sources

amiclaus:
## Pull Request Description

Add Sphinx documentation used for generating the main github pages of the no-OS repository.

## PR Type

## PR Checklist


amiclaus (commented):
add commit for sphinx build integration into the CI workflow.

amiclaus (commented):
v2: 
 - rebase with latest master branch

buha (commented):
Wouldn't it be better to have independent scripts for doxygen and sphinx? Yes they are both documentation related, but it just doesn't make sense to have them together in one script. Doxygen takes long to generate. Maybe I wanna quick check the sphinx output, I'd have to use the doxygen script and it wouldn't be quick.

amiclaus (commented):
> Wouldn't it be better to have independent scripts for doxygen and sphinx? Yes they are both documentation related, but it just doesn't make sense to have them together in one script. Doxygen takes long to generate. Maybe I wanna quick check the sphinx output, I'd have to use the doxygen script and it wouldn't be quick.

The script is intended mainly for CI builds. Both sphinx generated documentation and doxygen generated documentation have to be pushed on the same branch (`gh-pages`) under the same commit (see `update_gh_pages` function). Therefore splitting them into separate scripts would be a pain (they need to be in the same job anyways).

If you want to check sphinx documentation locally it should be pretty straight forward since the tool is very easy to use. (you can call only `build_sphinx` function  from the script or just type command`sphinx-build <path_to_sources> <output_path>`, for example `sphinx-build no-OS/doc/sphinx/source no-OS/doc/sphinx/build`)

For doxygen builds locally you can call only  `build_doc` function.

Doing the sphinx build in a separate script will imply to source it in this script anyways so the approach is similar to the current implementation with functions inside it).

I will do a patch with renames probably since now the `doxygen.sh` script does more that doxygen, etc. 

But adding separate scripts that would be sourced anyways in the same script doesn't add much value in comparison with splitting stuff in separate functions.

buha (approved):
fair enough

PR #1929, opened 2023-09-28
ci/release:delete 'last_commit' tag only if exists

SRaus:
## Pull Request Description

Add a condition to delete 'last_commit' tag only if it exists, so that the script won't fail if the tag doesn't exists.
Also, set 'Delete Release' task from Azure pipeline to continue on error - so to not fail the whole pipeline if there is no release associated to tag 'last_commit' or if the tag does not exists.

## PR Type

## PR Checklist


rbolboac (commented):
there are some unrelated changes which do not appear in the commit message, either create a new commit which contains those changes or update the commit message

SRaus:
v2: split in 2 commits - one that adds condition for deleting GitHub tag and one for deleting extra spaces.

AAndrisa (approved):

PR #1935, opened 2023-10-04
drivers: meter: ade7913: Add driver

danmois:
Add driver for the ADE7913 3-Channel, Isolated, Sigma Delta ADC with SPI.

## Pull Request Description

The ADE7913 are isolated, 3-channel Σ-Δ ADCs for a variety of applications such as shunt-based polyphase meters, power quality monitoring, solar inverters, process monitoring and protective devices.

## PR Type

## PR Checklist

### DIFF
--- a/drivers/meter/ade7913/ade7913.c
+++ b/drivers/meter/ade7913/ade7913.c
@@ -1,0 +1,2311 @@
+/***************************************************************************//**
+*   @file   ade7913.c
+*   @brief  Implementation of ADE7913 Driver.
+*   @author George Mois (george.mois@analog.com)
+********************************************************************************
+* Copyright 2023(c) Analog Devices, Inc.
+*
+* All rights reserved.
+*
+* Redistribution and use in source and binary forms, with or without
+* modification, are permitted provided that the following conditions are met:
+*  - Redistributions of source code must retain the above copyright
+*    notice, this list of conditions and the following disclaimer.
+*  - Redistributions in binary form must reproduce the above copyright
+*    notice, this list of conditions and the following disclaimer in
+*    the documentation and/or other materials provided with the
+*    distribution.
+*  - Neither the name of Analog Devices, Inc. nor the names of its
+*    contributors may be used to endorse or promote products derived
+*    from this software without specific prior written permission.
+*  - The use of this software may or may not infringe the patent rights
+*    of one or more patent holders.  This license does not release you
+*    from the requirement that you obtain separate licenses from these
+*    patent holders to use this software.
+*  - Use of the software either in source or binary form, must be run
+*    on or directly connected to an Analog Devices Inc. component.
+*
+* THIS SOFTWARE IS PROVIDED BY ANALOG DEVICES "AS IS" AND ANY EXPRESS OR
+* IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, NON-INFRINGEMENT,
+* MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED.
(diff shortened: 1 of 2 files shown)


buha (approved):
i've reviewed this offline, looks good to me

drivers/meter/ade7913/ade7913.c
```diff
@@ -0,0 +1,2308 @@
...
+int ade7913_hw_reset(struct ade7913_dev *dev)
+{
+	int ret;
+	uint8_t ver_product;
+
+	ret = no_os_gpio_set_value(dev->gpio_reset, NO_OS_GPIO_LOW);
+	if (ret)
+		return ret;
+
+	no_os_mdelay(10);
```
rbolboac, line 456: missing comment to explain this delay

drivers/meter/ade7913/ade7913.c
```diff
@@ -0,0 +1,2308 @@
...
+
+	ret = no_os_gpio_set_value(dev->gpio_reset, NO_OS_GPIO_HIGH);
+	if (ret)
+		return ret;
+
+	dev->crc_en = 1;
+
+	/* Wait for device to initialize */ // ToDo - check required timing
```
rbolboac, line 464: what about this todo?

drivers/meter/ade7913/ade7913.c
```diff
@@ -0,0 +1,2308 @@
...
+ * @param dev - The device structure.
+ */
+static void ade7913_irq_handler(void *dev)
+{
+	struct ade7913_dev *desc = dev;
+	uint8_t reg_val;
+	int ret;
+
+	if (!dev)
+		return;
```
rbolboac, line 226: is this check needed?

drivers/meter/ade7913/ade7913.c
```diff
@@ -0,0 +1,2308 @@
...
+ */
+int ade7913_remove(struct ade7913_dev *dev)
+{
+	int ret;
+
+	ret = no_os_spi_remove(dev->spi_desc);
+	if (ret)
+		return ret;
+
+	no_os_free(dev);
```
rbolboac, line 407: I think there are some memory leaks here. no_os_gpio_get_optional and no_os_irq_register_callback allocate memory which should be freed

drivers/meter/ade7913/ade7913.c
```diff
@@ -0,0 +1,2308 @@
...
+ * @return 0 in case of success, negative error code otherwise.
+ */
+int ade7913_get_stream_dbg_mode(struct ade7913_dev *dev,
+				enum ade7913_stream_debug_e *stream_dbg)
+{
+	int ret;
+	uint8_t reg_val;
+
+	if (!dev)
+		return	-ENODEV;
```
rbolboac, line 527: not sure if this check is needed, I would actually remove from all APIs which call in the end SPI read API because the SPI read API already checks for this. it would be better to check if stream_dbg is null

rbolboac (commented):
some remarks from my side

amiclaus (approved):

drivers/meter/ade7913/ade7913.c
danmois: No, I will remove it.

drivers/meter/ade7913/ade7913.c
danmois: Check removed from the functions that call the SPI read API.

drivers/meter/ade7913/ade7913.c
danmois: Fixed.

danmois:
V2

- remove unnecessary checks
- fix memory leaks in remove function
- remove check for dev = NULL from the functions that call the SPI read API

rbolboac (approved):

amiclaus (approved):

PR #1959, opened 2023-10-17
Call no_os_init() during board init

CiprianRegus:
## Pull Request Description

no_os_init() is supposed to be called for any project running on Maxim. The Board_Init() function is called in the SDK before main(), so we can also do no_os_init().

## PR Type

## PR Checklist


danmois (approved):

buha:
so this is good, it will work, but it will not be consistent with other platforms unless we find a way to do this automatically before main() on all platforms

if that's what the team decides, then we should merge and work incrementally on the other platforms to add the feature

rbolboac (approved):

amiclaus (approved):

PR #1978, opened 2023-10-25
ci: use all cpu cores

amiclaus:
## Pull Request Description

Add NUM_JOBS variable holding the available number of cpu cores on a build agent.

Use all avaliable cores for all ci builds.

## PR Type

## PR Checklist

### DIFF
--- a/ci/run_build.sh
+++ b/ci/run_build.sh
@@ -39,6 +39,7 @@ sudo apt-get update
 
 TOP_DIR="$(pwd)"
 DEPS_DIR="${TOP_DIR}/deps"
+NUM_JOBS=$(nproc)
 
 . ./ci/lib.sh
 
@@ -53,7 +54,7 @@ build_cppcheck() {
 build_drivers() {
     sudo apt-get install gcc-arm-none-eabi libnewlib-arm-none-eabi
     git submodule update --init libraries/iio/libtinyiiod
-    make -C ./drivers -f Makefile -j
+    make -j${NUM_JOBS} -C ./drivers -f Makefile
 }
 
 build_documentation() {
--- a/ci/astyle.sh
+++ b/ci/astyle.sh
@@ -96,7 +96,7 @@ build_astyle() {
 		wget --no-check-certificate "https://sourceforge.net/projects/astyle/files/astyle/astyle 3.1/astyle_3.1_linux.tar.gz"
 		tar -xzf astyle_3.1_linux.tar.gz
 		pushd ./astyle/build/gcc
-		make -j3
+		make -j${NUM_JOBS}
 		popd
 		popd
 	fi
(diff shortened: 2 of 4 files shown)


ci/run_build.sh
```diff
@@ -39,6 +39,7 @@ sudo apt-get update
 
 TOP_DIR="$(pwd)"
 DEPS_DIR="${TOP_DIR}/deps"
+NUM_JOBS=$(nproc)
```
rbolboac, line 42: just wondering if this won't freeze the system in some cases

SRaus (changes requested):
I don't think it's ok to use all cores for building projects. 
The No-OS agent is also used by small jobs from other Azure pipelines where internal agents are required (like uploading RPI boot files on Artifactory server etc). But even if would dedicated only for No-OS, there is required at least one core for pipeline and other background processes.
So please change "make -j${NUM_JOBS -1}" or  "NUM_JOBS=$(nproc -1)".

amiclaus:
> I don't think it's ok to use all cores for building projects. The No-OS agent is also used by small jobs from other Azure pipelines where internal agents are required (like uploading RPI boot files on Artifactory server etc). But even if would dedicated only for No-OS, there is required at least one core for pipeline and other background processes. So please change "make -j${NUM_JOBS -1}" or "NUM_JOBS=$(nproc -1)".

This has nothing to do with the no-OS agent.

If you take a look in the code, these scripts are run only on azure VMs which are handled **strictly** by https://github.com/analogdevicesinc/no-OS/blob/master/azure-pipelines.yml

ci/run_build.sh
amiclaus: Hmm, you mean calling the `nproc` command? or using all cores for azure VMs which btw were already forced in some cases with hardcoded values (see for example: https://github.com/analogdevicesinc/no-OS/blob/bc77edd35f9f42a9fb5289e21994dca8fdad7bbe/ci/astyle.sh#L99) ?

SRaus:
> > I don't think it's ok to use all cores for building projects. The No-OS agent is also used by small jobs from other Azure pipelines where internal agents are required (like uploading RPI boot files on Artifactory server etc). But even if would dedicated only for No-OS, there is required at least one core for pipeline and other background processes. So please change "make -j${NUM_JOBS -1}" or "NUM_JOBS=$(nproc -1)".
> 
> This has nothing to do with the no-OS agent.
> 
> If you take a look in the code, these scripts are run only on azure VMs which are handled **strictly** by https://github.com/analogdevicesinc/no-OS/blob/master/azure-pipelines.yml

Even so, aren't there any background processes - on "ubuntu-latest" container - for which we'll need to keep one CPU core?

ci/run_build.sh
rbolboac: I mean using all cores. but if we already do it in some cases, I guess there is no issue

SRaus (approved):

rbolboac (approved):

PR #1981, opened 2023-10-25
Platform SPI delays

CiprianRegus:
## Pull Request Description

## PR Type

## PR Checklist

### DIFF
--- a/drivers/platform/maxim/max32655/maxim_spi.c
+++ b/drivers/platform/maxim/max32655/maxim_spi.c
@@ -78,9 +78,15 @@ static void _max_delay_config(struct no_os_spi_desc *desc,
 	mxc_spi_regs_t *spi;
 	uint32_t clk_rate;
 	uint32_t ticks_ns;
+	uint32_t delay_first_ns;
+	uint32_t delay_last_ns;
 
-	if (msg->cs_delay_first == st->cs_delay_first &&
-	    msg->cs_delay_last == st->cs_delay_last)
+	delay_first_ns = msg->cs_delay_first * 1000 +
+			 desc->platform_delays.cs_delay_first;
+	delay_last_ns = msg->cs_delay_last * 1000 + desc->platform_delays.cs_delay_last;
+
+	if (delay_first_ns == st->cs_delay_first &&
+	    delay_last_ns == st->cs_delay_last)
 		return;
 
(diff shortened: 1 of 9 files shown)


buha:
I don't understand this PR, please explain in PR description what are you trying to achieve here.

There are already CS delays in the SPI transfer structure here: https://github.com/analogdevicesinc/no-OS/blob/5f046149967df706d1fba3587a4b0b66a351193a/include/no_os_spi.h#L106

Why move some of that functionality to the SPI peripheral descriptor structure when these delays are slave specific ?

drivers/platform/maxim/max32655/maxim_spi.c
```diff
@@ -78,25 +78,31 @@ static void _max_delay_config(struct no_os_spi_desc *desc,
 	mxc_spi_regs_t *spi;
 	uint32_t clk_rate;
 	uint32_t ticks_ns;
+	uint32_t delay_first_ns;
+	uint32_t delay_last_ns;
 
-	if (msg->cs_delay_first == st->cs_delay_first &&
-	    msg->cs_delay_last == st->cs_delay_last)
+	delay_first_ns = msg->cs_delay_first * 1000 +
```
buha, line 84: You seem to be using scenario A, there is also scenario B. Not sure which one is best.
B would result in less and simpler code.
```
                                 ┌──────────────────┐
                                 │    user          │
                                 └───────┬──────────┘
                                         │
                                         │
                                         │ use driver to get some data from ad1234
                                         │
                                         │
                                         │
                                         │
                                ┌────────▼───────────┐
                                │ ad1234 driver      │
                                └────────┬───────────┘
                                         │
                                         │
                                         │configure delays as in datasheet of ad1234
                                         │
                                         │
                                         │
                                 ┌───────▼───────────┐
                                 │ no_os_spi         │
                                 └───────┬───────────┘
                       A                 │               B
                                         │
                  add the above delays   │either use the above delays
to the board specific delays that may or │or override above delays with board specific delays
may not have been provided by the user   │if they were provided by the user at initialization
at initialization                        │
                                 ┌───────▼───────────┐
                                 │spi peripheral     │
                                 └───────────────────┘
```

buha (approved):
so Cipri wants to use scenario A, i'm fine with that

PR #1985, opened 2023-10-27
iio_axi_adc: Allow custom scan type

danmois:
## Pull Request Description

Allow the setting of custom scan type for the AXI ADC channels.

Until now, only a fixed setting of 16 realbists and storagebits was possible.

## PR Type

## PR Checklist


rbolboac (approved):
lgtm

rbolboac (commented):
maybe it's worth to consider if it makes sense to do this for iio_axi_dac as well

amiclaus (approved):

danmois:
> maybe it's worth to consider if it makes sense to do this for iio_axi_dac as well

Good point. However, maybe it would be better to find the specific case in which this applies, so that we are able to test it with actual hardware.

PR #1989, opened 2023-10-31
driver: ad7616: Add support for generic no-os SPI

Skallwar:
## Pull Request Description

This is preliminary work to get AD7616 working on SDP-K1 using STM32 hal. You can find the whole branch with a new project here, the hope is to get it merged too at some point (still some cleanup to do).

This PR is in draft state until I manage to test if `ad7616-sdz` project is still working.

## PR Type

## PR Checklist
- [x] I have build all projects affected by the changes in this PR (this is why it's in a draft state)
- [ ] I have tested in hardware affected projects, at the relevant boards (this is why it's in a draft state)

### DIFF
--- a/drivers/adc/ad7616/ad7616.c
+++ b/drivers/adc/ad7616/ad7616.c
@@ -3,8 +3,10 @@
  *   @brief  Implementation of AD7616 Driver.
  *   @author DBogdan (dragos.bogdan@analog.com)
  *   @author Antoniu Miclaus (antoniu.miclaus@analog.com)
+ *   @author Esteban Blanc (eblanc@baylibre.com)
 ********************************************************************************
  * Copyright 2020(c) Analog Devices, Inc.
+ * Copyright 2023(c) BayLibre, SAS.
  *
  * All rights reserved.
  *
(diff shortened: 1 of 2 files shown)


drivers/adc/ad7616/ad7616.h
```diff
@@ -235,7 +264,7 @@ int32_t ad7616_set_oversampling_ratio(struct ad7616_dev *dev,
 				      enum ad7616_osr osr);
 /* Read data in serial mode. */
 int32_t ad7616_read_data_serial(struct ad7616_dev *dev,
-				uint32_t *buf,
+				struct ad7616_conversion_result *results,
```
buha, line 267: When changing API, it's also expected to edit the existing project(s) that use the API, maybe as follow-up commits, so that we don't break what already works:
https://github.com/analogdevicesinc/no-OS/blob/f96a73637bbd994ffbe7281720055a619291afbf/projects/ad7616-sdz/src/ad7616_sdz.c#L144

drivers/adc/ad7616/ad7616.c
```diff
@@ -38,30 +39,301 @@
...
+ * @brief Compute the CRC for one channel. See CRC chapter in the datasheet
+ * @param data - Data to check
+ * @param crc - Initial crc
+ * @return The computed CRC
+ */
+static uint8_t crc_channel(uint16_t data, uint8_t crc)
+{
+	uint8_t crc_out;
+
+	uint8_t data_0 = GET_BIT(data, 0);
```
buha, line 80: no_os_test_bit from no_os_util.h is also an option here (although maybe slower)

drivers/adc/ad7616/ad7616.c
```diff
@@ -38,30 +39,301 @@
...
+	}
+
+	// Check CRC
+	if (dev->crc) {
+		ret = no_os_spi_write_and_read(dev->spi_desc, (void *)&tmp,
+					       sizeof(tmp));
+		if (ret != 0)
+			return ret;
+
+		tmp = __builtin_bswap16(tmp);
```
buha, line 278: We also have no_os_get_unaligned_le16 or no_os_get_unaligned_be16 as more portable alternatives than this gcc builtin.

drivers/adc/ad7616/ad7616.c
```diff
@@ -506,6 +830,95 @@ int32_t ad7616_core_setup(struct ad7616_dev *dev)
...
+		if (ret != 0)
+			return ret;
+	}
+
+	if (burst)
+		ret = ad7616_write_mask(dev, AD7616_REG_CONFIG, AD7616_BURSTEN,
+					AD7616_BURSTEN);
+
+	// Enable channel sequencer
+	ret |= ad7616_write_mask(dev, AD7616_REG_CONFIG, AD7616_SEQEN,
```
buha, line 867: lets avoid doing `red |=` as it really carries no meaning

buha: *comment applies throughout the whole file

drivers/adc/ad7616/ad7616.h
Skallwar: Yes, that's why this is still in draft. I did not do the change yet as I was not sure about how this change would be received. I found this better for the end user than "just" a buffer.
I had a discussion with @mike-bradley about maybe doing this. If you found it useless, it's trivial to revert. Let me know what you think about it

drivers/adc/ad7616/ad7616.c
Skallwar:
```suggestion
	ret = ret || ad7616_write_mask(dev, AD7616_REG_CONFIG, AD7616_SEQEN,
```

drivers/adc/ad7616/ad7616.c
Skallwar: [Here](https://godbolt.org/#g:!((g:!((g:!((h:codeEditor,i:(filename:'1',fontScale:14,fontUsePx:'0',j:1,lang:___c,selection:(endColumn:5,endLineNumber:17,positionColumn:5,positionLineNumber:17,selectionStartColumn:5,selectionStartLineNumber:17,startColumn:5,startLineNumber:17),source:'int+no_os_test_bit(int+pos,+const+volatile+void+*+addr)%0A%7B%0A%09return+(((const+int+*)addr)%5Bpos+/+32%5D+%3E%3E+pos)+%26+1UL%3B%0A%7D%0Aint+get_bit(int+pos,+const+volatile+unsigned+int+*+val)+%7B%0A++++return+(*val+%3E%3E+pos)+%26+0x1%3B%0A%7D%0A/*%0Aint+main+(void)+%7B%0A++++int+a+%3D+1+%2B+1%3B%0A++++int+b+%3D+1+%2B+1%3B%0A%0A++++a+%3D+no_os_test_bit(1,+%26a)%3B%0A++++b+%3D+get_bit(1,+%26b)%3B%0A%0A++++return+0%3B%0A%7D+*/'),l:'5',n:'0',o:'C+source+%231',t:'0')),k:50,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((g:!((h:compiler,i:(compiler:carmug1320,deviceViewOpen:'1',filters:(b:'0',binary:'1',binaryObject:'1',commentOnly:'0',debugCalls:'1',demangle:'0',directives:'0',execute:'0',intel:'0',libraryCode:'0',trim:'1'),flagsViewOpen:'1',fontScale:14,fontUsePx:'0',j:1,lang:___c,libs:!(),options:'-O2',overrides:!(),selection:(endColumn:1,endLineNumber:1,positionColumn:1,positionLineNumber:1,selectionStartColumn:1,selectionStartLineNumber:1,startColumn:1,startLineNumber:1),source:1),l:'5',n:'0',o:'+ARM+GCC+13.2.0+(unknown-eabi)+(Editor+%231)',t:'0')),k:50,l:'4',m:50,n:'0',o:'',s:0,t:'0'),(g:!((h:output,i:(editorid:1,fontScale:14,fontUsePx:'0',j:1,wrap:'1'),l:'5',n:'0',o:'Output+of+ARM+GCC+13.2.0+(unknown-eabi)+(Compiler+%231)',t:'0')),header:(),l:'4',m:50,n:'0',o:'',s:0,t:'0')),k:50,l:'3',n:'0',o:'',t:'0')),l:'2',n:'0',o:'',t:'0')),version:4) is a comparison. I would argue that we should keep my macro. Let me know what you to think

drivers/adc/ad7616/ad7616.c
Skallwar: I think the best way to deal with theses is with early returns

drivers/adc/ad7616/ad7616.c
buha: it's quite a difference! let's keep your macro then :+1:

drivers/adc/ad7616/ad7616.h
buha: I'm totally fine with changing it, it seems to add value. Just add a follow-up commit that edits the existing project (projects/ad7616-sdz) to compile with the new ad7616_read_data_serial() signature.

drivers/adc/ad7616/ad7616.c
```diff
@@ -416,13 +647,100 @@ int32_t ad7616_read_data_serial(struct ad7616_dev *dev,
...
+					   &crc);
+		if (ret)
+			return ret;
+
+		if (dev->crc) {
+			ret = check_crc(results + (i * dev->layers_nb),
+					dev->layers_nb, crc);
+			if (ret) {
+				return ret;
+			}
```
rbolboac, line 695: it looks like these lines can be removed and simply return ret instead of 0 below

drivers/adc/ad7616/ad7616.c
```diff
@@ -416,13 +647,100 @@ int32_t ad7616_read_data_serial(struct ad7616_dev *dev,
...
+	for (uint32_t i = 0; i < samples; i++) {
+		ret = ad7616_toggle_conv(dev);
+		if (ret != 0)
+			return ret;
+
+		uint8_t crc;
+		ret = ad7616_read_channels(dev, results + (i * dev->layers_nb),
+					   &crc);
+		if (ret)
+			return ret;
```
rbolboac, line 688: so this driver being an older driver uses if (ret !=0) syntax, I would use the same syntax everywhere or use if (ret) syntax in the whole driver.

drivers/adc/ad7616/ad7616.c
```diff
@@ -416,13 +647,100 @@ int32_t ad7616_read_data_serial(struct ad7616_dev *dev,
...
+{
+#ifdef PLATFORM_ZYNQMP
+	uint32_t read;
+
+	no_os_axi_io_write(dev->core_baseaddr, AD7616_REG_UP_WRITE_DATA,
+			   0x0000 | ((reg_addr & 0x3F) << 9));
+	no_os_udelay(50);
+	no_os_axi_io_read(dev->core_baseaddr, AD7616_REG_UP_READ_DATA, &read);
+	*reg_data = read & 0xFF;
+	no_os_mdelay(1);
```
rbolboac, line 719: hmmm, there is no return in this case. is the expected behavior to return -ENOSYS also for this platform or is there a missing return 0? see also in par_write API

rbolboac (commented):
just a few comments from my side

drivers/adc/ad7616/ad7616.c
Skallwar: Well spotted that, that's a mistake

drivers/adc/ad7616/ad7616.c
Skallwar: I don't think so. This is in a loop so if the CRC check failed at the first iteration, then you would still have x amount of iteration remaining and the user will not know that at least one values they've got is bad

rbolboac: Oops, right, missed the loop. Anyhow, I think the braces for if (ret) can be removed.

drivers/adc/ad7616/ad7616.h
buha: so can you please also modify the `projects/ad7616-sdz` to properly use the new ad7616_read_data_serial ?

Skallwar: Yes, I'm working on it. I have to build the setup for this project to test the changes. I will remove the draft state once I've tested the `projects/ad7616-sdz`

drivers/adc/ad7616/ad7616.c
```diff
@@ -38,20 +40,280 @@
...
+#ifdef PLATFORM_ZYNQMP
+#include "spi_engine.h"
+#include "axi_dmac.h"
+#include "no_os_axi_io.h"
+#else
+#include "no_os_spi.h"
+#endif
+
+#define GET_BIT(v, b) ((v >> b) & 1)
+#define TOGGLE_TIMEOUT_DELAY (8192 * 1024)
```
CiprianRegus, line 70: I would name these using the `AD7616_` prefix, since there might be a similar `GET_BIT` macro defined in the platform SDKs that we're using.

Skallwar: Good idea, fixed

drivers/adc/ad7616/ad7616.c
```diff
@@ -416,13 +666,103 @@ int32_t ad7616_read_data_serial(struct ad7616_dev *dev,
...
+ * PAR read from device.
+ * @param dev - The device structure.
+ * @param reg_addr - The register address.
+ * @param reg_data - The register data.
+ * @return 0 in case of success, negative error code otherwise.
+ */
+int32_t ad7616_par_read(struct ad7616_dev *dev, uint8_t reg_addr,
+			uint16_t *reg_data)
+{
+#ifdef PLATFORM_ZYNQMP
```
buha, line 729: We are trying to eliminate this kind of platform specific #ifdefs from driver code which should not know or care what platform it's running on. It should probably be the job of no_os_axi to return -ENOSYS.

We will address this in the future. For now it's ok.

buha (approved):

buha:
@Skallwar should this still be a draft ?

Skallwar:
> @Skallwar should this still be a draft ?

Yes. I'm currently checking `ad7616-sdz` project but I have trouble to get data out of ad7616, even on master branch.

PopPaul2021:
We tested with the previous version of the HDL design and the current implementation on the no-OS master branch and the data is correct only for half of the samples stored in the buffer.

drivers/adc/ad7616/ad7616.h
acostina: Given that the default SPI Engine based project on the main branch is not stable , I think testing in hardware is not necessary before merging generic SPI support. We're in the process of debugging the SPI Engine project

Skallwar:
Following @acostina [comment](https://github.com/analogdevicesinc/no-OS/pull/1989#discussion_r1407738436), I'm removing the draft status on this PR

rbolboac (changes requested):
there are still some things which should be worked on before merging (not functional).
also, please make sure to add a commit message as well.

drivers/adc/ad7616/ad7616.c
```diff
@@ -60,8 +322,7 @@
  * @param reg_data - The register data.
  * @return 0 in case of success, negative error code otherwise.
  */
-int32_t ad7616_read(struct ad7616_dev *dev,
-		    uint8_t reg_addr,
+int32_t ad7616_read(struct ad7616_dev *dev, uint8_t reg_addr,
```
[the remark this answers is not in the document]
Skallwar: I ran `astyle` for the CI to pass, I will move this to a new commit

Skallwar:
> there are still some things which should be worked on before merging (not functional)

What are you referring to?

drivers/adc/ad7616/ad7616.c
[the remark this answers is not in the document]
Skallwar: Ah it was clang-format, my bad

rbolboac:
> > there are still some things which should be worked on before merging (not functional)
> 
> What are you referring to?

I was referring to the other comments which were shown below (which I see you are already working on) and to the missing commit message.

drivers/adc/ad7616/ad7616.c
```diff
@@ -506,6 +844,99 @@ int32_t ad7616_core_setup(struct ad7616_dev *dev)
...
+
+	// Dummy CONV pulse as asked in datasheet
+	return ad7616_toggle_conv(dev);
+}
+
+int32_t ad7616_disable_sequencer(struct ad7616_dev *dev)
+{
+	dev->layers_nb = 1;
+	return ad7616_write_mask(dev, AD7616_REG_CONFIG,
+				 AD7616_SEQEN | AD7616_BURSTEN, 0);
```
rbolboac, line 899: it is kind of strange that for enabling you are using AD7616_SEQEN and for disabling it is simply 0. it is also kind of strange that you are using AD7616_SEQEN for mask and enable bit as well.
I would suggest to make things a little bit more clear for the reviewer by creating two different macros:
AD7616_SEQEN_MASK for the mask which can remain (1 << 5)
and AD7616_SEQEN(x) which can be ((x & 1)<< 5)
and then use it something like this:
```
ad7616_write_mask(dev, AD7616_REG_CONFIG, AD7616_SEQEN_MASK, AD7616_SEQEN(1));
```
similarly for AD7616_BURSTEN

drivers/adc/ad7616/ad7616.c
```diff
@@ -369,24 +604,33 @@ int32_t ad7616_set_oversampling_ratio(struct ad7616_dev *dev,
...
+		if (ret != 0)
+			return ret;
+
+		ret = no_os_gpio_set_value(dev->gpio_os1, ((osr & 0x02) >> 1));
+		if (ret != 0)
+			return ret;
+
+		ret = no_os_gpio_set_value(dev->gpio_os2, ((osr & 0x04) >> 2));
+		if (ret != 0)
+			return ret;
```
rbolboac, line 616: this change is not related to the commit message. 
either you leave it as it was, or you create a new commit in which you replace ret |= with a separate call each time it is used (currently there is another instance in which this logic is used but is kept as it is, see in ad7616_set_range). 
see all instances in which you performed this change: e.g. in ad7616_set_mode

drivers/adc/ad7616/ad7616.c
```diff
@@ -38,20 +40,280 @@
...
+ * @param ch_b - Pointer to store the current source for channel B
+ * @return 0 on success, an error code otherwise
+ */
+int32_t ad7616_read_channel_source(struct ad7616_dev *dev, enum ad7616_ch *ch_a,
+				   enum ad7616_ch *ch_b)
+{
+	uint16_t ret;
+	uint16_t val;
+
+	ret = ad7616_read_mask(dev, AD7616_REG_CHANNEL, 0xFF, &val);
```
rbolboac, line 308: instead of 0xFF you may create a macro AD7616_CHANNELS_MASK

drivers/adc/ad7616/ad7616.c
```diff
@@ -38,20 +40,280 @@
...
+/**
+ * @brief Enable CRC
+ * @param dev - The device structure.
+ * @return 0 on success, an error code otherwise
+ */
+static int32_t ad7616_enable_crc(struct ad7616_dev *dev)
+{
+	int32_t ret = 0;
+
+	ret = ad7616_write_mask(dev, AD7616_REG_CONFIG, 0b11, 0b11);
```
rbolboac, line 184: please use macros instead, similarly to what you are doing for the other configurations

rbolboac (commented):
I did not find any functional issues, however I think there are ways in which we can make the code easier to follow and the commits more compact (functional changes separated from styling changes).
If it is urgent to merge then it can be merged, otherwise I think it would be better to take into consideration these lasts remarks.

drivers/adc/ad7616/ad7616.c
```diff
@@ -595,8 +1041,15 @@ int32_t ad7616_setup(struct ad7616_dev **device,
 
 	*device = dev;
 
-	if (!ret)
-		printf("AD7616 successfully initialized\n");
```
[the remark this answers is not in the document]
Skallwar: It's because the UART interface is not enabled at this point for stm32 platform (it's initialized in iio_app). I could enable UART before in the new stm32 project #2000 but I feel like a print do not belong in a driver and it's redundant with the return value.

Let me know what you think

amiclaus (approved):

PR #2000, opened 2023-11-08
Add STM32 project for AD7616

Skallwar:
## Pull Request Description

Add a project to use AD7616 on SDP-K1 using the STM32 platform.

This depends on #1989 so this will be kept in draft until then

## PR Type

## PR Checklist

## `iio_info`
```
iio_info version: 0.25 (git tag:b6028fde)
Libiio version: 0.25 (git tag: b6028fd) backends: local xml ip usb serial
IIO context created with serial backend.
Backend version: 1.1 (git tag: 0000000)
Backend description string: no-OS/projects/ad7616-st eblanc/ad7616_stm32_project-d8a472b454db-modified
IIO context has 3 attributes:
	uri: serial:/dev/ttyACM2,230400,8n1n
	serial,port: /dev/ttyACM2
	serial,description: DAPLink CMSIS-DAP - 0604000062024e450029500434270048c001000197969900
IIO context has 1 devices:
	iio:device0: ad7616 (buffer capable)
		16 channels found:
			voltage0: voltage0 (input, index: 0, format: le:S16/16>>0)
			2 channel-specific attributes found:
				attr  0: raw value: 13639
				attr  1: scale value: 10
			voltage1: voltage1 (input, index: 1, format: le:S16/16>>0)
			2 channel-specific attributes found:
				attr  0: raw value: 6541
				attr  1: scale value: 10
			voltage2: voltage2 (input, index: 2, format: le:S16/16>>0)
			2 channel-specific attributes found:
				attr  0: raw value: 6566
				attr  1: scale value: 10
			voltage3: voltage3 (input, index: 3, format: le:S16/16>>0)
			2 channel-specific attributes found:
				attr  0: raw value: 6559
				attr  1: scale value: 10
			voltage4: voltage4 (input, index: 4, format: le:S16/16>>0)
			2 channel-specific attributes found:
				attr  0: raw value: 6562
				attr  1: scale value: 10
			voltage5: voltage5 (input, index: 5, format: le:S16/16>>0)
			2 channel-specific attributes found:
				attr  0: raw value: 6552
				attr  1: scale value: 10
			voltage6: voltage6 (input, index: 6, format: le:S16/16>>0)
			2 channel-specific attributes found:
				attr  0: raw value: 6553
				attr  1: scale value: 10
			voltage7: voltage7 (input, index: 7, format: le:S16/16>>0)
			2 channel-specific attributes found:
				attr  0: raw value: 6549
				attr  1: scale value: 10
			voltage8: voltage8 (input, index: 8, format: le:S16/16>>0)
			2 channel-specific attributes found:
				attr  0: raw value: 57896
				attr  1: scale value: 10
			voltage9: voltage9 (input, index: 9, format: le:S16/16>>0)
			2 channel-specific attributes found:
				attr  0: raw value: 6585
				attr  1: scale value: 10
			voltage10: voltage10 (input, index: 10, format: le:S16/16>>0)
			2 channel-specific attributes found:
				attr  0: raw value: 6594
				attr  1: scale value: 10
			voltage11: voltage11 (input, index: 11, format: le:S16/16>>0)
			2 channel-specific attributes found:
				attr  0: raw value: 6587
				attr  1: scale value: 10
			voltage12: voltage12 (input, index: 12, format: le:S16/16>>0)
			2 channel-specific attributes found:
				attr  0: raw value: 6572
				attr  1: scale value: 10
			voltage13: voltage13 (input, index: 13, format: le:S16/16>>0)
			2 channel-specific attributes found:
				attr  0: raw value: 6569
				attr  1: scale value: 10
			voltage14: voltage14 (input, index: 14, format: le:S16/16>>0)
			2 channel-specific attributes found:
				attr  0: raw value: 6578
				attr  1: scale value: 10
			voltage15: voltage15 (input, index: 15, format: le:S16/16>>0)
			2 channel-specific attributes found:
				attr  0: raw value: 6582
				attr  1: scale value: 10
		1 device-specific attributes found:
				attr  0: crc value: 0
		1 debug attributes found:
				debug attr  0: direct_reg_access value: 449
		No trigger on this device
```


drivers/adc/ad7616/iio_ad7616.h
```diff
@@ -0,0 +1,37 @@
+/***************************************************************************//**
+* @file   iio_ad7616.h
+* @brief  Header file for AD7616 IIO interface
+********************************************************************************
+* Copyright (c) 2023 Analog Devices, Inc.
+* Copyright (c) 2023 BayLibre, SAS.
+* All rights reserved.
+*
+* This software is proprietary to Analog Devices, Inc. and its licensors.
```
buha, line 9: Not sure about this copyright and license text, I have zero legal knowledge. I think the ADI SLA should be replicated in full here like in other files.

drivers/adc/ad7616/iio_ad7616.c
```diff
@@ -0,0 +1,293 @@
...
+			       const struct iio_ch_info *channel,
+			       intptr_t priv)
+{
+	struct ad7616_dev *drv_dev = device;
+
+	switch (priv) {
+	case CRC_ATTR_ID:
+		return sprintf(buf, "%u", drv_dev->crc);
+	default:
+		return -1;
```
buha, line 62: -EINVAL

drivers/adc/ad7616/iio_ad7616.c
```diff
@@ -0,0 +1,293 @@
...
+	}
+}
+
+static int iio_ad7616_attr_set(void *device, char *buf, uint32_t len,
+			       const struct iio_ch_info *channel,
+			       intptr_t priv)
+{
+	switch (priv) {
+	case CRC_ATTR_ID:
```
buha, line 71: so looks like this is a draft, could you please show an output of iio_info, to see how this device looks from libiio point of view ?

drivers/adc/ad7616/iio_ad7616.c
```diff
@@ -0,0 +1,293 @@
...
+		   intptr_t id)
+{
+	int ret;
+	uint8_t index = channel->ch_num < 8 ? channel->ch_num : channel->ch_num - 8;
+	struct ad7616_conversion_result res[8];
+	struct ad7616_dev *dev = device;
+
+	// TODO: we call this on every call, fetching all 16 inputs. Doesn't look smart
+	ret = ad7616_read_data_serial(dev, res, 1);
+	if (ret != 0)
```
amiclaus, line 121: `if (ret)` is preferred.

drivers/adc/ad7616/iio_ad7616.c
```diff
@@ -0,0 +1,293 @@
...
+	switch (range) {
+	case AD7616_10V:
+		return sprintf(buf, "10");
+	case AD7616_5V:
+		return sprintf(buf, "5");
+	case AD7616_2V5:
+		return sprintf(buf, "2.5");
+	}
+
+	return -EINVAL;
```
amiclaus, line 161: move in switch..case as default statement.

drivers/adc/ad7616/iio_ad7616.c
```diff
@@ -0,0 +1,293 @@
...
+ * @param	iio_dev_data[in] - Pointer to IIO device data structure
+ * @return	0 in case of success, negative error code otherwise
+ */
+static int32_t iio_ad7616_submit_buffer(struct iio_device_data *iio_dev_data)
+{
+	uint16_t ret;
+	struct ad7616_conversion_result results[8];
+	uint32_t bytes_per_scan = iio_dev_data->buffer->bytes_per_scan;
+	uint16_t data[2];
+	for (uint32_t i = 0; i < iio_dev_data->buffer->samples; i++) {
```
amiclaus, line 209: variable declaration at start of the function, not inline.

drivers/adc/ad7616/iio_ad7616.c
```diff
@@ -0,0 +1,293 @@
...
+ * @brief	Read buffer data corresponding to AD7616 IIO device
+ * @param	iio_dev_data[in] - Pointer to IIO device data structure
+ * @return	0 in case of success, negative error code otherwise
+ */
+static int32_t iio_ad7616_submit_buffer(struct iio_device_data *iio_dev_data)
+{
+	uint16_t ret;
+	struct ad7616_conversion_result results[8];
+	uint32_t bytes_per_scan = iio_dev_data->buffer->bytes_per_scan;
+	uint16_t data[2];
```
amiclaus, line 208: empty line after variable declaration.

drivers/adc/ad7616/iio_ad7616.c
```diff
@@ -0,0 +1,293 @@
...
+* All rights reserved.
+*
+* This software is proprietary to Analog Devices, Inc. and its licensors.
+* By using this software you agree to the terms of the associated
+* Analog Devices Software License Agreement.
+*******************************************************************************/
+#ifdef IIO_SUPPORT
+
+#include <stdio.h>
+#include <errno.h>
```
amiclaus, line 16: use "no_os_error.h" over errno.h

drivers/adc/ad7616/iio_ad7616.h
```diff
@@ -0,0 +1,37 @@
...
+/******************************************************************************/
+/***************************** Include Files **********************************/
+/******************************************************************************/
+#include <stdint.h>
+
+#include "iio.h"
+#include "iio_types.h"
+
+/******************************************************************************/
+/********************** Macros and Constants Definition ***********************/
```
amiclaus, line 25: drop this header comment. it's redundant.

projects/ad7616-st/src/app.c
```diff
@@ -0,0 +1,134 @@
...
+		printf("AD7616 failed to instantiate\r\n");
+		return 0;
+	}
+
+	printf("\r\nStarting iio fw app\r\n");
+	ret = iio_ad7616_initialize(ad7616_dev, &iio_desc);
+	if (ret != 0)
+		printf("IIO Initialization Failure!\r\n");
+
+	while (1) {
```
amiclaus, line 129: drop brackets.

projects/ad7616-st/src/app.c
```diff
@@ -0,0 +1,134 @@
...
+
+	ret = ad7616_setup(&ad7616_dev, &ad7616_ip);
+	if (ret) {
+		printf("AD7616 failed to instantiate\r\n");
+		return 0;
+	}
+
+	printf("\r\nStarting iio fw app\r\n");
+	ret = iio_ad7616_initialize(ad7616_dev, &iio_desc);
+	if (ret != 0)
```
amiclaus, line 126: `if (ret)` check occurrence throughout the entire code.

projects/ad7616-st/src/app.c
```diff
@@ -0,0 +1,134 @@
...
+{
+	struct ad7616_dev *ad7616_dev;
+	struct iio_desc *iio_desc;
+	int ret;
+
+	// ad7616_ip.crc = 1; // If you want to use CRC, uncomment this
+
+	ret = ad7616_setup(&ad7616_dev, &ad7616_ip);
+	if (ret) {
+		printf("AD7616 failed to instantiate\r\n");
```
amiclaus, line 120: use `https://github.com/analogdevicesinc/no-OS/blob/master/include/no_os_print_log.h` -> `pr_err` / `pr_info` depending on the case.

projects/ad7616-st/src/app.c
```diff
@@ -0,0 +1,134 @@
...
+
+	ad7616_setup_sequencer(dev, layers, NO_OS_ARRAY_SIZE(layers), 1);
+
+	struct ad7616_conversion_result results[8] = {0};
+	ret = ad7616_read_data_serial(dev, results, 1);
+	if (ret)
+		return ret;
+
+	printf("Sequencer + burst\r\n");
+	for (uint32_t i = 0; i < 8; i++) {
```
amiclaus, line 102: variable declaration at start of the function.

projects/ad7616-st/src/app.c
```diff
@@ -0,0 +1,134 @@
...
+		{ AD7616_VA5, AD7616_VB5 },
+		{ AD7616_VA6, AD7616_VB6 },
+		{ AD7616_VA7, AD7616_VB7 },
+	};
+
+	/* IIO interface init parameters */
+	static struct iio_init_param iio_init_params = {
+		.phy_type = USE_UART,
+		.nb_devs = 1,
+		// .uart_desc = uart_desc, // Not a constant...
```
amiclaus, line 88: drop these comments if not actually needed.

projects/ad7616-st/src/app.c
```diff
@@ -0,0 +1,134 @@
...
+
+	struct ad7616_conversion_result results[8] = {0};
+	ret = ad7616_read_data_serial(dev, results, 1);
+	if (ret)
+		return ret;
+
+	printf("Sequencer + burst\r\n");
+	for (uint32_t i = 0; i < 8; i++) {
+		printf("Channel A[%lu] = %x\r\n", i, results[i].channel_a);
+		printf("Channel B[%lu] = %x\r\n", i, results[i].channel_b);
```
amiclaus, line 104: same, use `pr_info`

projects/ad7616-st/src/parameters.h
```diff
@@ -0,0 +1,84 @@
...
+#define SPI_DEVICE_ID   4
+#define SPI_BAUDRATE    12000000
+#define SPI_CS          4
+#define SPI_CS_PORT     4
+#define SPI_OPS         &stm32_spi_ops
+#define SPI_EXTRA       &ad7616_spi_extra_ip
+
+#define GPIO_OPS        &stm32_gpio_ops
+
+#define GPIO_PORT_J 9
```
amiclaus, line 73: properly align the value assignment.

amiclaus (approved):

amiclaus (commented):
approved by mistake. my bad.

drivers/adc/ad7616/iio_ad7616.c
Skallwar: I've put a dump of `iio_info` in the PR description

drivers/adc/ad7616/iio_ad7616.c
```diff
@@ -0,0 +1,295 @@
...
+	struct ad7616_dev *dev = device;
+
+	// TODO: we call this on every call, fetching all 16 inputs. Doesn't look smart
+	ret = ad7616_read_data_serial(dev, res, 1);
+	if (ret)
+		return ret;
+
+	if (channel->ch_num < 8)
+		return sprintf(buf, "%i", res[index].channel_a);
+	else
```
rbolboac, line 126: else not needed, can simply return

drivers/adc/ad7616/iio_ad7616.c
```diff
@@ -0,0 +1,295 @@
...
+		return sprintf(buf, "%i", res[index].channel_b);
+}
+
+static int set_raw(void *device,
```
rbolboac, line 130: this API is not needed, simply do not populate the store reference

drivers/adc/ad7616/iio_ad7616.c
```diff
@@ -0,0 +1,295 @@
...
+ */
+int32_t ad7616_iio_param_init(struct iio_device **desc)
+{
+	struct iio_device *iio_ad7616_inst;
+
+	iio_ad7616_inst = calloc(1, sizeof(struct iio_device));
+	if (!iio_ad7616_inst)
+		return -ENOMEM;
+
+	*iio_ad7616_inst = ad7616_iio_device_template;
```
rbolboac, line 288: so in the newer drivers we have a different way of implementing iio drivers. the device driver is initialized here (and not in the example application). you may take a look at iio_adxl355 as an example

drivers/adc/ad7616/iio_ad7616.c
```diff
@@ -0,0 +1,295 @@
...
+static struct scan_type ad7616_iio_scan_type = {
+	.sign = 's',
+	.realbits = CHN_STORAGE_BITS,
+	.storagebits = CHN_STORAGE_BITS,
+	.shift = 0,
+	.is_big_endian = false
+};
+
+static struct iio_attribute ad7616_iio_ch_attributes[] = {
+	{ .name = "raw", .show = get_raw, .store = set_raw },
```
rbolboac, line 236: as I have mentioned above, .store can be omitted here

drivers/adc/ad7616/iio_ad7616.c
```diff
@@ -0,0 +1,295 @@
...
+
+	for (; i < iio_dev_data->buffer->samples; i++) {
+		ret = ad7616_read_data_serial(iio_dev_data->dev, results, 1);
+		if (ret)
+			return ret;
+
+		data[0] = results[0].channel_a;
+		data[1] = results[0].channel_b;
+
+		ret = no_os_cb_write(iio_dev_data->buffer->buf, data, bytes_per_scan);
```
rbolboac, line 219: you should not call no_os_cb related APIs, instead please use iio_buffer_push_scan for each sample

rbolboac (commented):
some comments from my side. I will re-review after iio driver is rewritten.
also, if there is no reason not to do this, I would advice to use the new project structure (it will allow to add support easily for new platforms and new examples)
eval-adis1650x project could be used as reference

drivers/adc/ad7616/iio_ad7616.c
Skallwar: I'm trying to use `iio_app_init` but at the moment I'm getting some random bytes out of the uart, even If I did not change the `uart_init_param`

rbolboac: the default message will always be shown at 115200, after that the uart is reconfigured with the chosen baudrate (in this case 230400). make sure to use 230400 in IIO Oscilloscope connection.
the alignment is not an issue. it is simply a matter of remapping LF to CR + LF in your terminal emulator.

Skallwar:
All comments should have been fixed

Skallwar:
I've added [520d3b2](https://github.com/analogdevicesinc/no-OS/pull/2000/commits/520d3b22fd54f305c08aa62bf33c0939b2e2835e) as it's needed for the project to compile properly. Once #1989 is merged I will remove it from here

drivers/adc/ad7616/iio_ad7616.c
```diff
@@ -0,0 +1,327 @@
...
+ * @param       readval[out]- Pointer to variable to read data into
+ * @return      0 in case of success, negative value otherwise
+ */
+static int32_t iio_ad7616_debug_reg_read(void *device, uint32_t reg,
+		uint32_t *val)
+{
+	struct ad7616_iio_dev *iio_dev = device;
+	struct ad7616_dev *dev = iio_dev->ad7616_dev;
+
+	return ad7616_read(dev,  (uint8_t) reg, (uint16_t *)val);
```
rbolboac, line 80: nit: remove extra space between

drivers/adc/ad7616/iio_ad7616.c
```diff
@@ -0,0 +1,327 @@
...
+{
+	struct ad7616_iio_dev *iio_dev = device;
+	struct ad7616_dev *dev = iio_dev->ad7616_dev;
+
+	return ad7616_write(dev, (uint8_t)reg, (uint16_t)val);
+}
+
+void iio_ad7616_event_handler(struct iio_desc *desc)
+{
+	iio_step(desc);
```
rbolboac, line 101: I don't think iio_step should be called from an iio driver

drivers/adc/ad7616/iio_ad7616.c
```diff
@@ -0,0 +1,327 @@
...
+	.submit = iio_ad7616_submit_buffer,
+};
+
+/**
+ * @brief	Init for reading/writing and parameterization of a
+ *			ad7616 IIO device
+ * @param	desc[in,out] - IIO device descriptor
+ * @return	0 in case of success, negative error code otherwise
+ */
+int32_t ad7616_iio_param_init(struct iio_device **desc)
```
rbolboac, line 271: is this API needed? the iio_init is already assigning the iio_dev structure

drivers/adc/ad7616/iio_ad7616.c
```diff
@@ -0,0 +1,327 @@
...
+		if (ret)
+			return ret;
+
+		data[0] = results[0].channel_a;
+		data[1] = results[0].channel_b;
+
+		ret = iio_buffer_push_scan(iio_dev_data->buffer, data);
+		if (ret)
+			return ret;
+	}
```
rbolboac, line 213: hmmm... so when working with buffers there is the possibility to receive request only for one channel, or two channels, or any combination of channels based on the scan mask. you will have to handle that scenario based on the scan mask and only push the data for the channels which are requested. the scan mask should be available in iio_dev_data structure. also, the data should be added in the buffer based on the channel's scan type.

rbolboac (commented):
some other remarks from my side

drivers/adc/ad7616/iio_ad7616.c
Skallwar: > the data should be added in the buffer based on the channel's scan type.

Not sure to understand this

drivers/adc/ad7616/iio_ad7616.c
Skallwar: No, sorry

drivers/adc/ad7616/iio_ad7616.c
rbolboac: in this case you are adding the data for all the channels in the buffer.
however, the data in the buffer should be added based on the channel scan_mask which is available in iio_dev_data structure.
please see https://wiki.analog.com/resources/tools-software/linux-software/libiio_internals for more information about LibIIO buffers.
you can see as an example here: https://github.com/analogdevicesinc/no-OS/blob/master/drivers/accel/adxl355/iio_adxl355.c#L1019 where the data of a channel is added in the buffer based on the scan mask

drivers/adc/ad7616/iio_ad7616.c
Skallwar: It's not even used anymore, I will remove it

drivers/adc/ad7616/iio_ad7616.c
Skallwar: Ah yes ok, thanks

drivers/adc/ad7616/iio_ad7616.c
```diff
@@ -0,0 +1,310 @@
...
+
+	return 0;
+}
+
+static struct scan_type ad7616_iio_scan_type = {
+	.sign = 's',
+	.realbits = CHN_STORAGE_BITS,
```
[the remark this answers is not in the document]
Skallwar: Reading the documentation of `realbits` in `iio/iio_types.h` ("Number of valid bits of data") it looks to me that it should be `16` as there is `16` bits (2 bytes). Same for `storagebits` as there is no padding.

image

drivers/adc/ad7616/iio_ad7616.c
```diff
@@ -0,0 +1,310 @@
...
+	uint32_t i;
+	uint32_t j;
+	uint32_t k;
+
+	for (i = 0; i < iio_dev_data->buffer->samples; i++) {
+		ret = ad7616_read_data_serial(dev, results, 1);
+		if (ret)
+			return ret;
+
+		for (j = 0, k = 0; j < 16; j++) {
```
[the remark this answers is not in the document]
Skallwar: I can propose this:
image

The built-in is supported by both gcc and clang (checked with godbolt) and we are iterating exactly on the number of requested channel

drivers/adc/ad7616/iio_ad7616.c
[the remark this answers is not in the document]
rbolboac: so first let's discuss what data is added in the buffer.
please correct me if I am wrong. there are 16 channels available in the iio driver. the output value of a channel is either 0 or 1. these values should be added in the buffer based on the scan_mask.

the scan type is actually only used for buffering.. it shows how the channel data is added in the buffer. it does not have anything to do with the way data is retrieved from AD7616.

seeing how the data added in the buffer is only one bit per channel, the realbits should be 1 (because we are referring to the scan_type of the channel). now, the storage bits can really be any multiple of a byte, but it does not make sense to use something higher than a byte, because that data is not useful.

so to sum up, the data in the buffer should contain the data based on the scan_mask in the scan_type format.

now let's take a look at ad7616_read_data_serial and how the data is added in the result.
values for channel 0 -> 7 are found in bits 0 - bits 7 from res.channel_a
values for channel 8 -> 15 are found in bits 0 - bits 7 from res.channel_b

with this in mind, it looks like the get raw and the submit APIs should be rewritten
here is my suggestion for get raw:
```
int ret;
	uint8_t index = channel->ch_num < 8 ? channel->ch_num : channel->ch_num - 8;
	struct ad7616_conversion_result res;
	struct ad7616_iio_dev *iio_dev = device;
	struct ad7616_dev *dev = iio_dev->ad7616_dev;

	// TODO: we call this on every call, fetching all 16 inputs. Doesn't look smart
	ret = ad7616_read_data_serial(dev, res, 1);
	if (ret)
		return ret;

	if (channel->ch_num < 8)
                // will return 0 is the bit according to the channel is 0, and 1 if the bit is 1
		return sprintf(buf, "%i", !!(res.channel_a & NO_OS_BIT(index)));

	return sprintf(buf, "%i", !!(res.channel_b & NO_OS_BIT(index)));
```

drivers/adc/ad7616/iio_ad7616.c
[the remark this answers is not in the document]
rbolboac: please see my remark in the comment below... and please be careful when using ad7616_read_data_serial(dev, results, 1); as shown in the implementation if the number of samples is equal to 1 only results[0] will be written. results[index] with index anything higher than 0 will not contain data read from the device.

drivers/adc/ad7616/iio_ad7616.c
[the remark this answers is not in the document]
Skallwar: > there are 16 channels available in the iio driver. the output value of a channel is either 0 or 1.

Yes there is 16 channels. However each channel is 16 bits wide. As stated by the [datasheet](https://www.analog.com/media/en/technical-documentation/data-sheets/AD7616.pdf):
```
The AD7616 is a 16-bit, DAS that supports dual simultaneous sampling of 16 channels
```
In find the quote a bit confusing but in the block diagram you can see that there is a MUX for A0 to A7 (same for B0..B7) and the output goes to a 16bit SAR (one for A and one for B). There is 2 data path returning 16 bit each, A in red, B in green:
image

The Linux driver also seem to agree that a channel returns 16 bits: https://elixir.bootlin.com/linux/latest/source/drivers/iio/adc/ad7606.h#L22

The input channel for both channel A and B is selected via the "Channel Register" (see page 43). In burst mode as configured here, each time a conversion is requested the device is sending A0[0..16], B0[0..16], A1[0..16], B1[0..16], ..., A7[0..16], B7[0..16] (all 16 input channels)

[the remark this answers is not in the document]
rbolboac: oh, alright. it all makes sense now. I apologize for the misunderstanding.

drivers/adc/ad7616/iio_ad7616.c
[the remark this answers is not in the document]
rbolboac: wouldn't no_os_find_first_set_bit do the job instead of ffs?
the code is a little bit hard to follow.

drivers/adc/ad7616/iio_ad7616.c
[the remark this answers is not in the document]
Skallwar: No problem I got confuse too the first time I read the datasheet :D

drivers/adc/ad7616/iio_ad7616.c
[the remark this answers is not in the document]
Skallwar: I would argue that the `__builtin` solution [looks](https://godbolt.org/#z:OYLghAFBqd5QCxAYwPYBMCmBRdBLAF1QCcAaPECAMzwBtMA7AQwFtMQByARg9KtQYEAysib0QXACx8BBAKoBnTAAUAHpwAMvAFYTStJg1AB9U8lJL6yAngGVG6AMKpaAVxYMQAZg2kHAGTwGTAA5dwAjTGIQACYuUgAHVAVCWwZnNw9vXySUmwFA4LCWSOi4i0wrfIYhAiZiAgz3Tx8KqrTa%2BoJC0Iio2PiFOoamrNahrp7i0oGASgtUV2Jkdg5XBhTgYPQAaiCCHZpiIeMlAmNwwgh1ze29wR2AdxJ0WZ2AUgB2ACF3jQBBHZAnY3PBbTC7faHPDHc5nC6ED5eAAiOw07y8vwBf0BwMeCDomB2EGexFeHx%2BOOB1L2VGJpN27xiADY0aouLMqTTucRMAQlgxobDTnyEQQMVjcdyudyGUjsBiFSidlwJTKaUcTvDLuKYr89WqAdSvsicerefziIKvDFDf8TWaAaDwZCHpq4aKdTFrhswXcoQy3l9JdTnf63TCtZ7ERjUaZwq46DYGMYqFQFCSXpzMY6pbTiWAwO6RecdZyjdydhaBTsbXbzXya8XtYiALQq%2BufZEcea0TgAVl4ng4WlIqE4jh2CkWyyJTK8PFIBE0PfmAGsQP3fH2OJIhyux5xeAoQL5lyOe6Q4LAYIgUKgWAlCWQKBA0I/nyh6ixmZJgFwbV8GhaAIKITwgcID0uZhiAAT04RdoPqWCAHlwm0TBrAQ3h3zYQQUIYWh4IvUgsATYBHDEWgT24XgsBYQxgHEEj8F5aw8AAN0wGjR0wVRMNcUCD32SoD1oPBwmIZDnCwA8CGIPAWGw%2BYqAMYAFAANTwTBHhQhJGGwmRBBEMR2CkIz5CUNQD10eIDCMEBTGMcxxPCE9IHmVAEmqGjWxQmIdlbAB1KjAqCvj5KYXhUC44gFKwdyIHmSxMOqewGCcFxmgkeIAiCXoSn6GJ%2B0SZJUgEUZPC4eJcnKhgpj6MoSpS9iBE6EYsqyaq2lSjphm6fLpiK5r%2BsqnKLH6hrCqa5KZxWCRewHfcSPHDgdn%2BAAlABZX8dgAcUcScAIAOhiY6NGJXBCBID4YgXWZeHPLRZnXTdt04PdSCUrdSGHUdVuPU8lxXeZrzvRYCASQTyHsfAiHivR%2BGM0RxHMpHLJUdQSNs0hHikhJlP0JbfoPVaUMEqGDlQOlNp2yR9sOlUvFO87iWcD96GIW77sekH5gQTAmCwaIkqJ3deG%2B3w/uio8LCBp7V1IDcfp3Lxlv%2B2WFZesWYnVmWOF5i9tdilI7EkIA%3D) faster. It could be a good idea to rework `no_os_find_first_set_bit` in the future. Anyway I will use it as it's more inline with the other projects. Thanks for pointing this out!

rbolboac:
please perform a re-base and resolve the conflicts, now that #1989 has been merged.

Skallwar:
Since I rebased the first compilation with `make` hangs at the `Configuring project`. If I kill it and restart it I manage to go further and get the HEX file in order to flash. It's a bit weird, I will try to find the issue

buha (approved):

amiclaus (commented):
Some general comments:
 - please add builds.json for your project. To make sure that it builds properly.
 - some commits lack the body. Please add proper descriptions.
 - there is already a ad7616 project on master branch. Do we want 2 separate projects for the same device?

drivers/adc/ad7616/iio_ad7616.c
```diff
@@ -0,0 +1,383 @@
...
+			channel->ch_num - CHANNEL_SOURCES_NUMBER;
+	struct ad7616_conversion_result res[RESULTS_NUMBER];
+	struct ad7616_iio_dev *iio_dev = device;
+	struct ad7616_dev *dev = iio_dev->ad7616_dev;
+
+	ret = ad7616_read_data_serial(dev, res, 1);
+	if (ret)
+		return ret;
+
+	if (is_channel_a(channel->ch_num))
```
rbolboac, line 120: is_channel_a is called twice for the same variable. it would be better to store the result of the first call and use it here, without calling the function again

drivers/adc/ad7616/iio_ad7616.c
```diff
@@ -0,0 +1,383 @@
...
+
+	switch (osr) {
+	case AD7616_OSR_0:
+	case AD7616_OSR_2:
+	case AD7616_OSR_4:
+	case AD7616_OSR_8:
+	case AD7616_OSR_16:
+	case AD7616_OSR_32:
+	case AD7616_OSR_64:
+	case AD7616_OSR_128:
```
rbolboac, line 167: can use the following syntax instead, to remove extra lines of code:
case AD7616_OSR_0 ... AD7616_OSR_128:

drivers/adc/ad7616/iio_ad7616.h
rbolboac: Any updated regarding this?

projects/ad7616-st/src/common/common_data.c
```diff
@@ -0,0 +1,102 @@
...
+	.pull = NO_OS_PULL_NONE,
+	.platform_ops = GPIO_OPS,
+};
+
+struct ad7616_init_param ad7616_ip = {
+	.gpio_reset_param = &ad7616_gpio_reset_ip,
+	.gpio_convst_param = &ad7616_gpio_convst_ip,
+	.gpio_busy_param = &ad7616_gpio_busy_ip,
+	.mode = AD7616_SW,
+	.osr = 0,
```
rbolboac, line 101: can use enum instead of 0

projects/ad7616-st/src/examples/examples_src.mk
```diff
@@ -0,0 +1,22 @@
+ifeq (y,$(strip $(IIO_EXAMPLE)))
+CFLAGS += -DIIO_EXAMPLE=1 -DIIO_SUPPORT -DINTC_DEVICE_ID=1
```
rbolboac, line 2: even if the result is the same,  would move the definition of INTC_DEVICE_ID in parameters.h. also, any reason why 1 was chosen? I think there is only one interrupt vector controller for stm32 and by default in the other projects it has id 0.

projects/ad7616-st/src/examples/iio/iio_example.c
```diff
@@ -0,0 +1,96 @@
...
+#include "no_os_gpio.h"
+#include "no_os_print_log.h"
+#include "iio_app.h"
+
+extern struct no_os_uart_desc *uart_desc;
+
+/******************************************************************************/
+/************************ Functions Declarations ******************************/
+/******************************************************************************/
+/******************************************************************************/
```
rbolboac, line 61: line is duplicated and can be removed

projects/ad7616-st/src/examples/iio/iio_example.c
```diff
@@ -0,0 +1,96 @@
...
+
+	iio_devices[0].dev = dev;
+	iio_devices[0].dev_descriptor = dev->iio_dev;
+	app_init_param.devices = iio_devices;
+	app_init_param.nb_devices = NO_OS_ARRAY_SIZE(iio_devices);
+	app_init_param.uart_init_params = ad7616_uart_ip;
+
+	ret = iio_app_init(&app, app_init_param);
+	if (ret)
+		return ret;
```
rbolboac, line 93: before returning, ad7616_iio_remove should be called

projects/ad7616-st/src/examples/iio/iio_example.c
```diff
@@ -0,0 +1,96 @@
...
+
+	return iio_app_run(app);
```
rbolboac, line 95: before returning, in case of an error iio_app_remove should be called, and also ad7616_iio_remove. you may take a look here and follow the example: https://github.com/analogdevicesinc/no-OS/blob/master/projects/max22200/src/examples/iio_example/iio_example.c

projects/ad7616-st/src/examples/iio/iio_example.h
```diff
@@ -0,0 +1,52 @@
...
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+*******************************************************************************/
+#ifndef __IIO_EXAMPLE_H__
+#define __IIO_EXAMPLE_H__
+
+/******************************************************************************/
+/***************************** Include Files **********************************/
+/******************************************************************************/
```
rbolboac, line 45: please remove, there are no include files

projects/ad7616-st/src/platform/stm32/parameters.h
```diff
@@ -0,0 +1,86 @@
...
+/********************** Macros and Constants Definitions **********************/
+/******************************************************************************/
+extern UART_HandleTypeDef huart5;
+
+#define UART_IRQ_ID			UART5_IRQn
+
+#define UART_DEVICE_ID		5
+#define UART_BAUDRATE		230400
+#define UART_EXTRA			&ad7616_uart_extra_ip
+#define UART_OPS			&stm32_uart_ops
```
rbolboac, line 65: please fix all alignment issues in this file

rbolboac (commented):
I did not see any functional issues. Only small remarks from my side regarding code style.
Also, there is a small typo in the following commit message:
```
drivers: adc: ad7616: Update osr field when changing oversampling ratio
This fix an issue 
```
Please use ``This fixes an issue`` instead.

Seeing how we are already in 2024, it would be better to have 2024 added in the header of the files. 
We still have to find out if the Copyright is correct, although some files in the master branch already contain the BayLibre copyright...

projects/ad7616-st/sdp-ck1z.ioc
```diff
@@ -0,0 +1,193 @@
...
+Mcu.Pin6=PE6
+Mcu.Pin7=PD2
+Mcu.Pin8=PH0/OSC_IN
+Mcu.Pin9=PH1/OSC_OUT
+Mcu.PinsNb=16
+Mcu.ThirdPartyNb=0
+Mcu.UserConstants=
+Mcu.UserName=STM32F469NIHx
+MxCube.Version=6.10.0
+MxDb.Version=DB.6.0.100
```
rbolboac, line 39: MxCube.Version=6.5.0
MxDb.Version=DB.6.0.50
Please use these values instead, these work on the CI

projects/ad7616-st/builds.json
```diff
@@ -0,0 +1,7 @@
+{
+  "stm32": {
+    "iio_example": {
+      "flags" : "IIO_EXAMPLE=y"
```
rbolboac, line 4: please specify also the HARDWARE=sdp-ck1z.ioc

rbolboac (commented):
regarding the build failure....

projects/ad7616-st/sdp-ck1z.ioc
Skallwar: I downgraded the version but now I'm getting errors regarding the size of the generated elf segments. I will have a look, but in case it's a known issue here are the logs:
> /usr/bin/../lib/gcc/arm-none-eabi/12.2.1/../../../../arm-none-eabi/bin/ld: /home/esteban/BayLibre/adi/no-os/ad7616_stm32_project/projects/ad7616-st/build/ad7616-st.elf section `.text' will not fit in region `FLASH'
/usr/bin/../lib/gcc/arm-none-eabi/12.2.1/../../../../arm-none-eabi/bin/ld: /home/esteban/BayLibre/adi/no-os/ad7616_stm32_project/projects/ad7616-st/build/ad7616-st.elf section `.data' will not fit in region `RAM'
/usr/bin/../lib/gcc/arm-none-eabi/12.2.1/../../../../arm-none-eabi/bin/ld: region `RAM' overflowed by 9576 bytes
/usr/bin/../lib/gcc/arm-none-eabi/12.2.1/../../../../arm-none-eabi/bin/ld: region `FLASH' overflowed by 109172 bytes

rbolboac: strange... I tried it on my machine and for me the build is successful. you may perform the changes and check whether the CI build is also successful. 
I have only encountered similar errors in bigger projects where indeed the available memory was not enough, but it is not the case here.
I don't know what to suggest, make sure the older version packages are downloaded completely before trying to build. you may also try and see if stm32cubemx is able to successfully generate the code using the .ioc with MxCube.Version=6.5.0
MxDb.Version=DB.6.0.50

Skallwar: I was not able to open the modified ioc file and there is no straight forward way to downgrade the stm32cubemx version (editing MxCube.Version and MxDb.Version leads to other problems from what I read on Stm32 forums). So I took the ioc file from ad74416h and modified it for this project and I got the exact same issue. I looked at the generated linker script and it's obvious that it's wrong:

```ld
/* Memories definition */
MEMORY
{
  CCMRAM    (xrw)    : ORIGIN = 0x10000000,   LENGTH = 64
  RAM    (xrw)    : ORIGIN = 0x20000000,   LENGTH = 320
  FLASH    (rx)    : ORIGIN = 0x8000000,   LENGTH = 2K
}
```
320 Bytes of RAM? 2K of flash? The [product description](https://www.st.com/en/microcontrollers-microprocessors/stm32f469ni.html) does not agree

rbolboac: The CI will not work with versions later than 6.5.0 simply because the STM32CubeMX will give a pop-up window when the project is built, and the CI cannot handle this pop-up. 
So the log file will not be useful, it will simply be a timeout error. 
You can simply modify the ioc in a text editor with the changes I mentioned and push it here to see what the build logs are saying.

projects/ad7616-st/src/examples/examples_src.mk
Skallwar: > any reason why 1 was chosen

I think it's a typo on my side. I never noticed it since it works like that. Using 0 also work

drivers/adc/ad7616/iio_ad7616.c
```diff
@@ -0,0 +1,392 @@
...
+ * @param       readval[out]- Pointer to variable to read data into
+ * @return      0 in case of success, negative value otherwise
+ */
+static int32_t iio_ad7616_debug_reg_read(void *device, uint32_t reg,
+		uint32_t *val)
+{
+	struct ad7616_iio_dev *iio_dev = device;
+	struct ad7616_dev *dev = iio_dev->ad7616_dev;
+
+	return ad7616_read(dev,  (uint8_t) reg, (uint16_t *)val);
```
rbolboac, line 84: also, please see this comment here about pointer casting, and how it was fixed: https://github.com/analogdevicesinc/no-OS/pull/2059#discussion_r1446018878

projects/ad7616-st/Makefile
```diff
@@ -0,0 +1,9 @@
+# Select the example you want to enable by chossing y for enabling and n for disabling
+IIO_EXAMPLE = y
+
+# Select the device you want to enable by choosing y for enabling and n for disabling
```
rbolboac, line 4: please remove this comment, it is not related

projects/ad7616-st/src/platform/stm32/main.c
```diff
@@ -59,7 +59,7 @@ struct no_os_uart_desc *uart_desc;
 int main()
 {
 	ad7616_ip.spi_param = &ad7616_spi_ip;
-	ad7616_spi_extra_ip.get_input_clock = HAL_RCC_GetPCLK1Freq;
+	ad7616_spi_extra_ip.get_input_clock = HAL_RCC_GetPCLK2Freq;
```
rbolboac, line 62: this is regarding the whole commit "Switch to SPI1". can be squashed in the previous commit where the project is added.

rbolboac (commented):
some new remarks from my side. Also please don't forget to change Copyright 2023(c) to Copyright 2024(c) for all the new files this PR adds.

drivers/adc/ad7616/iio_ad7616.h
Skallwar: Not on my side, sorry

rbolboac (approved):
all in all looks good to me.
I'm not sure how fast this needs to be merged... but it would be nice to also have at least the documentation of the project. similarly to what is done here maybe: https://github.com/analogdevicesinc/no-OS/tree/main/projects/eval-adis1650x#readme 
or whatever information you think it is useful such that a new customer is able to reproduce the setup and test the project. from my point of view it can be added with a different PR.
it would be even better if the driver + iio driver documentation would be made available, similarly to what we already have available on GitHub pages: http://analogdevicesinc.github.io/no-OS/drivers_doc.html

PR #2013, opened 2023-11-16
drivers:platform:stm32:Fixes to Prescaler calculation

jansunil:
The value of prescaler can be anything from 0 to 0xFFFF. Application of NO_OS_BIT() to the prescaler value changes the logic behind prescaler calculation.

Fixes: 9264d8983c4e434affd171c7e6b394826089fdb0 ("stm32 pwm support")

## Pull Request Description

## PR Type

## PR Checklist

### DIFF
--- a/drivers/platform/stm32/stm32_pwm.c
+++ b/drivers/platform/stm32/stm32_pwm.c
@@ -107,7 +107,7 @@ static int32_t stm32_init_timer(struct stm32_pwm_desc *desc,
 	if (sparam->get_timer_clock) {
 		timer_frequency_hz = sparam->get_timer_clock();
 		timer_frequency_hz *= sparam->clock_divider;
-		timer_frequency_hz /= NO_OS_BIT(sparam->prescaler);
+		timer_frequency_hz /= sparam->prescaler;
 		period = _compute_period_ticks(desc, timer_frequency_hz, param->period_ns);
 	} else {
 		period = PWM_DEFAULT_PERIOD - 1;
--- a/drivers/platform/stm32/stm32_pwm.h
+++ b/drivers/platform/stm32/stm32_pwm.h
@@ -69,7 +69,7 @@ enum TimOCMode {
  * @brief Structure holding the STM32 PWM parameters.
  */
 struct stm32_pwm_init_param {
-	/** Timer prescaler */
+	/** Timer prescaler (0 to 0xFFFF) */
 	uint32_t prescaler;
 	/** Timer autoreload enable */
 	bool timer_autoreload;


rbolboac (approved):
lgtm, but before merging please remove the extra line between Fixes and Signed-of-by tags in the commit message.

jansunil:
Changelog: Removed extra line in between Fixes and Signed-off tags.

drivers/platform/stm32/stm32_pwm.c
```diff
@@ -107,7 +107,7 @@ static int32_t stm32_init_timer(struct stm32_pwm_desc *desc,
 	if (sparam->get_timer_clock) {
 		timer_frequency_hz = sparam->get_timer_clock();
 		timer_frequency_hz *= sparam->clock_divider;
-		timer_frequency_hz /= NO_OS_BIT(sparam->prescaler);
+		timer_frequency_hz /= sparam->prescaler;
```
buha, line 110: Since you are at it, could you please also document here what this value should be ?
https://github.com/analogdevicesinc/no-OS/blob/master/drivers/platform/stm32/stm32_pwm.h#L72

It's unclear for the user, so specifying the range in the comment i linked above would be great.

jansunil: I've mentioned a range here. Does it look ok?

buha: Looking again at the stm32 pwm implementation, i think it's over complicated for the user.

Yes, it's configurable, the user can specify the stm32_pwm_init_param `prescaler` and `clock_divider`, but i think they should be computed by the stm32_pwm implementation based on the info in no_os_pwm_init_param.
https://github.com/analogdevicesinc/no-OS/blob/master/drivers/platform/stm32/stm32_pwm.c#L108C3-L110

This is probably a topic for another PR.

jansunil: I felt the same too. Few points on this
1)  We could probably have an implementation where the prescaler is calculated based on the PWM period.
2) The clock_divider parameter we set from the top layer does not get reflected into the HAL layer. It is being hardcoded [here](https://github.com/analogdevicesinc/no-OS/blob/master/drivers/platform/stm32/stm32_pwm.c#L157). So should we really expose that parameter via the [stm32_pwm.h](https://github.com/analogdevicesinc/no-OS/blob/master/drivers/platform/stm32/stm32_pwm.h#L85)?

amiclaus (approved):

drivers/platform/stm32/stm32_pwm.c
buha: 1. agreed
2. no need to expose it, agreed

I will merge this PR now and save this conclusion in a backlog. Would be great if you'd implement it but really it's up to you and your priorities.

buha: thanks!

PR #2015, opened 2023-11-16
drivers: digital-io: max22200 : Add driver support for MAX22200

RaduSabau1:
## Pull Request Description

Add driver support for MAX22200.

The MAX22200 is an octal 36V serial-controlled solenoid driver.
Each channel features a low impedance (200mΩ typ) push-pull output stage with sink-and-source driving
capability and up to 1ARMS driving current. A serial interface (SPI) that also supports daisy-chain configurations is
provided to individually control each channel

## PR Type

## PR Checklist

### DIFF
--- a/drivers/digital-io/max22200/iio_max22200.c
+++ b/drivers/digital-io/max22200/iio_max22200.c
@@ -1,0 +1,1096 @@
+/***************************************************************************//**
+ *   @file   iio_max22200.c
+ *   @brief  Source file of IIO MAX22200 Driver.
+ *   @author Radu Sabau (radu.sabau@analog.com)
+********************************************************************************
+ * Copyright 2023(c) Analog Devices, Inc.
+ *
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *  - Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ *  - Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in
+ *    the documentation and/or other materials provided with the
+ *    distribution.
+ *  - Neither the name of Analog Devices, Inc. nor the names of its
+ *    contributors may be used to endorse or promote products derived
+ *    from this software without specific prior written permission.
+ *  - The use of this software may or may not infringe the patent rights
+ *    of one or more patent holders.  This license does not release you
+ *    from the requirement that you obtain separate licenses from these
+ *    patent holders to use this software.
+ *  - Use of the software either in source or binary form, must be run
+ *    on or directly connected to an Analog Devices Inc. component.
+ *
+ * THIS SOFTWARE IS PROVIDED BY ANALOG DEVICES "AS IS" AND ANY EXPRESS OR
+ * IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, NON-INFRINGEMENT,
+ * MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED.
+ * IN NO EVENT SHALL ANALOG DEVICES BE LIABLE FOR ANY DIRECT, INDIRECT,
+ * INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, INTELLECTUAL PROPERTY RIGHTS, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+*******************************************************************************/
+#include <stdlib.h>
+#include <stdio.h>
+#include <string.h>
+#include "no_os_alloc.h"
+#include "no_os_error.h"
+#include "no_os_units.h"
+#include "no_os_util.h"
+
+#include "iio_max22200.h"
+
+#define MAX22200_CHANNEL(_addr)		\
+	{				\
+		.ch_type = IIO_VOLTAGE,	\
+		.indexed = true,	\
+		.channel = _addr,	\
+		.address = _addr,	\
+	}
+
+static int max22200_iio_read_raw(void *dev, char *buf, uint32_t len,
+				 const struct iio_ch_info *channel,
+				 intptr_t priv);
+
+static int max22200_iio_write_raw(void *dev, char *buf, uint32_t len,
+				  const struct iio_ch_info *channel,
+				  intptr_t priv);
+
+static int max22200_iio_read_scale(void *dev, char *buf, uint32_t len,
+				   const struct iio_ch_info *channel,
+				   intptr_t priv);
+
+static int max22200_iio_read_hit(void *dev, char *buf, uint32_t len,
+				 const struct iio_ch_info *channel,
+				 intptr_t priv);
+
+static int max22200_iio_read_hold(void *dev, char *buf, uint32_t len,
+				  const struct iio_ch_info *channel,
+				  intptr_t priv);
+
+static int max22200_iio_read_channel_trig(void *dev, char *buf, uint32_t len,
+		const struct iio_ch_info *channel,
+		intptr_t priv);
+
+static int max22200_iio_read_ch_mode(void *dev, char *buf, uint32_t len,
+				     const struct iio_ch_info *channel,
+				     intptr_t priv);
+
+static int max22200_iio_read_channel_freq(void *dev, char *buf, uint32_t len,
+		const struct iio_ch_info *channel,
+		intptr_t priv);
+
+static int max22200_iio_read_enable(void *dev, char *buf, uint32_t len,
+				    const struct iio_ch_info *channel,
+				    intptr_t priv);
+
+static int max22200_iio_read_fault_mask(void *dev, char *buf, uint32_t len,
+					const struct iio_ch_info *channel,
+					intptr_t priv);
+
+static int max22200_iio_read_chopping_freq(void *dev, char *buf, uint32_t len,
+		const struct iio_ch_info *channel,
+		intptr_t priv);
+
+static int max22200_iio_read_available(void *dev, char *buf, uint32_t len,
+				       const struct iio_ch_info *channel,
+				       intptr_t priv);
+
+static int max22200_iio_reg_read(struct max22200_iio_desc *, uint32_t,
+				 uint32_t *);
+
+static int max22200_iio_write_ch_mode(void *dev, char *buf, uint32_t len,
+				      const struct iio_ch_info *channel,
+				      intptr_t priv);
+
+static int max22200_iio_write_enable(void *dev, char *buf, uint32_t len,
+				     const struct iio_ch_info *channel,
+				     intptr_t priv);
+
+static int max22200_iio_write_fault_mask(void *dev, char *buf, uint32_t len,
+		const struct iio_ch_info *channel,
+		intptr_t priv);
+
+static int max22200_iio_write_trig_state(void *dev, char *buf, uint32_t len,
+		const struct iio_ch_info *channel,
+		intptr_t priv);
+
+static int max22200_iio_write_chopping_freq(void *dev, char *buf, uint32_t len,
+		const struct iio_ch_info *channel,
+		intptr_t priv);
+
+static int max22200_iio_write_ch_avail_attrs(void *dev, char *buf, uint32_t len,
+		const struct iio_ch_info *channel,
+		intptr_t priv);
+
+static int max22200_iio_write_channel_attrs(void *dev, char *buf, uint32_t len,
+		const struct iio_ch_info *channel,
+		intptr_t priv);
+
+static int max22200_iio_reg_write(struct max22200_iio_desc *, uint32_t,
+				  uint32_t);
+
+static struct iio_attribute max22200_attrs[] = {
+	{
+		.name = "raw",
+		.show = max22200_iio_read_raw,
+		.store = max22200_iio_write_raw
+	},
+	{
+		.name = "scale",
+		.show = max22200_iio_read_scale,
+		.store = max22200_iio_write_ch_avail_attrs,
+		.priv = MAX22200_IIO_SCALE
+	},
+	{
+		.name = "scale_available",
+		.show = max22200_iio_read_available,
+		.priv = MAX22200_IIO_SCALE_AVAILABLE,
+		.shared = IIO_SHARED_BY_ALL
+	},
+	{
+		.name = "hit_current",
+		.show = max22200_iio_read_hit,
+		.store = max22200_iio_write_channel_attrs,
+		.priv = MAX22200_IIO_HIT_CURRENT
+	},
+	{
+		.name = "hit_time",
+		.show = max22200_iio_read_hit,
+		.store = max22200_iio_write_channel_attrs,
+		.priv = MAX22200_IIO_HIT_TIME
+	},
+	{
+		.name = "hold_current",
+		.show = max22200_iio_read_hold,
+		.store = max22200_iio_write_channel_attrs,
+		.priv = MAX22200_IIO_HOLD_CURRENT
+	},
+	{
+		.name = "channel_trig",
+		.show = max22200_iio_read_channel_trig,
+		.store = max22200_iio_write_ch_avail_attrs,
+		.priv = MAX22200_IIO_CH_TRIG
+	},
+	{
+		.name = "channel_trig_available",
+		.show = max22200_iio_read_available,
+		.priv = MAX22200_IIO_CH_TRIG_AVAILABLE,
+		.shared = IIO_SHARED_BY_ALL
+	},
+	{
+		.name = "channel_drive",
+		.show = max22200_iio_read_ch_mode,
+		.store = max22200_iio_write_ch_mode,
+		.priv = MAX22200_IIO_CH_DRIVE
+	},
+	{
+		.name = "channel_drive_available",
+		.show = max22200_iio_read_available,
+		.priv = MAX22200_IIO_CH_DRIVE_AVAILABLE,
+		.shared = IIO_SHARED_BY_ALL
+	},
+	{
+		.name = "channel_side",
+		.show = max22200_iio_read_ch_mode,
+		.store = max22200_iio_write_ch_mode,
+		.priv = MAX22200_IIO_CH_SIDE
+	},
+	{
+		.name = "channel_side_available",
+		.show = max22200_iio_read_available,
+		.priv = MAX22200_IIO_CH_SIDE_AVAILABLE,
+		.shared = IIO_SHARED_BY_ALL
+	},
+	{
+		.name = "channel_op_mode",
+		.show = max22200_iio_read_ch_mode,
+		.store = max22200_iio_write_ch_mode,
+		.priv = MAX22200_IIO_CH_OP_MODE
+	},
+	{
+		.name = "channel_op_mode_available",
+		.show = max22200_iio_read_available,
+		.priv = MAX22200_IIO_CH_OP_MODE_AVAILABLE,
+		.shared = IIO_SHARED_BY_ALL
+	},
+	{
+		.name = "channel_freq",
+		.show = max22200_iio_read_channel_freq,
+		.store = max22200_iio_write_ch_avail_attrs,
+		.priv = MAX22200_IIO_CH_FREQ
+	},
+	{
+		.name = "channel_freq_available",
+		.show = max22200_iio_read_available,
+		.priv = MAX22200_IIO_CH_FREQ_AVAILABLE,
+		.shared = IIO_SHARED_BY_ALL
+	},
+	{
+		.name = "slew_rate_control",
+		.show = max22200_iio_read_enable,
+		.store = max22200_iio_write_enable,
+		.priv = MAX22200_SRC
+	},
+	{
+		.name = "open_load_detection",
+		.show = max22200_iio_read_enable,
+		.store = max22200_iio_write_enable,
+		.priv = MAX22200_OL_ENABLE
+	},
+	{
+		.name = "dpm_enable",
+		.show = max22200_iio_read_enable,
+		.store = max22200_iio_write_enable,
+		.priv = MAX22200_DPM_ENABLE
+	},
+	{
+		.name = "hit_currrent_check",
+		.show = max22200_iio_read_enable,
+		.store = max22200_iio_write_enable,
+		.priv = MAX22200_HHF_ENABLE
+	},
+	END_ATTRIBUTES_ARRAY
+};
+
+static struct iio_attribute max22200_debug_attrs[] = {
+	{
+		.name = "thermal_protection_mask",
+		.show = max22200_iio_read_fault_mask,
+		.store = max22200_iio_write_fault_mask,
+		.priv = MAX22200_M_OVT
+	},
+	{
+		.name = "overcurrent_protection_mask",
+		.show = max22200_iio_read_fault_mask,
+		.store = max22200_iio_write_fault_mask,
+		.priv = MAX22200_M_OCP
+	},
+	{
+		.name = "open_load_mask",
+		.show = max22200_iio_read_fault_mask,
+		.store = max22200_iio_write_fault_mask,
+		.priv = MAX22200_M_OLF
+	},
+	{
+		.name = "hit_current_fault_mask",
+		.show = max22200_iio_read_fault_mask,
+		.store = max22200_iio_write_fault_mask,
+		.priv = MAX22200_M_HHF
+	},
+	{
+		.name = "dpm_fault_mask",
+		.show = max22200_iio_read_fault_mask,
+		.store = max22200_iio_write_fault_mask,
+		.priv = MAX22200_M_DPM
+	},
+	{
+		.name = "com_err_mask",
+		.show = max22200_iio_read_fault_mask,
+		.store = max22200_iio_write_fault_mask,
+		.priv = MAX22200_M_COMF
+	},
+	{
+		.name = "uvlo_mask",
+		.show = max22200_iio_read_fault_mask,
+		.store = max22200_iio_write_fault_mask,
+		.priv = MAX22200_M_UVM
+	},
+	END_ATTRIBUTES_ARRAY
+};
+
+static struct iio_attribute max22200_global_attrs[] = {
+	{
+		.name = "chopping_frequency",
+		.show = max22200_iio_read_chopping_freq,
+		.store = max22200_iio_write_chopping_freq,
+		.shared = IIO_SHARED_BY_ALL
+	},
+	{
+		.name = "chopping_frequency_available",
+		.show = max22200_iio_read_available,
+		.priv = MAX22200_IIO_CHOP_FREQ_AVAILABLE,
+		.shared = IIO_SHARED_BY_ALL
+	},
+	{
+		.name = "external_trig_state",
+		.store = max22200_iio_write_trig_state,
+	},
+	END_ATTRIBUTES_ARRAY
+};
+
+static struct iio_device max22200_iio_dev = {
+	.debug_reg_read = (int32_t (*)())max22200_iio_reg_read,
+	.debug_reg_write = (int32_t (*)())max22200_iio_reg_write,
+	.debug_attributes = max22200_debug_attrs,
+	.attributes = max22200_global_attrs
+};
+
+/**
+ * @brief Read the raw attribute for a specific channel (state of the channel)
+ * @param dev - The iio device structure.
+ * @param buf - Buffer to be filled with requested data.
+ * @param len - Length of the received command buffer in bytes.
+ * @param channel - Command channel info.
+ * @param priv - Private descriptor
+ * @return 0 in case of success, error code otherwise
+ */
+static int max22200_iio_read_raw(void *dev, char *buf, uint32_t len,
+				 const struct iio_ch_info *channel,
+				 intptr_t priv)
+{
+	struct max22200_iio_desc *desc = dev;
+	bool ch_state;
+	int32_t val;
+	int ret;
+
+	ret = max22200_get_ch_state(desc->max22200_desc, channel->address,
+				    &ch_state);
+	if (ret)
+		return ret;
+
+	val = ch_state ? 1 : 0;
+
+	return iio_format_value(buf, len, IIO_VAL_INT, 1, &val);
+}
+
+/**
+ * @brief Read the scale attribute for a specific channel.
+ * @param dev - The iio device structure.
+ * @param buf - Buffer to be filled with requested data.
+ * @param len - Length of the received command buffer in bytes.
+ * @param channel - Command channel info.
+ * @param priv - Private descriptor
+ * @return 0 in case of success, error code otherwise
+ */
+static int max22200_iio_read_scale(void *dev, char *buf, uint32_t len,
+				   const struct iio_ch_info *channel,
+				   intptr_t priv)
+{
+	struct max22200_iio_desc *desc = dev;
+	enum max22200_scale ch_scale;
+	int ret;
+
+	ret = max22200_get_ch_scale(desc->max22200_desc, channel->address,
+				    &ch_scale);
+	if (ret)
+		return ret;
+
+	strcpy(buf, max22200_scale_avail[ch_scale]);
+
+	return strlen(buf);
+}
+
+/**
+ * @brief Read the hit attribute.
+ * @param dev - The iio device structure.
+ * @param buf - Buffer to be filled with requested data.
+ * @param len - Length of the received command buffer in bytes.
+ * @param channel - Command channel info.
+ * @param priv - Private descriptor
+ * @return 0 in case of success, error code otherwise
+*/
+static int max22200_iio_read_hit(void *dev, char *buf, uint32_t len,
+				 const struct iio_ch_info *channel,
+				 intptr_t priv)
+{
+	struct max22200_iio_desc *desc = dev;
+	int32_t hit_c, hit_t;
+	uint8_t hit_c_byte, hit_t_byte;
+	int ret;
+
+	ret = max22200_get_ch_hit(desc->max22200_desc, channel->address,
+				  &hit_c_byte, &hit_t_byte);
+	if (ret)
+		return ret;
+
+	hit_c = hit_c_byte;
+	hit_t = hit_t_byte;
+
+	switch (priv) {
+	case MAX22200_IIO_HIT_CURRENT:
+		return iio_format_value(buf, len, IIO_VAL_INT, 1, &hit_c);
+	case MAX22200_IIO_HIT_TIME:
+		return iio_format_value(buf, len, IIO_VAL_INT, 1, &hit_t);
+	default:
+		return -EINVAL;
+	}
+}
+
+/**
+ * @brief Read the hit attribute.
+ * @param dev - The iio device structure.
+ * @param buf - Buffer to be filled with requested data.
+ * @param len - Length of the received command buffer in bytes.
+ * @param channel - Command channel info.
+ * @param priv - Private descriptor
+ * @return 0 in case of success, error code otherwise
+*/
+static int max22200_iio_read_hold(void *dev, char *buf, uint32_t len,
+				  const struct iio_ch_info *channel,
+				  intptr_t priv)
+{
+	struct max22200_iio_desc *desc = dev;
+	int32_t val;
+	uint8_t hold_byte;
+	int ret;
+
+	ret = max22200_get_ch_hold(desc->max22200_desc, channel->address,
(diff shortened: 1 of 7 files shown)


drivers/digital-io/max22200/max22200.c
```diff
@@ -0,0 +1,958 @@
...
+
+	ret = no_os_gpio_set_value(desc->cmd_desc, NO_OS_GPIO_LOW);
+	if (ret)
+		return ret;
+
+	ret = no_os_spi_transfer(desc->comm_desc, &xfer, 1);
+	if (ret)
+		return ret;
+
+	if (one_byte) {
```
amiclaus, line 90: nit: can we do here an if else and return only once? In this manner we avoid duplicated `return 0`

drivers/digital-io/max22200/max22200.c
```diff
@@ -0,0 +1,958 @@
...
+int max22200_reg_read(struct max22200_desc *desc, uint32_t reg, uint32_t *val,
+		      bool one_byte)
+{
+	struct no_os_spi_msg xfer = {
+		.tx_buff = desc->buff,
+		.rx_buff = desc->buff,
+		.bytes_number = 1,
+		.cs_change = 1,
+	};
+	int ret;
```
amiclaus, line 65: nit: i'd put this declaration first, before the definition of struct.

drivers/digital-io/max22200/max22200.c
```diff
@@ -0,0 +1,958 @@
...
+ * MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED.
+ * IN NO EVENT SHALL ANALOG DEVICES BE LIABLE FOR ANY DIRECT, INDIRECT,
+ * INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, INTELLECTUAL PROPERTY RIGHTS, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+*******************************************************************************/
+#include <errno.h>
```
amiclaus, line 39: no_os_error.h

drivers/digital-io/max22200/max22200.c
```diff
@@ -0,0 +1,958 @@
...
+int max22200_reg_write(struct max22200_desc *desc, uint32_t reg, uint32_t val,
+		       bool one_byte)
+{
+	struct no_os_spi_msg xfer = {
+		.tx_buff = desc->buff,
+		.rx_buff = desc->buff,
+		.bytes_number = 1,
+		.cs_change = 1,
+	};
+	int ret;
```
amiclaus, line 120: ditto.

drivers/digital-io/max22200/max22200.c
```diff
@@ -0,0 +1,958 @@
...
+
+	ret = no_os_gpio_set_value(desc->cmd_desc, NO_OS_GPIO_HIGH);
+	if (ret)
+		return ret;
+
+	ret = no_os_spi_transfer(desc->comm_desc, &xfer, 1);
+	if (ret)
+		return ret;
+
+	if (!one_byte) {
```
amiclaus, line 137: either use brackets for both if and else statement, either you don't for both.

drivers/digital-io/max22200/max22200.c
```diff
@@ -0,0 +1,958 @@
...
+	int ret;
+	uint32_t reg_val = 0;
+
+	ret = max22200_reg_read(desc, reg, &reg_val, one_byte);
+	if (ret)
+		return ret;
+
+	reg_val &= ~mask;
+	reg_val |= mask & val;
+	return max22200_reg_write(desc, reg, reg_val, one_byte);
```
amiclaus, line 173: nit: i'd add extra line before last return of the function.

drivers/digital-io/max22200/max22200.c
```diff
@@ -0,0 +1,958 @@
...
+/**
+ * @brief Set device frequency value.
+ * @param desc - MAX22200 device descriptor.
+ * @param device_freq - Device frequency value.
+ * @return 0 in case of succes, negative error code otherwise.
+*/
+int max22200_set_dev_freq(struct max22200_desc *desc,
+			  enum max22200_device_freq device_freq)
+{
+	if (device_freq > 1)
```
amiclaus, line 248: since the `device_freq` is of type `enum max22200_device_freq` I don't think this check is useful.

drivers/digital-io/max22200/max22200.c
```diff
@@ -0,0 +1,958 @@
...
+
+	ret = max22200_reg_update(desc, MAX22200_CFG_CH(ch),
+				  MAX22200_HIT_MASK,
+				  no_os_field_prep(MAX22200_HIT_MASK,
+						  (uint32_t)hit_current), false);
+	if (ret)
+		return ret;
+
+	return max22200_reg_update(desc, MAX22200_CFG_CH(ch),
+				   MAX22200_HIT_T_MASK,
```
amiclaus, line 286: indentation is pretty odd. this applies to multiple returns throughout the code.

drivers/digital-io/max22200/max22200.c
```diff
@@ -0,0 +1,958 @@
...
+ * 		 negative error code.
+ * @return 0 in case of succes, negative error code otherwise.
+*/
+int max22200_set_ch_trig(struct max22200_desc *desc, uint32_t ch,
+			 enum max22200_trig trig)
+{
+	if (ch > MAX22200_CHANNELS - 1)
+		return -EINVAL;
+
+	if (trig > 1)
```
amiclaus, line 355: ditto. redundant check.

drivers/digital-io/max22200/max22200.c
```diff
@@ -0,0 +1,958 @@
...
+ * @param ch_freq - 80KHz/100KHz frequency.
+ * @return 0 in case of succes, negative error code otherwise.
+*/
+int max22200_set_ch_freq(struct max22200_desc *desc, uint32_t ch,
+			 enum max22200_ch_freq ch_freq)
+{
+	if (ch > MAX22200_CHANNELS - 1)
+		return -EINVAL;
+
+	if (ch_freq > 3)
```
amiclaus, line 428: same. ch_freq has a specific type.

drivers/digital-io/max22200/max22200.c
```diff
@@ -0,0 +1,958 @@
...
+ * @param enabled - Disabled/Enabled state.
+ * @return 0 in case of succes, negative error code otherwise.
+*/
+int max22200_set_ch_enable(struct max22200_desc *desc, uint32_t ch,
+			   enum max22200_ch_enable ch_enable, bool enabled)
+{
+	if (ch > MAX22200_CHANNELS - 1)
+		return -EINVAL;
+
+	if (ch_enable > 3)
```
amiclaus, line 451: ditto.

drivers/digital-io/max22200/max22200.c
```diff
@@ -0,0 +1,958 @@
...
+	uint32_t reg_val;
+
+	if (ch > MAX22200_CHANNELS - 1)
+		return -EINVAL;
+
+	ret = max22200_reg_read(desc, MAX22200_CFG_CH(ch), &reg_val, true);
+	if (ret)
+		return ret;
+
+	*enabled =  no_os_field_get(MAX22200_CH_ENABLE_MASK(ch_enable),
```
amiclaus, line 744: nit: extra space.

drivers/digital-io/max22200/max22200.c
```diff
@@ -0,0 +1,958 @@
...
+		if (ret)
+			goto fault_err;
+
+		ret = max22200_set_ch_hit(descriptor, i, MAX22200_HIT_MAX_VAL,
+					  MAX22200_HIT_NO_TIME);
+		if (ret)
+			goto fault_err;
+	}
+
+	no_os_udelay(2500);
```
amiclaus, line 892: meaning of this delay?

drivers/digital-io/max22200/max22200.c
```diff
@@ -0,0 +1,958 @@
...
+	ret = no_os_gpio_get_optional(&descriptor->enable_desc,
+				      init_param->enable_param);
+	if (ret)
+		goto cmd_err;
+
+	if (descriptor->enable_desc) {
+		ret = no_os_gpio_direction_output(descriptor->enable_desc,
+						  NO_OS_GPIO_HIGH);
+		if (ret)
+			goto enable_err;
```
rbolboac, line 824: cmd_err, enable_desc is not allocated in this case

drivers/digital-io/max22200/max22200.c
```diff
@@ -0,0 +1,958 @@
...
+		goto enable_err;
+
+	/* External trigger if used, is set logic LOW at initialization.
+	   Can be set logic HIGH in case of choosing to use the external
+	   trigger for Full-Bridge Mode. */
+	if (descriptor->trig_desc) {
+		ret = no_os_gpio_direction_output(descriptor->trig_desc,
+						  NO_OS_GPIO_LOW);
+		if (ret)
+			goto trig_err;
```
rbolboac, line 839: enable_err, trig_desc is not allocated in this case

drivers/digital-io/max22200/max22200.c
```diff
@@ -0,0 +1,958 @@
...
+
+	ret = no_os_gpio_get_optional(&descriptor->fault_desc,
+				      init_param->fault_param);
+	if (ret)
+		goto trig_err;
+
+	if (descriptor->fault_desc) {
+		ret = no_os_gpio_direction_input(descriptor->fault_desc);
+		if (ret)
+			goto fault_err;
```
rbolboac, line 850: trig_err, fault_desc is not allocated in this case

drivers/digital-io/max22200/iio_max22200.h
```diff
@@ -0,0 +1,130 @@
...
+};
+
+/** Initializes the MAXX22200 IIO descriptor. */
+int max22200_iio_init(struct max22200_iio_desc **,
+		      struct max22200_iio_desc_init_param *);
+
+/** Free resources allocated by the init function. */
+int max22200_iio_remove(struct max22200_iio_desc *);
+
+#endif /** IIO_MAX22200_H */
```
rbolboac, line 130: missing new line at the end of file

drivers/digital-io/max22200/iio_max22200.c
```diff
@@ -0,0 +1,1234 @@
...
+{
+	struct iio_channel *max22200_iio_channels;
+	uint32_t enabled_ch = 0;
+	uint32_t ch_offset = 0;
+	uint32_t i;
+	int ret;
+
+	for (i = 0; i < MAX22200_CHANNELS; i++)
+		if (ch_enable[i])
+			enabled_ch++;;;
```
rbolboac, line 1152: remove extra 2 semi-columns

drivers/digital-io/max22200/iio_max22200.h
```diff
@@ -0,0 +1,130 @@
...
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+*******************************************************************************/
+#ifndef IIO_MAX22200_H
+#define IIO_MAX22200_H
+
+#include <stdint.h>
+#include <stdbool.h>
+#include "iio.h"
+#include "iio_types.h"
```
rbolboac, line 45: not needed

drivers/digital-io/max22200/iio_max22200.c
```diff
@@ -0,0 +1,1234 @@
...
+		.show = max22200_iio_read_fault_mask,
+		.store = max22200_iio_write_fault_mask,
+		.priv = MAX22200_M_UVM
+	},
+	END_ATTRIBUTES_ARRAY
+};
+
+static struct iio_attribute max22200_global_attrs[] = {
+	{
+		.name = "device_frequency",
```
rbolboac, line 330: is this the sampling frequency of the device? let's try and use some generic attribute names if they apply. check this list: https://www.kernel.org/doc/Documentation/ABI/testing/sysfs-bus-iio

drivers/digital-io/max22200/max22200.c
RaduSabau1: Sure.

drivers/digital-io/max22200/max22200.c
RaduSabau1: Moving it.

drivers/digital-io/max22200/max22200.c
RaduSabau1: ofc.

drivers/digital-io/max22200/max22200.c
RaduSabau1: Adding brackets to the else statement.

drivers/digital-io/max22200/max22200.c
RaduSabau1: removing it.

drivers/digital-io/max22200/max22200.c
RaduSabau1: Removing it.

drivers/digital-io/max22200/max22200.c
RaduSabau1: removing it.

drivers/digital-io/max22200/max22200.c
RaduSabau1: Adding a comment specifying why this delay is needed.

drivers/digital-io/max22200/max22200.c
```diff
@@ -0,0 +1,958 @@
...
+		return -EINVAL;
+
+	mask = MAX22200_CH_MODE_MASK(ch);
+	val = no_os_field_prep(MAX22200_CH_MODE_MASK(ch), (uint32_t)ch_op_mode);
+
+	ret = max22200_reg_update(desc, MAX22200_STATUS_REG, mask, val, false);
+	if (ret)
+		return ret;
+
+	*(desc->ch_config + (ch / 2)) = ch_op_mode;
```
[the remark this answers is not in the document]
RaduSabau1: Channels are configured 2 by 2, will add a macro for it.

drivers/digital-io/max22200/max22200.c
RaduSabau1: the if condtion above checks if trig_desc is initialized there by allocated, same for the fault_err handling.

drivers/digital-io/max22200/iio_max22200.h
RaduSabau1: removing it.

drivers/digital-io/max22200/iio_max22200.c
RaduSabau1: it's the chopping frequency of the device, can switch name to chopping_frequency if needed.

RaduSabau1:
V2

- Fixed includes.
- Fixed typos and coding style.
- Removed unnecessary checks.
- Addressed other comments.

drivers/digital-io/max22200/max22200.c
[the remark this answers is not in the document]
CiprianRegus: `desc->ch_config[ch / 2] = ch_op_mode`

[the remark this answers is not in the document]
CiprianRegus: Also, a comment would help more. Using the macro doesn't make this easier to read.

drivers/digital-io/max22200/iio_max22200.c
rbolboac: yes, the name would be cleared

drivers/digital-io/max22200/max22200.c
```diff
@@ -0,0 +1,939 @@
...
+	if (ch > MAX22200_CHANNELS - 1)
+		return -EINVAL;
+
+	if (hit_current > MAX22200_HIT_MAX_VAL)
+		return -EINVAL;
+
+	ret = max22200_reg_update(desc, MAX22200_CFG_CH(ch),
+				  MAX22200_HIT_MASK,
+				  no_os_field_prep(MAX22200_HIT_MASK,
+						  (uint32_t)hit_current), false);
```
rbolboac, line 275: I would let astyle formatting take care of this. basically just write everything on 1 line and then use the VSC for formatting... see everywhere where it applies. it should look something like this:
```
	ret = max22200_reg_update(desc, MAX22200_CFG_CH(ch), MAX22200_HIT_MASK,
				  no_os_field_prep(MAX22200_HIT_MASK, (uint32_t)hit_current), false);
```

drivers/digital-io/max22200/max22200.c
```diff
@@ -0,0 +1,939 @@
...
+ * @return 0 in case of succes, negative error code otherwise.
+*/
+int max22200_fault_mask_get(struct max22200_desc * desc,
+			    enum max22200_fault_mask fault_mask, bool *enabled)
+{
+	int ret;
+	uint32_t reg_val;
+
+	if (fault_mask < 17 || fault_mask > 23)
+		return -EINVAL;
```
rbolboac, line 478: can be removed

drivers/digital-io/max22200/max22200.c
rbolboac: oops, my bad. I misread the code.

drivers/digital-io/max22200/max22200.c
RaduSabau1: Forgot about it the last time.

drivers/digital-io/max22200/iio_max22200.c
RaduSabau1: Changing to chopping_frequency then.

RaduSabau1:
V3

- Fixed coding style.
- Changed name of attribute from device frequency to chopping frequency.
- Addressed other comments.

RaduSabau1:
V4

- Rebased with master.

drivers/digital-io/max22200/max22200.c
```diff
@@ -0,0 +1,922 @@
...
+/**
+ * @brief Read data from desired register for MAX22200
+ * @param desc - MAX22200 device descriptor.
+ * @param reg - Register value to read data from.
+ * @param val - Value read from the register.
+ * @param one_byte - true - Only the register's LSB will be returned through val
+ * 		     false - Whole 32bit register will be returned through val.
+ * @return 0 in case of succes, negative error code otherwise.
+*/
+int max22200_reg_read(struct max22200_desc *desc, uint32_t reg, uint32_t *val,
```
amiclaus, line 56: having a second look, do we want to deviate from the "standard" read/write function prototypes by adding the `one_byte` parameter? Having a look at the datasheet, all registers are 32bits wide. Can't we just propely adjust the 32-bit register values prior to calling the read/write functions?

drivers/digital-io/max22200/max22200.c
```diff
@@ -0,0 +1,922 @@
...
+	ret = max22200_reg_read(descriptor, MAX22200_STATUS_REG, &reg_val,
+				false);
+	if (ret)
+		goto fault_err;
+
+	/* Getting channel mode info from init_param */
+	for (i = 0; i < MAX22200_CHANNELS_CONFIG; i++)
+		status_reg |= no_os_field_prep(MAX22200_CH_MODE_MASK(i),
+					       *(init_param->ch_config + i));
+
```
amiclaus, line 824: nit: extra line

drivers/digital-io/max22200/iio_max22200.c
```diff
@@ -0,0 +1,1234 @@
...
+	struct max22200_iio_desc *desc = dev;
+	uint8_t hit_c, hit_t;
+	int ret;
+
+	ret = max22200_get_ch_hit(desc->max22200_desc, channel->address, &hit_c,
+				  &hit_t);
+
+	switch (priv) {
+	case MAX22200_IIO_HIT_CURRENT:
+		if (ret)
```
amiclaus, line 454: this is odd. move it outside the switch case.

drivers/digital-io/max22200/max22200.c
RaduSabau1: Well with this chip you can either read/write the whole 32-bit register or only the LSB, and since the chip provides this functionality I thought the driver should as well.
There is also a possibility to use a bit in the reg value which can be used instead of the one_byte parameter, any thoughts about this?

RaduSabau1:
V5

- Removed one_byte parameter in read/write functions.

drivers/digital-io/max22200/iio_max22200.c
```diff
@@ -0,0 +1,1230 @@
...
+	switch (priv) {
+	case MAX22200_IIO_CH_DRIVE:
+		for (i = 0; i < NO_OS_ARRAY_SIZE(max22200_ch_drive_avail); i++) {
+			if (!strcmp(buf, max22200_ch_drive_avail[i])) {
+				ch_drive = i;
+				break;
+			}
+		}
+
+		if (i == NO_OS_ARRAY_SIZE(max22200_ch_drive_avail))
```
amiclaus, line 957: all these checks can be dropped if you return directly in the for loop. same applies in other functions of the driver. This is just a suggestion.

drivers/digital-io/max22200/max22200.c
```diff
@@ -0,0 +1,910 @@
...
+ * @brief Set channel state for specific channel
+ * @param desc - MAX22200 device descriptor.
+ * @param ch - Channel number.
+ * @param ch_state - Channel's state to be changed into.
+ * @return 0 in case of succces, negative error code otherwise.
+*/
+int max22200_set_ch_state(struct max22200_desc *desc, uint32_t ch,
+			  enum max22200_ch_state ch_state)
+{
+	if (ch > MAX22200_CHANNELS - 1)
```
rbolboac, line 196: I would rename this to MAX22200_MAX_CHN_IDX and give it a value of 7, and make the updates everywhere.

drivers/digital-io/max22200/iio_max22200.c
```diff
@@ -0,0 +1,1212 @@
...
+{
+	struct max22200_iio_desc *desc = dev;
+	uint32_t val;
+
+	iio_parse_value(buf, IIO_VAL_INT, (int32_t *)&val, NULL);
+
+	if (val > 1)
+		val = 1;
+
+	return max22200_set_trig_state(desc->max22200_desc, val);
```
rbolboac, line 1076: so here the function takes an enumeration as parameter and you are giving a uint32_t value. I would replace this code with something like:
```
return max22200_set_trig_state(desc->max22200_desc, val ? MAX22200_TRIG_ON : MAX22200_TRIG_OFF);
```
which shows better what is the intended purpose and what val represents
see in all APIs where applicable

drivers/digital-io/max22200/iio_max22200.c
```diff
@@ -0,0 +1,1212 @@
...
+{
+	struct max22200_iio_desc *desc = dev;
+	uint32_t val;
+
+	iio_parse_value(buf, IIO_VAL_INT, (int32_t *)&val, NULL);
+
+	if (val > 1)
+		val = 1;
+
+	return max22200_fault_mask_set(desc->max22200_desc, priv, val);
```
rbolboac, line 1052: return max22200_fault_mask_set(desc->max22200_desc, priv, val ? true : false);

drivers/digital-io/max22200/iio_max22200.c
RaduSabau1: commited.

drivers/digital-io/max22200/max22200.c
RaduSabau1: will do.

drivers/digital-io/max22200/iio_max22200.c
RaduSabau1: sure.

RaduSabau1:
V6

- Removed API call related warnings.
- Addressed other comments.

drivers/digital-io/max22200/max22200.c
```diff
@@ -0,0 +1,910 @@
...
+#include <string.h>
+#include "max22200.h"
+#include "no_os_util.h"
+#include "no_os_alloc.h"
+#include "no_os_delay.h"
+
+/**
+ * @brief Read data from desired register for MAX22200
+ * @param desc - MAX22200 device descriptor.
+ * @param reg - Register value to read data from.
```
buha, line 50: Rephrase to register number or register address

drivers/digital-io/max22200/max22200.c
```diff
@@ -0,0 +1,910 @@
...
+ * @param val - Value read from the register.
```
buha, line 51: something is odd about this phrasing

drivers/digital-io/max22200/iio_max22200.h
```diff
@@ -0,0 +1,131 @@
...
+	MAX22200_IIO_COMER_MASK,
+	MAX22200_IIO_DPM_MASK,
+	MAX22200_IIO_HHF_MASK,
+	MAX22200_IIO_OLF_MASK,
+	MAX22200_IIO_OCP_MASK,
+	MAX22200_IIO_OVT_MASK
+};
+
+static const char *const max22200_scale_avail[2] = {
+	"FULLSCALE",
```
buha, line 71: I'm not sure using caps is a good idea, i think i haven't come across of iio device drivers that use caps in attribute fields

drivers/digital-io/max22200/max22200.c
```diff
@@ -0,0 +1,910 @@
...
+
+	ret = no_os_spi_transfer(desc->comm_desc, &xfer, 1);
+	if (ret)
+		return ret;
+
+	if (!no_os_field_get(reg, MAX22200_R1B)) {
+		xfer.bytes_number = MAX22200_FRAME_SIZE;
+		no_os_put_unaligned_le32(val, desc->buff);
+	} else {
+		desc->buff[0] = val;
```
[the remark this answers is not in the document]
RaduSabau1: there is an if condition above that checks if val is greater than the max value for uint8_t in case of just one byte write operation.

drivers/digital-io/max22200/iio_max22200.c
```diff
@@ -0,0 +1,1221 @@
...
+		.show = max22200_iio_read_raw,
+	},
+	{
+		.name = "scale",
+		.show = max22200_iio_read_scale,
+		.store = max22200_iio_write_scale,
+	},
+	{
+		.name = "scale_available",
+		.show = max22200_iio_read_scale_available,
```
buha, line 179: You have a function for each attribute's show/store. This works but it can be written in a more compact way.

For example, with regards to these `...available` attributes, they all return a string directly, so you could have 1 single show function that handles all available attributes and differentiates them using an ID that you pass through `iio_attribute.priv`.

So instead of having 20 functions that do more or less the same think, you end up with 3/4/5, best case 1/2 for simple devices.

drivers/digital-io/max22200/max22200.c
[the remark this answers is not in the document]
buha: I mean, I briefly checked the datasheet register map and all registers are 4 byte long. What's the catch ? Can you do a partial write or read of a register?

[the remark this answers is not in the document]
RaduSabau1: Table 9 in max22200 datasheet explains the situation.

drivers/digital-io/max22200/iio_max22200.c
RaduSabau1: Tried to group as much as possible. I couldn't add ch_mode_available with the others since it needs its own separate priv.

RaduSabau1:
V7

- Grouped attributes to be showed by the same API using different priv for each one of them.
- Using bool instead of enum for on/off states.
- Addressed other comments.

buha (approved):

RaduSabau1:
V8

- Fixed name of API for chopping frequency.
- Fixed build error.

RaduSabau1:
V9

- Added iio_write_raw API.
- Using uint32_t variables instead of uint8_t for hit and hold attribute.
- Addressed comments related to unnecessary checks.

buha (approved):

drivers/digital-io/max22200/iio_max22200.c
```diff
@@ -0,0 +1,1090 @@
...
+	struct max22200_iio_desc *desc = dev;
+	bool val = 0;
+	int ret;
+
+	ret = max22200_get_ch_enable(desc->max22200_desc, channel->address,
+				     priv, &val);
+	if (ret)
+		return ret;
+
+	return iio_format_value(buf, len, IIO_VAL_INT, 1,
```
rbolboac, line 561: casting from bool to int32_t is not a good idea. please make the update also for this case

rbolboac (approved):
after fixing that, can be merged from my point of view

drivers/digital-io/max22200/iio_max22200.c
```diff
@@ -0,0 +1,1090 @@
...
+ * @param priv - Private descriptor
+ * @return 0 in case of success, error code otherwise
+*/
+static int max22200_iio_read_channel_trig(void *dev, char *buf, uint32_t len,
+		const struct iio_ch_info *channel,
+		intptr_t priv)
+{
+	struct max22200_iio_desc *desc = dev;
+	enum max22200_trig ch_trig;
+	uint32_t val;
```
CiprianRegus, line 464: Declare this variable as `int32_t` so you avoid casting

CiprianRegus (changes requested):

drivers/digital-io/max22200/iio_max22200.c
```diff
@@ -0,0 +1,1090 @@
...
+static int max22200_iio_read_hold(void *dev, char *buf, uint32_t len,
+				  const struct iio_ch_info *channel,
+				  intptr_t priv)
+{
+	struct max22200_iio_desc *desc = dev;
+	uint32_t val;
+	int ret;
+
+	ret = max22200_get_ch_hold(desc->max22200_desc, channel->address,
+				   (uint8_t)&val);
```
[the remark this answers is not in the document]
RaduSabau1: Fixing it.

drivers/digital-io/max22200/iio_max22200.c
RaduSabau1: Using different approach then.

RaduSabau1:
V10

- Fixed wrong casting.
- Addressed other comments.

drivers/digital-io/max22200/iio_max22200.c
```diff
@@ -0,0 +1,1091 @@
...
+
+	ret = max22200_get_ch_hit(desc->max22200_desc, channel->address,
+				  &hit_c, &hit_t);
+	if (ret)
+		return ret;
+
+	switch (priv) {
+	case MAX22200_IIO_HIT_CURRENT:
+		return iio_format_value(buf, len, IIO_VAL_INT, 1,
+					(int32_t *)&hit_c);
```
CiprianRegus, line 415: Casting to a pointer of a wider type is wrong. It will result in accessing memory beyond where `hit_c` is stored.

drivers/digital-io/max22200/iio_max22200.c
```diff
@@ -0,0 +1,1091 @@
...
+	enum max22200_trig ch_trig;
+	int32_t val;
+	int ret;
+
+	ret = max22200_get_ch_trig(desc->max22200_desc, channel->address,
+				   &ch_trig);
+	if (ret)
+		return ret;
+
+	val = (int32_t)ch_trig;
```
CiprianRegus, line 472: no need to cast

drivers/digital-io/max22200/iio_max22200.c
```diff
@@ -0,0 +1,1091 @@
...
+	struct max22200_iio_desc *desc = dev;
+	enum max22200_chopping_freq chopping_freq;
+	int32_t val;
+	int ret;
+
+	ret = max22200_get_chop_freq(desc->max22200_desc, &chopping_freq);
+	if (ret)
+		return ret;
+
+	val = (int32_t)chopping_freq;
```
CiprianRegus, line 616: useless cast

drivers/digital-io/max22200/iio_max22200.c
```diff
@@ -0,0 +1,1091 @@
...
+ * @param channel - Command channel info.
+ * @param priv - Private descriptor
+ * @return 0 in case of success, error code otherwise
+ */
+static int max22200_iio_write_raw(void *dev, char *buf, uint32_t len,
+				  const struct iio_ch_info *channel,
+				  intptr_t priv)
+{
+	struct max22200_iio_desc *desc = dev;
+	uint32_t val;
```
CiprianRegus, line 710: `int32_t`

drivers/digital-io/max22200/iio_max22200.c
```diff
@@ -0,0 +1,1091 @@
...
+ * @param channel - Command channel info.
+ * @param priv - Private descriptor
+ * @return 0 in case of success, error code otherwise
+ */
+static int max22200_iio_write_fault_mask(void *dev, char *buf, uint32_t len,
+		const struct iio_ch_info *channel,
+		intptr_t priv)
+{
+	struct max22200_iio_desc *desc = dev;
+	uint32_t val;
```
CiprianRegus, line 820: `int32_t`

drivers/digital-io/max22200/iio_max22200.c
RaduSabau1: Fixing and applying where it fits.

drivers/digital-io/max22200/iio_max22200.c
RaduSabau1: My bad, I misunderstood the earlier comment on this same API, fixing it.

amiclaus (approved):
a changelog would be nice after your latest push.

RaduSabau1:
V11

- Fixed bad memory access.
- Addressed other comments.

drivers/digital-io/max22200/iio_max22200.c
```diff
@@ -0,0 +1,1091 @@
...
+static int max22200_iio_read_hit(void *dev, char *buf, uint32_t len,
+				 const struct iio_ch_info *channel,
+				 intptr_t priv)
+{
+	struct max22200_iio_desc *desc = dev;
+	uint32_t hit_c, hit_t;
+	int ret;
+
+	ret = max22200_get_ch_hit(desc->max22200_desc, channel->address,
+				  (uint8_t *)&hit_c, (uint8_t *)&hit_t);
```
CiprianRegus, line 408: This will lead to different results based on the cpu's endianess. Please stop casting pointers to different widths, it's not safe to do so, and there is no point in doing it.

RaduSabau1: Won't cast then since there is no point in doing it.

drivers/digital-io/max22200/iio_max22200.c
```diff
@@ -0,0 +1,1091 @@
...
+*/
+static int max22200_iio_read_hit(void *dev, char *buf, uint32_t len,
+				 const struct iio_ch_info *channel,
+				 intptr_t priv)
+{
+	struct max22200_iio_desc *desc = dev;
+	uint32_t hit_c, hit_t;
+	int ret;
+
+	ret = max22200_get_ch_hit(desc->max22200_desc, channel->address,
```
rbolboac, line 407: get_ch_hit expect uin8_t* and you are giving uint32_t*. please use the same type of pointer. see everywhere where applicable

drivers/digital-io/max22200/iio_max22200.c
```diff
@@ -0,0 +1,1098 @@
...
+ * @param channel - Command channel info.
+ * @param priv - Private descriptor
+ * @return 0 in case of success, error code otherwise
+ */
+static int max22200_iio_write_channel_attrs(void *dev, char *buf, uint32_t len,
+		const struct iio_ch_info *channel,
+		intptr_t priv)
+{
+	struct max22200_iio_desc *desc = dev;
+	uint32_t val;
```
rbolboac, line 952: can be int32_t and cast can be removed

drivers/digital-io/max22200/iio_max22200.c
```diff
@@ -0,0 +1,1098 @@
...
+ * @param channel - Command channel info.
+ * @param priv - Private descriptor
+ * @return 0 in case of success, error code otherwise
+*/
+static int max22200_iio_read_hit(void *dev, char *buf, uint32_t len,
+				 const struct iio_ch_info *channel,
+				 intptr_t priv)
+{
+	struct max22200_iio_desc *desc = dev;
+	uint32_t hit_c, hit_t;
```
rbolboac, line 404: can be int32_t and cast can be removed

drivers/digital-io/max22200/iio_max22200.c
```diff
@@ -0,0 +1,1098 @@
...
+ * @param channel - Command channel info.
+ * @param priv - Private descriptor
+ * @return 0 in case of success, error code otherwise
+*/
+static int max22200_iio_read_hold(void *dev, char *buf, uint32_t len,
+				  const struct iio_ch_info *channel,
+				  intptr_t priv)
+{
+	struct max22200_iio_desc *desc = dev;
+	uint32_t val;
```
rbolboac, line 442: can be int32_t and cast can be removed

RaduSabau1:
V13

- Using variable casting instead of pointer casting.

CiprianRegus (approved):

amiclaus (approved):

PR #2032, opened 2023-11-23
proojects: ad-acevsecrdset-sl: AddADE9113 rename

danmois:
The ADE9113 is a precision simultaneously sampling Σ-Δ analog-to-digital converter (ADCs) for both DC and polyphase shunt-based energy metering applications. The ADE9113 integrates safety certified signal and power galvanic isolation with three simultaneously sampling fully differential 24-bit Σ-Δ ADC channels. 

## Pull Request Description

## PR Type

## PR Checklist


rbolboac (commented):
are there any functional changes involved here? it looks like there are only renames performed.
anyhow, if there are any functional changes, it would be a good idea to split this commit in 2 commits. first one would be to simply perform the renaming from ade7913 to ade79113, which would not imply any real functional changes. 
then for a second commit add the actual functional changes.
if no functional changes are present then maybe just specify in the commit message that only renaming is performed.

danmois:
> are there any functional changes involved here? it looks like there are only renames performed. anyhow, if there are any functional changes, it would be a good idea to split this commit in 2 commits. first one would be to simply perform the renaming from ade7913 to ade79113, which would not imply any real functional changes. then for a second commit add the actual functional changes. if no functional changes are present then maybe just specify in the commit message that only renaming is performed.

There are no functional changes. Will update commit message.

rbolboac:
> > are there any functional changes involved here? it looks like there are only renames performed. anyhow, if there are any functional changes, it would be a good idea to split this commit in 2 commits. first one would be to simply perform the renaming from ade7913 to ade79113, which would not imply any real functional changes. then for a second commit add the actual functional changes. if no functional changes are present then maybe just specify in the commit message that only renaming is performed.
> 
> There are no functional changes. Will update commit message.

Maybe just state in the commit that a rename is performed. The support is already present so the commit message does not really fit to what the commit is actually doing.

danmois:
V2

- change commit title and message

rbolboac (approved):
lgtm

amiclaus (approved):

buha (approved):

PR #2033, opened 2023-11-23
Release build

AAndrisa:
## Pull Request Description

Cherry-pick a few commits from master and changed the hardware files to be downloaded from 2022_R2 hdl branch also.

## PR Type

## PR Checklist


SRaus (approved):

tools/scripts/build_projects.py
```diff
@@ -81,6 +81,25 @@ def shell_source(script):
 
 	os.environ.update(env)
 
+def re_run_stm32(cmd):
+	global ERR
+	log("Project first failed, rebuild to check if the error persists")
```
rbolboac, line 86: nit: instead of Project first failed maybe rewrite to First build of the project failed, rebuild ...

tools/scripts/build_projects.py
```diff
@@ -81,6 +81,25 @@ def shell_source(script):
...
+	err = os.system(cmd_reset + ' >> %s 2>&1' % log_file)
+	if err != 0:
+		ERR = 1
+		log_err("ERROR")
+		return err
+
+	log("make all")
+	err = os.system(cmd + ' >> %s 2>&1' % log_file)
+	if err != 0:
+		log("Error persits, not a random fail, please check!")
```
rbolboac, line 98: I mean, chances are small but the project can fail a second time randomly. I am not sure these logs are that useful. I would simply return err. If the project fails the second time you will have to check the error logs anyways to see what kind of error ocurred. 
if the project does not fail a second time you will just see the successful build and you will not check the logs because the build was successful.

rbolboac (commented):
there is a commit without commit message, I think we require a message for a commit.
also, there is the cherry-pick message after the SIgned-off-by tag which I think it should be removed.

SRaus:
> there is a commit without commit message, I think we require a message for a commit. also, there is the cherry-pick message after the Signed-off-by tag which I think it should be removed.

Agree with the no-message on commit.
`(cherry picked from commit <commit_number>)` is added automatically at the end if cherry-pick is run with option "-x" and we usually use it to correlate more easily from where the changes were cherry-picked. I think we can change the order, so to be before sign-off, only if is added manually.

tools/scripts/build_projects.py
SRaus: If the project is failing 2 consecutive times, we just consider that that's not a random failure anymore.

amiclaus (commented):
stm32 build still fails randomly according to the build logs, although 1addee02535fe0f6042e3a8c772987274250d6c5 was added,

amiclaus (approved):

PR #2047, opened 2023-12-08
include: Add constant GENMASK macro

gastmaier:
Replace non-constant GENMASK macro with constant alternative.
This is the [implementation in the Linux kernel](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/include/linux/bits.h?h=v6.7-rc4#n33).

## Pull Request Description

Allows to use GENMASK in header files and the assembly is much simpler.
Assembly comparison:
* [Master](https://godbolt.org/z/nTW1ndvc3)
* [PR](https://godbolt.org/z/j8djah9o4)

## PR Type

## PR Checklist

### DIFF
--- a/include/no_os_util.h
+++ b/include/no_os_util.h
@@ -83,12 +83,10 @@
 
 #define NO_OS_BITS_PER_LONG 32
 
-#define NO_OS_GENMASK(h, l) ({ 					\
-		uint32_t t = (uint32_t)(~0UL);			\
-		t = t << (NO_OS_BITS_PER_LONG - (h - l + 1));		\
-		t = t >> (NO_OS_BITS_PER_LONG - (h + 1));		\
-		t;						\
-})
+#define NO_OS_GENMASK(h, l) (					\
+	(((uint32_t)(~0UL) << (l))) &				\
+	(((uint32_t)(~0UL) >> (NO_OS_BITS_PER_LONG - (h + 1))))	\
+)
 
 #define no_os_bswap_constant_32(x) \
 	((((x) & 0xff000000) >> 24) | (((x) & 0x00ff0000) >>  8) | \


amiclaus (commented):
looks good at first sight. was the new genmask tested properly?

gastmaier:
Tested on hw (STM32 Nucleo board).
Here is some more assembly inspection https://godbolt.org/z/YsTrTfY96
The last one is an invalid use of the macro (`l > h`), where it always resolves to an empty mask `'h0`. 

The Linux version includes [a safe guard](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/include/linux/bits.h?h=v6.7-rc4#n22) to throw a compiler error when the macros resolves into 0 ([BUILD_BUG_ON_ZERO](https://docs.huihoo.com/doxygen/linux/kernel/3.7/include_2linux_2bug_8h.html#ab57fd92ea606d15c3291e6bc0396c8af) macro), but I don't see such method in no-os.

amiclaus (approved):

buha:
Hi @gastmaier this is good in principle, but i'm not sure it is good licensing-wise because it's too similar to linux implementation.

gastmaier:
Hi @buha, thanks for the feedback.
I updated the commit to be a direct improvement from the previous implementation, keeping it as similar as possible.
The update assembly inspection is available at https://godbolt.org/z/a1oz9EfEq

The main motivation is to evaluate during compile-time into a constant value, allowing GENMASK to be used in headers files.
Please compare the assembly of the implementation on [master](https://godbolt.org/z/nTW1ndvc3) and the [pr](https://godbolt.org/z/Wa1hK67sa).
With this implementation, it is possible for example:
```
// some_header.h
#define MAX_ADDRESS             GENMASK(6,0)
#define ENCODED_ADDR_LENGTH     MAX_ADDRESS/8
uint8_t some_encoded_addr_buffer[ENCODED_ADDR_LENGTH];
```
The [openbsd](https://github.com/openbsd/src/blob/dc44c75f13fd4b8e7f5f5c288bc98bf1099cf917/sys/dev/pci/drm/include/linux/bitops.h#L33) implementation is pretty much the same, also.

buha:
I'm going to close this because of possible licensing issues, specifically GPL infection.

PR #2051, opened 2023-12-15
condition the vs code project configuration

buha:
## Pull Request Description
This is currently only supported on stm32, so limit it to that
platform. This fixes some issues with building for xilinx platform
which has a funky way of detecting the architecture (build/tmp/arch.txt)
which makes the builds fail. The failure is due to a which command
call without arguments and it all comes from vs code support.

https://ez.analog.com/microcontroller-no-os-drivers/f/q-a/576135/no-os-build-ad9081-vitis-2022-2

## PR Type

## PR Checklist

### DIFF
--- a/tools/scripts/vsc_intellisense.mk
+++ b/tools/scripts/vsc_intellisense.mk
@@ -1,6 +1,7 @@
 #
 #	Makefile to create project dependent configs files to enable VS Code IntelliSense
 #
+ifeq 'yes' '$(VSCODE_SUPPORT)'
 
 # Definitions
 CPP_PROP_TEMPLATE		= $(NO-OS)/tools/scripts/platform/template_c_cpp_properties.json
@@ -33,4 +34,6 @@ INCLUDEPATH_INTELLI		:= $(subst ${INCLUDE_NOOS_PATTERN},$(INCLUDE_NOOS_CORRECTED
 CPP_FINAL_CONTENT		:= $(file < $(CPP_PROP_TEMPLATE))
 CPP_FINAL_CONTENT		:= $(subst VSCODE_INCLUDEPATH_INTELLI,$(INCLUDEPATH_INTELLI),$(CPP_FINAL_CONTENT))
 CPP_FINAL_CONTENT 		:= $(subst VSCODE_DEFINES_INTELLI,$(DEFINES_INTELLI),$(CPP_FINAL_CONTENT))
+
+endif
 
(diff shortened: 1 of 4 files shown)


tools/scripts/vsc_intellisense.mk
```diff
@@ -34,3 +35,4 @@ CPP_FINAL_CONTENT		:= $(file < $(CPP_PROP_TEMPLATE))
 CPP_FINAL_CONTENT		:= $(subst VSCODE_INCLUDEPATH_INTELLI,$(INCLUDEPATH_INTELLI),$(CPP_FINAL_CONTENT))
 CPP_FINAL_CONTENT 		:= $(subst VSCODE_DEFINES_INTELLI,$(DEFINES_INTELLI),$(CPP_FINAL_CONTENT))
 
+endif
```
amiclaus, line 38: nit: new line.

tools/scripts/vsc_openocd_debug.mk
```diff
@@ -24,3 +25,4 @@ VSC_LAUNCH_CONTENT	:= $(subst VSCODE_CMSISCFG_FILE,"$(BINARY).openocd-cmsis",$(V
 VSC_LAUNCH_CONTENT	:= $(subst VSCODE_STLINKCFG_FILE,"$(BINARY).openocd",$(VSC_LAUNCH_CONTENT))
 VSC_LAUNCH_CONTENT	:= $(subst VSCODE_SVD_FILE,"$(OPENOCD_SVD)/$(TARGETSVD).svd",$(VSC_LAUNCH_CONTENT))
 
+endif
```
amiclaus, line 28: nit: new line.

amiclaus (commented):
Isn't vs code supported also by the maxim platform?

buha:
@amiclaus 
V2:
- i added new lines at the end of files

Regarding the question, yes we have VS Code support with maxim platform but it's different, it doesn't use these two vsc_*.mk files. The current support that we have should be modified to use this new infrastructure but we didn't have time to do it yet I guess.

amiclaus (approved):
> @amiclaus V2:
> 
>     * i added new lines at the end of files
> 
> 
> Regarding the question, yes we have VS Code support with maxim platform but it's different, it doesn't use these two vsc_*.mk files. The current support that we have should be modified to use this new infrastructure but we didn't have time to do it yet I guess.

Thanks for the clarification! The naming `VSCODE_SUPPORT` was a bit confusing, since I know that we use vs code for other platforms too.

PR #2059, opened 2023-12-29
Add ad738x for stm32

ahaslam2:
## Pull Request Description

This adds stm32 project for the ad738x. The stategy for this code
was based of the recently approved PR for ad7616

## PR Type

## PR Checklist


projects/ad738x-st/src/platform/stm32/parameters.h
```diff
@@ -0,0 +1,79 @@
...
+#include "stm32_uart_stdio.h"
+
+/******************************************************************************/
+/********************** Macros and Constants Definitions **********************/
+/******************************************************************************/
+extern UART_HandleTypeDef huart5;
+
+#define UART_IRQ_ID			UART5_IRQn
+
+#define UART_DEVICE_ID		5
```
amiclaus, line 62: ditto: please properly align all values for the macro definitions.

projects/ad738x-st/src/platform/stm32/platform_src.mk
```diff
@@ -0,0 +1,16 @@
+INCS += $(PLATFORM_DRIVERS)/stm32_delay.h     \
+	$(PLATFORM_DRIVERS)/stm32_gpio.h      \
+	$(PLATFORM_DRIVERS)/stm32_hal.h       \
+	$(PLATFORM_DRIVERS)/stm32_spi.h       \
+	$(PLATFORM_DRIVERS)/stm32_irq.h       \
+	$(PLATFORM_DRIVERS)/stm32_gpio_irq.h  \
+	$(PLATFORM_DRIVERS)/stm32_uart.h      \
+	$(PLATFORM_DRIVERS)/stm32_uart_stdio.h
+
+	SRCS += $(PLATFORM_DRIVERS)/stm32_delay.c     \
```
amiclaus, line 10: nit: wrong indentation.

amiclaus (commented):
looks good at first sight, some minor comments on my side.

ahaslam2:
Changes in V2:

- fix indentation issues
- use mV instead of float for reference voltage

drivers/adc/ad738x/iio_ad738x.c
```diff
@@ -0,0 +1,207 @@
...
+	struct iio_device *iio_dev;
+};
+
+/**
+ * @brief       Read the debug register value
+ * @param       dev[in, out]- Pointer to IIO device instance
+ * @param       reg[in]- Register address to read from
+ * @param       readval[out]- Pointer to variable to read data into
+ * @return      0 in case of success, negative value otherwise
+ */
```
rbolboac, line 52: the text added near param label should be the same as the one used in the function definition - please check in the whole file. also, we do not really use in/out for parameter description

drivers/adc/ad738x/iio_ad738x.c
```diff
@@ -0,0 +1,207 @@
...
+	struct ad738x_dev *dev = iio_dev->ad738x_dev;
+
+	return ad738x_spi_reg_write(dev, (uint8_t)reg, (uint16_t)val);
+}
+
+static int get_raw(void *device,
+		   char *buf,
+		   uint32_t len,
+		   const struct iio_ch_info *channel,
+		   intptr_t id)
```
rbolboac, line 82: no need to add each parameter on a new line, just write the parameters on the same line and let astyle take care of the formatting for you. see everywhere where it applies.

drivers/adc/ad738x/iio_ad738x.c
```diff
@@ -0,0 +1,207 @@
...
+	return snprintf(buf, len, "%10f", ((double)ADC_REF_VOLTAGE / ADC_MAX_COUNT_UNIPOLAR));
+}
+
+static int get_offset(void *device,
+		      char *buf,
+		      uint32_t len,
+		      const struct iio_ch_info *channel,
+		      intptr_t id)
+{
+	return snprintf(buf, len, "%d", 0);
```
rbolboac, line 110: if offset is 0 and cannot be modified... there is no need to have an offset attribute

drivers/adc/ad738x/iio_ad738x.c
```diff
@@ -0,0 +1,207 @@
...
+static struct scan_type ad738x_iio_scan_type = {
+	.sign = 'u',
+	.realbits = CHN_STORAGE_BITS,
+	.storagebits = CHN_STORAGE_BITS,
+};
+
+static struct iio_attribute ad738x_iio_ch_attributes[] = {
+	{ .name = "raw", .show = get_raw },
+	{ .name = "scale", .show = get_scale, },
+	{ .name = "offset", .show = get_offset, },
```
rbolboac, line 159: as mentioned above, can be removed

drivers/adc/ad738x/iio_ad738x.h
```diff
@@ -0,0 +1,40 @@
...
+struct ad738x_iio_dev {
+	struct ad738x_dev *ad738x_dev;
+	struct iio_device *iio_dev;
+};
+
+/* Init the IIO interface */
+int ad738x_iio_init(struct ad738x_iio_dev **dev,
+		    struct ad738x_init_param *init_param);
+
+void iio_ad738x_event_handler(struct iio_desc *desc);
```
rbolboac, line 38: I don't see a definition for this... what is the purpose of this declaration?

projects/ad738x-st/Makefile
```diff
@@ -0,0 +1,11 @@
+# Select the example you want to enable by chossing y for enabling and n for disabling
+IIO_EXAMPLE = y
+
+# Select the device you want to enable by choosing y for enabling and n for disabling
```
rbolboac, line 4: remove comment

projects/ad738x-st/Makefile
```diff
@@ -0,0 +1,11 @@
...
+include ../../tools/scripts/generic_variables.mk
+
+include src.mk
+
+include ../../tools/scripts/generic.mk
+
+CFLAGS += -O2
```
rbolboac, line 11: any reason why this is added?

projects/ad738x-st/src/examples/examples_src.mk
```diff
@@ -0,0 +1,22 @@
+ifeq (y,$(strip $(IIO_EXAMPLE)))
+CFLAGS += -DIIO_EXAMPLE=1 -DIIO_SUPPORT -DINTC_DEVICE_ID=1 -DUSE_STANDARD_SPI
```
rbolboac, line 2: where is USE_STANDARD_SPI used and why is it defined?

even if the result is the same, would move the definition of INTC_DEVICE_ID in parameters.h. also, any reason why 1 was chosen? I think there is only one interrupt vector controller for stm32 and by default in the other projects it has id 0.

projects/ad738x-st/src/examples/iio/iio_example.c
```diff
@@ -0,0 +1,92 @@
...
+#include "iio_ad738x.h"
+#include "no_os_util.h"
+#include "no_os_gpio.h"
+#include "no_os_print_log.h"
+#include "iio_app.h"
+
+/******************************************************************************/
+/************************ Functions Declarations ******************************/
+/******************************************************************************/
+/******************************************************************************/
```
rbolboac, line 59: remove duplicated line

projects/ad738x-st/src/examples/iio/iio_example.c
```diff
@@ -0,0 +1,92 @@
...
+	iio_devices[0].dev_descriptor = dev->iio_dev;
+	app_init_param.devices = iio_devices;
+	app_init_param.nb_devices = NO_OS_ARRAY_SIZE(iio_devices);
+	app_init_param.uart_init_params = ad738x_uart_ip;
+
+	ret = iio_app_init(&app, app_init_param);
+	if (ret)
+		return ret;
+
+	return iio_app_run(app);
```
rbolboac, line 91: before returning, in case of an error iio_app_remove should be called, and also ad7616_iio_remove. you may take a look here and follow the example: https://github.com/analogdevicesinc/no-OS/blob/master/projects/max22200/src/examples/iio_example/iio_example.c

projects/ad738x-st/src/examples/iio/iio_example.h
```diff
@@ -0,0 +1,52 @@
...
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+*******************************************************************************/
+#ifndef __IIO_EXAMPLE_H__
+#define __IIO_EXAMPLE_H__
+
+/******************************************************************************/
+/***************************** Include Files **********************************/
+/******************************************************************************/
```
rbolboac, line 45: can be removed because there are no includes

projects/ad738x-st/sdp-ck1z.ioc
```diff
@@ -0,0 +1,158 @@
...
+Mcu.Pin5=PH0/OSC_IN
+Mcu.Pin6=PH1/OSC_OUT
+Mcu.Pin7=PA7
+Mcu.Pin8=VP_SYS_VS_Systick
+Mcu.PinsNb=9
+Mcu.ThirdPartyNb=0
+Mcu.UserConstants=
+Mcu.UserName=STM32F469NIHx
+MxCube.Version=6.10.0
+MxDb.Version=DB.6.0.100
```
rbolboac, line 32: MxCube.Version=6.5.0
MxDb.Version=DB.6.0.50
Please use these values instead, these work on the CI and please add the builds.json file for this project so that the CI is building it and the binaries are added in the release

projects/ad738x-st/src/platform/stm32/main.c
```diff
@@ -0,0 +1,75 @@
...
+	ad738x_ip.spi_param = &ad738x_spi_ip;
+	ad738x_ip.conv_mode = ONE_WIRE_MODE;
+	ad738x_ip.ref_sel = INT_REF;
+	ad738x_spi_extra_ip.get_input_clock = HAL_RCC_GetPCLK2Freq;
+
+	stm32_init();
+#ifdef IIO_EXAMPLE
+	ret = iio_example_main();
+	if (ret)
+		return ret;
```
rbolboac, line 69: these can be removed and instead of returning 0, ret can be returned at line 74

rbolboac (commented):
some small remarks from my side.
Also. seeing how we are already in 2024, it would be better to have 2024 added in the header of the files.

drivers/adc/ad738x/iio_ad738x.h
ahaslam2: we dont need this, got copied from the ad7616 project that i used as a base.

projects/ad738x-st/Makefile
ahaslam2: better performance. but ill remove it, and will add it in another PR with performance improvements.

projects/ad738x-st/src/examples/examples_src.mk
ahaslam2: the ./drivers/adc/ad738x/ad738x.c file uses for offload spi. in the k1/stm32 we use standard spi
so we need to define this flag. 

https://github.com/analogdevicesinc/no-OS/blob/master/drivers/adc/ad738x/ad738x.c#L348

no idea why INTC_DEVICE_ID is 1, this was from the project i copied to start. will test with 0 and change if i dont see any issue.

projects/ad738x-st/Makefile
rbolboac: in general, we keep the default -O0 seeing how these are project examples which are expected to be built, run and debugged on targets by our customers. 
if better performance is needed by a release (for a client) I would advice using RELEASE=y in builds.json which would generate a binary compiled with O2 which can be used to program the target directly.
see https://github.com/analogdevicesinc/no-OS/blob/master/tools/scripts/generic.mk#L169C1-L172C1

projects/ad738x-st/builds.json
```diff
@@ -0,0 +1,7 @@
+{
+  "stm32": {
+    "basic_example": {
```
rbolboac, line 3: instead of basic example, rename to iio_example

ahaslam2: ok, not sure what the CI was looking for in this file..

projects/ad738x-st/builds.json
```diff
@@ -0,0 +1,7 @@
...
+      "flags" : "IIO_EXAMPLE=y"
```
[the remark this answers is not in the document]
ahaslam2: ill add this flag with another patch set for performance improvements.

ahaslam2:
changes v4:
- update copyright date to 2024
- rename builds.json test
- More verbose  commit message

ahaslam2:
changes v4:

update copyright date to 2024
rename builds.json test and add HARDWARE flag
More verbose commit message

projects/ad738x-st/builds.json
rbolboac: well, you have the platform (stm32), the name of the example (iio_example) and then the compilation flags
the name of the binary in the release file will use these values, the final result being:
nameoftheprojectfolder_platform_nameoftheexample
so basically in your case it would be
ad738x-st_stm32_basic_example for a binary which is not the standard example, that is why it is better to use ad738x-st_stm32_iio_example to make it clear that this example uses iio.

drivers/adc/ad738x/iio_ad738x.c
```diff
@@ -0,0 +1,200 @@
...
+ * @param val - Pointer to variable to read data into
+ * @return 0 in case of success, negative value otherwise
+ */
+static int32_t iio_ad738x_debug_reg_read(void *device, uint32_t reg,
+		uint32_t *val)
+{
+	struct ad738x_iio_dev *iio_dev = device;
+	struct ad738x_dev *dev = iio_dev->ad738x_dev;
+
+	return ad738x_spi_reg_read(dev,  (uint8_t) reg, (uint16_t *)val);
```
buha, line 59: I'm pretty sure this may create bugs on processor with a certain endianess. When you cast that pointer to uint16_t * you're essentially saying the ad738x_spi_reg_read to fill in the uppermost or lowermost (endianess based) 2 bytes of the 4 bytes that is being used to represent val. And when val is interpreted as uint32_t, in one of the scenarios you'll end up with a wrong value. You might need to create an temporary variable here to achieve what you want.

drivers/adc/ad738x/iio_ad738x.c
```diff
@@ -0,0 +1,200 @@
...
+	.ch_type = IIO_VOLTAGE,\
+	.ch_out = IIO_DIRECTION_INPUT,\
+	.indexed = true,\
+	.channel = _idx,\
+	.scan_index = _idx,\
+	.scan_type = &ad738x_iio_scan_type,\
+	.attributes = ad738x_iio_ch_attributes,\
+}
+
+#define CHN_STORAGE_BITS	16
```
buha, line 36: in general it is a good practice to put some prefix to these macros like AD738X_*

drivers/adc/ad738x/iio_ad738x.c
```diff
@@ -0,0 +1,200 @@
...
+#define ADC_RESOLUTION		16
+#define ADC_REF_VOLTAGE		3300
```
buha, line 38: I guess this reference voltage may change from one project to another, from one hardware to another.
In this case it's probably better to initialize it from project through init param and not hardcode it into the driver (unless it's the only possible ref voltage, i didn't check the datasheet).

ahaslam2:
changes v5:
- avoid pointer cast in reg_debug_read
- add prefix in macro names
- move vref voltage to init param

drivers/adc/ad738x/ad738x.c
```diff
@@ -351,6 +351,7 @@ int32_t ad738x_init(struct ad738x_dev **device,
 #endif
 	dev->conv_mode = init_param->conv_mode;
 	dev->ref_sel = init_param->ref_sel;
+	dev->ref_voltage = init_param->ref_voltage;
```
rbolboac, line 354: I would move the change from ad738x driver, regarding the ref_voltage in a different commit. It is not really related to this commit about IIO support.

drivers/adc/ad738x/ad738x.h
```diff
@@ -177,6 +178,7 @@ struct ad738x_init_param {
 	/* Device Settings */
 	enum ad738x_conv_mode	conv_mode;
 	enum ad738x_ref_sel		ref_sel;
+	uint32_t		ref_voltage;
```
rbolboac, line 181: some documentation should be given here regarding the measurement unit (volts or microvolts)

drivers/adc/ad738x/iio_ad738x.c
```diff
@@ -0,0 +1,210 @@
...
+}
+
+static int get_scale(void *device, char *buf, uint32_t len,
+		     const struct iio_ch_info *channel, intptr_t id)
+{
+	struct ad738x_iio_dev *iio_dev = device;
+	struct ad738x_dev *dev = iio_dev->ad738x_dev;
+
+	return snprintf(buf, len, "%10f",
+			((double)dev->ref_voltage / AD738X_ADC_MAX_COUNT_UNIPOLAR));
```
rbolboac, line 106: In general, we do not make floating point operations... there are other ways to return the scale in a string format using iio_format_value using only integers.
you may take a look at other projects which are using iio_format_value for scale and you may also take a look at the ad738x linux driver to see how the scale is computes.

rbolboac (commented):
just a few remarks

drivers/adc/ad738x/ad738x.h
ahaslam2: how about i rename to  ref_voltage_mv?

rbolboac: even better!

ahaslam2:
changes v6:

- rename ref_voltage to ref_voltage_mv and move to separte patch
- use iio_format_value instead of double operation

drivers/adc/ad738x/iio_ad738x.c
```diff
@@ -0,0 +1,212 @@
...
+static int32_t iio_ad738x_debug_reg_write(void *device, uint32_t reg,
+		uint32_t val)
+{
+	struct ad738x_iio_dev *iio_dev = device;
+	struct ad738x_dev *dev = iio_dev->ad738x_dev;
+
+	return ad738x_spi_reg_write(dev, (uint8_t)reg, (uint16_t)val);
+}
+
+static int get_raw(void *device, char *buf, uint32_t len,
```
rbolboac, line 84: missing API description, see everywhere where it occurs

drivers/adc/ad738x/iio_ad738x.h
```diff
@@ -0,0 +1,38 @@
...
+
+struct ad738x_iio_dev {
+	struct ad738x_dev *ad738x_dev;
+	struct iio_device *iio_dev;
+};
+
+/* Init the IIO interface */
+int ad738x_iio_init(struct ad738x_iio_dev **dev,
+		    struct ad738x_init_param *init_param);
+int ad738x_iio_remove(struct ad738x_iio_dev *dev);
```
rbolboac, line 37: missing doxygen comment

rbolboac (approved):
all in all looks good to me.
I'm not sure how fast this needs to be merged... but it would be nice to also have at least the documentation of the project. similarly to what is done here maybe: https://github.com/analogdevicesinc/no-OS/tree/main/projects/eval-adis1650x#readme 
or whatever information you think it is useful such that a new customer is able to reproduce the setup and test the project. from my point of view it can be added with a different PR.
it would be even better if the driver + iio driver documentation would be made available, similarly to what we already have available on GitHub pages: http://analogdevicesinc.github.io/no-OS/drivers_doc.html

PR #2062, opened 2024-01-08
drivers: api: Fix variable assignment

CiprianRegus:
The intent is to set i2c_table[bus_number] to NULL, so that we know the memory has been freed. However, the current implementation just makes the bus variable point to NULL, while i2c_table[bus_number] remains unchaged.

Running the init(), remove(), init() sequence will result in using freed memory after the second init. Fix this by correctly setting the i2c_table array element to NULL.

Fixes: d4a9723 ("spi and i2c: added slave counter for deallocation check")

## Pull Request Description

## PR Type

## PR Checklist

### DIFF
--- a/drivers/api/no_os_i2c.c
+++ b/drivers/api/no_os_i2c.c
@@ -147,7 +147,7 @@ void no_os_i2cbus_remove(uint32_t bus_number)
 
 		if (bus != NULL) {
 			no_os_free(bus);
-			bus = NULL;
+			i2c_table[bus_number] = NULL;
 		}
 	}
 }


amiclaus (approved):

rbolboac (approved):

drivers/api/no_os_i2c.c
```diff
@@ -147,7 +147,7 @@ void no_os_i2cbus_remove(uint32_t bus_number)
 
 		if (bus != NULL) {
 			no_os_free(bus);
-			bus = NULL;
+			i2c_table[bus_number] = NULL;
```
buha, line 150: It could be that the reason this doesn't behave as expected is that first, bus is not assigned correctly at line 140, it should be (not the & character):
```
&i2c_table[bus_number];
```

And on line 150 where you made the change, `*bus = NULL` should do.

Could you please try this ?

PR #2065, opened 2024-01-09
Allow for IP setting at compile time

CiprianRegus:
## Pull Request Description
Add compile flags for IP, netmask and gateway. These take effect when using the networking layer based on lwip sockets and allow the user to configure these parameters, instead of using the predefined ones we currently have. Change project makefiles in order to keep the current IP settings.

The format should be (similar for the other ones): NO_OS_IP = A.B.C.D

## PR Type

## PR Checklist


CiprianRegus:
Renamed the `NO_OS_STATIC_IP` macro to `SWIOT1L_STATIC_IP` and `APARD32690_STATIC_IP`, since it's scope is now limited to single projects (the lwip networking layer doesn't use it directly anymore).

buha (approved):

amiclaus (approved):

PR #2075, opened 2024-01-15
iio:iio.c: Add cyclic mode size checks in pop scan

ribdp:
## Pull Request Description

1. Use no_os_cb_read even in cyclic mode to check if there are samples available to be read
2. Refactor the function implementation

## PR Type

## PR Checklist

### DIFF
--- a/iio/iio.c
+++ b/iio/iio.c
@@ -1423,18 +1423,16 @@ int iio_buffer_pop_scan(struct iio_buffer *buffer, void *data)
 	if (!buffer)
 		return -EINVAL;
 
-	if(!buffer->cyclic_info.is_cyclic)
-		return no_os_cb_read(buffer->buf, data, buffer->bytes_per_scan);
+	int ret;
 
-	memcpy(data,
-	       &buffer->buf->buff[buffer->cyclic_info.buff_index],
-	       buffer->bytes_per_scan);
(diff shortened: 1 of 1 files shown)


iio/iio.c
```diff
@@ -1423,18 +1423,16 @@ int iio_buffer_pop_scan(struct iio_buffer *buffer, void *data)
 	if (!buffer)
 		return -EINVAL;
 
-	if(!buffer->cyclic_info.is_cyclic)
-		return no_os_cb_read(buffer->buf, data, buffer->bytes_per_scan);
+	uint32_t ret;
```
rbolboac, line 1426: should be int,, otherwise lgtm

ribdp: Fixed this now

buha:
I think @rbolboac meant 'int', not 'int32' to match the function return type.

rbolboac:
> I think @rbolboac meant 'int', not 'int32' to match the function return type.

Yes, so I think a while ago we decided to use int instead of int32_t for new APIs so it will be best to use int also in this case, even though the result is the same because int = int32_t

ribdp:
Changelog - 
1. Changed int32_t -> int

rbolboac (approved):

ribdp:
is the PR good to be merged?

PR #2078, opened 2024-01-18
Update ad7616-sdz project

Skallwar:
## Pull Request Description

The `ad7616-sdz` project was found broken recently. This PR aims to make it work again using the latest version of the HDL.

## PR Type

## PR Checklist

### DIFF
--- a/drivers/adc/ad7616/ad7616.c
+++ b/drivers/adc/ad7616/ad7616.c
@@ -66,8 +66,9 @@
 #include "no_os_spi.h"
 #endif
 
-#define AD7616_GET_BIT(v, b) ((v >> b) & 1)
-#define AD7616_TOGGLE_TIMEOUT_DELAY (8192 * 1024)
+#define AD7616_GET_BIT(v, b)		((v >> b) & 1)
+#define AD7616_TOGGLE_TIMEOUT_DELAY	(8192 * 1024)
+#define SPI_ENGINE_WORDS_PER_READ	1
 
 /**
  * @brief Compute the CRC for one channel. See CRC chapter in the datasheet
@@ -623,9 +624,13 @@ static int32_t ad7616_read_data_serial_zynqmp(struct ad7616_dev *dev,
 	struct spi_engine_offload_message msg;
 	uint32_t spi_eng_msg_cmds[3] = {
 		CS_LOW,
-		READ(2),
+		READ(SPI_ENGINE_WORDS_PER_READ),
 		CS_HIGH,
 	};
+
+	ret = no_os_pwm_enable(dev->trigger_pwm_desc);
+	if (ret != 0)
+		return ret;
 
 	dev->spi_desc->mode = NO_OS_SPI_MODE_3;
 	spi_engine_set_speed(dev->spi_desc, dev->spi_desc->max_speed_hz);
@@ -649,7 +654,7 @@ static int32_t ad7616_read_data_serial_zynqmp(struct ad7616_dev *dev,
 	no_os_axi_io_write(dev->core_baseaddr, AD7616_REG_UP_CTRL, AD7616_CTRL_RESETN);
 
 	if (dev->dcache_invalidate_range)
-		dev->dcache_invalidate_range(msg.rx_addr, samples * 2);
+		dev->dcache_invalidate_range(msg.rx_addr, samples * 4);
 
 	return ret;
 }
@@ -952,6 +957,18 @@ int32_t ad7616_setup(struct ad7616_dev **device,
 	}
 
 #ifdef XILINX_PLATFORM
+	ret = axi_clkgen_init(&dev->clkgen, init_param->clkgen_init);
+	if (ret != 0)
+		goto cleanup;
+
+	ret = axi_clkgen_set_rate(dev->clkgen, init_param->axi_clkgen_rate);
+	if (ret != 0)
+		goto cleanup;
+
+	ret = no_os_pwm_init(&dev->trigger_pwm_desc, init_param->trigger_pwm_init);
+	if (ret != 0)
+		goto cleanup;
+
 	dev->core_baseaddr = init_param->core_baseaddr;
 	dev->offload_init_param = init_param->offload_init_param;
 	dev->reg_access_speed = init_param->reg_access_speed;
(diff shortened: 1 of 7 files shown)


drivers/adc/ad7616/ad7616.c
```diff
@@ -952,6 +956,18 @@ int32_t ad7616_setup(struct ad7616_dev **device,
 	}
 
 #ifdef XILINX_PLATFORM
+	ret = axi_clkgen_init(&dev->clkgen, init_param->clkgen_init);
+	if (ret != 0)
+		return ret;
```
rbolboac, line 961: dev should be freed before returning

drivers/adc/ad7616/ad7616.c
```diff
@@ -952,6 +956,18 @@ int32_t ad7616_setup(struct ad7616_dev **device,
...
+
+	ret = axi_clkgen_set_rate(dev->clkgen, init_param->axi_clkgen_rate);
+	if (ret != 0)
+		return ret;
```
rbolboac, line 965: dev should be freed before returning and axi_clkgen_remove should be called.

drivers/adc/ad7616/ad7616.c
```diff
@@ -952,6 +956,18 @@ int32_t ad7616_setup(struct ad7616_dev **device,
...
+
+	ret = no_os_pwm_init(&dev->trigger_pwm_desc, init_param->trigger_pwm_init);
+	if (ret != 0)
+		return ret;
```
rbolboac, line 969: dev should be freed before returning and axi_clkgen_remove should be called.

rbolboac: also, no_os_pwm_remove should be called extra if an error occurs in the lines below.
I propose to refactor this function to treat all the potential memory leaks.

rbolboac (commented):
some comments from my side, also, the last commit is missing a commit message... and the project build is failing for some reason, I am attaching the logs from the CI here:
[ad7616-sdz_xilinx_demo_ad7616_sdz_zed.txt](https://github.com/analogdevicesinc/no-OS/files/13985925/ad7616-sdz_xilinx_demo_ad7616_sdz_zed.txt)
[ad7616-sdz_xilinx_demo_ad7616_sdz_zc706.txt](https://github.com/analogdevicesinc/no-OS/files/13985931/ad7616-sdz_xilinx_demo_ad7616_sdz_zc706.txt)

drivers/adc/ad7616/ad7616.c
```diff
@@ -623,10 +623,14 @@ static int32_t ad7616_read_data_serial_zynqmp(struct ad7616_dev *dev,
 	struct spi_engine_offload_message msg;
 	uint32_t spi_eng_msg_cmds[3] = {
 		CS_LOW,
-		READ(2),
+		READ(1),
```
danmois, line 626: Maybe using a define here instead of  1 would increase readability of the code.

Skallwar:
> the project build is failing for some reason, I am attaching the logs from the CI here

I think it's because the `system_top.xsa` used by the CI is not the one this PR is compatible with. The one that should be can be found [here](https://github.com/analogdevicesinc/no-OS/blob/update_ad7616_noos_spi_engine_v2/projects/ad7616-sdz/system_top.xsa). I'm not sure where to push it to make the CI work

rbolboac:
> > the project build is failing for some reason, I am attaching the logs from the CI here
> 
> I think it's because the `system_top.xsa` used by the CI is not the one this PR is compatible with. The one that should be can be found [here](https://github.com/analogdevicesinc/no-OS/blob/update_ad7616_noos_spi_engine_v2/projects/ad7616-sdz/system_top.xsa). I'm not sure where to push it to make the CI work

Currently, the CI is running with an older version of .xsa. 
When we update the artifacts the CI is using we will re-trigger a build on this PR to make sure everything works as expected.

drivers/adc/ad7616/ad7616.c
```diff
@@ -1039,19 +1052,31 @@ int32_t ad7616_setup(struct ad7616_dev **device,
...
-			return ret;
+			goto cleanup;
 	} else {
 		dev->crc = 0;
 	}
 
 	return ad7616_self_test(dev);
+
+cleanup:
+	ad7616_remove(dev);
```
rbolboac, line 1064: so...please take a look at which APIs are called and which ones require a remove operation.
calling remove in this case (for memory which has not been allocated yet) leads to undefined behavior. 
so, for each API call the according remove function should be called only if the memory was allocated previously.

Skallwar: That was the idea at first but then I found out that all the remove function called are doing null checks before freeing the memory (except for `axi_clkgen_remove`, hence the check at line 1075):
- [no_os_pwm_remove](https://github.com/analogdevicesinc/no-OS/blob/0b437cd6b8145324dbac3dd2d59dabe2a67284f3/drivers/api/no_os_pwm.c#L87)
- [no_os_spi_remove](https://github.com/analogdevicesinc/no-OS/blob/0b437cd6b8145324dbac3dd2d59dabe2a67284f3/drivers/api/no_os_spi.c#L127). On this one I think there might be a bug because `desc` is dereferenced *before* the null check. I don't have the full context so I might be wrong also
- [no_os_gpio_remove](https://github.com/analogdevicesinc/no-OS/blob/0b437cd6b8145324dbac3dd2d59dabe2a67284f3/drivers/api/no_os_gpio.c#L112)

I can add null check everywhere in the remove function if you prefer, but I will avoid the goto tag cascade as much as possible in the init function IMAO.

During the preparation of my commit it seems that I forgot to include the switch from `no_os_malloc` to `no_os_calloc` during the device creation. I will add it now if we want to keep only one goto tag

CiprianRegus:
Please update the commit message for this one 2a9c1c03f. Provide some additional details about what was broken and what you did to fix it.

drivers/adc/ad7616/ad7616.c
rbolboac: using calloc will do the trick, so the remove call can be kept as is.

drivers/adc/ad7616/iio_ad7616.c
```diff
@@ -383,18 +383,12 @@ int ad7616_iio_init(struct ad7616_iio_dev **dev,
...
 
-	ret = ad7616_remove(dev->ad7616_dev);
-	if (ret)
-		return ret;
+	ad7616_remove(dev->ad7616_dev);
 
 	no_os_free(dev);
-
-	return ret;
 }
```
rbolboac, line 392: please add a return here

drivers/adc/ad7616/ad7616.c
Skallwar: Ok then, perfect.

> On this one I think there might be a bug because desc is dereferenced before the null check. I don't have the full context so I might be wrong also

I still think this should be address later

rbolboac: I will take care of the spi remove issue.

rbolboac:
please rebase with main to confirm that the build for the project passes. thank you!

rbolboac (approved):

drivers/adc/ad7616/ad7616.c
```diff
@@ -1050,7 +1050,7 @@ int32_t ad7616_setup(struct ad7616_dev **device,
 int ad7616_remove(struct ad7616_dev *dev)
 {
 	if (!dev)
-		return -ENODEV;
+		return;
```
rbolboac, line 1072: I just saw this, this change should be in the commit.

Skallwar: I will revert this, my bad

projects/ad7616-sdz/src/ad7616_sdz.c
```diff
@@ -70,9 +74,29 @@ struct spi_engine_offload_init_param spi_engine_offload_init_param = {
...
+struct axi_clkgen_init clkgen_init = {
+	.name = "rx_clkgen",
+	.base = AD7616_RX_CLKGEN_BASEADDR,
+	.parent_rate = 100000000,
+};
+
+struct axi_pwm_init_param axi_pwm_init_params = {
+	.base_addr = AXI_PWMGEN_BASEADDR,
+	.ref_clock_Hz = 100000000,
+	//.channel = 0,
```
rbolboac, line 91: in general we do not leave commented code in commits, if this is not needed simply remove it.

danmois (approved):

Skallwar:
The CI failed for other projects than AD7616. I will address the remaining comment shortly

drivers/adc/ad7616/ad7616.c
```diff
@@ -623,7 +624,7 @@ static int32_t ad7616_read_data_serial_zynqmp(struct ad7616_dev *dev,
 	struct spi_engine_offload_message msg;
 	uint32_t spi_eng_msg_cmds[3] = {
 		CS_LOW,
-		READ(1),
+		READ(SPI_ENGINE_WORDS_PER_READ),
```
rbolboac, line 627: same here

rbolboac:
errors are the same as the ones present on main, merging this PR

PR #2082, opened 2024-01-23
Dev/i2c spi remove fix

rbolboac:
## Pull Request Description
Call i2cbus_remove only if desc is not null to avoid undefined behavior.
Call spibus_remove only if desc is not null to avoid undefined behavior.
## PR Type

## PR Checklist

### DIFF
--- a/drivers/api/no_os_i2c.c
+++ b/drivers/api/no_os_i2c.c
@@ -118,11 +118,11 @@ int32_t no_os_i2cbus_init(const struct no_os_i2c_init_param *param)
  */
 int32_t no_os_i2c_remove(struct no_os_i2c_desc *desc)
 {
-	// Remove I2C bus
-	no_os_i2cbus_remove(desc->bus->device_id);
-
 	if (!desc || !desc->platform_ops)
 		return -EINVAL;
+
+	if (desc->bus)
+		no_os_i2cbus_remove(desc->bus->device_id);
 
 	if (!desc->platform_ops->i2c_ops_remove)
 		return -ENOSYS;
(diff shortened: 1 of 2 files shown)


drivers/api/no_os_i2c.c
```diff
@@ -118,12 +118,12 @@ int32_t no_os_i2cbus_init(const struct no_os_i2c_init_param *param)
  */
 int32_t no_os_i2c_remove(struct no_os_i2c_desc *desc)
 {
-	// Remove I2C bus
-	no_os_i2cbus_remove(desc->bus->device_id);
-
 	if (!desc || !desc->platform_ops)
 		return -EINVAL;
 
+	// Remove I2C bus
```
CiprianRegus, line 124: This may be removed.

drivers/api/no_os_spi.c
```diff
@@ -121,12 +121,12 @@ int32_t no_os_spibus_init(const struct no_os_spi_init_param *param)
  */
 int32_t no_os_spi_remove(struct no_os_spi_desc *desc)
 {
-	// Remove SPI bus
-	no_os_spibus_remove(desc->bus->device_id);
-
 	if (!desc || !desc->platform_ops)
 		return -EINVAL;
 
+	// Remove SPI bus
```
CiprianRegus, line 127: This may be removed.

drivers/api/no_os_spi.c
```diff
@@ -121,12 +121,12 @@ int32_t no_os_spibus_init(const struct no_os_spi_init_param *param)
...
+	no_os_spibus_remove(desc->bus->device_id);
```
amiclaus, line 128: wondering if it is worth checking desc->bus == NULL prior to calling this function. or calling this function if bus != NULL.

amiclaus (commented):
commits are a bit messed up: https://github.com/analogdevicesinc/no-OS/pull/2082/commits/4a8c2ff281a8ef1b000adcb92f25fe164f188047 has changes in both i2c and spi for example.

CiprianRegus (approved):

Skallwar (approved):

amiclaus (approved):

PR #2083, opened 2024-01-24
Performance improvement for the ad7616 drivers and its SDP-K1 project 

Skallwar:
## Pull Request Description

This aims to improve the sample rate of the ad7616 projects.

## PR Type

## PR Checklist

### DIFF
--- a/drivers/adc/ad7616/iio_ad7616.c
+++ b/drivers/adc/ad7616/iio_ad7616.c
@@ -249,6 +249,51 @@ static int iio_ad7616_attr_set(void *device, char *buf, uint32_t len,
 	}
 }
 
+/*
+ * @brief Setup AD7616's sequencer to ask the channel found in the iio buffer active_mask
+ * @param dev - The ad7616 device structure
+ * @param active_mask - IIO buffer active_mask
+ * @return Result of call to ad7616_setup_sequencer
+ */
+static int32_t setup_sequencer_layers_from_active_mask(struct ad7616_dev *dev,
+		uint32_t active_mask)
+{
+	uint32_t i;
+	uint32_t index_a, index_b;
+	uint8_t active_mask_a = active_mask & 0xFF;
+	uint8_t active_mask_b = (active_mask >> 8) & 0xFF;
+	uint32_t nb_layers = no_os_max(no_os_hweight8(active_mask_a),
+				       no_os_hweight8(active_mask_b));
+	enum ad7616_ch ch_a, ch_b;
+	struct ad7616_sequencer_layer layers[nb_layers];
+
+	for (i = 0; i < nb_layers; i++) {
+		index_a = no_os_find_first_set_bit(active_mask_a);
+		if (index_a == 32) {
+			// No more A channel requested, default to self test channel
+			ch_a = AD7616_VA_SELF_TEST;
+		} else {
+			active_mask_a = active_mask_a & ~NO_OS_BIT(index_a);
+			ch_a = AD7616_VA0 + index_a;
+		}
+
+		index_b = no_os_find_first_set_bit(active_mask_b);
+		if (index_b == 32) {
+			// No more B channel requested, default to self test channel
+			ch_b = AD7616_VB_SELF_TEST;
+		} else {
+			active_mask_b = active_mask_b & ~NO_OS_BIT(index_b);
+			ch_b = AD7616_VB0 + index_b;
+		}
+
+		layers[i] = (struct ad7616_sequencer_layer) {
+			ch_a, ch_b
+		};
+	}
+
+	return ad7616_setup_sequencer(dev, layers, nb_layers, 1);
+}
+
 /**
  * @brief	Read buffer data corresponding to AD7616 IIO device
  * @param	iio_dev_data - Pointer to IIO device data structure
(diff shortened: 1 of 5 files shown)


Skallwar:
The cppcheck error looks like a false positive to me. Leaking an integer on return?

I don't have this warning on my laptop with cppcheck 2.12.0

CiprianRegus:
I'd say it's because you're previously allocating memory in the same function and you might return without calling `free()`.

drivers/adc/ad7616/ad7616.c
```diff
@@ -909,10 +911,12 @@ int32_t ad7616_self_test(struct ad7616_dev *dev)
 	if (ret != 0)
 		return ret;
 
-	// First toggle to remove pending garbage
-	ret = ad7616_toggle_conv(dev);
-	if (ret != 0)
-		return ret;
+	// First couple of toggles return garbage values, discard them
```
rbolboac, line 914: any idea why this is happening? there should be an explanation from the data-sheet here

drivers/adc/ad7616/iio_ad7616.c
```diff
@@ -249,38 +249,108 @@ static int iio_ad7616_attr_set(void *device, char *buf, uint32_t len,
 	}
 }
 
+/*
+ * @brief Setup AD7616's sequencer to ask the channel found in the iio buffer active_mask
+ * @param dev - The ad7616 device structure
+ * @param nb_channels - Number of requested channels
```
rbolboac, line 255: this parameter is not present, remove from documentation

drivers/adc/ad7616/iio_ad7616.c
```diff
@@ -249,38 +249,108 @@ static int iio_ad7616_attr_set(void *device, char *buf, uint32_t len,
...
+	uint8_t active_mask_a;
+	uint8_t active_mask_b;
+	uint32_t index_a;
+	uint32_t index_b;
 	int32_t ret;
 	uint32_t i;
 	uint32_t j;
+	uint32_t k;
+
+	results = no_os_malloc(dev->layers_nb * iio_dev_data->buffer->samples * sizeof(
```
rbolboac, line 318: please fix the memory leaks in this API and use no_os_calloc instead to set the memory to 0 by default

drivers/adc/ad7616/ad7616.c
Skallwar: The first one is expected when reading the datasheet. But when I compile the project in release mode (-O2) the second one is also not correct.
On a side note a similar issue exist on the `ad7616_sdz` where the first 4 values are not correct (either 0 or value that does not correspond to the signal being sampled)

rbolboac: it sounds like a timing issue, I think it should be investigated what is the root-cause of this... does adding a delay after a single ad7616_toggle_conv call produce similar results?

Skallwar: > it sounds like a timing issue

I agree.

> does adding a delay after a single ad7616_toggle_conv call produce similar results?

I tested that without success at some point. The configuration is a bit different now so I will test again

Skallwar: I'm not able to reproduce the issue at the moment. I definitely had the issue in previous versions where without this fix it was working 3 out of 4 times.

As I can't reproduce the bug at the moment I think I can remove the fix for now but I would not be surprised if we encounter this issue again

drivers/adc/ad7616/iio_ad7616.c
```diff
@@ -249,46 +249,120 @@ static int iio_ad7616_attr_set(void *device, char *buf, uint32_t len,
...
-		if (ret)
-			return ret;
+		active_mask_a = iio_dev_data->buffer->active_mask & 0xFF;
+		active_mask_b = (iio_dev_data->buffer->active_mask >> 8) & 0xFF;
+		k = 0;
 
-		active_mask = iio_dev_data->buffer->active_mask;
+		for (j = 0; active_mask_a; j++) {
+			index_a = no_os_find_first_set_bit(active_mask_a);
+			if (index_a == 32)
```
rbolboac, line 339: instead of 32 maybe use a macro to better understand what this value represents

rbolboac:
- also please rebase with main and resolve conflicts

drivers/adc/ad7616/ad7616.c
rbolboac: I would say to remove the fix. As long as we do not know the root-cause (and cannot reproduce the issue currently) we can not say for sure that this is the fix, it might do the trick for your setup, but for another setup it might not.

drivers/adc/ad7616/iio_ad7616.c
rbolboac: Alright, so I took a look... and 32 is returned when no bit is set.
If that is the case, does the check really makes sense? The for loop will stop (because the mask is 0) and that if condition will never be true, right?

CiprianRegus (approved):

drivers/adc/ad7616/iio_ad7616.c
```diff
@@ -249,46 +249,114 @@ static int iio_ad7616_attr_set(void *device, char *buf, uint32_t len,
...
-			return ret;
+			goto cleanup;
 	}
 
 	return 0;
+
+cleanup:
+	no_os_free(results);
+	return ret;
+
```
amiclaus, line 359: nitpick: this line can be dropped.

amiclaus (approved):
lgtm. one nit.

Skallwar:
The CI is failing but that does not seem to be related to this PR. I rebased on top of main just in case, but that did not fixed it

drivers/adc/ad7616/iio_ad7616.c
Skallwar: Done

amiclaus (commented):
> The CI is failing but that does not seem to be related to this PR. I rebased on top of main just in case, but that did not fixed it

Indeed the CI fails are unrelated. We will have soon a fix for that.

PR #2085, opened 2024-01-25
Maxim IRQ fixes

CiprianRegus:
## Pull Request Description
- set event lists to NULL in order to mark them as free.
- implement reference counting for IRQ descriptors 

## PR Type

## PR Checklist

### DIFF
--- a/drivers/api/no_os_irq.c
+++ b/drivers/api/no_os_irq.c
@@ -63,6 +63,7 @@ int32_t no_os_irq_ctrl_init(struct no_os_irq_ctrl_desc **desc,
 	if (ret)
 		return ret;
 
+	(*desc)->ref++;
 	(*desc)->platform_ops = param->platform_ops;
 
 	return 0;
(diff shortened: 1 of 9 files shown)


rbolboac (approved):
tested and works

buha:
I've tested iio_demo on max78000 and max32650 fthr boards, it does fix things for both in the sense that the uart reinit succeeds. However, on max78000 the project actually works by connecting with an iio client but on max32650 it doesn't. This issue may or may not be related to the symptom you fixed in the 1st commit, I don't have time to investigate more now.

drivers/api/no_os_irq.c
```diff
@@ -63,6 +63,7 @@ int32_t no_os_irq_ctrl_init(struct no_os_irq_ctrl_desc **desc,
 	if (ret)
 		return ret;
 
+	(*desc)->ref++;
```
buha, line 66: can ref be anything else than 1 and 0 ?

CiprianRegus: Depending on the platform specific implementation, yes. STM32 for example, has only one reference to the IRQ descriptor. If you call init() multiple times, it will use the same pointer which is only allocated once. So in this case `ref` can be anything. I think multiple platforms can benefit from this reference counting mechanism, since they can use only one pointer to an IRQ descriptor.

buha: ok makes sense

buha (approved):

PR #2086, opened 2024-01-26
drivers: digital-io: max14919 : Add MAX14919 driver support

RaduSabau1:
## Pull Request Description

The MAX14919/MAX14919A industrial-protected quad-channel lowside switch features 140mΩ (typ) on-resistance (RON) per channel with integrated ±1kV/42Ω surge protection for robust operation.

Resistor-settable accurate current limiting provides guaranteed operating currents in the range of 100mA to 800mA. Loads that draw large activation or inrush currents are supported using the 2x inrush load-current option. The outputs can be connected in parallel to achieve higher load currents. The four switches are pin-controlled to allow for simple and fast switching of up to 500kHz.

The MAX14919/MAX14919A features reverse current detection. The MAX14919 implements reverse-current protection by driving an external FET for non-capacitive loads. The MAX14919A has reverse current indication.

The device itself doesn't have any SPI/I2C communication protocol, so the driver consists off no-OS GPIOs structures commanding GPIO pins to be high or low in order to set the OUTPUT channels states, get FAULT information, get current polarity information or set current limiting.

## PR Type

## PR Checklist

### DIFF
--- a/drivers/digital-io/max14919/max14919.c
+++ b/drivers/digital-io/max14919/max14919.c
@@ -1,0 +1,213 @@
+/***************************************************************************//**
+ *   @file   max14919.c
+ *   @brief  Source file of MAX14919 Driver.
+ *   @author Radu Sabau (radu.sabau@analog.com)
+********************************************************************************
+ * Copyright 2024(c) Analog Devices, Inc.
+ *
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *  - Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ *  - Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in
+ *    the documentation and/or other materials provided with the
+ *    distribution.
+ *  - Neither the name of Analog Devices, Inc. nor the names of its
+ *    contributors may be used to endorse or promote products derived
+ *    from this software without specific prior written permission.
+ *  - The use of this software may or may not infringe the patent rights
+ *    of one or more patent holders.  This license does not release you
+ *    from the requirement that you obtain separate licenses from these
+ *    patent holders to use this software.
+ *  - Use of the software either in source or binary form, must be run
+ *    on or directly connected to an Analog Devices Inc. component.
+ *
+ * THIS SOFTWARE IS PROVIDED BY ANALOG DEVICES "AS IS" AND ANY EXPRESS OR
+ * IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, NON-INFRINGEMENT,
+ * MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED.
+ * IN NO EVENT SHALL ANALOG DEVICES BE LIABLE FOR ANY DIRECT, INDIRECT,
+ * INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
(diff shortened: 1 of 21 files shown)


drivers/digital-io/max14919/max14919.c
```diff
@@ -0,0 +1,204 @@
...
+ * @brief Set the OUT channels state
+ * @param desc - MAX14919 device descriptor
+ * @param state - OUT channels state array containing information about the to
+ * 		  be set channels state.
+ * @return 0 in case of succes, negative error code otherwise
+*/
+int max14919_set_out(struct max14919_desc *desc, enum max14919_out_state *state)
+{
+	int ret, i;
+
```
rbolboac, line 52: as an extra precaution it would be good to check whether desc, desc->in_desc and desc->out_state are not null seeing how they are accessed here.

drivers/digital-io/max14919/max14919.c
```diff
@@ -0,0 +1,204 @@
...
+ * @brief Set current limiting
+ * @param desc - MAX14919 devce descriptor
+ * @param state - State to be set.
+ * 		  TRUE - on.
+ * 		  FALSE - off.
+ * @return 0 in case of succes, negative error code otherwise
+*/
+int max14919_set_climit(struct max14919_desc *desc, bool state)
+{
+	if (!desc->inrush_desc)
```
rbolboac, line 75: also check desc

drivers/digital-io/max14919/max14919.c
```diff
@@ -0,0 +1,204 @@
...
+ * @param desc - MAX14919 device descriptor
+ * @return 0 - a fault has occurred.
+ * 	   1 - no faults have been reported.
+*/
+int max14919_get_fault(struct max14919_desc *desc)
+{
+	uint8_t fault;
+	int ret;
+
+	ret = no_os_gpio_get_value(desc->fault_desc, &fault);
```
rbolboac, line 93: check desc pointer before and return fault with pointer, we always return the status of the API execution with return, and any other values over parameters with pointers.
see everywhere where applicable

drivers/digital-io/max14919/max14919.c
```diff
@@ -0,0 +1,204 @@
...
+	return rev;
+}
+
+/**
+ * @brief Initialize the MAX14919 device descriptor
+ * @param desc - MAX14919 device descriptor
+ * @param init_param - MAX14919 initialization parameter
+ * @return 0 in case of succes, negative error code otherwise
+*/
+int max14919_init(struct max14919_desc **desc,
```
rbolboac, line 127: fix all memory leaks in this API, similar to what you did in other drivers

rbolboac (changes requested):
some remarks from my side, also please fix the build issues.

drivers/digital-io/max14919/max14919.c
rbolboac: this has not been resolved, please return fault value as pointer

RaduSabau1:
V3

- Fixed cppcheck that caused it to fail.
- get_fault and detect_rev API now return the state as a pointer.

drivers/digital-io/max14919/max14919.c
RaduSabau1: Solved it.

RaduSabau1:
V4

- Approached different way of freeing the memory in case of an error being returned at init function.

RaduSabau1:
V5

- Returning ret so the error code doesn't get lost.

rbolboac:
For the project documentation you should also create a .rst file in sphinx/projects similarly to what you did for the driver and you should also update the projects_doc.rst accordingly. this will probably solve the Documentation build error.

rbolboac (approved):

rbolboac:
- please solve the conflicts

CiprianRegus (approved):

PR #2102, opened 2024-02-14
doc: Add Makefile and repository to conf.py

gastmaier:
## Pull Request Description

Adds the default Sphinx Makefile with vars `SOURCEDIR` and `BUILDDIR` set.
I need this file to exist because [doctools](https://github.com/analogdevicesinc/doctools/blob/8192817d818b9fe0cd34dc69034619f9936d930d/adi_doctools/cli/aggregate.py#L157) needs to know the source and build paths.
Set values are consistent with the ones at [documentation.sh](https://github.com/analogdevicesinc/no-OS/blob/0f275e683906480b9478e4a01f2d8d983f328ff9/ci/documentation.sh#L113).
Created it in the `doc/sphinx/source` path instead of `doc/sphinx`  because it would be the only file at the latter path.
Also, includes `repository` var to `conf.py`, also used by doctools.

## PR Type
- [X] New feature (change that adds new functionality)

## PR Checklist

### DIFF
--- a/doc/sphinx/source/conf.py
+++ b/doc/sphinx/source/conf.py
@@ -17,6 +17,7 @@
 
 # -- Project information -----------------------------------------------------
 
+repository = 'no-OS'
 project = 'no-OS'
 copyright = '2023, Analog Devices, Inc.'
 author = 'Analog Devices, Inc.'
--- a/doc/sphinx/source/Makefile
+++ b/doc/sphinx/source/Makefile
@@ -1,0 +1,13 @@
+SHELL         = /bin/bash
+SPHINXOPTS   ?=
+SPHINXBUILD  ?= sphinx-build
+SOURCEDIR     = .
+BUILDDIR      = ../build
+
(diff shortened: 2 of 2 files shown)


doc/sphinx/source/conf.py
```diff
@@ -17,6 +17,7 @@
 
 # -- Project information -----------------------------------------------------
 
+repository = 'no-os'
```
amiclaus, line 20: maybe use uppercase for `OS` to match the actual name of the repository.

amiclaus (commented):
look good to me. please update your commit description since it is empty. you can use the PR description.

amiclaus (commented):
also i think there is a typo in your PR description: `Created it in the doc/sphinx/source path instead of doc/sphinx/source`

amiclaus (approved):
thanks, this looks good to me. after the merge I will add the Makefile usage also in the ci procedure.

PR #2105, opened 2024-02-20
ad400x: Support for pulsar HDL and STM32

ahaslam2:
## Pull Request Description

This PR adds:
* support for stm32 with IIO
* changes for  new HDL that contains:
  - fixes for simple spi, 
  - pwm
  - clkgen
 
HDL is not merged yet on main repo, but code was tested with:
https://github.com/adi-ses/sea-misc/tree/main/AD4000-PRJ/debug_single_conversion

## PR Type

## PR Checklist


drivers/adc/ad400x/ad400x.h
```diff
@@ -79,6 +79,8 @@ struct ad400x_dev {
 	uint32_t reg_access_speed;
 	/* Device Settings */
 	enum ad400x_supported_dev_ids dev_id;
+	/* Voltage reference */
+	uint32_t ref_voltage_mv;
```
amiclaus, line 89: if you add this new parameter, i would add also it's usage in the same commit. Moreover, where do you pass the init param for this to the device descriptor? I see dev->ref_voltage_mv is used in the iio driver.

projects/ad400x-st/README.rst
```diff
@@ -0,0 +1,104 @@
+Evaluating AD400x on SDP-K1
```
amiclaus, line 1: documentation usually goes in separate commit.

drivers/adc/ad400x/ad400x.h
ahaslam2: the ref_voltage_mv shoud have been copied to the dev descriptor on the driver, but this got deleted in one of my manipulations. will fix.  and ill squash it o the iio driver patch so that the usage is included.

projects/ad400x-st/README.rst
ahaslam2: ok.

drivers/adc/ad400x/iio_ad400x.c
```diff
@@ -0,0 +1,226 @@
...
+
+	desc->iio_dev = &ad400x_iio_device_template;
+
+	/* fill scan_type based on device id */
+	memcpy(&ad400x_iio_scan_type, &ad400x_iio_scan_types[init_param->dev_id],
+	       sizeof(ad400x_iio_scan_type));
+	ad400x_iio_scan_type.is_big_endian = true;
+	ad400x_iio_scan_type.storagebits = 32;
+	ad400x_iio_scan_type.shift =
+		ad400x_iio_scan_type.storagebits - ad400x_iio_scan_type.realbits;
```
rbolboac, line 201: all this data is knows and can be added in ad400x_iio_scan_types... then you can simply do
desc->iio_dev->channels[0].scan_type  = &ad400x_iio_scan_types[init_param->dev_id] and remove the extra variable ad400x_iio_scan_type

drivers/adc/ad400x/iio_ad400x.c
```diff
@@ -0,0 +1,226 @@
...
+#include "iio_ad400x.h"
+#include "iio.h"
+#include "no_os_util.h"
+#include "no_os_alloc.h"
+#include "no_os_error.h"
+
+struct ad400x_iio {
+	struct ad400x_dev *drv_dev;
+	struct iio_device *iio_dev;
+};
```
rbolboac, line 28: this is not being used, please remove it

drivers/adc/ad400x/iio_ad400x.c
```diff
@@ -0,0 +1,226 @@
...
+	struct ad400x_iio_dev *iio_dev = device;
+	struct ad400x_dev *dev = iio_dev->ad400x_dev;
+	uint32_t results = 0;
+	int32_t ret;
+
+	ret = ad400x_spi_single_conversion(dev, &results);
+	if (ret)
+		return ret;
+
+	return sprintf(buf, "%lx", results);
```
rbolboac, line 97: please use iio_format_value

drivers/adc/ad400x/iio_ad400x.c
```diff
@@ -0,0 +1,226 @@
...
+
+	return ad400x_spi_reg_write(dev, (uint8_t)val);
+}
+
+static int get_raw(void *device, char *buf, uint32_t len,
+		   const struct iio_ch_info *channel, intptr_t id)
+{
+	struct ad400x_iio_dev *iio_dev = device;
+	struct ad400x_dev *dev = iio_dev->ad400x_dev;
+	uint32_t results = 0;
```
rbolboac, line 90: no need to initialize it

drivers/adc/ad400x/iio_ad400x.c
```diff
@@ -0,0 +1,226 @@
...
+	int32_t ret;
```
rbolboac, line 91: please use int, instead of int32_t everywhere in this file

drivers/adc/ad400x/iio_ad400x.c
```diff
@@ -0,0 +1,226 @@
...
+}
+
+static int get_scale(void *device, char *buf, uint32_t len,
+		     const struct iio_ch_info *channel, intptr_t id)
+{
+	struct ad400x_iio_dev *iio_dev = device;
+	struct ad400x_dev *dev = iio_dev->ad400x_dev;
+	int32_t vals[2];
+
+	vals[0] = 5000;
```
rbolboac: the reference should be given by the user (it can vary between 2.4 and 5V), so add it in iio_init_param (which has to be created and give in ad400x_iio_init function)

drivers/adc/ad400x/iio_ad400x.c
```diff
@@ -0,0 +1,226 @@
...
+	vals[1] = ad400x_iio_scan_types[dev->dev_id].realbits;
+
+	return iio_format_value(buf, len, IIO_VAL_FRACTIONAL, 1, vals);
```
rbolboac, line 110: this is incorrect, the correct scale is v_ref/2^(realbits), in this case you are returining v_ref/realbits. use IIO_VAL_LOG2_FRACTIONAL instead of IIO_VAL_FRACTIONAL to obtain the correct result

drivers/adc/ad400x/iio_ad400x.c
```diff
@@ -0,0 +1,226 @@
...
+	struct ad400x_iio_dev *iio_dev = iio_dev_data->dev;
+	struct ad400x_dev *dev = iio_dev->ad400x_dev;
+	uint32_t samples = iio_dev_data->buffer->samples;
+	uint32_t *results;
+	uint32_t i;
+	int32_t ret;
+
+	results = no_os_malloc(sizeof(*results) * samples);
+	if (!results)
+		return -ENOMEM;
```
rbolboac, line 129: you do not need to allocate memory... a simple uint32_t result; variable will suffice.
simply have a single for loop in which you read and push 
ret = iio_buffer_push_scan(iio_dev_data->buffer, &result);

drivers/adc/ad400x/ad400x.c
```diff
@@ -182,6 +182,8 @@ int32_t ad400x_init(struct ad400x_dev **device,
 	if (!dev)
 		return -1;
 
+	dev->ref_voltage_mv = init_param->ref_voltage_mv;
```
rbolboac, line 185: As I have already mentioned... this parameter doe not have anything to do with the driver, it should be a parameter in the iio_driver only. and you should add it in the same commit in which you add the iio_driver

projects/ad400x-st/src/common/common_data.c
```diff
@@ -73,5 +73,6 @@ struct ad400x_init_param ad400x_ip = {
 		.number = GPIO_CNV,
 		.platform_ops = GPIO_OPS,
 	},
+	.ref_voltage_mv = AD400X_ADC_REF_VOLTAGE,
```
rbolboac, line 76: after adding ref_voltage_mv in iio_driver, please add these changes in the project commit file (this commit should not be needed if the changes are performed in the other commits accordingly)

rbolboac (commented):
I still have to think a little bit about the spi chip select changes you have submitted, until then here are my remarks.

drivers/adc/ad400x/iio_ad400x.c
ahaslam2: yes, this is done on the last patch of this PR.. i got a separate comment that the patch that adds this should show the a user. that why its hardcoded right now but is replaced on the patch that adds ref_voltage_ms.

drivers/adc/ad400x/iio_ad400x.c
ahaslam2: you are right. plus i need to take into account if the signal is "signed" or "unsigned"
will fix.

drivers/adc/ad400x/iio_ad400x.c
ahaslam2: originally i had it like that,  but there is a  performance gain if we remove
form the capture loop the iio_buffer_push_scan. 

this was done on the ad7616 too. BTW.

drivers/adc/ad400x/ad400x.c
ahaslam2: ok

projects/ad400x-st/src/common/common_data.c
ahaslam2: ok,

drivers/adc/ad400x/iio_ad400x.h
```diff
@@ -0,0 +1,37 @@
...
+
+struct ad400x_iio_dev {
+	struct ad400x_dev *ad400x_dev;
+	struct iio_device *iio_dev;
+};
+
+/* Init the IIO interface */
+int ad400x_iio_init(struct ad400x_iio_dev **dev,
+		    struct ad400x_init_param *init_param);
+int ad400x_iio_remove(struct ad400x_iio_dev *dev);
```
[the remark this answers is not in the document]
ahaslam2: done

drivers/adc/ad400x/iio_ad400x.c
ahaslam2: ill change it to get the basic stuff merged, and we will see about performance later.

ahaslam2:
> I still have to think a little bit about the spi chip select changes you have submitted, until then here are my remarks.

There is onother option with these chips is  leave the CS go low, but this would require SDI to be high.
this could be done by setting buf=0xFFFFFFFF before the spi operation.

drivers/adc/ad400x/iio_ad400x.h
```diff
@@ -0,0 +1,37 @@
...
+* @file   iio_ad400x.h
+* @brief  Header file for AD400X IIO interface
+********************************************************************************
+* Copyright (c) 2024 Analog Devices, Inc.
+* All rights reserved.
+*
+* This software is proprietary to Analog Devices, Inc. and its licensors.
+* By using this software you agree to the terms of the associated
+* Analog Devices Software License Agreement.
+*******************************************************************************/
```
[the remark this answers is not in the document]
ahaslam2: i changed it to the same as iio_ad463x

drivers/adc/ad400x/iio_ad400x.c
ahaslam2: ok.. i had it like this originally, but i thought it would be better to avoid the big table.  but will revert back to this way.

ahaslam2:
changes in v2:

- voltage ref patch squashed into iio/project patches.
- changed licence agreement text got iio driver
- expanded iio scan type table, and removed global scan_type variable
- fixed scale factor
- removed alloc from capture loop, used single 32 bit variable
- add driver remove on iio remove
- other minor fixes.

ahaslam2:
changes in v2:

voltage ref patch squashed into iio/project patches.
changed licence agreement text in iio driver
expanded iio scan type table, and removed global scan_type variable
fixed scale factor
removed alloc from capture loop, used single 32 bit variable
add driver remove on iio remove
other minor fixes.

ahaslam2:
changes in v2:

voltage ref patch squashed into iio/project patches.
changed licence agreement text in iio driver
expanded iio scan type table, and removed global scan_type variable
fixed scale factor
removed alloc from capture loop, used single 32 bit variable
add driver remove on iio remove
other minor fixes.

rbolboac:
> > I still have to think a little bit about the spi chip select changes you have submitted, until then here are my remarks.
> 
> There is onother option with these chips is leave the CS go low, but this would require SDI to be high. this could be done by setting buf=0xFFFFFFFF before the spi operation.

So please correct me if I am wrong: all you actually need for this driver is to have chip-select active high instead of active low, is that right?
If this is the case, I think it would be better to have an cs_active_high flag in the spi driver. the chip select is actual part of the SPI protocol and I find strange to make it optional in a spi driver. however, I think there are cases in which the polarity is inverted... this would ma a more acceptable approach.

drivers/adc/ad400x/ad400x.c
```diff
@@ -195,9 +195,9 @@ int32_t ad400x_init(struct ad400x_dev **device,
 	if (ret < 0)
 		goto error;
 
+	dev->dev_id = init_param->dev_id;
```
rbolboac, line 198: this changes does not look related to this commit

drivers/adc/ad400x/ad400x.h
```diff
@@ -96,6 +96,8 @@ struct ad400x_init_param {
...
```
rbolboac, line 100: this is not needed here

drivers/adc/ad400x/iio_ad400x.c
```diff
@@ -0,0 +1,236 @@
...
+ * MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED.
+ * IN NO EVENT SHALL ANALOG DEVICES BE LIABLE FOR ANY DIRECT, INDIRECT,
+ * INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, INTELLECTUAL PROPERTY RIGHTS, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+*******************************************************************************/
+#ifdef IIO_SUPPORT
```
rbolboac, line 39: I think this is not necessary

drivers/adc/ad400x/iio_ad400x.c
```diff
@@ -0,0 +1,236 @@
...
+	struct ad400x_dev *dev = iio_dev->ad400x_dev;
+	struct scan_type *scan_type = &ad400x_iio_scan_types[dev->dev_id];
+	uint32_t results;
+	int ret;
+
+	ret = ad400x_spi_single_conversion(dev, &results);
+	if (ret)
+		return ret;
+
+	no_os_memswap64(&results, 4, 4);
```
rbolboac, line 118: why are you choosing this API? this is for memory blocks, simply use no_os_get_unaligned_be32.

drivers/adc/ad400x/iio_ad400x.c
```diff
@@ -0,0 +1,236 @@
...
+	results >>= scan_type->shift;
+
+	return sprintf(buf, "%ld", results);
```
rbolboac, line 121: please use iio_format_value

drivers/adc/ad400x/iio_ad400x.c
```diff
@@ -0,0 +1,236 @@
...
+	struct scan_type *scan_type = &ad400x_iio_scan_types[dev->dev_id];
+	uint32_t results;
+	int ret;
+
+	ret = ad400x_spi_single_conversion(dev, &results);
+	if (ret)
+		return ret;
+
+	no_os_memswap64(&results, 4, 4);
+	results >>= scan_type->shift;
```
rbolboac, line 119: based on the raw data type format (2's complement or binary), you will also have to get the sign and append the result with the sign bit. 
this is correct for binary data, for 2's complement you will have to use sign_extend (see examples in other drivers)

drivers/adc/ad400x/iio_ad400x.h
```diff
@@ -0,0 +1,74 @@
...
+ */
+struct ad400x_iio_dev {
+	/** AD400x driver handler */
+	struct ad400x_dev *ad400x_dev;
+	/** Generic IIO device handler */
+	struct iio_device *iio_dev;
+	/* Voltage reference */
+	uint32_t ref_voltage_mv;
+};
+
```
rbolboac, line 68: you should create another structure here:
```
struct ad400x_iio_ip {
  struct ad400x_init_param *init_param;
  /* Voltage reference */
  uint32_t ref_voltage_mv;
}
```
and have the following template for init:
```
int ad400x_iio_init(struct ad400x_iio_dev **dev,
		    struct ad400x_iio_ip *init_param);
```
there is no reason to have ref_voltage_mv in the ad400x_init_param, because it is not used in ad400x driver, it is only used in the scope of iio driver, so it only should be defined in the scope of iio driver.

drivers/adc/ad400x/iio_ad400x.c
```diff
@@ -0,0 +1,236 @@
...
+
+	*dev = desc;
+
+	return 0;
+
+error_setup:
+	no_os_free(desc);
+	return ret;
+}
+
```
rbolboac, line 227: missing API description

projects/ad400x-st/README.rst
```diff
@@ -0,0 +1,104 @@
...
+* Industrial Automation Technology
+* Instrumentation and Measurement Solutions
+* Intelligent Building Solutions
+* Healthcare Solutions
+* Aerospace and Defense Systems
+* Wireless Communication Solutions
+
+Hardware Specifications
+-----------------------
+
```
rbolboac, line 43: nothing added in this section, please remove it if it is not needed

drivers/adc/ad400x/ad400x.c
ahaslam2: ok,

drivers/adc/ad400x/ad400x.h
ahaslam2: ok

drivers/adc/ad400x/iio_ad400x.c
ahaslam2: ok

drivers/adc/ad400x/iio_ad400x.h
ahaslam2: ok.

projects/ad400x-st/README.rst
ahaslam2: ok

ahaslam2:
changes v4:
- toggle cnv gpio on register read/write
- remove ref_voltage from driver init param
- extend sign in case of signed channel

rbolboac:
> > > > I still have to think a little bit about the spi chip select changes you have submitted, until then here are my remarks.
> > > 
> > > 
> > > There is onother option with these chips is leave the CS go low, but this would require SDI to be high. this could be done by setting buf=0xFFFFFFFF before the spi operation.
> > 
> > 
> > So please correct me if I am wrong: all you actually need for this driver is to have chip-select active high instead of active low, is that right? If this is the case, I think it would be better to have an cs_active_high flag in the spi driver. the chip select is actual part of the SPI protocol and I find strange to make it optional in a spi driver. however, I think there are cases in which the polarity is inverted... this would ma a more acceptable approach.
> 
> this is not exactly right,he chip does not have a CS pin. its a conversion pin.
> 
> the issue with just inverting the polarity of the CS pin on the spi driver, is that in fact to read/write the control register the CNV pin has to be low :( (just realized need to add a toggle in register read/write too)
> 
> if you want to avoid handling the gpio on the driver: the other option: setting buf to 0xFFFFFFFFF so SDI is high would allow for CS pin to be low and read samples. the issue with this is that the conversion starts on a rising edge of CNV, so if i read 1 sample for example: the sample would be from past: from when the last SPI read finished and CS toggled from low to high. would this be acceptable?
> 
> the above is what i see on the datasheet and from experiments: https://www.analog.com/media/en/technical-documentation/data-sheets/ad4000-4004-4008.pdf

So we would like to avoid changing the SPI driver like that (seeing how for other platforms we will have to adapt this kind of behavior and it might not be possible to not have a chip select configured)... A workaround would be to simply have a dummy chip select in the project (which will not be connected to the chip) and then gpio_cnv pin can be used however is needed directly in the driver.

ahaslam2:
> > > > > I still have to think a little bit about the spi chip select changes you have submitted, until then here are my remarks.
> > > > 
> > > > 
> > > > There is onother option with these chips is leave the CS go low, but this would require SDI to be high. this could be done by setting buf=0xFFFFFFFF before the spi operation.
> > > 
> > > 
> > > So please correct me if I am wrong: all you actually need for this driver is to have chip-select active high instead of active low, is that right? If this is the case, I think it would be better to have an cs_active_high flag in the spi driver. the chip select is actual part of the SPI protocol and I find strange to make it optional in a spi driver. however, I think there are cases in which the polarity is inverted... this would ma a more acceptable approach.
> > 
> > 
> > this is not exactly right,he chip does not have a CS pin. its a conversion pin.
> > the issue with just inverting the polarity of the CS pin on the spi driver, is that in fact to read/write the control register the CNV pin has to be low :( (just realized need to add a toggle in register read/write too)
> > if you want to avoid handling the gpio on the driver: the other option: setting buf to 0xFFFFFFFFF so SDI is high would allow for CS pin to be low and read samples. the issue with this is that the conversion starts on a rising edge of CNV, so if i read 1 sample for example: the sample would be from past: from when the last SPI read finished and CS toggled from low to high. would this be acceptable?
> > the above is what i see on the datasheet and from experiments: https://www.analog.com/media/en/technical-documentation/data-sheets/ad4000-4004-4008.pdf
> 
> So we would like to avoid changing the SPI driver like that (seeing how for other platforms we will have to adapt this kind of behavior and it might not be possible to not have a chip select configured)... A workaround would be to simply have a dummy chip select in the project (which will not be connected to the chip) and then gpio_cnv pin can be used however is needed directly in the driver.

ok, ill do that.

drivers/adc/ad400x/iio_ad400x.c
```diff
@@ -0,0 +1,243 @@
...
+	struct ad400x_dev *dev = iio_dev->ad400x_dev;
+	struct scan_type *scan_type = &ad400x_iio_scan_types[dev->dev_id];
+	uint32_t results;
+	int ret;
+
+	ret = ad400x_spi_single_conversion(dev, &results);
+	if (ret)
+		return ret;
+
+	results = no_os_get_unaligned_be32((uint8_t *)&results);
```
rbolboac, line 116: actually, this is not entirely correct. so the ad400x_spi_single_conversion is expected to return the conversion result as read from the device but in the processor endianess. so this no_os_get_unaligned_be32 should actually be called in ad400x_spi_single_conversion, after reading the SPI data... further more it should also be shifted (in my opinion).

drivers/adc/ad400x/iio_ad400x.c
```diff
@@ -0,0 +1,243 @@
...
+	results >>= scan_type->shift;
+
+	if (scan_type->sign == 's')
+		results = no_os_sign_extend32(results, scan_type->realbits);
```
rbolboac, line 120: you can use a int32_t extra variable here to store the result

drivers/adc/ad400x/iio_ad400x.c
```diff
@@ -0,0 +1,243 @@
...
+
+/**
+ * @brief Release resources.
+ * @param iio_dev - iio device.
+ * @return 0 in case of success, -1 otherwise.
+ */
+int ad400x_iio_remove(struct ad400x_iio_dev *iio_dev)
+{
+	if (!iio_dev)
+		return -1;
```
rbolboac, line 237: -EINVAL

drivers/adc/ad400x/iio_ad400x.c
ahaslam2: is it not up to the final application to check the scan_type of iio_info, and apply all these bit operations to display the raw_data correctly?

i would think that the driver would want to get sample data out as fast as possible. if we do bit operations in ad400x_spi_single_conversion it will affect also the loop in iio_ad400x_submit_buffer.. which will have a performance impact.

ahaslam2:
changes in v5:
- dont mask high nibble of buffer (new)
- return EINVAL instead of -1
- fix parameter documentation.

ahaslam2:
changes in v5:

1. dont mask high nibble of buffer (new)
2. fix sign of read_raw 
3. return EINVAL instead of -1
4. fix parameter documentation.

ahaslam2:
changes in v5:

- dont mask high nibble of buffer (new)
- fix sign of read_raw
- return EINVAL instead of -1
- fix parameter documentation.
- fix ad4000 realbit size

drivers/adc/ad400x/iio_ad400x.c
ahaslam2: @rbolboac are you ok with leaving this bit operations in this function rather than moving them to ad400x_spi_single_conversion? 

if i do move the bit operation i would have to:

- wrap the bit operations in #if flags for simple spi as HDL does not return the same byte/bit order.
- crate a new function that does not apply bit operation for use in the get data buffer loop

rbolboac: It is fine for me but please describe in ad400x_spi_single_conversion API the exact format of the result

drivers/adc/ad400x/ad400x.c
```diff
@@ -164,7 +164,7 @@ int32_t ad400x_spi_single_conversion(struct ad400x_dev *dev,
 	if (ret)
 		return ret;
 #endif
-	*adc_data = buf & 0xFFFFF;
+	*adc_data = buf & 0xFFFFFF;
```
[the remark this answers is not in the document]
ahaslam2: bit endianess i meant bit order.  ill changed it.  
the datasheet only specifies that 20 bits of data are sent MSB first. because i was always just seeing  16bits, i removed the mask and i confirmed that the higher nibble of the 3rd byte is the one that contains the LSB bits (which i do see when probing the spi bus).

drivers/adc/ad400x/iio_ad400x.c
ahaslam2: ill double check. (need to build HDL for 400x) but from what i see in the code the exact format of the result seems to depend if we use HDL or simple spi. I think HDL might be doing some magic manipulations.

rbolboac: please keep in mind that the result should be the same for both platforms

rbolboac (commented):
it looks like there are still some things which are not clear.
some things I want to point out:
- the driver and the iio driver should be platform independent -> in this particular case it means that the ad400x_spi_single_conversion API should return exactly the same data format in both cases so the IIO driver can handle it correctly.
- do you have the means to test this code? it should be tested and we should make sure the behavior is the same on both platforms. iio_info command can be used to compare the behavior

ahaslam2:
> it looks like there are still some things which are not clear. some things I want to point out:
> 
> * the driver and the iio driver should be platform independent -> in this particular case it means that the ad400x_spi_single_conversion API should return exactly the same data format in both cases so the IIO driver can handle it correctly.
> * do you have the means to test this code? it should be tested and we should make sure the behavior is the same on both platforms. iio_info command can be used to compare the behavior

Since HDL and standard SPI are not returning the same byte order, we will need to do bit operation 
in  ad400x_spi_single_conversion just for the get_raw attribute, which i wanted to avoid for the buffer data case.
perhaps we need to change HDL to return the same as standard SPI?

i did test on the k1 (stm32) with the ad400x boards.
for the bare metal xilinx platform (with HDL) we see from the code that they print the value as it comes
so it must be little endiann. but i did not actually run this.

rbolboac:
> > it looks like there are still some things which are not clear. some things I want to point out:
> > 
> > * the driver and the iio driver should be platform independent -> in this particular case it means that the ad400x_spi_single_conversion API should return exactly the same data format in both cases so the IIO driver can handle it correctly.
> > * do you have the means to test this code? it should be tested and we should make sure the behavior is the same on both platforms. iio_info command can be used to compare the behavior
> 
> Currently there is no stand-alone HDL with iio project. its either xilinx HDL no IIO or stm32 standard SPI with IIO.
> 
> Since HDL and standard SPI are not returning the same byte order, we will need to do bit operation in ad400x_spi_single_conversion just for the get_raw attribute, which i wanted to avoid for the buffer data case. perhaps we need to change HDL to return the same as standard SPI? or create a new driver function for standard spi? perhaps ill do that but in the iio dirver there would still have to be a way to make the difference between the platforms
> 
> why not just add "#if defined(USE_STANDARD_SPI)" for these bit operations in the iio driver?
> 
> here is what the xilinx based platfrom is doing:
> 
> ```
>                 while(1) {
>                         ad400x_spi_single_conversion(dev, &adc_data);
>                         xil_printf("ADC: %d\n\r", adc_data);
>                 }
> ```
> 
> i did test on the k1 (stm32) with the ad400x boards. for the bare metal xilinx platform (with HDL) we see from the code that they print the value as it comes so it must be little endiann. but i did not actually run this.

the IIO driver should be platform independent... I understand why the driver is not (because SPI transactions are handled differently in HDL), but the result should be the same. 

it does not really matter that there is no IIO project with HDL, if there would be one done, still the same iio driver would be used.
so let's make ad400x_spi_single_conversion API return the same result, regardless of the platform, and handle it accordingly in IIO driver, even if the buffered data will take longer. then we do not need to make a difference between platform in the iio driver.

ahaslam2:
> > > it looks like there are still some things which are not clear. some things I want to point out:
> > > 
> > > * the driver and the iio driver should be platform independent -> in this particular case it means that the ad400x_spi_single_conversion API should return exactly the same data format in both cases so the IIO driver can handle it correctly.
> > > * do you have the means to test this code? it should be tested and we should make sure the behavior is the same on both platforms. iio_info command can be used to compare the behavior
> > 
> > 
> > Currently there is no stand-alone HDL with iio project. its either xilinx HDL no IIO or stm32 standard SPI with IIO.
> > Since HDL and standard SPI are not returning the same byte order, we will need to do bit operation in ad400x_spi_single_conversion just for the get_raw attribute, which i wanted to avoid for the buffer data case. perhaps we need to change HDL to return the same as standard SPI? or create a new driver function for standard spi? perhaps ill do that but in the iio dirver there would still have to be a way to make the difference between the platforms
> > why not just add "#if defined(USE_STANDARD_SPI)" for these bit operations in the iio driver?
> > here is what the xilinx based platfrom is doing:
> > ```
> >                 while(1) {
> >                         ad400x_spi_single_conversion(dev, &adc_data);
> >                         xil_printf("ADC: %d\n\r", adc_data);
> >                 }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > i did test on the k1 (stm32) with the ad400x boards. for the bare metal xilinx platform (with HDL) we see from the code that they print the value as it comes so it must be little endiann. but i did not actually run this.
> 
> the IIO driver should be platform independent... I understand why the driver is not (because SPI transactions are handled differently in HDL), but the result should be the same.
> 
> it does not really matter that there is no IIO project with HDL, if there would be one done, still the same iio driver would be used. so let's make ad400x_spi_single_conversion API return the same result, regardless of the platform, and handle it accordingly in IIO driver, even if the buffered data will take longer. then we do not need to make a difference between platform in the iio driver.

Ok.. will do that, we are  paying a performance impact per sample because HDL is flipping bytes.

ahaslam2:
> > > > it looks like there are still some things which are not clear. some things I want to point out:
> > > > 
> > > > * the driver and the iio driver should be platform independent -> in this particular case it means that the ad400x_spi_single_conversion API should return exactly the same data format in both cases so the IIO driver can handle it correctly.
> > > > * do you have the means to test this code? it should be tested and we should make sure the behavior is the same on both platforms. iio_info command can be used to compare the behavior
> > > 
> > > 
> > > Currently there is no stand-alone HDL with iio project. its either xilinx HDL no IIO or stm32 standard SPI with IIO.
> > > Since HDL and standard SPI are not returning the same byte order, we will need to do bit operation in ad400x_spi_single_conversion just for the get_raw attribute, which i wanted to avoid for the buffer data case. perhaps we need to change HDL to return the same as standard SPI? or create a new driver function for standard spi? perhaps ill do that but in the iio dirver there would still have to be a way to make the difference between the platforms
> > > why not just add "#if defined(USE_STANDARD_SPI)" for these bit operations in the iio driver?
> > > here is what the xilinx based platfrom is doing:
> > > ```
> > >                 while(1) {
> > >                         ad400x_spi_single_conversion(dev, &adc_data);
> > >                         xil_printf("ADC: %d\n\r", adc_data);
> > >                 }
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > i did test on the k1 (stm32) with the ad400x boards. for the bare metal xilinx platform (with HDL) we see from the code that they print the value as it comes so it must be little endiann. but i did not actually run this.
> > 
> > 
> > the IIO driver should be platform independent... I understand why the driver is not (because SPI transactions are handled differently in HDL), but the result should be the same.
> > it does not really matter that there is no IIO project with HDL, if there would be one done, still the same iio driver would be used. so let's make ad400x_spi_single_conversion API return the same result, regardless of the platform, and handle it accordingly in IIO driver, even if the buffered data will take longer. then we do not need to make a difference between platform in the iio driver.
> 
> Ok.. will do that, we are paying a performance impact per sample because HDL is flipping bytes.

For info, the HDL and driver is being worked on after the issue was raised: 
https://github.com/analogdevicesinc/no-OS/commit/5a56faeab4731b2a0df855b5e7be201af42b00a3

once those fixes are fixes are finalized and merged.

ahaslam2:
@amiclaus @rbolboac 

The HDL was fixed for simple spi case, and this comes with mandatory support of PWM and CLKGEN. 
the HDL dev branch was:  https://github.com/analogdevicesinc/no-OS/compare/update_ad400x_noos_spi_engine
the HDL dev binary is https://github.com/adi-ses/sea-misc/tree/main/AD4000-PRJ/debug_single_conversion

This series now integrates the changes for this, so single conversion function works
for both HDL and simple SPI.

the "new patches" are:
[projects: ad400x: Compile fix for new HDL](https://github.com/analogdevicesinc/no-OS/pull/2105/commits/61789ec6b95743d9b6e7aa2cb580aa10a291c414)
[drivers: ad400x: Fix xfer with when offload disabled](https://github.com/analogdevicesinc/no-OS/pull/2105/commits/5455f9fddb971c560875800d179c42b4833f50f4)
[drivers: ad400x: Fix single conversion transfer](https://github.com/analogdevicesinc/no-OS/pull/2105/commits/d17535e6e66ab80d9418bfceec540a512a8f160a)
[drivers: ad400x: Add clkgen and pwm](https://github.com/analogdevicesinc/no-OS/pull/2105/commits/d962c96fe5504fed1aec0ba418ac5f13d527da8d)

i tested in the following cases: 
zed + spi offload,  
zed + spi no offload, 
k1 + simple spi

these parts: 
ad4000 (16bit unsigned) , 
ad4001 (16bit signed) ,  
ad4002 (18bit unsigned) , 
ad4020 (20bit signed)

of those the only test that is failing right now is:
ad4020 in the offload case. 

I need to figure out this case (probably with the HDL team), while this goes on
im updating this PR to perhaps get some  feedback on the new changes,

projects/ad400x-st/src/common/common_data.h
```diff
@@ -0,0 +1,56 @@
...
+#ifndef __COMMON_DATA_H__
+#define __COMMON_DATA_H__
+
+/******************************************************************************/
+/***************************** Include Files **********************************/
+/******************************************************************************/
+#include "platform_includes.h"
+#include "ad400x.h"
+#include "iio_ad400x.h"
+/******************************************************************************/
```
rbolboac, line 48: add a new blank line here

projects/ad400x-st/src/examples/iio/iio_example.c
```diff
@@ -0,0 +1,100 @@
...
+#include "ad400x.h"
+#include "iio_ad400x.h"
+#include "no_os_util.h"
+#include "no_os_gpio.h"
+#include "no_os_print_log.h"
+#include "iio_app.h"
+
+/******************************************************************************/
+/************************ Functions Declarations ******************************/
+
```
rbolboac, line 57: add missing comment bloack
`/******************************************************************************/`

rbolboac (commented):
all in all, this looks much better... I did not see any obvious functional issues.
please make sure all commits have a commit description and fix astyle.
for this commit `drivers: ad400x: Fix single conversion transfer` the last sentence looks unfinished.
for this commit `drivers: ad400x: Add clkgen and pwm` please add a more detailed description explaining why was the support added

One other thing, in No-OS we have a project for a device, which can support multiple platforms. I would suggest to add support for stm platform in already existing ad400x project, for that, first you would have to modify ad400x-fmcz to use the new project structure, then you can easily add support for stm and maybe in the future, other platform.

rbolboac:
> @rbolboac ok, about your your suggestion to moving to the new format, ist it ok we first merge the changes as they are and then do the conversion in a separate PR?

I would say it is alright if it makes things easier for you...
However, the CI build should be green before merging these changes.

ahaslam2:
@rbolboac 

Projects were merged using the new directory structure.

Testing done: ad4020, ad4000, ad4001, ad4002.

zedboard:
 - iio_example and basic_example
 - spi engine offload enabled and disabled

-stm32
  - iio_example using standard spi. 
  -  basic_example: not done, no print on uart console.

ahaslam2:
I dont know what this error is. Could someone explain why its failing:

 -> Building iio_demo (iio_zed) -- xilinx -- ad40xx_fmc_zed
 -> Same hardware from last build, use existing bsp
 -> make update
 -> make clean
 -> make all
 -> DONE
 -> test -d latest_build/iio_demo/iio_demo_xilinx_iio_zed_ad40xx_fmc_zed || mkdir -p latest_build/iio_demo/iio_demo_xilinx_iio_zed_ad40xx_fmc_zed
test -d latest_build/iio_demo/iio_demo_xilinx_iio_zed_ad40xx_fmc_zed
 -> cp /no-OS_builds/builds_main/build_xilinx_ad40xx_fmc_zed/bootgen_sysfiles.tar.gz latest_build/iio_demo/iio_demo_xilinx_iio_zed_ad40xx_fmc_zed
 -> cp /no-OS_builds/builds_main/build_xilinx_ad40xx_fmc_zed/output_boot_bin/BOOT.BIN latest_build/iio_demo/iio_demo_xilinx_iio_zed_ad40xx_fmc_zed
 -> rm -r latest_build/iio_demo
Error occured: 1
##[error]Bash exited with code '1'.
Finishing: Run project build

drivers/adc/ad400x/ad400x.c
```diff
@@ -206,6 +240,25 @@ int32_t ad400x_spi_single_conversion(struct ad400x_dev *dev,
 	return no_os_spi_write_and_read(dev->spi_desc, adc_data, bytes_number);
 }
```
ahaslam2, line 242: i need to  add API doc for this function.

drivers/adc/ad400x/ad400x.c
```diff
@@ -172,6 +172,40 @@ int32_t ad400x_spi_reg_write(struct ad400x_dev *dev,
 	return ret;
 }
 
+#if !defined(USE_STANDARD_SPI)
+static int32_t ad400x_read_data_offload(struct ad400x_dev *dev,
```
ahaslam2, line 176: missing doc will do

rbolboac (commented):
some small remarks from my side.

projects/ad400x-fmcz/src/platform/stm32/main.c
```diff
@@ -0,0 +1,67 @@
...
+#ifdef IIO_EXAMPLE
+#include "iio_example.h"
+#endif
+
+/***************************************************************************//**
+ * @brief Main function execution for STM32 platform.
+ *
+ * @return ret - Result of the enabled examples execution.
+*******************************************************************************/
+
```
rbolboac, line 56: remove extra blank line

projects/ad400x-fmcz/src/platform/xilinx/main.c
```diff
@@ -64,6 +66,8 @@ int main()
 
 #ifdef BASIC_EXAMPLE
 	ret = basic_example_main();
+#elif IIO_EXAMPLE
+	ret = iio_example_main();
```
rbolboac, line 70: could be useful to continue in a similar way as with stm32 main file:
```
#else
#error At least one example has to be selected using y value in Makefile.
#endif
```

rbolboac (commented):
also, there is a typo in the last commit `usies`.

ahaslam2:
The STM32 CI builds are failing. i dont know why. They build ok on my machine.
is there any way we can see the log files to see why the fail happens

CI mentions that some more info might be at:
See log logs/ad400x-fmcz_stm32_basic.txt 
See log logs/ad400x-fmcz_stm32_iio.txt 

```
test -d latest_build/ad400x-fmcz
 -> Building ad400x-fmcz (basic) -- stm32 
 -> make reset
 -> make all
 -> Project first failed, rebuild to check if the error persists
 -> make reset
 -> make all
 -> Error persits, not a random fail, please check!
 -> ERROR
 -> See log logs/ad400x-fmcz_stm32_basic.txt -- Use cat (linux) or type (windows) to see colored output
 -> test -d latest_build/ad400x-fmcz || mkdir -p latest_build/ad400x-fmcz
test -d latest_build/ad400x-fmcz
 -> Building ad400x-fmcz (iio) -- stm32 
 -> make reset
 -> make all
 -> Project first failed, rebuild to check if the error persists
 -> make reset
 -> make all
 -> Error persits, not a random fail, please check!
 -> ERROR
 -> See log logs/ad400x-fmcz_stm32_iio.txt -- Use cat (linux) or type (windows) to see colored output

```

ahaslam2:
after discussion with @rbolboac 
The issue was due to recent push for spi_dma that requires dma includes for sm32 projects.

drivers/adc/ad400x/ad400x.c
```diff
@@ -124,22 +172,104 @@ int32_t ad400x_spi_reg_write(struct ad400x_dev *dev,
...
+	ret = no_os_gpio_set_value(dev->gpio_cnv, 1);
+	if (ret)
+		return ret;
+
+	ret = no_os_gpio_set_value(dev->gpio_cnv, 0);
+	if (ret)
+		return ret;
+#endif
+	/* SDI must remain high */
+	memset(adc_data, 0xFF, bytes_number);
```
rbolboac, line 246: I am just now seeing this... it looks it is unsafe... I think this should be reworked (the function template), such that the bytes_number are given as a parameter... 
If someone calls this function with a simple uint8_t variable, instead of an array, we would end-up writing wrong memory locations.

projects/ad400x-fmcz/src/ad400x_fmcz.c
```diff
@@ -70,6 +70,12 @@ int main()
 	Xil_ICacheEnable();
 	Xil_DCacheEnable();
 
+	/* data must be byte alligned when offload is disabled */
+	if (SPI_ENGINE_OFFLOAD_EXAMPLE == 0) {
```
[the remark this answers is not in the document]
ahaslam2: this file is removed in a later patch.. but ok ill fix it on the patch.

drivers/adc/ad400x/ad400x.c
```diff
@@ -172,8 +172,49 @@ int32_t ad400x_spi_reg_write(struct ad400x_dev *dev,
...
+	struct spi_engine_offload_message msg;
+	uint32_t commands_data[2] = {0xFF, 0xFF};
+	int ret;
+	uint32_t spi_eng_msg_cmds[3] = {
+		CS_LOW,
+		READ(2),
+		CS_HIGH
+	};
+
+	ret = spi_engine_offload_init(dev->spi_desc, dev->offload_init_param);
```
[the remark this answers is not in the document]
ahaslam2: since this is static to this file, and only called in one place. ill add the check in ad400x_read_data thats the entry point for this.

ahaslam2:
changes:
- remove the pointer cast on spi read,
this means a per sample copy, so return data same format as offload.
(offload returns LE, chip data is BE)

- add dev_info struct to hold per device sign and resolution properties.

drivers/adc/ad400x/ad400x.c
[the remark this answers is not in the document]
ahaslam2: Other functions are missing check for null in dev.  like reg read/write and remove.
will do a patch that adds it for all.

drivers/adc/ad400x/iio_ad400x.c
```diff
@@ -0,0 +1,237 @@
...
+	dev_id = iio_init_param->init_param->dev_id;
+
+	/* fill scan_type based on device id */
+	desc->scan_type.sign = ad400x_device_sign[dev_id];
+	desc->scan_type.realbits = ad400x_device_resol[dev_id];
+	desc->scan_type.storagebits = 32;
+	desc->scan_type.is_big_endian = false;
+	desc->scan_type.shift = 0;
+
+
```
rbolboac: remove extra lines

rbolboac (commented):
overall looks good

commit: https://github.com/analogdevicesinc/no-OS/pull/2105/commits/e5210d3667da205749bcec335f2abe9eb63214e2  has a typo `add add a device`
commit: https://github.com/analogdevicesinc/no-OS/pull/2105/commits/3b83e8f379f32cea8ad6cae75fbf96be03022b21 has a typo 'usies'
also, I see all commits have really short lines for the commit message, you can have up to 72-characters per line for the commit description

drivers/adc/ad400x/iio_ad400x.c
ahaslam2: this extra line is not part of the latest push.

ahaslam2:
Change log:

- fix 2 commit message typos.
- use 72 line indent for commit message.
-

drivers/adc/ad400x/iio_ad400x.c
rbolboac: it is part of this commit: https://github.com/analogdevicesinc/no-OS/pull/2105/commits/5e09916ac4bbffcf776f241d2fdb8684b814824e 
in general, in the same PR, we avoid fixing a problem we introduced with a later commit, rather we remove it in the commit it was introduced by rebasing.
this makes the review process easier, commit by commit

ahaslam2: ok,  ill fix it there.

ahaslam2:
changelog:
- fixed white line in iio driver patch.

ahaslam2:
> seeing how there is no ad400x driver documentation, do you plan on adding that here?

I added the driver README.

rbolboac (approved):

PR #2107, opened 2024-02-22
drivers:platform:stm32:Update STM32 I2C Drivers

jansunil:
An STM32F4 macro check has been added for usage two structure members in the stm32_i2c.c, that aren't a part of STM32L5, STM32H5 family.

## Pull Request Description

## PR Type

## PR Checklist

### DIFF
--- a/drivers/platform/stm32/stm32_i2c.c
+++ b/drivers/platform/stm32/stm32_i2c.c
@@ -55,6 +55,7 @@ int32_t stm32_i2c_init(struct no_os_i2c_desc **desc,
 	int32_t ret;
 	struct no_os_i2c_desc *descriptor;
 	struct stm32_i2c_desc *xdesc;
+	struct stm32_i2c_init_param *i2cinit;
 	I2C_TypeDef *base = NULL;
 
 	if (!desc || !param)
@@ -71,6 +72,7 @@ int32_t stm32_i2c_init(struct no_os_i2c_desc **desc,
 		goto error_1;
 	}
 
+	i2cinit = param->extra;
 	descriptor->extra = xdesc;
 
 	switch (param->device_id) {
@@ -95,8 +97,12 @@ int32_t stm32_i2c_init(struct no_os_i2c_desc **desc,
 	};
 
 	xdesc->hi2c.Instance = base;
+#if defined (STM32F4) || defined (STM32F1) || defined (STM32F2) || defined (STM32L1)
 	xdesc->hi2c.Init.ClockSpeed = param->max_speed_hz;
 	xdesc->hi2c.Init.DutyCycle = I2C_DUTYCYCLE_2;
+#else
+	xdesc->hi2c.Init.Timing = i2cinit->i2c_timing;
+#endif
 	xdesc->hi2c.Init.OwnAddress1 = 0;
 	xdesc->hi2c.Init.AddressingMode = I2C_ADDRESSINGMODE_7BIT;
 	xdesc->hi2c.Init.DualAddressMode = I2C_DUALADDRESS_DISABLE;
--- a/drivers/platform/stm32/stm32_i2c.h
+++ b/drivers/platform/stm32/stm32_i2c.h
@@ -53,6 +53,16 @@ struct stm32_i2c_desc {
 };
 
 /**
+ * @struct stm32_i2c_init_param
+ * @brief Structure holding the initialization parameters for stm32 platform
+ * specific I2C parameters.
+ */
+struct stm32_i2c_init_param {
+	/** I2C Timing */
+	uint32_t i2c_timing;
+};
+
+/**
  * @brief stm32 specific I2C platform ops structure
  */
 extern const struct no_os_i2c_platform_ops stm32_i2c_ops;


drivers/platform/stm32/stm32_i2c.c
```diff
@@ -95,8 +95,10 @@ int32_t stm32_i2c_init(struct no_os_i2c_desc **desc,
 	};
 
 	xdesc->hi2c.Instance = base;
+#if defined (STM32F4)
```
rbolboac, line 98: how is the speed configured in STM32L5/H5 families? similar question for duty cycle

jansunil: There is an "I2C Speed Frequency (KHz) " option to be configured via the .ioc file.
I assume that the duty cycle is taken as a default 50% since it's a clock pulse?

rbolboac: the speed should be configured in code based on the init parameter (we are overwriting the settings in the .ioc), we should be consistent  for all families...so we need to find the equivalent for ClockInit and configure it in the code for L5/H5

jansunil: @rbolboac - The speed is configured via the I2C timing register in some parts. The HAL driver does not provide separate bit masks for each bit in the timing register. Hence, I've added a value in the stm32_i2c_init_params so that the user can configure it according to the required frequency and duty ratio. Does this look ok?

drivers/platform/stm32/stm32_i2c.c
```diff
@@ -95,8 +97,13 @@ int32_t stm32_i2c_init(struct no_os_i2c_desc **desc,
...
 	xdesc->hi2c.Init.ClockSpeed = param->max_speed_hz;
 	xdesc->hi2c.Init.DutyCycle = I2C_DUTYCYCLE_2;
+#else
+	xdesc->hi2c.Init.Timing = i2cinit->i2c_timing;
+#endif
+	stm32_calculate_i2c_timing(param->max_speed_hz, hi2c);
```
rbolboac, line 106: where is this defined? I suppose you forgot to add it in the commit

drivers/platform/stm32/stm32_i2c.c
```diff
@@ -71,6 +72,7 @@ int32_t stm32_i2c_init(struct no_os_i2c_desc **desc,
 		goto error_1;
 	}
 
+	sinit = param->extra;
```
rbolboac, line 75: I think you mean i2cinit, right?

jansunil: Apologies, the wrong file got checked in. I've updated the code now

drivers/platform/stm32/stm32_i2c.c
jansunil: This function is not required. It was added by mistake. I've corrected the file now

rbolboac:
Alright, so the timing parameter will be calculated accordingly by the application.
I think some documentation will be needed for this. Maybe here would be the right place for it: https://wiki.analog.com/resources/no-os/drivers/i2c
I know that for the timer, I added platform specific documentation: https://wiki.analog.com/resources/no-os/drivers/timer
Maybe you could update, in a similar way, the documentation for i2c, but only for stm32 platform.

rbolboac (approved):

drivers/platform/stm32/stm32_i2c.c
rbolboac: One thing I did not think about is this check for STM32F4... are these registers available only for this family?

jansunil: The timing register is not present in STM32F4. 
The F4 handles the frequency and duty ratio via the "I2C Clock Control Register"

rbolboac: yes, but not only F4, from here https://github.com/search?q=org%3ASTMicroelectronics+hi2c-%3EInit.ClockSpeed&type=code&p=1 it looks like there a few other mcu families which use the same approach

rbolboac: when building for STM32F1, STM32F2, STM32L1 the current implementation will probably lead to a compilation error. So I would suggest to also add these definitions checks besides STM32F4

rbolboac: hopefully these are all....

drivers/platform/stm32/stm32_i2c.c
```diff
@@ -95,8 +97,12 @@ int32_t stm32_i2c_init(struct no_os_i2c_desc **desc,
 	};
 
 	xdesc->hi2c.Instance = base;
+#if defined (STM32F4) || defined (STM32F1) ||defined (STM32F2) || defined (STM32L1)
```
rbolboac, line 100: one tiny nit, please add a space between `||defined`

rbolboac:
And also, a more detailed commit description would be great.

jansunil:
@rbolboac  - I've added a detailed description for this

rbolboac:
> @rbolboac - I've added a detailed description for this

looks good, but please make sure you do not exceed 72 characters per line in the commit message.

rbolboac (approved):

jansunil:
@rbolboac  - I've added an internal wiki page to explain this- https://wiki.analog.com/resources/no-os/drivers/platform/stm32/stm32_i2c

jansunil:
@rbolboac  - I've made the suggested updates now- https://wiki.analog.com/resources/no-os/drivers/i2c#stm32_platform
Does it look gtg?

PR #2111, opened 2024-02-23
Add support to Adaq4224 and ad4030

jonathanns:
## Pull Request Description
This adds AD4030 and ADAQ4224 features (such as PGIA) to ad463x driver and ad463x_fmcz project.
It also adds raw and scale related attributes to IIO.

## PR Type

## PR Checklist


drivers/adc/ad463x/ad463x.c
```diff
@@ -380,10 +439,53 @@ int32_t ad463x_read_data(struct ad463x_dev *dev,
...
+ * @brief Multiply a interger number by a fraction.
+ * @param number - interger multiplier.
+ * @param numerator - numerator part of the fraction.
+ * @param denominator - denominator part of the fraction.
+ * @return Result of the multiplication.
+ */
+int mult_frac (int number, int numerator, int denominator)
+{
+	/** Check if the denominator is not zero to avoid division by zero */
+	double temp;
```
amiclaus, line 452: drop please `float/double`.

drivers/adc/ad463x/ad463x.c
```diff
@@ -161,7 +192,7 @@ int32_t ad463x_spi_reg_write_masked(struct ad463x_dev *dev,
 int32_t ad463x_set_pwr_mode(struct ad463x_dev *dev, uint8_t mode)
 {
 	if ((mode != AD463X_NORMAL_MODE) && (mode != AD463X_LOW_POWER_MODE))
-		return -1;
+		return -EINVAL;
```
amiclaus, line 195: don't do these kind of changes in the same patch. this patch adds a new feature as stated in the commit title/body.

drivers/adc/ad463x/ad463x.c
```diff
@@ -187,7 +218,7 @@ int32_t ad463x_set_avg_frame_len(struct ad463x_dev *dev, uint8_t mode)
 					   AD463X_REG_AVG,
 					   AD463X_AVG_FILTER_RESET);
 	} else {
-		if(mode < 1 || mode > 16)
+		if (mode < 1 || mode > 16)
```
amiclaus, line 221: same here. unrelated.

drivers/adc/ad463x/ad463x.c
```diff
@@ -306,42 +337,66 @@ int32_t ad463x_set_ch_offset(struct ad463x_dev *dev, uint8_t ch_idx,
...
+	return ret;
 }
 
 /**
  * @brief Read from device.
  *        Enter register mode to read/write registers
- * @param [in] dev - ad469x_dev device handler.
- * @param [out] buf - data buffer.
- * @param [in] samples - sample number.
+ * @param dev - ad469x_dev device handler.
```
amiclaus, line 397: same here.

projects/ad463x_fmcz/src/parameters.h
```diff
@@ -47,8 +47,8 @@
 /******************************************************************************/
 /********************** Macros and Constants Definitions **********************/
 /******************************************************************************/
-#define AD463x_DMA_BASEADDR             XPAR_AXI_AD463X_DMA_BASEADDR
-#define AD463x_SPI_ENGINE_BASEADDR      XPAR_SPI_AD463X_SPI_AD463X_AXI_REGMAP_BASEADDR
+#define AD463x_DMA_BASEADDR             XPAR_AXI_AD4X3X_DMA_BASEADDR
```
amiclaus, line 50: i think this change breaks the Xilinx build.

amiclaus (commented):
Some initial comments that will ease the second round of reivew.
- please clean 6af73c8cb384cd044d29a53fe3c9581ca4671825 from unrelated changes (I left only some comments, but there is more thoughout the commit).
- 7bfcddfd38c2216e0bcabd167894ac04a569e63a lacks commit body.

drivers/adc/ad463x/ad463x.c
```diff
@@ -529,23 +660,108 @@ int32_t ad463x_init(struct ad463x_dev **device,
...
+	}
+
+	int32_t closest_index = 0;
+	int32_t closest_diff = abs((int32_t)(target - array[0]));
+
+	for (size_t i = 1; i < size; ++i) {
+		uint64_t current_diff = abs((int32_t)(target - array[i]));
+
+		if (current_diff < closest_diff) {
+			closest_index = (int32_t)i;
```
CiprianRegus, line 692: you don't need to cast

drivers/adc/ad463x/ad463x.h
```diff
@@ -150,7 +150,7 @@
 
 /**
  * @enum ad463x_id
- * @brief Device type
+ * @brief Device type.
```
CiprianRegus, line 153: unrelated

drivers/adc/ad463x/ad463x.c
```diff
@@ -529,23 +660,108 @@ int32_t ad463x_init(struct ad463x_dev **device,
...
+	if (gain_idx > 3)
+		return -EINVAL;
+
+	/** update gain index value in the device handler */
+	dev->pgia_idx = gain_idx;
+	/** Set A0 and A1 pins acoording to gain index value */
+	ret = no_os_gpio_set_value(dev->gpio_pgia_a0, gain_idx & NO_OS_BIT(0));
+	if (ret != 0)
+		return ret;
+	ret = no_os_gpio_set_value(dev->gpio_pgia_a1, (gain_idx>>1) & NO_OS_BIT(0));
```
CiprianRegus, line 750: replace the bit operations with `no_os_field_get()`

drivers/adc/ad463x/iio_ad463x.c
```diff
@@ -91,31 +151,184 @@ static int32_t _iio_ad463x_prepare_transfer(struct iio_ad463x *desc,
...
+static int ad463x_iio_read_raw(void *dev, char *buf, uint32_t len,
+			       const struct iio_ch_info *channel, intptr_t priv)
+{
+	uint32_t temp[20];
+	int ret;
+	int32_t temp2;
+	struct iio_ad463x *iio_desc = dev;
+
+	/* Exit register configuration mode */
+	ret = ad463x_exit_reg_cfg_mode(iio_desc->ad463x_desc);
```
CiprianRegus, line 173: `ret` is not checked

drivers/adc/ad463x/iio_ad463x.c
```diff
@@ -91,31 +151,184 @@ static int32_t _iio_ad463x_prepare_transfer(struct iio_ad463x *desc,
...
+ * @param len     - Length of the received command buffer in bytes.
+ * @param channel - Command channel info.
+ * @param priv    - Command attribute id.
+ * @return ret    - Result of the reading procedure.
+ *		    In case of success, the size of the read data is returned.
+*/
+static int ad463x_iio_read_raw(void *dev, char *buf, uint32_t len,
+			       const struct iio_ch_info *channel, intptr_t priv)
+{
+	uint32_t temp[20];
```
CiprianRegus, line 167: Shouldn't the size be 2 instead?

drivers/adc/ad463x/iio_ad463x.c
```diff
@@ -91,31 +151,184 @@ static int32_t _iio_ad463x_prepare_transfer(struct iio_ad463x *desc,
...
+	struct iio_ad463x *iio_desc = dev;
+	struct ad463x_dev *ad463x_dev;
+	ad463x_dev = iio_desc->ad463x_desc;
+
+	int32_t vals[2];
+	/** Scale only configurable if PGIA is available */
+	if(ad463x_dev->has_pgia) {
+		iio_parse_value(buf, IIO_VAL_FRACTIONAL, &vals[0], &vals[1]);
+
+		int32_t gain_idx = ad463x_calc_pgia_gain(vals[0], vals[1], ad463x_dev->vref,
```
CiprianRegus, line 244: declare `gain_idx` at the start of the function.

drivers/adc/ad463x/iio_ad463x.c
```diff
@@ -91,31 +151,184 @@ static int32_t _iio_ad463x_prepare_transfer(struct iio_ad463x *desc,
...
+static int ad463x_iio_store_scale(void *dev, char *buf, uint32_t len,
+				  const struct iio_ch_info *channel, intptr_t priv)
+{
+	struct iio_ad463x *iio_desc = dev;
+	struct ad463x_dev *ad463x_dev;
+	ad463x_dev = iio_desc->ad463x_desc;
+
+	int32_t vals[2];
+	/** Scale only configurable if PGIA is available */
+	if(ad463x_dev->has_pgia) {
```
CiprianRegus, line 241: We usually use negative logic in such situations.

```
if (!ad463x_dev->has_pgio)
    return 0;
...
```

drivers/adc/ad463x/iio_ad463x.c
```diff
@@ -91,31 +151,184 @@ static int32_t _iio_ad463x_prepare_transfer(struct iio_ad463x *desc,
...
+		iio_parse_value(buf, IIO_VAL_FRACTIONAL, &vals[0], &vals[1]);
+
+		int32_t gain_idx = ad463x_calc_pgia_gain(vals[0], vals[1], ad463x_dev->vref,
+				   ad463x_dev->real_bits_precision);
+		return ad463x_set_pgia_gain(ad463x_dev, (uint8_t)gain_idx);
```
CiprianRegus, line 246: You don't need to cast

drivers/adc/ad463x/iio_ad463x.c
```diff
@@ -91,31 +151,184 @@ static int32_t _iio_ad463x_prepare_transfer(struct iio_ad463x *desc,
...
+	if(ad463x_dev->has_pgia) {
+		/** Remove data existing in the buffer */
+		strcpy(buf, "");
+
+		/** Go through the values in ODR table and add them to the buffer */
+		for (uint8_t i = 0; i < NO_OS_ARRAY_SIZE(ad463x_dev->scale_table); i++) {
+			vals[0] = ad463x_dev->scale_table[i][0];
+			vals[1] = ad463x_dev->scale_table[i][1];
+			iio_format_value(buffer, sizeof(buffer), IIO_VAL_INT_PLUS_NANO, 2, vals);
+			strcat(buf, buffer);
```
CiprianRegus, line 279: `sprintf(buf, "%s ", buffer);`, the you only need 1 function call.

drivers/adc/ad463x/iio_ad463x.c
```diff
@@ -91,31 +151,184 @@ static int32_t _iio_ad463x_prepare_transfer(struct iio_ad463x *desc,
...
+{
+	int32_t vals[2];
+	char buffer[20];
+	struct iio_ad463x *iio_desc = dev;
+	struct ad463x_dev *ad463x_dev;
+	ad463x_dev = iio_desc->ad463x_desc;
+
+	if(ad463x_dev->has_pgia) {
+		/** Remove data existing in the buffer */
+		strcpy(buf, "");
```
CiprianRegus, line 272: Is there any situation in which we have something in this buffer?

drivers/adc/ad463x/iio_ad463x.c
```diff
@@ -91,31 +151,184 @@ static int32_t _iio_ad463x_prepare_transfer(struct iio_ad463x *desc,
...
+
+		/** Go through the values in ODR table and add them to the buffer */
+		for (uint8_t i = 0; i < NO_OS_ARRAY_SIZE(ad463x_dev->scale_table); i++) {
+			vals[0] = ad463x_dev->scale_table[i][0];
+			vals[1] = ad463x_dev->scale_table[i][1];
+			iio_format_value(buffer, sizeof(buffer), IIO_VAL_INT_PLUS_NANO, 2, vals);
```
CiprianRegus, line 278: You can use `buf` directly here, thus you don't need to declare `buffer` anymore. Also don't you need to increase the reference to the `buffer` array as you write into it?

drivers/adc/ad463x/iio_ad463x.h
```diff
@@ -62,7 +62,8 @@ struct iio_ad463x {
 	struct ad463x_dev *ad463x_desc;
 };
 
-extern struct iio_device ad463x_iio_desc;
+extern struct iio_device ad463x_iio_desc_one_chn;
```
CiprianRegus, line 65: why do these need to be extern?

drivers/adc/ad463x/iio_ad463x.c
```diff
@@ -91,31 +151,184 @@ static int32_t _iio_ad463x_prepare_transfer(struct iio_ad463x *desc,
...
 	if(ret)
 		return ret;
 
-	for(i = 0, j = 0; i < nb_samples; i++)
-		for (ch = 0; ch < desc->iio_dev_desc.num_ch; ch++)
-			if (desc->mask & NO_OS_BIT(ch)) {
-				buff[j++] = data[i];
-			}
+	for (i = 0, j = 0; i < NO_OS_ARRAY_SIZE(data); i++)
+		data[i] = buff[i];
```
CiprianRegus, line 317: `memcpy()`

drivers/adc/ad463x/iio_ad463x.c
```diff
@@ -91,31 +151,184 @@ static int32_t _iio_ad463x_prepare_transfer(struct iio_ad463x *desc,
...
-	uint32_t data[nb_samples], i, j, ch;
+	uint16_t total_samples = nb_samples * 2;
+	uint32_t data[total_samples], i, j, ch;
 
 	if (!desc)
 		return -EINVAL;
 
-	ret = ad463x_read_data(desc->ad463x_desc, data, nb_samples);
+	/* Exit register configuration mode */
+	ret = ad463x_exit_reg_cfg_mode(desc->ad463x_desc);
```
CiprianRegus, line 309: check `ret`

drivers/adc/ad463x/iio_ad463x.c
```diff
@@ -91,31 +151,184 @@ static int32_t _iio_ad463x_prepare_transfer(struct iio_ad463x *desc,
...
+ * @return ret    - Result of the reading procedure.
+ * 					In case of success, the size of the read data is returned.
+*/
 static int32_t _iio_ad463x_read_dev(struct iio_ad463x *desc, uint32_t *buff,
 				    uint32_t nb_samples)
 {
 	int ret;
-	uint32_t data[nb_samples], i, j, ch;
+	uint16_t total_samples = nb_samples * 2;
+	uint32_t data[total_samples], i, j, ch;
```
CiprianRegus, line 303: It's not usually that great to use the stack for potentially large arrays like that. If you really need the `data` array, allocate the memory on the heap using `no_os_calloc()`.

projects/ad463x_fmcz/src/parameters.h
jonathanns: how to solve this? since this change is needed to work with the latest HDL build

drivers/adc/ad463x/ad463x.c
jonathanns: noted.

drivers/adc/ad463x/iio_ad463x.c
jonathanns: yes, i increased for tests and forgot to change it back.

drivers/adc/ad463x/iio_ad463x.c
jonathanns: Noted.

drivers/adc/ad463x/iio_ad463x.c
jonathanns: Other drivers do the same, so i just followed. but i will upload a different approach.

drivers/adc/ad463x/iio_ad463x.c
jonathanns: not in this case, since Buffer is used to hold a temporary string that will be append to buf and then be overwritten.

drivers/adc/ad463x/iio_ad463x.h
jonathanns: Also don't know, just kept as it was before. but i will remove

jonathanns:
Updates:
* Unrelated changes removed;
* Dropped double/float usage;
* Unnecessary casts removed;
* Fixed Astyle error and misspellings;
* Improved commit body;
* Used another approach in ad463x_iio_read_scale_avail function;
* Fixed other minor problems stated in the review comments;

drivers/adc/ad463x/ad463x.c
```diff
@@ -314,26 +350,50 @@ static int32_t ad463x_init_gpio(struct ad463x_dev *dev,
...
 		return ret;
 
 	/** Reset to configure pins */
-	if (dev->gpio_resetn) {
-		ret = no_os_gpio_direction_output(dev->gpio_resetn, NO_OS_GPIO_LOW);
-		if (ret != 0)
-			return ret;
+	ret = no_os_gpio_direction_output(dev->gpio_resetn, NO_OS_GPIO_LOW);
+	if (ret != 0)
+		goto error_gpio_reset;
```
rbolboac, line 361: the previous implementation looks correct to me.
resetn is not allocated in this function so it should not be removed here, simply return without calling the remove function,

drivers/adc/ad463x/ad463x.c
```diff
@@ -314,26 +350,50 @@ static int32_t ad463x_init_gpio(struct ad463x_dev *dev,
...
 
-		no_os_mdelay(100);
-		ret = no_os_gpio_set_value(dev->gpio_resetn, NO_OS_GPIO_HIGH);
-		if (ret != 0)
-			return ret;
+	no_os_mdelay(100);
+	ret = no_os_gpio_set_value(dev->gpio_resetn, NO_OS_GPIO_HIGH);
+	if (ret != 0)
+		goto error_gpio_reset;
```
rbolboac, line 366: same remark as above

drivers/adc/ad463x/ad463x.c
```diff
@@ -314,26 +350,50 @@ static int32_t ad463x_init_gpio(struct ad463x_dev *dev,
...
 
-		no_os_mdelay(100);
+	no_os_mdelay(100);
+
+	/** Configure PGIA pins, if available */
+	if (dev->has_pgia) {
+		ret = no_os_gpio_get_optional(&dev->gpio_pgia_a0,init_param->gpio_pgia_a0);
+		if (ret != 0)
+			goto error_gpio_reset;
```
rbolboac, line 374: return directly

drivers/adc/ad463x/ad463x.c
```diff
@@ -314,26 +350,50 @@ static int32_t ad463x_init_gpio(struct ad463x_dev *dev,
...
+		ret = no_os_gpio_get_optional(&dev->gpio_pgia_a1, init_param->gpio_pgia_a1);
+		if (ret != 0)
+			goto error_gpio_reset;
```
rbolboac, line 377: in this case gpio_pgia_a0 should be removed

drivers/adc/ad463x/ad463x.c
```diff
@@ -380,6 +444,49 @@ int32_t ad463x_read_data(struct ad463x_dev *dev,
...
+ * @param denominator - denominator part of the fraction.
+ * @return Result of the multiplication, or -EINVAL if denominator is zero.
+ */
+int mult_frac(int number, int numerator, int denominator)
+{
+	/** Check if the denominator is not zero to avoid division by zero */
+	if (denominator != 0) {
+		int temp = number * numerator;
+		return temp / denominator;
+	} else {
```
rbolboac, line 460: else can be removed, simply leave return -EINVAL

drivers/adc/ad463x/ad463x.c
```diff
@@ -380,6 +444,49 @@ int32_t ad463x_read_data(struct ad463x_dev *dev,
...
 }
 
+/**
+ * @brief Multiply an integer number by a fraction.
+ * @param number - integer multiplier.
+ * @param numerator - numerator part of the fraction.
+ * @param denominator - denominator part of the fraction.
+ * @return Result of the multiplication, or -EINVAL if denominator is zero.
+ */
+int mult_frac(int number, int numerator, int denominator)
```
rbolboac, line 454: this could be moved to no_os_util

drivers/adc/ad463x/ad463x.c
```diff
@@ -467,17 +576,45 @@ int32_t ad463x_init(struct ad463x_dev **device,
...
 	if (dev->data_rate == AD463X_DDR_MODE)
 		dev->capture_data_width /= 2;
 
 	dev->read_bytes_no = dev->capture_data_width / 8;
 
+	if (dev->device_id == ID_ADAQ4224) {
+		dev->has_pgia = true;
+		ad463x_fill_scale_tbl(dev);
+	} else
+		dev->has_pgia = false;
```
rbolboac, line 611: brackets are needed in else condition if the if condition has them.

drivers/adc/ad463x/ad463x.c
```diff
@@ -529,12 +665,100 @@ int32_t ad463x_init(struct ad463x_dev **device,
...
 }
 
+/**
+ * @brief Find closest value of an array of intergers.
+ * @param target - Target/reference value.
+ * @param array - Input Array.
+ * @param size - array size.
+ * @return Gain control index, in case of success, negative number otherwise
+ */
+static int32_t ad463x_find_closest(int32_t target, const int32_t array[],
```
rbolboac, line 683: this could be moved to no_os_util

drivers/adc/ad463x/ad463x.c
```diff
@@ -529,12 +665,100 @@ int32_t ad463x_init(struct ad463x_dev **device,
...
+				   size_t size)
+{
+	if (size == 0) {
```
rbolboac, line 686: !size, brackets not needed

drivers/adc/ad463x/ad463x.c
```diff
@@ -314,26 +350,50 @@ static int32_t ad463x_init_gpio(struct ad463x_dev *dev,
...
+	ret = no_os_gpio_set_value(dev->gpio_resetn, NO_OS_GPIO_HIGH);
+	if (ret != 0)
+		goto error_gpio_reset;
 
-		no_os_mdelay(100);
+	no_os_mdelay(100);
+
+	/** Configure PGIA pins, if available */
+	if (dev->has_pgia) {
+		ret = no_os_gpio_get_optional(&dev->gpio_pgia_a0,init_param->gpio_pgia_a0);
```
rbolboac, line 372: all in all, I would move allocation for gpio_pgia_a0 and gpio_pgia_a1 in init, similar to resetn, and simply handle the gpio value here. it will make the error handling easier.

drivers/adc/ad463x/iio_ad463x.c
```diff
@@ -91,31 +151,191 @@ static int32_t _iio_ad463x_prepare_transfer(struct iio_ad463x *desc,
...
+			       const struct iio_ch_info *channel, intptr_t priv)
+{
+	uint32_t temp[2];
+	int ret;
+	int32_t temp2;
+	struct iio_ad463x *iio_desc = dev;
+
+	/* Exit register configuration mode */
+	ret = ad463x_exit_reg_cfg_mode(iio_desc->ad463x_desc);
+	if (ret < 0)
```
rbolboac, line 174: I think this driver uses if (ret != 0) syntax, please use this as well

drivers/adc/ad463x/ad463x.c
jonathanns: Actualy I found an existing implementation of that. i will check that

drivers/adc/ad463x/ad463x.c
jonathanns: Isn't resetn allocated in `no_os_gpio_get_optional` ?  `no_os_gpio_direction_output` and `no_os_gpio_set_value` only fails when allocated, right? 
To summarize: the gpio optional is to keep the driver working even when this GPIOs are not set (in this case nothing will be allocate and it won't return error). but once it is set and allocated, it need to be removed after an error. correct me if I am wrong or if you meant something else.

drivers/adc/ad463x/ad463x.c
```diff
@@ -314,26 +350,50 @@ static int32_t ad463x_init_gpio(struct ad463x_dev *dev,
...
+		if (ret != 0)
+			goto error_gpio_reset;
+		ret = no_os_gpio_get_optional(&dev->gpio_pgia_a1, init_param->gpio_pgia_a1);
+		if (ret != 0)
+			goto error_gpio_reset;
+		/* set default values */
+		ret = no_os_gpio_direction_output(dev->gpio_pgia_a0, NO_OS_GPIO_LOW);
+		if (ret != 0)
+			goto error_gpio_pgia;
```
rbolboac, line 381: in this case dev->dev->gpio_pgia_a1 is not allocated but you are calling remove for it. this will work only if calloc is used (otherwise you will remove some random memory), and I see this driver uses malloc...
I suggest using calloc and then you can use a single label in which you call all remove functions, if the memory is not allocated, it will be null and the remove function will do nothing.

drivers/adc/ad463x/ad463x.c
rbolboac: for some reason I thought dev->gpio_resetn is allocated in ad463x_init (because it is already removed there)...
but I see that it is allocated in  ad463x_init_gpio, so it is alright. I deleted all my comments related to this.

projects/ad463x_fmcz/src/parameters.h
rbolboac: I will re-run the build, maybe the new HDL artifacts were not yet available in the build system

jonathanns: I talked to the HDL guys and they said to remove this commit and keep as it is on main, so I will do that.

jonathanns:
Updates v2:
* Removed find_closest function and replaced by the no_os_util implementation.
* Removed mult_frac implementation.
* Revert DMA and SPI-Engine base address changes to comply with the newest HDL parameters.
* Fix memory leak handling.
* Other minor fixes.
* Tested again with the latest HDL build.

Observation: Axi_pwm needs to be updated to accommodate the new PWM IP that supports up to 16 channels.

drivers/adc/ad463x/ad463x.c
```diff
@@ -53,6 +54,25 @@
...
+
+/*
+ * Gains computed as fractions of 1000 so they can be expressed by integers.
+ */
+static const int32_t ad463x_gains[4] = {
+	330, 560, 2220, 6670
+};
+
+/*
+ * Gains stored and computed as fractions to avoid introducing rounding erros.
```
buha, line 67: comment typo: `erros`

drivers/adc/ad463x/iio_ad463x.c
```diff
@@ -91,31 +151,191 @@ static int32_t _iio_ad463x_prepare_transfer(struct iio_ad463x *desc,
...
+		goto error_comm;
 
-	for(i = 0, j = 0; i < nb_samples; i++)
-		for (ch = 0; ch < desc->iio_dev_desc.num_ch; ch++)
-			if (desc->mask & NO_OS_BIT(ch)) {
-				buff[j++] = data[i];
-			}
+	/** Fill IIO Buffer */
+	for (i = 0, j = 0; i < total_samples; i++)
+		if (desc->mask & NO_OS_BIT(i % 2)) {
```
buha, line 324: I would avoid `%` operation in a loop, it can slow execution by a lot. Outside a loop nobody would care, but the ineffectivness is multiplied here by total_samples. If you just wanna check if i is a multiple of 2, only check the least significant bit.

buha: if (i & 0x1)
    odd
else
    even

projects/ad463x_fmcz/Makefile
```diff
@@ -1,3 +1,8 @@
+IIOD=y
+ADAQ4224=n
```
buha, line 2: So i guess no more than one can be `y` at the same time, in which case a variable like AD463X_ID that can take either one of these 3 values would make more sense. This is not critical but I think it's cleaner.

buha (dismissed):
approving with some optional remarks

drivers/adc/ad463x/ad463x.c
```diff
@@ -529,12 +647,70 @@ int32_t ad463x_init(struct ad463x_dev **device,
...
+		return -EINVAL;
+
+	/** update gain index value in the device handler */
+	dev->pgia_idx = gain_idx;
+	/** Set A0 and A1 pins according to gain index value */
+	ret = no_os_gpio_set_value(dev->gpio_pgia_a0, no_os_field_get(NO_OS_BIT(0),
+				   gain_idx));
+	if (ret != 0)
+		return ret;
+	ret = no_os_gpio_set_value(dev->gpio_pgia_a1, no_os_field_get(NO_OS_BIT(1),
```
rbolboac, line 709: return directly

projects/ad463x_fmcz/src/parameters.h
```diff
@@ -54,13 +54,13 @@
 #define AD463x_SPI_CS                   0
 #define GPIO_OFFSET			54
 #define GPIO_RESETN_1			GPIO_OFFSET + 32
+#define GPIO_PGIA_0			    GPIO_OFFSET + 33
+#define GPIO_PGIA_1			    GPIO_OFFSET + 34
```
rbolboac, line 58: fix alignment

projects/ad463x_fmcz/src/ad463x_fmc.c
```diff
@@ -207,12 +265,15 @@ int main()
 	/* Read data */
 	while (true) {
 		ret = ad463x_read_data(dev, buf, AD463x_EVB_SAMPLE_NO);
+
 		if (ret != 0)
 			return ret;
-		for (i = 0; i < AD463x_EVB_SAMPLE_NO; i+=2)
+		for (i = 0; i < AD463x_EVB_SAMPLE_NO; i+=2) {
```
rbolboac, line 271: braces not needed

projects/ad463x_fmcz/src/ad463x_fmc.c
```diff
@@ -207,12 +265,15 @@ int main()
...
 			pr_info("ADC sample ch1: %lu : %lu \n", i, buf[i]);
-		for (i = 1; i < AD463x_EVB_SAMPLE_NO; i+=2)
-			pr_info("ADC sample ch2: %lu : %lu \n", i-1, buf[i]);
+		}
+		for (i = 1; i < AD463x_EVB_SAMPLE_NO; i+=2) {
```
rbolboac, line 274: braces not needed

projects/ad463x_fmcz/src/ad463x_fmc.c
```diff
@@ -207,12 +265,15 @@ int main()
 	/* Read data */
 	while (true) {
 		ret = ad463x_read_data(dev, buf, AD463x_EVB_SAMPLE_NO);
+
```
rbolboac, line 268: remove new line

drivers/adc/ad463x/iio_ad463x.c
```diff
@@ -91,31 +151,191 @@ static int32_t _iio_ad463x_prepare_transfer(struct iio_ad463x *desc,
...
 
 	if (!desc)
 		return -EINVAL;
 
-	ret = ad463x_read_data(desc->ad463x_desc, data, nb_samples);
-	if(ret)
-		return ret;
+	data = (uint32_t *)no_os_calloc(1, sizeof(uint32_t) * total_samples * 2);
+	if (!data)
+		return -ENOMEM;
```
rbolboac, line 310: why are you allocating memory dynamically?
declaring uint32 data[total_samples*2] should work as well

drivers/adc/ad463x/iio_ad463x.c
```diff
@@ -91,31 +151,191 @@ static int32_t _iio_ad463x_prepare_transfer(struct iio_ad463x *desc,
...
+{
+	struct iio_ad463x *iio_desc = dev;
+	struct ad463x_dev *ad463x_dev;
+	int32_t gain_idx;
+	ad463x_dev = iio_desc->ad463x_desc;
+
+	int32_t vals[2];
+	/** Scale only configurable if PGIA is available */
+	if(!ad463x_dev->has_pgia)
+		return 0;
```
rbolboac, line 245: not sure returning 0 is the best option, -EINVAL might fit better

rbolboac (commented):
some remarks from my side

drivers/adc/ad463x/ad463x.c
```diff
@@ -314,26 +350,50 @@ static int32_t ad463x_init_gpio(struct ad463x_dev *dev,
...
 		return ret;
 
 	/** Reset to configure pins */
-	if (dev->gpio_resetn) {
-		ret = no_os_gpio_direction_output(dev->gpio_resetn, NO_OS_GPIO_LOW);
-		if (ret != 0)
-			return ret;
+	ret = no_os_gpio_direction_output(dev->gpio_resetn, NO_OS_GPIO_LOW);
+	if (ret != 0)
+		return ret;
```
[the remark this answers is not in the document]
jonathanns: Done.

drivers/adc/ad463x/ad463x.c
```diff
@@ -529,12 +647,70 @@ int32_t ad463x_init(struct ad463x_dev **device,
...
+	return gain_idx;
+}
+
+/**
+ * @brief Set the PGIA gain.
+ * @param dev - Pointer to the device handler.
+ * @param gain_idx - Gain control index, consult values in ad463x_pgia_gain.
+ * @return 0 in case of success, -1 otherwise
+ */
+int32_t ad463x_set_pgia_gain(struct ad463x_dev *dev,
```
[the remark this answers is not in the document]
jonathanns: Yes, I can do that.

drivers/adc/ad463x/ad463x.c
```diff
@@ -529,12 +647,70 @@ int32_t ad463x_init(struct ad463x_dev **device,
...
+			     uint8_t gain_idx)
+{
+	int32_t ret;
+	/** Check if gain available in the ADC */
+	if (!dev->has_pgia)
```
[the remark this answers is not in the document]
jonathanns: done.

projects/ad463x_fmcz/src/ad463x_fmc.c
```diff
@@ -151,44 +171,66 @@ int main()
...
+#endif
 		.dcache_invalidate_range =
 		(void (*)(uint32_t, uint32_t))Xil_DCacheInvalidateRange,
 	};
 
+#ifndef IIO_SUPPORT
+	/** Initialize UART when IIO is not enabled */
+	ret = no_os_uart_init(&uart_desc, &iio_uart_ip);
+
+	if (ret <0)
```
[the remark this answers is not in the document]
jonathanns: done.

projects/ad463x_fmcz/src/ad463x_fmc.c
```diff
@@ -151,44 +171,66 @@ int main()
...
 		.device_id = ID_AD4630_24, /* dev_id */
+#endif
 		.dcache_invalidate_range =
 		(void (*)(uint32_t, uint32_t))Xil_DCacheInvalidateRange,
 	};
 
+#ifndef IIO_SUPPORT
+	/** Initialize UART when IIO is not enabled */
+	ret = no_os_uart_init(&uart_desc, &iio_uart_ip);
+
```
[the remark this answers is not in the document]
jonathanns: done.

drivers/adc/ad463x/ad463x.c
```diff
@@ -529,12 +647,70 @@ int32_t ad463x_init(struct ad463x_dev **device,
...
 
+/**
+ * @brief Calculate the PGIA gain.
+ * @param gain_int - Interger part of the gain.
+ * @param gain_fract - Fractional part of the gain.
+ * @param vref - Reference Voltage.
+ * @param precision - Precision  value shifter.
+ * @return Gain control index, in case of success, -1 otherwise
+ */
+int32_t ad463x_calc_pgia_gain(int32_t gain_int, int32_t gain_fract,
```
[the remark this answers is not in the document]
jonathanns: done.

drivers/adc/ad463x/ad463x.c
jonathanns: done.

projects/ad463x_fmcz/src/ad463x_fmc.c
jonathanns: Done.

projects/ad463x_fmcz/src/ad463x_fmc.c
jonathanns: Done.

projects/ad463x_fmcz/src/ad463x_fmc.c
jonathanns: Done.

drivers/adc/ad463x/iio_ad463x.c
jonathanns: actually it was recommended to do like this in previous reviews.

drivers/adc/ad463x/iio_ad463x.c
jonathanns: Done.

drivers/adc/ad463x/iio_ad463x.c
```diff
@@ -91,31 +151,194 @@ static int32_t _iio_ad463x_prepare_transfer(struct iio_ad463x *desc,
...
+	uint32_t *data;
+	uint32_t i, j, ch;
 
 	if (!desc)
 		return -EINVAL;
 
-	ret = ad463x_read_data(desc->ad463x_desc, data, nb_samples);
-	if(ret)
-		return ret;
+	data = (uint32_t *)no_os_calloc(1, sizeof(uint32_t) * total_samples * 2);
```
CiprianRegus, line 311: the first parameter of `calloc()` should be the number of element, and the second one is the size of each element.
So, the proper way to write this would be:

`no_os_calloc(total_samples * 2, sizeof(*data));`

jonathanns: noted, thank you.

drivers/adc/ad463x/iio_ad463x.c
jonathanns: ok!

rbolboac (commented):
some nits from my side, but overall looks good... can you confirm when you have tested this successfully?

jonathanns:
Updates v3:

* Adressed @rbolboac , @CiprianRegus and @buha comments and sugestions.
* Tested with setup ZedBoard + ADAQ4224.

drivers/adc/ad463x/ad463x.c
```diff
@@ -101,6 +121,22 @@ int32_t ad463x_spi_reg_write(struct ad463x_dev *dev,
...
 }
 
+/**
+ * @brief Enter register configuration mode.
+ * @param dev - The device structure.
+ * @return 0 in case of success, negative error code otherwise.
+ */
+int32_t ad463x_enter_config_mode(struct ad463x_dev *dev)
+{
+	uint8_t buf[3];
```
rbolboac, line 131: dev should be checked for null pointer

drivers/adc/ad463x/ad463x.c
```diff
@@ -380,6 +444,31 @@ int32_t ad463x_read_data(struct ad463x_dev *dev,
...
+ * @return None.
+ */
+static void ad463x_fill_scale_tbl(struct ad463x_dev *dev)
+{
+	int val, val2, tmp0;
+	int32_t tmp1;
+	unsigned int i;
+	int64_t tmp2;
+
+	val2 = dev->real_bits_precision;
```
rbolboac, line 459: dev should be checked for null pointer

drivers/adc/ad463x/ad463x.c
```diff
@@ -529,12 +647,73 @@ int32_t ad463x_init(struct ad463x_dev **device,
...
+ * @return Zero in case of success, negative number otherwise
+ */
+int32_t ad463x_calc_pgia_gain(int32_t gain_int, int32_t gain_fract,
+			      int32_t vref,
+			      int32_t precision,
+			      enum ad463x_pgia_gain *gain_idx)
+{
+	uint64_t gain_nano, tmp;
+
+	gain_nano = (uint64_t)gain_int * NANO + gain_fract;
```
rbolboac, line 674: gain_idx should be checked for null pointer

rbolboac (approved):
some null checks might be useful... but all in all looks good to me

rbolboac:
not sure why build did not start, please perform a force-push (no changes needed) to re-trigger the build

rbolboac (approved):
in my opinion, we can merge this.

amiclaus (approved):

PR #2115, opened 2024-03-05
tools: scripts: make TARGET case insensitive

amiclaus:
## Pull Request Description

This is a small suggestion for improvement.

Currently when using uppercase letters for specifying the maxim target an unrelated error is thrown.

Allow users to use both lower case and uppercase letters when specifying the target.

For maxim targets we already have uppercase and lowercase variables initialized with `TARGET_UCASE`, `TARGET_LCASE` so they can be directly used for this feature.

With this patch both versions work, for example:
`make TARGET=max32650`
`make TARGET=MAX32650`

## PR Type

## PR Checklist


buha (approved):
i tested it, couldn't detect any issues

PR #2116, opened 2024-03-05
adis scale rework

rbolboac:
## Pull Request Description

- Rename adis_chan_type to adis_iio_chan_type and adis_debug_attrs to adis_iio_debug_attrs to match the file name.
- Currently the scale implementation is not optimal: the arrays used for scale which are defined in the chip specific drivers have a bigger size than needed because device id is used. This change removes this problem by using a device id with no offset in each chip specific driver.

## PR Type

## PR Checklist


amiclaus (commented):
do the unit tests also need an update after this rework?

rbolboac:
> do the unit tests also need an update after this rework?

yes, I am working on that

amiclaus (approved):

rbolboac:
- changed second commit message to better explain what was done

CiprianRegus (approved):

PR #2120, opened 2024-03-11
axi_dmac.c: Clear CYCLIC flag if transfer not cyclic

danmois:
Clear the CYCLIC flag in the DMA flags register if the transfer is set to not cyclic.

## PR Type

## PR Checklist


rbolboac (dismissed):
lgtm

drivers/axi_core/axi_dmac/axi_dmac.c
```diff
@@ -410,18 +410,20 @@ int32_t axi_dmac_transfer_start(struct axi_dmac *dmac,
...
 
-	/* Cyclic transfers set to HW or SW depending on size for MEM to DEV. */
+	/* Clear the DMA_CYCLIC flag for all transfers */
+	axi_dmac_read(dmac, AXI_DMAC_REG_FLAGS, &reg_val);
+	reg_val = reg_val & ~DMA_CYCLIC;
+	axi_dmac_write(dmac, AXI_DMAC_REG_FLAGS, reg_val);
+
+	/* Cyclic transfers set to HW for MEM to DEV if smaller than maximum transfer size
+	 * and DMA has this feature. */
 	if ((dmac->direction == DMA_MEM_TO_DEV) && (dmac->transfer.cyclic == CYCLIC)) {
```
buha, line 420: combine these conditions with && instead of nested ifs

drivers/axi_core/axi_dmac/axi_dmac.c
```diff
@@ -410,18 +410,20 @@ int32_t axi_dmac_transfer_start(struct axi_dmac *dmac,
 		}
 	}
 
-	/* Cyclic transfers set to HW or SW depending on size for MEM to DEV. */
+	/* Clear the DMA_CYCLIC flag for all transfers */
+	axi_dmac_read(dmac, AXI_DMAC_REG_FLAGS, &reg_val);
```
CiprianRegus, line 414: Do you need to clear this flag even if the transfer is cyclic (and the flag would be set again after this)? If not, you can write this as:

```
if (cyclic) {
    set flag
} else {
    clear flag
}
```

Also, even though it's not in the scope of this PR, maybe it's worth to refactor the cyclic condition in the `if` bellow, which would make the code cleaner. I would create a `can_offload_cyclic` (or a similar name) variable which would be set if those conditions are met.

danmois: Yes, there are two ways of performing a cyclic transfer: hardware or software. This complicates the if condition.

There is also the hw_cyclic property of the DMA that indicates if hardware cyclic transfers are supported.

drivers/axi_core/axi_dmac/axi_dmac.c
danmois: Will do this.

CiprianRegus (approved):

buha (approved):

PR #2125, opened 2024-03-14
build_projects.yml: change group variable for next_stable

AAndrisa:
## Pull Request Description

A new group of variables was created in azure just for next_stable branch and now we point to that one.

## PR Type

## PR Checklist

### DIFF
--- a/build_projects.yml
+++ b/build_projects.yml
@@ -9,7 +9,7 @@ pr:
 - 20*
 
 variables:
-- group: noos_releases_group_variables
+- group: noos_next_stable_group_variables
 - name: isMain
   value: $[eq(variables['Build.SourceBranch'], 'refs/heads/main')]
 - name: isNextStable


SRaus:
On line 324-25, you have: 
  _```
- name: branch
    value: $(Build.SourceBranchName)
```_
Couldn't you use 'branch' variable instead of hardcoding it, since thre is 'next_stable' in both - No-OS and HDL ?

AAndrisa:
> On line 324-25, you have: _```
> 
> * name: branch
>   value: $(Build.SourceBranchName)
> 
> ```
> Couldn't you use 'branch' variable instead of hardcoding it, since thre is 'next_stable' in both - No-OS and HDL ?
> ```

In this case I can, but it won't work for release branches, so if you consider the variable `branch` from line 25 is a better fit, I can change it.

AAndrisa:
v2: updated PR description and name because the way we set the variable HDL_BRANCH was changed using a new group of variables just for next_stable branch.

amiclaus (approved):
this looks better now :)

SRaus (approved):

PR #2132, opened 2024-03-21
ADIS improvements

rbolboac:
## Pull Request Description

This PR contains adis improvements and preparations for adding support for new devices.

## PR Type

## PR Checklist


drivers/imu/adis.c
```diff
@@ -113,10 +113,12 @@ STATIC const uint8_t burst_size_bytes[2][2] = {
...
  * @brief Initialize adis device.
  * @param adis - The adis device.
- * @param info - Initialization data.
+ * @param info - Chip specific initialization data.
+ * @param ip   - User specific initialization.
  * @return 0 in case of success, error code otherwise.
  */
-int adis_init(struct adis_dev **adis, const struct adis_chip_info *info)
+int adis_init(struct adis_dev **adis, struct adis_chip_info *info,
+	      const struct adis_init_param *ip)
```
CiprianRegus, line 104: Is there any reason for this besides maybe convenience? Is the `ip` field of `adis_chip_info` used after init?

rbolboac: the user should not have to assign the ip to adis_chip_info information. he would only be concerned about giving the ip as parameter. now the init function from adis is taking care of it instead of relying on the user application to do it (if you take a look at the basic_example you'll see the need for the user to do something like `adis1646x_chip_info.ip = init_param;` before calling `adis_init`).

rbolboac: an for the second question, it is not used, I will remove

rbolboac:
removed ip from adis_chip_info since is no longer needed

drivers/imu/adis.c
CiprianRegus: Since each ADIS16x driver defines it's own `adis_chip_info`, (which after this change the user shouldn't change at all) I wonder if the init function can be defined as:

```
int adis_init(struct adis_dev **adis, const struct adis_init_param *ip)
```

And then add the `adis_chip_info` field in the `adis_init_param` struct. Also, the driver defined `adis_chip_info` structs may now be declared as `(static) const`.

CiprianRegus (dismissed):

amiclaus (dismissed):

rbolboac:
- re-pushed because build was not triggered with the previous push (no changes done)

CiprianRegus (approved):

amiclaus (approved):

PR #2136, opened 2024-04-02
drivers:api:Add an API to clear pending irq

D-Disha:
No-os API to clear pending irq is added.
Corresponding STM32 platform layer api is added and calls to
HAL Layer is made.

## Pull Request Description

## PR Type
- [x ] New feature (change that adds new functionality)

## PR Checklist
- [ x] I have followed the [Coding style guidelines](http://analogdevicesinc.github.io/no-OS/drivers_guide.html#coding-style)
- [x ] I have performed a self-review of the changes
- [ x] I have commented my code, at least hard-to-understand parts
- [ x] I have build all projects affected by the changes in this PR
- [ x] I have tested in hardware affected projects, at the relevant boards
- [x ] I have signed off all commits from this PR
- [ x] I have updated the documentation (wiki pages, ReadMe etc), if applies


D-Disha:
Hi @CiprianRegus , 
Yes, it is added for ad7124 project for stm32h563zi support use case for clearing pending interrupts.

rbolboac:
Hi @D-Disha ,
Can you please describe the current behavior without this change? Is the interrupt not cleared?
If you take a look here: https://github.com/STMicroelectronics/stm32h5xx_hal_driver/blob/main/Src/stm32h5xx_hal_gpio.c#L577 the interrupt is cleared before calling the callback so this functionality should already be implemented in the HAL.

D-Disha:
@rbolboac Yes, it is already implemented.

However, I was facing issues with spurious interrupts. Interrupts are configured to trigger at a falling edge in my application, but the interrupt was getting triggered even without a falling edge occurring.

So, it is entering the ISR at an unexpected condition. The wrong interrupt is fired immediately after a "legitimate" interrupt, if the flag in peripheral causing the interrupt is cleared late in the ISR. 

The fix here was to again clear the interrupt gpio before enabling back the interrupt. 

Again, this might not occur in all the applications, my sampling frequency was low and I have somewhat of a lengthy ISR stemming to issues.

rbolboac:
> @rbolboac Yes, it is already implemented.
> 
> However, I was facing issues with spurious interrupts. Interrupts are configured to trigger at a falling edge in my application, but the interrupt was getting triggered even without a falling edge occurring.
> 
> So, it is entering the ISR at an unexpected condition. The wrong interrupt is fired immediately after a "legitimate" interrupt, if the flag in peripheral causing the interrupt is cleared late in the ISR.
> 
> The fix here was to again clear the interrupt gpio before enabling back the interrupt.
> 
> Again, this might not occur in all the applications, my sampling frequency was low and I have somewhat of a lengthy ISR stemming to issues.

So... this would be needed for a workaround in your project. I am not sure I would agree 100% with this. The interrupt is cleared in the HAL, I would not interfere with this implementation. 
I cannot think about a normal use-case in which the user would want to also clear the interrupt if it is already cleared. Let's see what my colleagues have to say as well.
Maybe this issue can be solved another way, this fix does not seem to be fixing the root-cause of the behavior you are seeing.. Maybe the interrupt line is too noisy or the ad7124 chip is not working as expected and is sending spurious impulses.

D-Disha:
@rbolboac No there is nothing wrong with the eval-ad7124 adc I am using and it completely works fine with mbed, only with stm32 this issue is observed. I understand that it is a workaround for my project. I could directly call from the HAL_Layer for now, but for any future projects within the team if this issue is observed, it'd be helpful for us to have an API in No-Os. I understand that this is not needed in most cases.

CiprianRegus:
Are both the mbed and stm32 examples running on the same hardware (eval board for stm32h563zi)?

D-Disha:
> Are both the mbed and stm32 examples running on the same hardware (eval board for stm32h563zi)?

The mbed example is running for SDP-K1 and STM32H563ZI is for the STM support

rbolboac:
@buha , any input here?

CiprianRegus:
I would suggest to debug this further and fix the root cause of the issue instead of applying a work around. From what you're saying this seems to be a problem with the microcontroller's interrupt setting. Since we don't even have official support for the STM32H5, there might be some configurations missing.

You said that you have a long interrupt. Is there a possibility of getting another falling edge while it's still processing? Can you also add the commits which make use of this functionality?

gprasadbng:
> I would suggest to debug this further and fix the root cause of the issue instead of applying a work around. From what you're saying this seems to be a problem with the microcontroller's interrupt setting. Since we don't even have official support for the STM32H5, there might be some configurations missing.
> 
> You said that you have a long interrupt. Is there a possibility of getting another falling edge while it's still processing? Can you also add the commits which make use of this functionality?

I guess this is the basic API we used in many project and may be first time being used  in the PCG firmware. It is a new API and hence will not break any other module.  Having this API in  the no_os layer is really helpful in my opinion.

drivers/platform/stm32/stm32_gpio_irq.c
```diff
@@ -449,6 +449,25 @@ static int32_t stm32_gpio_irq_set_priority(struct no_os_irq_ctrl_desc *desc,
...
+ * @param irq_id         - Not used, pin id is already present in desc.
+ * @return 0
+ */
+static int32_t stm32_irq_clear_pending(struct no_os_irq_ctrl_desc* desc,
+				       uint32_t irq_id)
+{
+	if (!desc || !desc->extra || !IS_EXTI_GPIO_PIN(desc->irq_ctrl_id))
+		return -EINVAL;
+
+	if (__HAL_GPIO_EXTI_GET_IT(1 << (desc->irq_ctrl_id)) != 0U) {
```
CiprianRegus, line 464: remove `!= 0` and the brackets. Otherwise, this looks good.

CiprianRegus (approved):

CiprianRegus (changes requested):
There is still some internal discussion over this.

include/no_os_irq.h
```diff
@@ -188,6 +188,8 @@ struct no_os_irq_platform_ops {
 				uint32_t priority_level);
 	/** IRQ remove function pointer */
 	int32_t (*remove)(struct no_os_irq_ctrl_desc *desc);
+	/** Clear pending interrupt */
+	int32_t(*clear_pending_irq)(struct no_os_irq_ctrl_desc* desc, uint32_t irq_id);
```
buha, line 192: remove the _irq suffix so that name of this op is `clear_pending`

drivers/platform/stm32/stm32_gpio_irq.c
```diff
@@ -462,5 +481,6 @@ const struct no_os_irq_platform_ops stm32_gpio_irq_ops = {
 	.enable = &stm32_gpio_irq_enable,
 	.disable = &stm32_gpio_irq_disable,
 	.set_priority = &stm32_gpio_irq_set_priority,
-	.remove = &stm32_gpio_irq_ctrl_remove
+	.remove = &stm32_gpio_irq_ctrl_remove,
+	.clear_pending_irq = &stm32_irq_clear_pending
```
buha, line 485: same here, remove the `_irq suffix`

buha (dismissed):
i'm ok with this, we will need to add dummy (return -ENOSYS) implementations of this op for each platform that doesn't store pending gpio interrupts and proper implementations for those that do store pending gpio interrupts

CiprianRegus (dismissed):
After @buha's comments are addressed, I think we can merge this.

drivers/platform/aducm3029/aducm3029_gpio_irq.c
```diff
@@ -542,6 +542,18 @@ static int aducm_gpio_irq_set_priority(struct no_os_irq_ctrl_desc *desc,
...
+/**
+ * @brief Unused
+ * @param desc - GPIO interrupt controller descriptor.
+ * @return -ENOSYS
+ */
+static int32_t aducm_gpio_irq_clear_pending(struct no_os_irq_ctrl_desc *desc,
+		uint32_t irq_id)
+{
+	return -ENOSYS;
+}
```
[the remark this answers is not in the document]
buha: yeah my bad, sorry for that, it was just an afterthought after review and i wrote it so that we keep that in mind, not to implement it in this review, as @rbolboac is saying it turns out those dummy implementations are not needed

D-Disha:
Hi all, are there any more comments on this PR?

buha (approved):

Issue #2135, opened 2024-03-28
Symbolic links break Windows MSDK builds 


Brandon-Hurst:
The symbolic links for the maxim_dma .c/.h files break builds out-of-the-box on Windows. Windows does not easily import symlinks from Git.

![image](https://github.com/analogdevicesinc/no-OS/assets/82959533/8a63e701-7bf2-448d-95dc-488293172280)

![image](https://github.com/analogdevicesinc/no-OS/assets/82959533/ee02a8eb-6c45-487f-b1e3-f7c2d8262720)

Suggested fix: Compile common HAL files via the Makefile includes rather than creating symbolic links.

dbogdan:
This should fix the issue: https://github.com/analogdevicesinc/no-OS/commit/b2775b22b47ac1109adf166863bb0a033ef8de92

Brandon-Hurst:
Hi Dragos,

Thanks, that's great! I noticed that the commit does not make the changes for MAX78000. This MCU should have the same changes done as well.

PR #2137, opened 2024-04-03
Remove maxim_dma symlinks

CiprianRegus:
This is a workaround for this issue https://github.com/analogdevicesinc/no-OS/issues/2135

In my opinion this doesn't look that good, since from now on you'd have to remember the different path for the maxim_dma files. On top of that, the path when included in a makefile would look something like this: `$(PLATFORM_DRIVERS)/../common/maxim_dma.c`.

The alternative would be for everyone to enable symlinks in their git config `git config core.symlink=true` (looks like this is set to false on Windows by default) or clone the no-OS repository using `git -c core.symlinks=true ...`.

What do you think about this?

## Pull Request Description

## PR Type

## PR Checklist


rbolboac:
It looks like there are some project failing now.

rbolboac (approved):

dbogdan (approved):

PR #2139, opened 2024-04-04
ad796x: Add project and driver for ad796x-fmc

ahaslam2:
This adds driver files and a project test the ad796x series of pulsar Differential ADC.

The HDL for this project is still not merged, and probably its why CI is failing.
perhaps review can start without the HDL 

this was tested using the HDL binary at https://github.com/adi-ses/sea-misc/tree/main/AD796x-PRJ

## PR Type

## PR Checklist

### DIFF
--- a/drivers/adc/ad796x/README.rst
+++ b/drivers/adc/ad796x/README.rst
@@ -1,0 +1,218 @@
+AD796X no-OS driver
+====================
+
+Supported Devices
+-----------------
+
+`AD7960 <https://www.analog.com/en/products/ad7960.html>`_
+
+Overview
+--------
+
+The EVAL-AD7960FMCZ is a fully featured evaluation Kit designed to demonstrate
+the performance of the low power 18-bit,5 MSPS PulSAR® Differential ADC AD7960.
+
+This board operates in conjunction with the Zedboard. The evaluation software
+is provided to enable the user to perform detailed analysis of the AD7960's
+performance. The technical user guide includes the detailed description of
+the operation and set up of the evaluation board and software when operated
+with the System Demonstration Platform board.
+
+
+AD796X Device Configuration
+----------------------------
+
+Driver Initialization
+^^^^^^^^^^^^^^^^^^^^^^
+
+In order to be able to use the device, the pulsar lvds HDL is used. The usage
+of the driver is:
+
+**ad796x_init** is called with the settings dma, pwm, clkgen, and the 4 gpios to
+set the operation mode.
+
+**ad796x_read_data** called to fill a buffer with n number of samples.
+
+**ad796x_remove** Free the allocated resources
+
+AD796X Driver Initialization Example
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+.. code-block:: bash
+
+	uint32_t *buf = (uintptr_t)ADC_DDR_BASEADDR;
+	int i, ret;
+	struct ad796x_dev *adc_dev;
+
+	struct axi_adc_init ad796x_core_ip = {
+		.name = "ad796x_core",
+		.num_channels = 1,
+		.base = RX_CORE_BASEADDR
+	};
+
+	struct axi_clkgen_init clkgen_ip = {
+		.name = "rx_clkgen",
+		.base = RX_CLKGEN_BASEADDR,
+		.parent_rate = 100000000,
+	};
+
+	struct axi_dmac_init dmac_ip = {
+		.name = "ad796x_dmac",
+		.base = RX_DMA_BASEADDR,
+		.irq_option = IRQ_DISABLED
+	};
+
+	struct axi_pwm_init_param axi_pwm_0_extra = {
+		.base_addr = AXI_PWMGEN_BASEADDR,
+		.ref_clock_Hz = 125000000,
+		.channel = 0
+	};
+
+	struct no_os_pwm_init_param axi_pwm_0_ip = {
+		.period_ns = PWM_0_PERIOD_NS,
+		.duty_cycle_ns = PWM_0_DUTY_NS,
+		.phase_ns = PWM_0_PHASE,
+		.platform_ops = &axi_pwm_ops,
+		.extra = &axi_pwm_0_extra
+	};
+
+	struct axi_pwm_init_param axi_pwm_1_extra = {
+		.base_addr = AXI_PWMGEN_BASEADDR,
+		.ref_clock_Hz = 125000000,
+		.channel = 1
+	};
+
+	struct no_os_pwm_init_param axi_pwm_1_ip = {
+		.period_ns = PWM_1_PERIOD_NS,
+		.duty_cycle_ns = PWM_1_DUTY_NS,
+		.phase_ns = PWM_1_PHASE,
+		.platform_ops = &axi_pwm_ops,
+		.extra = &axi_pwm_1_extra
+	};
+
+	struct xil_gpio_init_param xil_gpio_init = {
+		.device_id = GPIO_DEVICE_ID,
+		.type = GPIO_PS
+	};
+
+	struct no_os_gpio_init_param gpio_adc_en3_fmc_ip = {
+		.number = GPIO_EN3_FMC,
+		.platform_ops = GPIO_OPS,
+		.extra = GPIO_EXTRA
+	};
+
+	struct no_os_gpio_init_param gpio_adc_en2_fmc_ip = {
+		.number = GPIO_EN2_FMC,
+		.platform_ops = GPIO_OPS,
+		.extra = GPIO_EXTRA
+	};
+
+	struct no_os_gpio_init_param gpio_adc_en1_fmc_ip = {
+		.number = GPIO_EN1_FMC,
+		.platform_ops = GPIO_OPS,
+		.extra = GPIO_EXTRA
+	};
+
+	struct no_os_gpio_init_param gpio_adc_en0_fmc_ip = {
+		.number = GPIO_EN0_FMC,
+		.platform_ops = GPIO_OPS,
+		.extra = GPIO_EXTRA
+	};
+
+	struct ad796x_init_param ad796x_init_param = {
+		.ad796x_core_ip = &ad796x_core_ip,
+		.clkgen_ip = &clkgen_ip,
+		.dmac_ip = &dmac_ip,
+		.axi_pwm_0_ip = &axi_pwm_0_ip,
+		.axi_pwm_1_ip = &axi_pwm_1_ip,
+		.gpio_ip[3] = &gpio_adc_en3_fmc_ip,
+		.gpio_ip[2] = &gpio_adc_en2_fmc_ip,
+		.gpio_ip[1] = &gpio_adc_en1_fmc_ip,
+		.gpio_ip[0] = &gpio_adc_en0_fmc_ip,
+		.mode = AD796X_MODE1_EXT_REF_5P0,
+	};
+
+	ret = ad796x_init(&adc_dev, &ad796x_init_param);
+	if (ret)
+		return ret;
+
+	pr_info("Capture start.\n");
+	ret = ad796x_read_data(adc_dev, buf, SAMPLES_PER_CHANNEL);
+	if (ret) {
+		pr_err("read data error %d\n", ret);
+		ad796x_remove(adc_dev);
+		return ret;
+	}
+
+	for (i = 0; i < SAMPLES_PER_CHANNEL; i++, buf++)
+		printf("CH1: %ld\n", *buf);
+
+
+	return ad796x_remove(adc_dev);
+
+AD796X no_OS IIO Support
+-------------------------
+
+The AD796X uses the iio_axi_adc driver to provide iio support.
+
+
+AD796X IIO Driver Initialization Example
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+.. code-block:: bash
+
+	struct ad796x_dev *adc_dev;
+	struct iio_app_desc *app;
+	struct iio_app_init_param app_init_param = { 0 };
+	struct iio_axi_adc_desc *iio_axi_adc_desc;
+	struct iio_device *dev_desc;
+	struct iio_data_buffer adc_buff = {
+		.buff = (void *)ADC_DDR_BASEADDR,
+		.size = MAX_SIZE_BASE_ADDR
+	};
+	struct scan_type init_scan_type = {
+		.sign = 's',
+		.realbits = 32,
+		.storagebits = 32,
+		.shift = 0,
+		.is_big_endian = false
+	};
+
+	ret = ad796x_init(&adc_dev, &ad796x_init_param);
+	if (ret) {
+		pr_err("Error: ad796x_init: %d\n", ret);
+		return ret;
+	}
+
+	struct iio_axi_adc_init_param iio_axi_adc_init_par = {
+		.rx_adc = adc_dev->ad796x_core,
+		.rx_dmac = adc_dev->axi_dmac,
+		.scan_type_common = &init_scan_type,
+		.dcache_invalidate_range = (void (*)(uint32_t,
+						     uint32_t))Xil_DCacheInvalidateRange,
+	};
+
+	ret = iio_axi_adc_init(&iio_axi_adc_desc, &iio_axi_adc_init_par);
+	if (ret) {
+		pr_err("Error: iio_axi_adc_init: %d\n", ret);
+		goto err_adc_init;
+	}
+
+	iio_axi_adc_get_dev_descriptor(iio_axi_adc_desc, &dev_desc);
+
+	struct iio_app_device devices[] = {
+		IIO_APP_DEVICE("ad796x", iio_axi_adc_desc,
+			       dev_desc, &adc_buff, NULL, NULL),
+	};
+
+	app_init_param.devices = devices;
+	app_init_param.nb_devices = NO_OS_ARRAY_SIZE(devices);
+	app_init_param.uart_init_params = iio_uart_ip;
+
+	ret = iio_app_init(&app, app_init_param);
+	if (ret) {
+		pr_err("Error: iio_app_init: %d\n", ret);
+		goto err_app_init;
+	}
+
+	ret = iio_app_run(app);
--- a/drivers/adc/ad796x/ad796x.c
+++ b/drivers/adc/ad796x/ad796x.c
@@ -1,0 +1,259 @@
+/***************************************************************************//**
+ *   @file   ad796x.c
+ *   @brief  Implementation of AD796X Driver.
+ *   @author Axel Haslam (ahaslam@baylibre.com)
+********************************************************************************
+ * Copyright 2024(c) Analog Devices, Inc.
+ *
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *  - Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ *  - Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in
+ *    the documentation and/or other materials provided with the
+ *    distribution.
+ *  - Neither the name of Analog Devices, Inc. nor the names of its
+ *    contributors may be used to endorse or promote products derived
+ *    from this software without specific prior written permission.
+ *  - The use of this software may or may not infringe the patent rights
+ *    of one or more patent holders.  This license does not release you
+ *    from the requirement that you obtain separate licenses from these
+ *    patent holders to use this software.
+ *  - Use of the software either in source or binary form, must be run
+ *    on or directly connected to an Analog Devices Inc. component.
+ *
+ * THIS SOFTWARE IS PROVIDED BY ANALOG DEVICES "AS IS" AND ANY EXPRESS OR
+ * IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, NON-INFRINGEMENT,
+ * MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED.
+ * IN NO EVENT SHALL ANALOG DEVICES BE LIABLE FOR ANY DIRECT, INDIRECT,
+ * INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, INTELLECTUAL PROPERTY RIGHTS, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+*******************************************************************************/
+
+/******************************************************************************/
+/***************************** Include Files **********************************/
+/******************************************************************************/
+#include <stdlib.h>
+#include <errno.h>
+#include "ad796x.h"
+#include "no_os_alloc.h"
+#include "no_os_print_log.h"
+
+/******************************************************************************/
+/********************** Macros and Constants Definitions **********************/
+/******************************************************************************/
+
+#define AD796X_BYTES_PER_SAMPLE		4
+
+/******************************************************************************/
+/************************ Functions Definitions *******************************/
+/******************************************************************************/
+
+/* gpio values for the respective mode {en0, en1, en2, en3} */
+static uint8_t ad796x_modes[][AD796X_NUM_GPIOS] = {
+	[AD796X_MODE1_EXT_REF_5P0] =		{1, 0, 0, 1},
+	[AD796X_MODE2_INT_REF_4P0] =		{1, 0, 0, 1},
+	[AD796X_MODE3_EXT_REF_4P0] =		{0, 1, 0, 1},
+	[AD796X_MODE4_SNOOZE] =			{1, 1, 0, 1},
+	[AD796X_MODE5_TEST] =			{0, 0, 1, 0},
+	[AD796X_MODE6_INVALID] =		{0, 0, 1, 1},
+	[AD796X_MODE7_EXT_REF_5P0_9MHZ] =	{1, 0, 1, 1},
+	[AD796X_MODE8_INT_REF_4P0_9MHZ] =	{1, 0, 1, 1},
+	[AD796X_MODE9_EXT_REF_4P0_9MHZ] =	{0, 1, 1, 1},
+	[AD796X_MODE10_SNOOZE2] =		{1, 1, 1, 1},
+};
+
+/**
+ * @brief Read data samples from adc
+ * @param dev - The device structure.
+ * @param buf - The buffer to fill.
+ * @param samples - number of samples to read.
+ * @return 0 in case of success, negative error code otherwise.
+ */
+int32_t ad796x_read_data(struct ad796x_dev *dev,
+			 uint32_t *buf,
+			 uint16_t samples)
+{
+	int ret;
+	struct 	axi_dma_transfer read_transfer = {
+		.size = samples * AD796X_BYTES_PER_SAMPLE,
+		.transfer_done = 0,
+		.cyclic = NO,
+		.src_addr = 0,
+		.dest_addr = (uintptr_t)buf,
+	};
+
+	ret = axi_dmac_transfer_start(dev->axi_dmac, &read_transfer);
+	if (ret) {
+		pr_err("axi_dmac_transfer_start() failed!\n");
+		return ret;
+	}
+
+	return axi_dmac_transfer_wait_completion(dev->axi_dmac, 3000);
+}
+
+/**
+ * @brief Remove the device gpio resources.
+ * @param dev - The device structure.
+ * @return 0 in case of success, negative error code otherwise.
+ */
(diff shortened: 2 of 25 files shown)


drivers/adc/ad796x/ad796x.c
```diff
@@ -0,0 +1,168 @@
...
+/******************************************************************************/
+#include <stdlib.h>
+#include <errno.h>
+#include "ad796x.h"
+#include "no_os_alloc.h"
+
+/******************************************************************************/
+/************************ Functions Definitions *******************************/
+/******************************************************************************/
+/******************************************************************************/
```
rbolboac, line 51: remove extra comment line

drivers/adc/ad796x/ad796x.c
```diff
@@ -0,0 +1,168 @@
...
+
```
rbolboac, line 52: remove extra blank line

drivers/adc/ad796x/ad796x.h
```diff
@@ -0,0 +1,92 @@
...
+#define __AD796X_H__
+
+/******************************************************************************/
+/***************************** Include Files **********************************/
+/******************************************************************************/
+#include "no_os_gpio.h"
+
+/******************************************************************************/
+/********************** Macros and Constants Definitions **********************/
+/******************************************************************************/
```
rbolboac, line 49: remove this seeing how there are no macros or constants definitions

drivers/adc/ad796x/ad796x.h
```diff
@@ -0,0 +1,92 @@
...
+ * @struct ad796x_dev
+ * @brief AD796X Device structure.
+ */
+struct ad796x_dev {
+	/** Device communication descriptor */
+	struct no_os_gpio_desc		*gpio_adc_en3_fmc;
+	struct no_os_gpio_desc		*gpio_adc_en2_fmc;
+	struct no_os_gpio_desc		*gpio_adc_en1_fmc;
+	struct no_os_gpio_desc		*gpio_adc_en0_fmc;
+
```
rbolboac, line 77: remove extra blank line

rbolboac (commented):
some remarks from my side, also:
please create a different commit for the driver and in the same commit please also add the driver documentation
please use the new project structure

ahaslam2:
@rbolboac 

ok ill separate driver and project. but 
what do you mean by "new project structure"?

rbolboac:
> @rbolboac
> 
> ok ill separate driver and project. but what do you mean by "new project structure"?

By new project structure I mean the src folder should have platform independent examples, with common folder and platform specific folders. This is an example from iio_demo project:
```
├── src
│   ├── common
│   │   ├── app_config.h
│   │   ├── common_data.c
│   │   └── common_data.h
│   ├── examples
│   │   ├── examples_src.mk
│   │   ├── iio_example
│   │   │   ├── iio_example.c
│   │   │   └── iio_example.h
│   │   ├── iio_sw_trigger_example
│   │   │   ├── iio_sw_trigger_example.c
│   │   │   └── iio_sw_trigger_example.h
│   │   └── iio_timer_trigger_example
│   │       ├── iio_timer_trigger_example.c
│   │       └── iio_timer_trigger_example.h
│   └── platform
│       ├── aducm3029
│       │   ├── main.c
│       │   ├── parameters.c
│       │   ├── parameters.h
│       │   └── platform_src.mk
│       ├── linux
│       │   ├── main.c
│       │   ├── parameters.c
│       │   ├── parameters.h
│       │   └── platform_src.mk
│       ├── maxim
│       │   ├── main.c
│       │   ├── parameters.c
│       │   ├── parameters.h
│       │   └── platform_src.mk
│       ├── pico
│       │   ├── main.c
│       │   ├── parameters.c
│       │   ├── parameters.h
│       │   └── platform_src.mk
│       ├── platform_includes.h
│       ├── stm32
│       │   ├── main.c
│       │   ├── parameters.c
│       │   ├── parameters.h
│       │   └── platform_src.mk
│       └── xilinx
│           ├── main.c
│           ├── parameters.c
│           ├── parameters.h
│           └── platform_src.mk

```

ahaslam2:
Ok, i did not know this structure was a requirement, will do.

ahaslam2:
one question @rbolboac 

i would suppose the structure requirement is to keep the examples platform
independant? 

i see things like axi_dac_init_begin(&phy->tx_dac, &tx_dac_init);
that i think are tied to some specific fpga/platform ?

https://github.com/analogdevicesinc/no-OS/blob/main/projects/adrv902x/src/examples/iio_example/iio_example.c#L304

are there some rules as to what goes into platform/common/example?

rbolboac:
well, the idea is to have the examples generic and platform independent, but if that is not possible I suppose it is alright.
in that case adrv902x project can be used as an example.

ahaslam2:
The failure of CI is currently due to the missing HDL:

```test -d latest_build/ad796x_fmcz
 -> Building ad796x_fmcz (iio) -- xilinx -- ad796x_fmc_zed
 -> cp /no-OS_builds/builds_main/new_hardware/ad796x_fmc_zed/system_top.xsa /no-OS_builds/builds_main/hardware/ad796x_fmc_zed.xsa
 -> ERROR
 -> See log logs/ad796x_fmcz_xilinx_iio_ad796x_fmc_zed.txt -- Use cat (linux) or type (windows) to see colored output
Error occured: 1
````

File: drivers/adc/ad796x/ad796x.c [hunk 5]
```diff
@@ -0,0 +1,297 @@
...
+ * @param buf - The buffer to fill.
+ * @param samples - number of samples to read.
+ * @return 0 in case of success, negative error code otherwise.
+ */
+int32_t ad796x_read_data(struct ad796x_dev *dev,
+			 uint32_t *buf,
+			 uint16_t samples)
+{
+	int ret;
+
```
rbolboac, line 85: blank line can be removed, no reason to leave blank lines between declarations. see everywhere where it applies

drivers/adc/ad796x/ad796x.c
```diff
@@ -0,0 +1,297 @@
...
+	int ret, i;
+
+	struct no_os_gpio_desc *desc[] = {
+		dev->gpio_adc_en3_fmc,
+		dev->gpio_adc_en2_fmc,
+		dev->gpio_adc_en1_fmc,
+		dev->gpio_adc_en0_fmc,
+	};
+
+	for (i = 0; i < 4; i++) {
```
rbolboac, line 119: create a macro value for that magic value

drivers/adc/ad796x/ad796x.c
```diff
@@ -0,0 +1,297 @@
...
+	ret = axi_clkgen_init(&dev->clkgen, init_param->clkgen_ip);
+	if (ret) {
+		pr_err("error: axi_clkgen_init error: %d\n", ret);
+		goto err_clkgen;
+	}
+
+	ret = axi_clkgen_set_rate(dev->clkgen, 125000000);
+	if (ret) {
+		pr_err("error: axi_clkgen_set_rate error: %d\n", ret);
+		goto err_set_rate;
```
rbolboac, line 213: no need for an extra label, can use err_clkgen

drivers/adc/ad796x/ad796x.c
```diff
@@ -0,0 +1,297 @@
...
+	no_os_pwm_remove(dev->axi_pwm_1);
+err_pwm_1:
+	no_os_pwm_remove(dev->axi_pwm_0);
+err_pwm_0:
+err_set_rate:
+	axi_clkgen_remove(dev->clkgen);
+err_clkgen:
+	ad796x_gpio_remove(dev);
+error_gpio:
+	no_os_free(dev);
```
rbolboac, line 256: I think it is safe to have a single error label (but please check before).
basically, if the init function has not been called, the desc will be null and the remove functions should check whether the desc is null or not. if it is null nothing will happen, if it is not null it will be freed.

drivers/adc/ad796x/ad796x.h
```diff
@@ -0,0 +1,119 @@
...
+/******************************************************************************/
+
+/* Initialize the device. */
+int ad796x_init(struct ad796x_dev **device,
+		struct ad796x_init_param *init_param);
+
+/* Remove the device and release resources. */
+int ad796x_remove(struct ad796x_dev *dev);
+
+int32_t ad796x_read_data(struct ad796x_dev *dev, uint32_t *buf,
```
rbolboac, line 115: missing doxygen description

projects/ad796x_fmcz/builds.json
```diff
@@ -0,0 +1,10 @@
+{
+  "xilinx": {
+    "iio": {
+      "flags": "IIO_EXAMPLE=y BASIC_EXAMPLE=n",
+      "hardware": [
+        "ad796x_fmc_zed"
```
rbolboac, line 6: the build is probably failing because of this... the hdl project is here from whay I have been tols: https://github.com/analogdevicesinc/hdl/tree/main/projects/pulsar_lvds_adc
so the hardware artifact is actually named pulsar_lvds_adc_zed
let's give this a try and see if the build is successful

projects/ad796x_fmcz/readme.txt
```diff
@@ -0,0 +1 @@
+TODO
```
rbolboac, line 1: remove this file from this commit if it is not useful

projects/ad796x_fmcz/src/examples/basic_example/basic_example.c
```diff
@@ -0,0 +1,95 @@
...
+*******************************************************************************/
+
+/******************************************************************************/
+/***************************** Include Files **********************************/
+/******************************************************************************/
+#include "basic_example.h"
+#include "common_data.h"
+#include "ad796x.h"
+#include "no_os_util.h"
+
```
rbolboac, line 47: remove extra blank line

projects/ad796x_fmcz/src/examples/basic_example/basic_example.c
```diff
@@ -0,0 +1,95 @@
...
+
+/******************************************************************************/
+/************************ Functions Definitions *******************************/
+/******************************************************************************/
+/***************************************************************************//**
```
rbolboac, line 52: add blank line in between

projects/ad796x_fmcz/src/examples/basic_example/basic_example.c
```diff
@@ -0,0 +1,95 @@
...
+ * @brief BASIC example main execution.
+ *
+ * @return ret - Result of the example execution. If working correctly, will
+ *               execute continuously function basic_app_run and will not return.
+*******************************************************************************/
+int basic_example_main()
+{
+
```
rbolboac, line 60: remove blank line

projects/ad796x_fmcz/src/examples/basic_example/basic_example.c
```diff
@@ -0,0 +1,95 @@
...
+	uint32_t *buf = (uintptr_t)ADC_DDR_BASEADDR;
+	int i, ret;
+
```
rbolboac, line 63: remove blank line

projects/ad796x_fmcz/src/examples/basic_example/basic_example.c
```diff
@@ -0,0 +1,95 @@
...
+	printf("\n init\n");
+
+	ret = ad796x_init(&adc_dev, &ad796x_init_param);
+	if (ret)
+		return ret;
+	printf("\n Capture start.\n");
+
+	ret = ad796x_read_data(adc_dev, buf, SAMPLES_PER_CHANNEL);
+	if (ret)
+		return ret;
```
rbolboac, line 75: ad796x_remove should be called before returning to avoid memory leaks and also an error message might be useful

projects/ad796x_fmcz/src/examples/basic_example/basic_example.c
```diff
@@ -0,0 +1,95 @@
...
+
+	for (i = 0; i < SAMPLES_PER_CHANNEL; i++, buf++)
+		printf("CH1: %ld\n", *buf);
+
+	printf("\n Capture done.\n");
+
+	return 0;
```
rbolboac, line 82: ad796x_remove should be called before returning to avoid memory leaks

projects/ad796x_fmcz/src/examples/basic_example/basic_example.c
```diff
@@ -0,0 +1,95 @@
...
+
+
+
+
+
+
+
+
+
+
```
rbolboac, line 95: remove all extra blank lines

projects/ad796x_fmcz/src/examples/basic_example/basic_example.h
```diff
@@ -0,0 +1,51 @@
...
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+*******************************************************************************/
+#ifndef __BASIC_EXAMPLE_H__
+#define __BASIC_EXAMPLE_H__
+
+/******************************************************************************/
+/***************************** Include Files **********************************/
+/******************************************************************************/
+
```
rbolboac, line 45: remove this seeing how there are no includes

projects/ad796x_fmcz/src/examples/examples_src.mk
```diff
@@ -0,0 +1,21 @@
+ifeq (y,$(strip $(IIO_EXAMPLE)))
+CFLAGS += -DIIO_EXAMPLE=1
+SRCS += $(PROJECT)/src/examples/iio_example/iio_example.c
+INCS += $(PROJECT)/src/examples/iio_example/iio_example.h
+IIOD=y
+SRC_DIRS += $(NO-OS)/iio/iio_app
+endif
+
```
rbolboac, line 8: remove extra blank line

projects/ad796x_fmcz/src/examples/iio_example/iio_example.c
```diff
@@ -0,0 +1,123 @@
...
+#include "common_data.h"
+#include "ad796x.h"
+#include "no_os_util.h"
+#include "iio_app.h"
+#include "iio_axi_adc.h"
+
+/******************************************************************************/
+/************************ Functions Definitions *******************************/
+/******************************************************************************/
+/***************************************************************************//**
```
rbolboac, line 53: add a blank line in between

projects/ad796x_fmcz/src/examples/iio_example/iio_example.c
```diff
@@ -0,0 +1,123 @@
...
+		.rx_adc = adc_dev->ad796x_core,
+		.rx_dmac = adc_dev->axi_dmac,
+		.scan_type_common = &init_scan_type,
+		.dcache_invalidate_range = (void (*)(uint32_t,
+						     uint32_t))Xil_DCacheInvalidateRange,
+	};
+
+	ret = iio_axi_adc_init(&iio_axi_adc_desc, &iio_axi_adc_init_par);
+	if (ret)
+		return ret;
```
rbolboac, line 103: ad796x_remove should be called in case of an error

projects/ad796x_fmcz/src/examples/iio_example/iio_example.c
```diff
@@ -0,0 +1,123 @@
...
+			       dev_desc, &adc_buff, NULL, NULL),
+	};
+
+	app_init_param.devices = devices;
+	app_init_param.nb_devices = NO_OS_ARRAY_SIZE(devices);
+	app_init_param.uart_init_params = iio_uart_ip;
+
+	ret = iio_app_init(&app, app_init_param);
+	if (ret)
+		return ret;
```
rbolboac, line 118: the according remove functions should be called in case of an error

projects/ad796x_fmcz/src/examples/iio_example/iio_example.c
```diff
@@ -0,0 +1,123 @@
...
+
+	return iio_app_run(app);
```
rbolboac, line 120: the according remove functions should be called in case of an error

projects/ad796x_fmcz/src/examples/iio_example/iio_example.h
```diff
@@ -0,0 +1,51 @@
...
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+*******************************************************************************/
+#ifndef __IIO_EXAMPLE_H__
+#define __IIO_EXAMPLE_H__
+
+/******************************************************************************/
+/***************************** Include Files **********************************/
+/******************************************************************************/
+
```
rbolboac, line 45: remove this

projects/ad796x_fmcz/src/platform/xilinx/parameters.c
```diff
@@ -0,0 +1,55 @@
...
+#ifdef XPAR_XUARTLITE_NUM_INSTANCES
+	.type = UART_PL,
+#else
+	.type = UART_PS,
+	.irq_id = UART_IRQ_ID
+#endif
+};
+
+
+
```
rbolboac, line 55: remove extra blank lines

projects/ad796x_fmcz/src/platform/xilinx/parameters.h
```diff
@@ -0,0 +1,112 @@
...
+#define __PARAMETERS_H__
+
+/******************************************************************************/
+/***************************** Include Files **********************************/
+/******************************************************************************/
+#include <xparameters.h>
+#include <xil_cache.h>
+#include <xilinx_uart.h>
+#include "xilinx_gpio.h"
+
```
rbolboac, line 50: remove extra blank line

projects/ad796x_fmcz/src/platform/xilinx/parameters.h
```diff
@@ -0,0 +1,112 @@
...
+#ifdef XPS_BOARD_ZCU102
+#define UART_IRQ_ID		XPAR_XUARTPS_0_INTR
+#else
+#define UART_IRQ_ID		XPAR_XUARTPS_1_INTR
+#endif
+
+#else // _XPARAMETERS_PS_H_
+#define UART_DEVICE_ID	XPAR_AXI_UART_DEVICE_ID
+#define INTC_DEVICE_ID	XPAR_INTC_SINGLE_DEVICE_ID
+#define UART_IRQ_ID		XPAR_AXI_INTC_AXI_UART_INTERRUPT_INTR
```
rbolboac, line 68: use the same indentation

projects/ad796x_fmcz/src/platform/xilinx/parameters.h
```diff
@@ -0,0 +1,112 @@
...
+
+/* This value can be modified based on the number
+of samples needed to be stored in the device buffer
+and based on the available RAM memory of the platform */
+#define SAMPLES_PER_CHANNEL_PLATFORM    20000
+#define UART_BAUDRATE	                115200
+
+#define MAX_SIZE_BASE_ADDR		(SAMPLES_PER_CHANNEL_PLATFORM * 4)
+
+
```
rbolboac, line 98: remove extra blank line

rbolboac (commented):
some remarks from my side

drivers/adc/ad796x/ad796x.c
ahaslam2: Ok, that just means that the convention is that  if the desc returns something that
something has to be freed (even if the int func fails). 

i guess i can add those checks in remove and just call remove then.

rbolboac: Well... the convention is that if something is allocated and ret is not 0, it should be freed before returning.
However, seeing how we use calloc, all pointers are initialized with null, so if we call the remove function, even if memory has not been allocated, then the remove function should check for null and free the memory only if the pointer is not null.
Assuming that all remove functions check for null pointer before attempting to free it, it is safe to call the remove function even if the init has not been called. This will make the code easier to follow and shorter (a single label can be used).

projects/ad796x_fmcz/builds.json
ahaslam2: i tested using prebuilt a HDL shared sea-misc. i dont think this was the pulsar_lvds
as the readme in there does not mention anything about pulsar. Also the defines in parameters
do not not mention PULSAR. 

I think the HDL for this part still needs to be merged. i will ask

drivers/adc/ad796x/ad796x.c
ahaslam2: RE: if something is allocated and ret is not 0, it should be freed before returning

What i see from the no os-code is a bit confusing:

axi_dmac does not free before returning -1
(https://github.com/analogdevicesinc/no-OS/blob/main/drivers/axi_core/axi_dmac/axi_dmac.c#L349)

axi_adc_core frees the allocation before returning -1 But the the pointer is not set to null after the free. 
https://github.com/analogdevicesinc/no-OS/blob/main/drivers/axi_core/axi_adc_core/axi_adc_core.c#L677


So in case one of the int function call fails, im not sure we can trust the pointer to be freed, or
to be freed and null.

rbolboac: regarding https://github.com/analogdevicesinc/no-OS/blob/main/drivers/axi_core/axi_dmac/axi_dmac.c#L349, it looks like the remove function should be called if ret is not 0. so the suggestion works great in this case. I do not know why this implementation was done like that. personally I would rework it into something like:
```
struct axi_dmac *dmac;

dmac = (struct axi_dmac *)no_os_calloc(1, sizeof(*dmac));
if (!dmac)
	return -1;

dmac->name = init->name;
dmac->base = init->base;
dmac->irq_option = init->irq_option;

int32_t status = axi_dmac_detect_caps(dmac);
if (status < 0)
	goto free;

*dmac_core = dmac;

return 0;

free:
	no_os_free(dmac);
	return -1;
```

regarding https://github.com/analogdevicesinc/no-OS/blob/main/drivers/axi_core/axi_adc_core/axi_adc_core.c#L677,
the descriptor remains null in case of an error, only in case of success the descriptor is set: https://github.com/analogdevicesinc/no-OS/blob/main/drivers/axi_core/axi_adc_core/axi_adc_core.c#L673

so basically, dev->ad796x_core will be null if init is not successful, and when calling the remove function nothing will happen. 
the problem I see is that axi_adc_remove does not check for null pointer, we should add that check I think.

ahaslam2: yes, ok. with those changes. 

BTW i have done it this way because the logic i was following was
to keep balanced number of init/remove. 

this would mean that:
- init will free its resources in case of fail, a failed init is not a init  (like you mention)
- remove should not be called if init was not called before
- remove should not be called for a failed init
- should not assume the state of the pointer

ahaslam2:
changes:

- use remove function instead of err_ lables in driver init error path
- call driver remove in case of error on project main  
- misc style fixes.

drivers/adc/ad796x/ad796x.c
rbolboac: I understand... this is not mandatory. in my case the code would look cleaner...
also, to me there is no real value in adding a duplicate label: `err_pwm_0`

rbolboac: also, regarding my previous comment `the problem I see is that axi_adc_remove does not check for null pointer, we should add that check I think.` there is no problem with that API, it is safe to call free for null pointers.

drivers/adc/ad796x/ad796x.c
```diff
@@ -0,0 +1,297 @@
...
+/**
+ * @brief Remove the device and release resources.
+ * @param dev - The device structure.
+ * @return 0 in case of success, negative error code otherwise.
+ */
+int ad796x_remove(struct ad796x_dev *dev)
+{
+	int ret;
+
+	if (dev->ad796x_core) {
```
rbolboac, line 265: the only thing I would check for null here is dev.
freeing a null pointer is allowed

drivers/adc/ad796x/ad796x.c
```diff
@@ -0,0 +1,297 @@
...
+ * @brief Remove the device gpio resources.
+ * @param dev - The device structure.
+ * @return 0 in case of success, negative error code otherwise.
+ */
+static int ad796x_gpio_remove(struct ad796x_dev *dev)
+{
+	int ret, i;
+
+	struct no_os_gpio_desc *desc[] = {
+		dev->gpio_adc_en3_fmc,
```
rbolboac, line 120: dev should be checked for null before trying to access its members

drivers/adc/ad796x/ad796x.c
```diff
@@ -0,0 +1,297 @@
...
+		dev->gpio_adc_en2_fmc,
+		dev->gpio_adc_en1_fmc,
+		dev->gpio_adc_en0_fmc,
+	};
+
+	for (i = 0; i < AD796X_NUM_GPIOS; i++) {
+		if (!desc[i])
+			continue;
```
rbolboac, line 128: no_os_gpio_remove already checks for null pointer, no need to do it here

drivers/adc/ad796x/ad796x.c
ahaslam2: ill use a table index instead to loop though the gpios.

drivers/adc/ad796x/ad796x.c
```diff
@@ -0,0 +1,260 @@
...
+
+	ret = ad796x_gpio_remove(dev);
+	if (ret)
+		return ret;
+
+	no_os_free(dev);
+
+	return ret;
+}
+
```
amiclaus, line 260: drop extra line

drivers/adc/ad796x/ad796x.h
```diff
@@ -0,0 +1,116 @@
...
+int ad796x_init(struct ad796x_dev **device,
+		struct ad796x_init_param *init_param);
+
+/* Remove the device and release resources. */
+int ad796x_remove(struct ad796x_dev *dev);
+
+/* Read samples from the device */
+int32_t ad796x_read_data(struct ad796x_dev *dev, uint32_t *buf,
+			 uint16_t samples);
+
```
amiclaus, line 114: drop extra line

amiclaus (commented):
minor comments, otherwise lgtm

projects/ad796x_fmcz/src/examples/iio_example/iio_example.c
```diff
@@ -0,0 +1,125 @@
...
+
+	ret = iio_app_init(&app, app_init_param);
+	if (ret)
+		goto err_app_init;
+
+	ret = iio_app_run(app);
+	if (ret)
+		goto err_app_run;
+
+	return 0;
```
rbolboac, line 116: I would remove these lines...if fore some reason iio_app_run returns (either with 0 or another value), the remove APIs should call before exiting this example.

projects/ad796x_fmcz/src/examples/iio_example/iio_example.c
```diff
@@ -0,0 +1,125 @@
...
+
+err_app_run:
+	iio_app_remove(app);
+err_app_init:
+	iio_axi_adc_remove(iio_axi_adc_desc);
+err_adc_init:
+	ad796x_remove(adc_dev);
+	return ret;
```
rbolboac, line 124: if (ret) I would print an error message

projects/ad796x_fmcz/src/platform/xilinx/main.c
```diff
@@ -0,0 +1,72 @@
...
+
+	/* Enable the instruction cache. */
+	Xil_ICacheEnable();
+	/* Enable the data cache. */
+	Xil_DCacheEnable();
+#ifdef IIO_EXAMPLE
+	ret = iio_example_main();
+#elif BASIC_EXAMPLE
+	ret = basic_example_main();
+#endif
```
rbolboac, line 70: maybe add an `#error` message in case no example is defined

rbolboac (commented):
some remarks from my side, also ad796x driver documentation is missing
same remark regarding commit message length, can have up to 72 characters per line.

projects/ad796x_fmcz/src/examples/iio_example/iio_example.c
ahaslam2: Ok ill put print in after the ret that fails were i can show who produced the error.

rbolboac:
do you plan on adding the driver documentation in this PR or in a new one?

ahaslam2:
> do you plan on adding the driver documentation in this PR or in a new one?

will do.

drivers/adc/ad796x/README.rst
```diff
@@ -0,0 +1,218 @@
...
+This board operates in conjunction with the Zedboard. The evaluation software
+is provided to enable the user to perform detailed analysis of the AD7960's
+performance. The technical user guide includes the detailed description of
+the operation and set up of the evaluation board and software when operated
+with the System Demonstration Platform board.
+
+
+AD796X Device Configuration
+----------------------------
+
```
rbolboac, line 24: looks like device configuration documentation is missing...
in general, I think it will be useful to also describe the available APIs, just in case someone checks the documentation before the code... there are various examples here: http://analogdevicesinc.github.io/no-OS/drivers_doc.html

ahaslam2: i took the example from https://analogdevicesinc.github.io/no-OS/drivers/adm1177.html
where "driver initialization " is a subtopic of "device configuration "

in fact there is no device configuration for this driver. only driver initialization and reading data,
which i mention, but perhaps not clearly enough?

rbolboac: It is not clear that driver initialization is a subtopic of device configuration. they are using the same indentation. use `^^^^`, instead of `----` for driver initialization if that is the case

ahaslam2: ok , i  just followed what was done in the other docs: 

```
AD405X Device Configuration
---------------------------

Driver Initialization
---------------------

In order to be able to use the device, you will have to provide the support for
the communication protocol (SPI) as well as 3 external GPIOs for the CNV pin and two
```

```
ADM1177 Device Configuration
----------------------------

Driver Initialization
---------------------

In order to be able to use the device, you will have to provide the support for
the communication protocol (I2C) as mentioned above.

The first API to be called is **adm1177_init**. Make sure that it returns 0,
which means that the driver was initialized correctly.

```

rbolboac (approved):

amiclaus (approved):

PR #2140, opened 2024-04-04
Add support for AXI PWM Generator v2

jonathanns:
## Pull Request Description

This introduces support for the AXI PWM Gen. IP Core v2, which has been released in the HDL repository. It includes a runtime check within the driver to ensure compatibility with v1.  This does not affect the existing projects.

## PR Type

## PR Checklist

### DIFF
--- a/drivers/axi_core/axi_pwmgen/axi_pwm.c
+++ b/drivers/axi_core/axi_pwmgen/axi_pwm.c
@@ -57,20 +57,22 @@
 #define AXI_PWMGEN_REG_CORE_MAGIC	0x0C
 #define AXI_PWMGEN_REG_CONFIG		0x10
 #define AXI_PWMGEN_REG_NPWM		0x14
-#define AXI_PWMGEN_CH_PERIOD_BASE	0x40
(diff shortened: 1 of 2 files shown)


jonathanns:
v1:
* Fix memory leak

include/no_os_pwm.h
```diff
@@ -99,6 +99,8 @@ struct no_os_pwm_init_param {
 struct no_os_pwm_desc {
 	/** PWM mutex*/
 	void *mutex;
+	/** Hardware version necessary for checking for the right register offsets  */
+	uint32_t hw_major_ver;
```
buha, line 103: So i'm wondering about adding this here, it seems unfit. This is a top level API header and such a change has implications in all pwm implementations.

Either rename it to something very generically named like `version` which would apply to other implementations as well, or move this field to the axi_pwm implementation of this API, here: https://github.com/analogdevicesinc/no-OS/blob/main/drivers/axi_core/axi_pwmgen/axi_pwm_extra.h#L69

jonathanns:
v2:
* Moved `hw_major_ver` Field from `no_os_pwm_desc` to `axi_pwm_desc` struct as requested.

PR #2154, opened 2024-04-15
drivers:platform:stm32:Added DMA Driver

jansunil:
Added STM32 DMA platform driver

## Pull Request Description

## PR Type

## PR Checklist

### DIFF
--- a/drivers/platform/stm32/stm32_dma.c
+++ b/drivers/platform/stm32/stm32_dma.c
@@ -1,0 +1,297 @@
+/***************************************************************************//**
+ *   @file   stm32/stm32_dma.c
+ *   @brief  Implementation of stm32 DMA functionality.
+ * 	 @author Janani Sunil (janani.sunil@analog.com)
+********************************************************************************
+ * Copyright 2024(c) Analog Devices, Inc.
+ *
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *  - Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ *  - Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in
+ *    the documentation and/or other materials provided with the
+ *    distribution.
+ *  - Neither the name of Analog Devices, Inc. nor the names of its
+ *    contributors may be used to endorse or promote products derived
+ *    from this software without specific prior written permission.
+ *  - The use of this software may or may not infringe the patent rights
+ *    of one or more patent holders.  This license does not release you
+ *    from the requirement that you obtain separate licenses from these
+ *    patent holders to use this software.
+ *  - Use of the software either in source or binary form, must be run
+ *    on or directly connected to an Analog Devices Inc. component.
+ *
+ * THIS SOFTWARE IS PROVIDED BY ANALOG DEVICES "AS IS" AND ANY EXPRESS OR
+ * IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, NON-INFRINGEMENT,
+ * MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED.
+ * IN NO EVENT SHALL ANALOG DEVICES BE LIABLE FOR ANY DIRECT, INDIRECT,
+ * INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, INTELLECTUAL PROPERTY RIGHTS, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+*******************************************************************************/
+
+#include "stm32_dma.h"
+#include "no_os_error.h"
+#include "no_os_alloc.h"
+#include "stm32_irq.h"
+
+/******************************************************************************/
+/************************ Functions Definitions *******************************/
+/******************************************************************************/
+
+/**
+* @brief There is only one DMA controller
+*/
+static struct no_os_dma_desc *dma_descriptor;
+
+/**
+ * @brief Configure a DMA channel for a transfer.
+ * @param channel - The DMA channel descriptor.
+ * @param xfer - Descriptor for the transfer.
+ * @return 0 in case of success, negative error codes otherwise
+ */
+int stm32_dma_config_xfer(struct no_os_dma_ch *channel,
+			  struct no_os_dma_xfer_desc *xfer)
+{
+	struct stm32_dma_channel* sdma_ch;
+	int ret;
+
+	if (!channel || !xfer || !channel->extra)
+		return -EINVAL;
+
+	sdma_ch = channel->extra;
+
+	/* Note: Channel number is assigned via the Instance for MCU
+	 * families other than STM32F2, STM32F4 and STM32F7 */
+#if defined (STM32F2) || defined (STM32F4) || defined (STM32F7)
+	sdma_ch->hdma->Init.Channel = sdma_ch->ch_num;
+#else
+	sdma_ch->hdma->Instance = sdma_ch->ch_num;
+#endif
+	sdma_ch->hdma->Init.MemInc = sdma_ch->mem_increment ? DMA_MINC_ENABLE :
+				     DMA_MINC_DISABLE;
+	sdma_ch->hdma->Init.PeriphInc = sdma_ch->per_increment ? DMA_PINC_ENABLE :
+					DMA_PINC_DISABLE;
+
+	switch (sdma_ch->mem_data_alignment) {
+	case DATA_ALIGN_BYTE:
+		sdma_ch->hdma->Init.MemDataAlignment = DMA_MDATAALIGN_BYTE;
+		break;
+	case DATA_ALIGN_HALF_WORD:
+		sdma_ch->hdma->Init.MemDataAlignment = DMA_MDATAALIGN_HALFWORD;
+		break;
+	case DATA_ALIGN_WORD:
+		sdma_ch->hdma->Init.MemDataAlignment = DMA_MDATAALIGN_WORD;
+		break;
+	default:
+		return -EINVAL;
+	}
+
+	switch (sdma_ch->per_data_alignment) {
+	case DATA_ALIGN_BYTE:
+		sdma_ch->hdma->Init.PeriphDataAlignment = DMA_PDATAALIGN_BYTE;
+		break;
+	case DATA_ALIGN_HALF_WORD:
+		sdma_ch->hdma->Init.PeriphDataAlignment = DMA_PDATAALIGN_HALFWORD;
+		break;
+	case DATA_ALIGN_WORD:
+		sdma_ch->hdma->Init.PeriphDataAlignment = DMA_PDATAALIGN_WORD;
+		break;
+	default:
+		return -EINVAL;
+	}
+
+	switch (sdma_ch->dma_mode) {
+	case DMA_NORMAL_MODE:
+		sdma_ch->hdma->Init.Mode = DMA_NORMAL;
+		break;
+	case DMA_CIRCULAR_MODE:
+		sdma_ch->hdma->Init.Mode = DMA_CIRCULAR;
+		break;
+	default:
+		return -EINVAL;
+	}
+
+	switch (xfer->xfer_type) {
+	case MEM_TO_MEM:
+		sdma_ch->hdma->Init.Direction = DMA_MEMORY_TO_MEMORY;
+		break;
+	case MEM_TO_DEV:
+		sdma_ch->hdma->Init.Direction = DMA_MEMORY_TO_PERIPH;
+		break;
+	case DEV_TO_MEM:
+		sdma_ch->hdma->Init.Direction = DMA_PERIPH_TO_MEMORY;
+		break;
+	default:
+		return -EINVAL;
+	}
+
+	sdma_ch->src = xfer->src;
+	sdma_ch->dst = xfer->dst;
+	sdma_ch->length = xfer->length;
+
+	ret = HAL_DMA_Init(sdma_ch->hdma);
+	if (ret != HAL_OK)
+		return -EINVAL;
+
+	return 0;
+}
+
+/**
+ * @brief Initialize a DMA controller.
+ * @param desc - Descriptor to be initialized.
+ * @param param - Initialization parameter for the decriptor.
+ * @return 0 in case of success
+ * 	   -ENOMEM if there is not enough free memory
+ */
+int stm32_dma_init(struct no_os_dma_desc** desc,
+		   const struct no_os_dma_init_param* param)
+{
+
+	struct no_os_dma_desc* descriptor;
+	int ret;
+	uint8_t i;
+	struct no_os_irq_init_param irq_param = {
+		.irq_ctrl_id = 0,
+		.platform_ops = &stm32_irq_ops,
+		.extra = NULL
+	};
+
+	if (!desc)
+		return -EINVAL;
+
+	if (dma_descriptor) {
+		/*
+		 * There is only one DMA controller and it's already
+		 * initialized.
+		 */
+		*desc = dma_descriptor;
+		return 0;
+	}
+
+	descriptor = no_os_calloc(1, sizeof(*descriptor));
+	if (!descriptor)
+		return -ENOMEM;
+
+	descriptor->channels = no_os_calloc(param->num_ch,
+					    sizeof(*descriptor->channels));
+	if (!descriptor->channels) {
+		ret = -ENOMEM;
+		goto free_descriptor;
+	}
+
+	descriptor->id = param->id;
+	descriptor->num_ch = param->num_ch;
+
+	for (i = 0; i < param->num_ch; i++) {
+		descriptor->channels[i].free = true;
+	}
+
+	no_os_irq_ctrl_init(&descriptor->irq_ctrl, &irq_param);
+
+	*desc = descriptor;
+
+	return 0;
+
+free_descriptor:
+	no_os_free(descriptor);
+	descriptor = NULL;
+
+	return ret;
+}
+
+/**
+ * @brief Start a DMA transfer for a specific channel .
+ * @param desc - Descriptor for the DMA controller.
+ * @param chan - The DMA Channel
+ * @return 0 in case of success, negative error codes otherwise
+ */
+int stm32_dma_xfer_start(struct no_os_dma_desc *desc,
+			 struct no_os_dma_ch *chan)
+{
+	int ret;
+	struct stm32_dma_channel* sdma_chan;
+
+	if (!desc || !chan || !chan->extra)
+		return -EINVAL;
+
+	sdma_chan = chan->extra;
+
+	if (chan->irq_num) {
+		ret = HAL_DMA_Start_IT(sdma_chan->hdma, (uint32_t)sdma_chan->src,
+				       (uint32_t )sdma_chan->dst, sdma_chan->length);
+	} else {
+		ret = HAL_DMA_Start(sdma_chan->hdma, (uint32_t )sdma_chan->src,
+				    (uint32_t )sdma_chan->dst, sdma_chan->length);
+	}
+
+	if (ret != HAL_OK)
+		return -EINVAL;
+
+	return 0;
+}
+
+/**
+ * @brief Disable a DMA channel.
+ * @param desc - Descriptor for the DMA controller.
+ * @param chan - The DMA Channel
+ * @return 0 in case of success, negative error codes otherwise
+ */
+int stm32_dma_xfer_abort(struct no_os_dma_desc* desc,
+			 struct no_os_dma_ch* chan)
+{
+	int ret;
+	struct stm32_dma_channel* sdma_chan;
+
+	if (!desc || !chan || !chan->extra)
+		return -EINVAL;
+
+	sdma_chan = chan->extra;
+
+	if (chan->irq_num)
+		ret = HAL_DMA_Abort_IT(sdma_chan->hdma);
+	else
+		ret = HAL_DMA_Abort(sdma_chan->hdma);
+
+	if (ret != HAL_OK)
+		return -EINVAL;
+
+	return 0;
+}
+
+/**
+ * @brief Free the resources allocated for a DMA descriptor.
+ * @param desc - Descriptor to be freed.
+ * @return 0 in case of success, negative error code otherwise
+ */
+int stm32_dma_remove(struct no_os_dma_desc* desc)
+{
+	no_os_irq_ctrl_remove(desc->irq_ctrl);
+
+	no_os_free(desc->channels);
+
+	no_os_free(desc);
+
+	dma_descriptor = NULL;
+
+	return 0;
+}
+
+/**
+ * @brief stm32 platform specific DMA platform ops structure
+ */
+const struct no_os_dma_platform_ops stm32_dma_ops = {
+	.dma_init = stm32_dma_init,
+	.dma_config_xfer = stm32_dma_config_xfer,
+	.dma_xfer_start = stm32_dma_xfer_start,
+	.dma_xfer_abort = stm32_dma_xfer_abort,
+	.dma_remove = stm32_dma_remove
+};
+
--- a/drivers/platform/stm32/stm32_dma.h
+++ b/drivers/platform/stm32/stm32_dma.h
@@ -1,0 +1,106 @@
+/***************************************************************************//**
+ *   @file   stm32_dma.h
+ *   @brief  Platform independent function definitions and data types
+ * 	     for the DMA API.
+ * 	 @author Janani Sunil (janani.sunil@analog.com)
+********************************************************************************
+ * Copyright 2024(c) Analog Devices, Inc.
+ *
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *  - Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ *  - Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in
+ *    the documentation and/or other materials provided with the
+ *    distribution.
+ *  - Neither the name of Analog Devices, Inc. nor the names of its
+ *    contributors may be used to endorse or promote products derived
+ *    from this software without specific prior written permission.
+ *  - The use of this software may or may not infringe the patent rights
+ *    of one or more patent holders.  This license does not release you
+ *    from the requirement that you obtain separate licenses from these
+ *    patent holders to use this software.
+ *  - Use of the software either in source or binary form, must be run
+ *    on or directly connected to an Analog Devices Inc. component.
+ *
+ * THIS SOFTWARE IS PROVIDED BY ANALOG DEVICES "AS IS" AND ANY EXPRESS OR
+ * IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, NON-INFRINGEMENT,
+ * MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED.
+ * IN NO EVENT SHALL ANALOG DEVICES BE LIABLE FOR ANY DIRECT, INDIRECT,
+ * INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, INTELLECTUAL PROPERTY RIGHTS, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+*******************************************************************************/
+
+#ifndef STM32_DMA_H_
+#define STM32_DMA_H_
+
+#include "no_os_dma.h"
+#include "stm32_hal.h"
+
+/**
+ * @enum stm32_dma_data_alignment
+ * @brief DMA Data alignment
+ */
+enum stm32_dma_data_alignment {
+	DATA_ALIGN_BYTE = 0,
+	DATA_ALIGN_HALF_WORD = 1,
+	DATA_ALIGN_WORD = 2
+};
+
+/**
+ * @enum stm32_dma_mode
+ * @brief DMA Data Modes
+ */
+enum stm32_dma_mode {
+	DMA_NORMAL_MODE = 0,
+	DMA_CIRCULAR_MODE = 1
+};
+
+/**
+ * @struct stm32_dma_channel
+ * @brief STM32 DMA Channels
+ */
+struct stm32_dma_channel {
+	/* DMA Handle */
+	DMA_HandleTypeDef *hdma;
+	/* Channel Number */
+	uint32_t ch_num;
+	/* Memory Increment */
+	bool mem_increment;
+	/* Peripheral Increment */
+	bool per_increment;
+	/* Memory Data Alignment */
+	enum stm32_dma_data_alignment mem_data_alignment;
+	/* Peripheral Data Alignment */
+	enum stm32_dma_data_alignment per_data_alignment;
+	/* DMA Mode */
+	enum stm32_dma_mode dma_mode;
+	/* Source Address for the data */
+	uint8_t* src;
+	/* Destination Address for the data */
+	uint8_t* dst;
+	/* Transfer length in Bytes */
+	uint32_t length;
+};
+
+struct stm32_dma_init_param {
+	/* DMA Channel descriptor */
+	struct stm32_dma_channel *chn;
+};
+
+struct stm32_dma_desc {
+	/* DMA Channel Descriptor */
+	struct stm32_dma_channel* chn;
+};
+
+extern const struct no_os_dma_platform_ops stm32_dma_ops;
+
+#endif
+
--- a/drivers/platform/stm32/stm32_spi.c
+++ b/drivers/platform/stm32/stm32_spi.c
@@ -45,6 +45,10 @@
 #include "stm32_spi.h"
 #include "no_os_delay.h"
 #include "no_os_alloc.h"
+#include "stm32_dma.h"
+#ifdef HAL_TIM_MODULE_ENABLED
+#include "no_os_pwm.h"
+#endif
 
 static int stm32_spi_config(struct no_os_spi_desc *desc)
 {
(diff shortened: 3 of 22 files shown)


drivers/platform/stm32/stm32_dma.c
```diff
@@ -0,0 +1,99 @@
...
+	if (!desc || !param)
+		return -EINVAL;
+
+	dma_desc = (struct no_os_dma_desc*)no_os_calloc(1, sizeof(*dma_desc));
+	if (!dma_desc)
+		return -ENOMEM;
+
+	ddesc = (struct stm32_dma_desc*)no_os_calloc(1, sizeof(*ddesc));
+	if (!ddesc)
+		return -ENOMEM;
```
rbolboac, line 67: why is this needed? after all ddesc is a pointer, which points to the already allocated DMA handler in generated_main.c (I would assume a similar handling is done for DMA as with UART and other peripherals).

drivers/platform/stm32/stm32_dma.h
```diff
@@ -0,0 +1,72 @@
...
+	DMA_Stream_TypeDef* dma_stream;
+	uint32_t priority;
+	bool mem_increment;
+	bool per_increment;
+	uint32_t mem_data_alignment;
+	uint32_t periph_data_alignment;
+	uint32_t mode;
+	uint32_t dma_channel;
+	uint32_t direction;
+	enum no_os_irq_event event;
```
rbolboac, line 59: the init param can be directly the HandleTypeDef pointer for DMA

rbolboac (commented):
This still needs some work.
First of all, I would recommend to take a look at how other stm32 peripherals are handled and handle the DMA in a similar way, UART can be such an example where the UART handle in generated_main is used.
Secondly, please take a look at what was done for maxim platform, the API you currently have simply initializes the peripheral, without taking into consideration any of no_os_dma_init_param fields. Those fields should be mapped to the stm32 hdma instance.
For example the instance can be the id, the mode can be mapped to xfer_type and so on.
I would say it's no easy task to add support for dma...and it will require a lot of API implementations and functionality mapping.
Of course, implementation for the other platform ops will be necessary, and I think maxim_dmac.c drivers can serve as an example for this.
Lastly, it would be nice to have a little context, to see how you want to use the DMA.
For example, in no_os there is support for SPI transfers with dma, see no_os_spi_transfer_dma_sync and no_os_spi_transfer_dma_async which were also implemented in maxim_spi.c driver (can serve as an example if this is the final goal for DMA usage).

drivers/platform/stm32/stm32_dma.h
jansunil: I've added a handletypedef to the inti params now

drivers/platform/stm32/stm32_dma.c
```diff
@@ -0,0 +1,99 @@
...
+	*desc = dma_desc;
+
+	return 0;
+}
+
+/**
+ * @brief stm32 platform specific DMA platform ops structure
+ */
+const struct no_os_dma_platform_ops stm32_dma_ops = {
+	.dma_init = stm32_dma_init,
```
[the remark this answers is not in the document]
jansunil: Added a remove function now

jansunil:
@rbolboac -  Thanks for the suggestions. There are some updates made to the driver in order to use the other structure made available in the no_os_dma_init_params. The driver flow is also aligned to the maxim implementation.

In order to get a context of how the drivers can be used, an example implementation has been added in the stm32_spi_dma_transfer_sync() implementation in the existing SPI driver. 

Please let me know on your inputs further.

drivers/platform/stm32/stm32_spi.h
```diff
@@ -66,6 +75,14 @@ struct stm32_spi_desc {
 	uint32_t input_clock;
 	/** Chip select gpio descriptor */
 	struct no_os_gpio_desc *chip_select;
+	/** DMA Descriptor */
+	struct no_os_dma_desc* dma_desc;
```
buha, line 79: that's good progress, i see fewer members in this structure but it can be improved further

this field shouldn't be here, actually the parent descriptor of this `stm32_spi_desc` is a `no_os_dma_desc` and you have direct access to that in the DMA API functions

jansunil: @buha  I am not sure I got the point right here. Do you mean to add the no_os_dma_desc to the no_os_spi_desc? We might need access to this DMA descriptor somewhere in the SPI descriptors so that it can be used in the async functions

jansunil:
Hi all - are there any further review comments on this PR?

rbolboac:
> Hi all - are there any further review comments on this PR?

First request would be to please fix CI errors.

drivers/platform/stm32/stm32_spi.c
```diff
@@ -45,6 +45,9 @@
 #include "stm32_spi.h"
 #include "no_os_delay.h"
 #include "no_os_alloc.h"
+#include "stm32_dma.h"
+#include "no_os_pwm.h"
+#include "stm32_pwm.h"
```
buha, line 50: is this pwm dependency really needed ?

jansunil: So the dependency we have is to control the chip select.
Since this is a DMA operation and cannot toggle chip select individually for each byte/word, we need a PWM module here

drivers/platform/stm32/stm32_dma.c
```diff
@@ -0,0 +1,284 @@
...
+ */
+int stm32_dma_init(struct no_os_dma_desc** desc,
+		   const struct no_os_dma_init_param* param)
+{
+
+	struct no_os_dma_desc* descriptor;
+	int ret;
+	uint8_t i;
+
+	if (!desc)
```
CiprianRegus, line 153: The current DMA driver for maxim uses shared references for `no_os_dma_desc` descriptors (meaning that if you call the dma_init function with the same `id` field you'll get the same `no_os_dma_desc *` pointer). Is there a reason for not implementing the same here?

drivers/platform/stm32/stm32_dma.c
```diff
@@ -0,0 +1,284 @@
...
+ */
+int stm32_dma_remove(struct no_os_dma_desc* desc)
+{
+	int ret;
+	uint8_t i;
+
+	no_os_irq_ctrl_remove(desc->irq_ctrl);
+
+	for (i = 0; i < desc->num_ch; i++)
+		no_os_free(&desc->channels[i]);
```
CiprianRegus, line 268: This is incorrect since you'll end up double freeing some memory. Should be `no_os_free(desc->channels)`

drivers/platform/stm32/stm32_dma.c
```diff
@@ -0,0 +1,284 @@
...
+}
+
+/**
+ * @brief Free the resources allocated for a DMA descriptor.
+ * @param desc - Descriptor to be freed.
+ * @return 0 in case of success, negative error code otherwise
+ */
+int stm32_dma_remove(struct no_os_dma_desc* desc)
+{
+	int ret;
```
CiprianRegus, line 262: Unused

drivers/platform/stm32/stm32_dma.h
```diff
@@ -0,0 +1,106 @@
...
+enum stm32_dma_mode {
+	DMA_NORMAL_MODE = 0,
+	DMA_CIRCULAR_MODE = 1
+};
+
+/**
+ * @struct stm32_dma_channel
+ * @brief STM32 DMA Channels
+ */
+struct stm32_dma_channel {
```
CiprianRegus, line 69: I'm a bit confused by this... Is this struct specific to a DMA instance or channel? Why do we need fields such as `src`, `dst` or `length` during init?

drivers/platform/stm32/stm32_spi.c
```diff
@@ -212,6 +216,40 @@ int32_t stm32_spi_init(struct no_os_spi_desc **desc,
 	if (ret)
 		goto error;
 
+	if (sinit->dma_init) {
+		ret = no_os_dma_init(&sdesc->dma_desc, sinit->dma_init);
+		if (ret) {
+			return ret;
+		}
```
CiprianRegus, line 223: remove the brackets

drivers/platform/stm32/stm32_spi.c
```diff
@@ -212,6 +216,40 @@ int32_t stm32_spi_init(struct no_os_spi_desc **desc,
 	if (ret)
 		goto error;
 
+	if (sinit->dma_init) {
+		ret = no_os_dma_init(&sdesc->dma_desc, sinit->dma_init);
+		if (ret) {
+			return ret;
```
CiprianRegus, line 222: `goto error`

drivers/platform/stm32/stm32_spi.c
```diff
@@ -386,12 +424,134 @@ int32_t stm32_spi_write_and_read(struct no_os_spi_desc *desc,
...
+		rx_ch_xfer[i].dst = msgs[i].rx_buff;
+		rx_ch_xfer[i].src = &(SPIx->DR);
+		rx_ch_xfer[i].periph = NO_OS_DMA_IRQ;
+		rx_ch_xfer[i].xfer_type = DEV_TO_MEM;
+		rx_ch_xfer[i].length = msgs[i].bytes_number;
+	}
+
+	rx_ch->id = sdma_rx->hdma;
+	rx_ch->sg_list = NULL;
+	if (sdma_rx->irq_num > 0)
```
CiprianRegus, line 477: The `irq_num` is `uint32_t`. Moreover, I think this adds unnecessary complexity. I don't see why would the user care to assign a specific interrupt number.

projects/ad74413r/src.mk
```diff
@@ -37,7 +38,8 @@ SRCS += $(DRIVERS)/api/no_os_gpio.c \
 		$(NO-OS)/util/no_os_crc8.c \
 		$(NO-OS)/util/no_os_util.c \
 		$(NO-OS)/util/no_os_alloc.c \
-		$(NO-OS)/util/no_os_mutex.c
+		$(NO-OS)/util/no_os_mutex.c \
+		$(NO-OS)/util/no_os_dma.c \
```
CiprianRegus, line 42: No such file

drivers/platform/stm32/stm32_spi.c
```diff
@@ -386,12 +424,134 @@ int32_t stm32_spi_write_and_read(struct no_os_spi_desc *desc,
...
+ * 	  completion before returning.
+ * @param desc - The SPI descriptor.
+ * @param msgs - The messages array.
+ * @param len - Number of messages.
+ * @param callback - Function to be invoked once the transfers are done.
+ * @param ctx - User defined parameter for the callback function.
+ * @param is_async - Whether or not the function should wait for the completion.
+ * @return 0 in case of success, errno codes otherwise.
+ */
+int32_t stm32_spi_dma_transfer_sync(struct no_os_spi_desc* desc,
```
CiprianRegus, line 539: What is the behavior of this transfer function? I see it doesn't wait for the transfer to finish before returning, but `callback` is ignored as well.

drivers/platform/stm32/stm32_spi.c
CiprianRegus: As far as I know this functionality is not supported even for regular SPI transfers. Is there a use case for this? Why wouldn't you chain multiple byte/word transfers and handle the CS control in the transfer done interrupt?

drivers/platform/stm32/stm32_dma.h
```diff
@@ -0,0 +1,106 @@
...
+	/* Peripheral Increment */
+	bool per_increment;
+	/* Memory Data Alignment */
+	enum stm32_dma_data_alignment mem_data_alignment;
+	/* Peripheral Data Alignment */
+	enum stm32_dma_data_alignment per_data_alignment;
+	/* DMA Mode */
+	enum stm32_dma_mode dma_mode;
+	/* IRQ ID */
+	uint32_t irq_num;
```
rbolboac, line 85: already present in struct no_os_dma_ch, should use that instead

drivers/platform/stm32/stm32_spi.c
jansunil: The goal of using a DMA is to read huge chunks of samples without interference of the core. The rate at which the read happens is at around 1MSPS, where in I would end up reading 2 bytes per conversion. Triggering interrupts at that high a rate just to control the state of the CS GPIO alone would be too much overhead on the core.

Hence, I am trying to control the CS via a PWM, where in I read n number of samples and get into the interrupt handler after reception of 400 samples and not after every byte

drivers/platform/stm32/stm32_dma.c
```diff
@@ -0,0 +1,284 @@
...
+	sdma_ch = channel->extra;
+
+	sdma_ch->hdma->Init.Channel = sdma_ch->ch_num;
+	sdma_ch->hdma->Init.MemInc = sdma_ch->mem_increment ? DMA_MINC_ENABLE :
+				     DMA_MINC_DISABLE;
+	sdma_ch->hdma->Init.PeriphInc = sdma_ch->per_increment ? DMA_PINC_ENABLE :
+					DMA_PINC_DISABLE;
+	sdma_ch->hdma->Init.MemDataAlignment = sdma_ch->mem_data_alignment;
+	sdma_ch->hdma->Init.PeriphDataAlignment = sdma_ch->per_data_alignment;
+	sdma_ch->hdma->Init.Mode = sdma_ch->dma_mode;
```
rbolboac, line 72: don't really understand the purpose of these and also the switch statement below. basically they realize the same thing without the invalid value check

drivers/platform/stm32/stm32_dma.c
```diff
@@ -0,0 +1,284 @@
...
+	if (!desc || !chan || !chan->extra)
+		return -EINVAL;
+
+	sdma_chan = chan->extra;
+
+	if (sdma_chan->irq_num) {
+		ret = HAL_DMA_Start_IT(sdma_chan->hdma, (uint32_t)sdma_chan->src,
+				       (uint32_t )sdma_chan->dst, sdma_chan->length);
+		if (ret != HAL_OK)
+			return -EINVAL;
```
rbolboac, line 214: can be taken out of if / else

drivers/platform/stm32/stm32_dma.c
```diff
@@ -0,0 +1,284 @@
...
+	} else {
+		ret = HAL_DMA_Start(sdma_chan->hdma, (uint32_t )sdma_chan->src,
+				    (uint32_t )sdma_chan->dst, sdma_chan->length);
+		if (ret != HAL_OK)
+			return -EINVAL;
```
rbolboac, line 219: can be taken out of if / else

drivers/platform/stm32/stm32_dma.c
```diff
@@ -0,0 +1,284 @@
...
+
+	if (!desc || !chan || !chan->extra)
+		return -EINVAL;
+
+	sdma_chan = chan->extra;
+
+	if (sdma_chan->irq_num) {
+		ret = HAL_DMA_Abort_IT(sdma_chan->hdma);
+		if (ret != HAL_OK)
+			return -EINVAL;
```
rbolboac, line 245: can be taken out of if / else

drivers/platform/stm32/stm32_dma.c
```diff
@@ -0,0 +1,284 @@
...
+	} else {
+		ret = HAL_DMA_Abort(sdma_chan->hdma);
+		if (ret != HAL_OK)
+			return -EINVAL;
```
rbolboac, line 249: can be taken out of if / else

drivers/platform/stm32/stm32_dma.h
```diff
@@ -0,0 +1,106 @@
+/***************************************************************************//**
+ *   @file   stm32_dma.h
+ *   @brief  Platform independent function definitions and data types
+ * 	     for the DMA API.
```
rbolboac, line 4: missing author

drivers/platform/stm32/stm32_spi.c
jansunil: I am trying to implement a non blocking function, the goal is to return from this API irrespective of whether the transfer/receive is complete or not. 
The callback is being registeredwithin the no_os_dma_config_xfer() and enabled via the no_os_dma_xfer_start()

drivers/platform/stm32/stm32_spi.c
```diff
@@ -191,6 +194,7 @@ int32_t stm32_spi_init(struct no_os_spi_desc **desc,
 	csip.pull = NO_OS_PULL_NONE;
 	csip.extra = &csip_extra;
 	csip.platform_ops = &stm32_gpio_ops;
+
```
rbolboac, line 197: unrelated change

drivers/platform/stm32/stm32_dma.h
jansunil: This struct is specific to the channel. 
We do not really need them in the init, but its required in the stm32_dma_xfer_start(). The HAL_DMA_Start apis need to be provided with the src/dst addresses in order to initiate the DMA transfer

drivers/platform/stm32/stm32_spi.c
```diff
@@ -212,6 +216,40 @@ int32_t stm32_spi_init(struct no_os_spi_desc **desc,
 	if (ret)
 		goto error;
 
+	if (sinit->dma_init) {
+		ret = no_os_dma_init(&sdesc->dma_desc, sinit->dma_init);
+		if (ret) {
```
rbolboac, line 221: brackets not needed

drivers/platform/stm32/stm32_spi.c
```diff
@@ -212,6 +216,40 @@ int32_t stm32_spi_init(struct no_os_spi_desc **desc,
...
+			return ret;
+		}
+
+		if (sinit->rxdma_ch) {
```
rbolboac, line 225: brackets not needed for single line statements, see everywhere where applicable

drivers/platform/stm32/stm32_spi.c
jansunil: irq_num is a field required in the no_os_irq_register_callback()/ no_os_irq_set_priority(). These are being called in the no_os_dma_config_xfer() for configuring transfer complete callbacks

rbolboac (commented):
some other comments from my side,
also please squash the CI fix commit into the first one.
I would also suggest to have 2 commits:
- one to add stm32_dma support
- the second one to add dma support for SPI transfers

jansunil:
Changelog: 
Addressed comments from @CiprianRegus

drivers/platform/stm32/stm32_dma.c
jansunil: I am not sure I got this right..- what is the shared references you are referring to here?

drivers/platform/stm32/stm32_dma.c
jansunil: I've removed free of channels. I suppose freeing up the desc should take care of this as well?

jansunil:
Hi All- Please hold the review. I shall address all the comments and add a comment here

drivers/platform/stm32/stm32_dma.c
CiprianRegus: You still need to free `desc->channels`. What I was referring to is to replace

```
for (i = 0; i < desc->num_ch; i++)
		no_os_free(&desc->channels[i]);
```

with

```
no_os_free(desc->channels);
```

drivers/platform/stm32/stm32_dma.h
```diff
@@ -0,0 +1,106 @@
...
+
+/**
+ * @struct stm32_dma_channel
+ * @brief STM32 DMA Channels
+ */
+struct stm32_dma_channel {
+	/* DMA Handle */
+	DMA_HandleTypeDef *hdma;
+	/* Channel Number */
+	uint32_t ch_num;
```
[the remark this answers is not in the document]
jansunil: no_os_dma_ch consists of the ``id``, which is interpreted as the dma handle in the no_os_dma_config_xfer() - https://github.com/analogdevicesinc/no-OS/blob/main/drivers/api/no_os_dma.c#L285

The ch_num parameter needs to be a 32 bit number.

Please let me know if my understanding is wrong

drivers/platform/stm32/stm32_dma.c
jansunil: Intention is to configure the DMA init parameters via the platform drivers. 
A switch case has been added since there are enum elements in the stm32_dma.h to realize the parameters, so that the user doesn't really have to look into the HAL drivers to configure the metrics. Do you suggest any other alternate approach here?

rbolboac: what I mean is that you should remove the 3 lines above, because the switch blocks are assigning MemDataAlignment, PeriphDataAlignment and Mode fields.

jansunil: Oh yeah, you are right here. I had overlooked the duplicated assignments- I've removed the 3 lines above

drivers/platform/stm32/stm32_pwm.c
```diff
@@ -56,6 +56,8 @@
 /******************************************************************************/
 /************************ Functions Definitions *******************************/
 /******************************************************************************/
+
+#ifdef HAL_TIM_MODULE_ENABLED
```
rbolboac, line 60: please remove these from stm32_pwm, it is not necessary

drivers/platform/stm32/stm32_spi.c
```diff
@@ -229,6 +228,7 @@ int32_t stm32_spi_init(struct no_os_spi_desc **desc,
 		sdesc->dma_desc->sg_handler = sinit->dma_init->sg_handler;
 	}
 
+#ifdef HAL_TIM_MODULE_ENABLED
```
rbolboac, line 231: all changes added in stm32_spi should be moved in the previous commit

projects/max14906/src/platform/stm32/platform_src.mk
```diff
@@ -10,6 +12,8 @@ INCS += $(PLATFORM_DRIVERS)/stm32_delay.h     \
 SRCS += $(PLATFORM_DRIVERS)/stm32_delay.c     \
 	$(PLATFORM_DRIVERS)/stm32_gpio.c      \
 	$(PLATFORM_DRIVERS)/stm32_spi.c       \
+	$(PLATFORM_DRIVERS)/stm32_pwm.c       \
```
rbolboac, line 15: I think the whole idea for using HAL_TIM_MODULE_ENABLED is to not have to  include stm32_pwm.c, unless the ioc contains a timer module and it actually uses PWM for CS. this is not valid for any of the projects is No-OS. so please remove all stm32_pwm files, and no_os_pwm files from projects (added in this commit).
the same remarks applies for dma. currently no project is using spi with DMA, so why would we include those files? stm32_spi should compile without stm32_dma, if HAL_DMA_MODULE_ENABLED is not defined.

of course, you can add a project which showcases how to use SPI with DMA transfers and PWM CS. this will be use as an example (should contain DMA init and PWM init structures and the ioc should contain the modules as well).

drivers/platform/stm32/stm32_spi.c
```diff
@@ -45,6 +45,10 @@
 #include "stm32_spi.h"
 #include "no_os_delay.h"
 #include "no_os_alloc.h"
+#ifdef HAL_DMA_MODULE_ENABLED
+#include "stm32_dma.h"
+#endif
+#include "no_os_pwm.h"
```
rbolboac, line 51: include only if HAL_TIM_MODULE_ENABLED is defined

drivers/platform/stm32/stm32_dma.c
```diff
@@ -60,6 +60,7 @@ static struct no_os_dma_desc *dma_descriptor;
 int stm32_dma_config_xfer(struct no_os_dma_ch *channel,
 			  struct no_os_dma_xfer_desc *xfer)
 {
+#ifdef HAL_DMA_MODULE_ENABLED
```
rbolboac, line 63: why would you use stm32_dma.c if HAL_DMA_MODULE_ENABLED is not defined? I think this check is not needed...
only use this check in stm32_spi, because the spi driver can be used without the DMA module.
also, this change is not related to this commit and should be part of the first commit.

drivers/platform/stm32/stm32_dma.c
```diff
@@ -68,7 +69,11 @@ int stm32_dma_config_xfer(struct no_os_dma_ch *channel,
 
 	sdma_ch = channel->extra;
 
+#if defined (STM32L4) || defined (STM32H5)
```
rbolboac, line 72: a comment here would be nice... why only these families? I see there are other families which have hdma->Instance available. I think the problem is that the Init.Channels is not available for most of them.
so maybe change this check to only use Init.Channel for stm32f2, f4 and f7: https://github.com/search?q=org%3ASTMicroelectronics+hdma-%3EInit.Channel&type=code
also, this change is not related to this commit and should be part of the first commit.

drivers/platform/stm32/stm32_pwm.h
```diff
@@ -64,6 +64,8 @@ enum TimOCMode {
 	TIM_OC_PWM2 = 2,
 };
 
+#ifdef HAL_TIM_MODULE_ENABLED
```
rbolboac, line 67: please remove these from stm32_pwm, it is not necessary. same remark applies as for DMA, why would you use stm32_pwm without a timer module?
there is no HAL_SPI_MODULE_ENABLED check in stm32_spi...

drivers/platform/stm32/stm32_spi.c
```diff
@@ -212,9 +216,45 @@ int32_t stm32_spi_init(struct no_os_spi_desc **desc,
 	if (ret)
 		goto error;
 
+	if (sinit->dma_init) {
```
rbolboac, line 219: everything related to DMA from stm32_spi shoul use #ifdef HAL_DMA_MODULE_ENABLED and everything related to PWM should use HAL_TIM_MODULE_ENABLED

rbolboac (commented):
it still needs a little bit of work from my point of view.

drivers/platform/stm32/stm32_spi.c
```diff
@@ -386,12 +430,134 @@ int32_t stm32_spi_write_and_read(struct no_os_spi_desc *desc,
...
+ * @param callback - Function to be invoked once the transfers are done.
+ * @param ctx - User defined parameter for the callback function.
+ * @param is_async - Whether or not the function should wait for the completion.
+ * @return 0 in case of success, errno codes otherwise.
+ */
+int32_t stm32_spi_dma_transfer_async(struct no_os_spi_desc* desc,
+				     struct no_os_spi_msg* msgs,
+				     uint32_t len,
+				     void (*callback)(void*),
+				     void* ctx, bool is_async)
```
CiprianRegus, line 549: Same as before, this callback is totally ignored and the `stm32_spi_dma_transfer_async` function doesn't take the same parameters as `no_os_spi_transfer_dma_async`, so there is no way this can work.

jansunil: The callback is being registeredwithin the no_os_dma_config_xfer() and enabled via the no_os_dma_xfer_start()

drivers/platform/stm32/stm32_spi.c
```diff
@@ -212,9 +218,47 @@ int32_t stm32_spi_init(struct no_os_spi_desc **desc,
...
+	}
+#endif
+
+#ifdef HAL_TIM_MODULE_ENABLED
+	if (sinit->pwm_init) {
+		struct stm32_pwm_desc* spwm_desc;
+		/* Initialize CS PWM */
+		ret = no_os_pwm_init(&sdesc->pwm_desc, sinit->pwm_init);
+		if (ret)
+			goto error;
```
rbolboac, line 243: missing no_os_dma_remove in case of an error

drivers/platform/stm32/stm32_spi.c
```diff
@@ -386,12 +430,134 @@ int32_t stm32_spi_write_and_read(struct no_os_spi_desc *desc,
...
+	return stm32_config_dma_and_start(desc, msgs, len);
+}
+
 /**
  * @brief stm32 platform specific SPI platform ops structure
  */
 const struct no_os_spi_platform_ops stm32_spi_ops = {
 	.init = &stm32_spi_init,
 	.write_and_read = &stm32_spi_write_and_read,
 	.remove = &stm32_spi_remove,
```
rbolboac, line 560: the spi remove should remove everyting that was added in spi_config (dma and pwm)

rbolboac (commented):
also, second commit message has a typo, replace 4) with 2)

drivers/platform/stm32/stm32_spi.c
jansunil: @CiprianRegus - what is the difference between the callback that's in the argument here and the one in the no_os_dma_init param?

jansunil:
Hi All- The builds are all passing now. I have also addressed to the comments and replied back to the conversations. Does this look good to go?

drivers/platform/stm32/stm32_spi.c
CiprianRegus: The callback in the `no_os_dma_init_param` struct is meant to be called once a DMA transfer finished (doesn't matter if the transfer was for the SPI peripheral or it was a memory to memory transfer). The callback in the `no_os_spi_transfer_dma_async()` function should be called once the DMA SPI transfer issued by the function call is done.

projects/adxrs290-pmdz/src.mk
```diff
@@ -29,7 +29,8 @@ INCS += $(INCLUDE)/no_os_delay.h        \
         $(INCLUDE)/no_os_util.h         \
         $(INCLUDE)/no_os_units.h        \
         $(INCLUDE)/no_os_alloc.h        \
-        $(INCLUDE)/no_os_mutex.h
+        $(INCLUDE)/no_os_mutex.h         \
+        $(INCLUDE)/no_os_pwm.h
```
rbolboac, line 33: is this needed?

projects/adxrs290-pmdz/src.mk
```diff
@@ -41,4 +42,5 @@ SRCS += $(DRIVERS)/api/no_os_gpio.c     \
         $(NO-OS)/util/no_os_list.c      \
         $(NO-OS)/util/no_os_util.c      \
         $(NO-OS)/util/no_os_alloc.c     \
+        $(DRIVERS)/api/no_os_pwm.c     \
```
rbolboac, line 45: is this needed?

projects/adxrs290-pmdz/src/platform/stm32/platform_src.mk
```diff
@@ -2,13 +2,17 @@ INCS += $(PLATFORM_DRIVERS)/stm32_delay.h     \
         $(PLATFORM_DRIVERS)/stm32_gpio.h      \
         $(PLATFORM_DRIVERS)/stm32_hal.h       \
         $(PLATFORM_DRIVERS)/stm32_spi.h       \
+        $(PLATFORM_DRIVERS)/stm32_pwm.h       \
```
rbolboac, line 5: is this needed?

projects/eval-adis1646x/src.mk
```diff
@@ -26,7 +26,7 @@ INCS += $(INCLUDE)/no_os_delay.h     \
 		$(INCLUDE)/no_os_util.h \
 		$(INCLUDE)/no_os_units.h \
 		$(INCLUDE)/no_os_alloc.h \
-        	$(INCLUDE)/no_os_mutex.h
+        $(INCLUDE)/no_os_mutex.h
```
rbolboac, line 29: unrelated change

projects/eval-adis1647x/src.mk
```diff
@@ -26,7 +26,7 @@ INCS += $(INCLUDE)/no_os_delay.h     \
 		$(INCLUDE)/no_os_util.h \
 		$(INCLUDE)/no_os_units.h \
 		$(INCLUDE)/no_os_alloc.h \
-        	$(INCLUDE)/no_os_mutex.h
```
rbolboac, line 29: unrelated change

projects/eval-adxl313z/src.mk
```diff
@@ -26,7 +26,7 @@ INCS += $(INCLUDE)/no_os_delay.h     \
 		$(INCLUDE)/no_os_units.h \
 		$(INCLUDE)/no_os_init.h \
 		$(INCLUDE)/no_os_alloc.h \
-        	$(INCLUDE)/no_os_mutex.h
+        $(INCLUDE)/no_os_mutex.h
```
rbolboac, line 29: unrelated change

projects/eval-adxl313z/src.mk
```diff
@@ -38,7 +38,7 @@ SRCS += $(DRIVERS)/api/no_os_gpio.c \
 		$(DRIVERS)/api/no_os_uart.c \
 		$(NO-OS)/util/no_os_util.c \
 		$(NO-OS)/util/no_os_alloc.c \
-        	$(NO-OS)/util/no_os_mutex.c
+        $(NO-OS)/util/no_os_mutex.c
```
rbolboac, line 41: unrelated change

projects/eval-adxl355-pmdz/src.mk
```diff
@@ -27,7 +27,7 @@ INCS += $(INCLUDE)/no_os_delay.h     \
 		$(INCLUDE)/no_os_units.h \
 		$(INCLUDE)/no_os_init.h \
 		$(INCLUDE)/no_os_alloc.h \
-        	$(INCLUDE)/no_os_mutex.h
```
rbolboac, line 30: unrelated change

projects/eval-adxl355-pmdz/src.mk
```diff
@@ -40,7 +40,7 @@ SRCS += $(DRIVERS)/api/no_os_gpio.c \
 		$(NO-OS)/util/no_os_list.c \
 		$(NO-OS)/util/no_os_util.c \
 		$(NO-OS)/util/no_os_alloc.c \
-        	$(NO-OS)/util/no_os_mutex.c
+        $(NO-OS)/util/no_os_mutex.c
```
rbolboac, line 43: unrelated change

rbolboac (commented):
also, the third commit should be squashed in the second one because the changes depend on each other.

drivers/platform/stm32/stm32_spi.c
jansunil: As discussed in the call, I've added a condition to overwrite the desc->sg_handler if the user provides a callback in the function argument of the ``no_os_spi_transfer_dma_async()`` .

projects/adxrs290-pmdz/src.mk
jansunil: The adxrs290-pmdz project has the TIM module enabled in the .ioc file. Hence added the PWM drivers here as we are making a check for the HAL_TIM_MODULE_ENABLED in the stm32_spi.c driver

projects/adxrs290-pmdz/src.mk
jansunil: The adxrs290-pmdz project has the TIM module enabled in the .ioc file. Hence added the PWM drivers here as we are making a check for the HAL_TIM_MODULE_ENABLED in the stm32_spi.c driver

projects/adxrs290-pmdz/src/platform/stm32/platform_src.mk
jansunil: The adxrs290-pmdz project has the TIM module enabled in the .ioc file. Hence added the PWM drivers here as we are making a check for the HAL_TIM_MODULE_ENABLED in the stm32_spi.c driver

projects/eval-adis1646x/src.mk
jansunil: Removed this

projects/eval-adis1647x/src.mk
jansunil: Removed this

projects/eval-adxl313z/src.mk
jansunil: Removed this

projects/eval-adxl355-pmdz/src.mk
jansunil: Removed this

projects/eval-adxl355-pmdz/src.mk
jansunil: Removed this

jansunil:
Changelog:
1) Removed unrelated spacing changes in the src/stm32_src.mk
2) Added snippet to overwrite the callback function if the user specifies one via the no_os_spi_transfer_dma_async()

jansunil:
Changelog:
Fixed LinuxBuilds Documentation errors

projects/eval-adxl355-pmdz/src.mk
```diff
@@ -40,7 +40,7 @@ SRCS += $(DRIVERS)/api/no_os_gpio.c \
 		$(NO-OS)/util/no_os_list.c \
 		$(NO-OS)/util/no_os_util.c \
 		$(NO-OS)/util/no_os_alloc.c \
-        	$(NO-OS)/util/no_os_mutex.c
+		$(NO-OS)/util/no_os_mutex.c
```
rbolboac, line 43: for some reason these whitespace changes still appear here in the commit

rbolboac (approved):
looks good from my point of view, although @CiprianRegus should have the final word.
also, please do not forget to add (in the near future) a new PR with a project example using STM32 DMA with SPI (with and without PWM for CS would be nice) and the necessary documentation on wiki.

jansunil:
Changelog: Addressed offline comment from @CiprianRegus - Removed the is_async parameter from the ``stm32_spi_dma_transfer_async()`` in order to match the no_os_spi.h definition

CiprianRegus (approved):

rbolboac (approved):

PR #2158, opened 2024-04-18
drivers:uart: Register user-supplied rxcallback

ribdp:
## Pull Request Description

1. Add rx callback function pointer to no-OS uart desc
2. Register user-supplied callback instead of default, if supplied

## PR Type

## PR Checklist


drivers/platform/stm32/stm32_uart.c
```diff
@@ -142,7 +142,11 @@ static int32_t stm32_uart_init(struct no_os_uart_desc **desc,
 		if (ret < 0)
 			goto error;
 
-		sud->rx_callback.callback = uart_rx_callback;
+		if (param->rx_callback) {
+			sud->rx_callback.callback = param->rx_callback;
+		} else {
+			sud->rx_callback.callback = uart_rx_callback;
+		}
```
rbolboac, line 149: alright so... I think you should explain why you want to do this. why do you want to stop using the uart_rx_callback altogether? 
if indeed a user callback is needed to do something else besides what is already done in uart_rx_callback,  I would add the call inside uart_rx_callback. 
however, if that is not the case, I would like a little context for this change. why does the current callback does not work for your use-case?

buha: I agree with @rbolboac, if you really want a user callback, call it at the end of uart_rx_callback in stm32_uart.c, don't replace the callback that feeds characters into lf256fifo.

ribdp: > alright so... I think you should explain why you want to do this. why do you want to stop using the uart_rx_callback altogether? if indeed a user callback is needed to do something else besides what is already done in uart_rx_callback, I would add the call inside uart_rx_callback. however, if that is not the case, I would like a little context for this change. why does the current callback does not work for your use-case?

Makes sense. Adding the callback inside the uart_rx_callback is definitely better aligned with what I wanted to do. However, thinking about my own use case some more, I'm not sure if even this would go far enough in supporting that use case. I'll have to look into it a bit more. Will change this to a draft PR for now.

CiprianRegus:
I'll close this PR for now due to inactivity

PR #2159, opened 2024-04-23
drivers: imu: adis: Rework burst read API

rbolboac:
## Pull Request Description
Rework burst read API such that it can be used for devices which have 32-bit temperature data and 32-bit data counter. Furthermore, some devices support burst reading with 16-bit LSB first, while other support burst reading with 16-bit MSB first in burst data, thus memcpy cannot be used anymore to simply copy the whole burst data, instead, a burst data structure has been created which should be populated accordingly by each driver. The iio_adis generic implementation now populates the iio buffer using this structure.
Rework the unit tests accordingly.

## PR Type

## PR Checklist


drivers/imu/adis.c
```diff
@@ -2532,29 +2528,49 @@ int adis_read_burst_data(struct adis_dev *adis, uint8_t buff_size,
...
+	uint8_t axis_data_size = 12;
+	if (burst32)
+		axis_data_size = 24;
+
+	uint8_t axis_data_offset = ADIS_READ_BURST_DATA_CMD_SIZE + 2;
+	uint8_t temp_offset = axis_data_offset + axis_data_size;
+	uint8_t data_cntr_offset = temp_offset + 2;
+
+	if (burst32) {
+		memcpy(&data->x_gyro_lsb, &buffer[axis_data_offset], 2);
```
CiprianRegus, line 2540: Can't we just have 32 bit variables for each of these values and just arrange the two 16 bit parts? It would simplify things a bit if the `adis_burst_data` struct is meant to be used directly in an application.

rbolboac: I would say no. if the user is interested only in 16 bit burst reads, then he should only be concerned about the msb part and only handle those values. furthermore, there are devices which support only 16-bit burst reads so this structure would be useful if support is needed in the future.

CiprianRegus (approved):

PR #2177, opened 2024-05-08
Ad7091r updates

SaikiranGudla:
## Pull Request Description

Add updates to the driver & Fix set_port api

    Add sleep mode part of the init_param, device structures.
    Update init function to configure sleep mode.
    Update set_sleep_mode API.
    Add spi_write_mask function declaration to the header file.
    Fix ad7091r8_set_port api

## PR Type

## PR Checklist

### DIFF
--- a/drivers/adc/ad7091r8/ad7091r8.c
+++ b/drivers/adc/ad7091r8/ad7091r8.c
@@ -200,11 +200,11 @@ int ad7091r8_set_port(struct ad7091r8_dev *dev, enum ad7091r8_port port,
 	switch (port) {
 	case AD7091R8_GPO0:
 		mask = REG_CONF_GPO0_MASK;
-		val = no_os_field_get(REG_CONF_GPO0_MASK, value);
+		val = no_os_field_prep(REG_CONF_GPO0_MASK, value);
 		break;
 	case AD7091R8_GPO1:
 		mask = REG_CONF_GPO1_MASK;
-		val = no_os_field_get(REG_CONF_GPO1_MASK, value);
+		val = no_os_field_prep(REG_CONF_GPO1_MASK, value);
 		break;
 	default:
 		return -EINVAL;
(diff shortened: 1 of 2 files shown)


drivers/adc/ad7091r8/ad7091r8.c
```diff
@@ -420,7 +431,11 @@ int ad7091r8_init(struct ad7091r8_dev **device,
 			goto err_release_reset;
 	}
 
-	/* Device powers-up in normal mode */
+	/* Set device sleep-mode */
+	ret = ad7091r8_set_sleep_mode(dev, init_param->sleep_mode);
+	if (ret)
+		goto err_release_spi;
```
rbolboac, line 437: I think you should `goto err_release_reset` instead to avoid memory leaks.

drivers/adc/ad7091r8/ad7091r8.c
```diff
@@ -420,7 +431,11 @@ int ad7091r8_init(struct ad7091r8_dev **device,
 			goto err_release_reset;
 	}
 
-	/* Device powers-up in normal mode */
+	/* Set device sleep-mode */
+	ret = ad7091r8_set_sleep_mode(dev, init_param->sleep_mode);
```
rbolboac, line 435:
```
dev->vref_mv = init_param->vref_mv;
	if (!dev->vref_mv) {
		dev->vref_mv = 2500;
		ret = ad7091r8_set_sleep_mode(dev, AD7091R8_SLEEP_MODE_1);
		if (ret)
			goto err_release_reset;
	}
```
the code above: will be overwritten by this line, is this the intended behavior?

rbolboac (commented):
also, it would be nice to have a fixes tag in the first commit.

drivers/adc/ad7091r8/ad7091r8.c
SaikiranGudla: Rearranged the set_sleep_mode function call above this so that user can set sleep mode / device mode using the init parameters and if vref is not set, the device will go to sleep mode 1.

SaikiranGudla:
Changelog: 
Addressed review comments
Added fixes tag to the fix commit

rbolboac (approved):

amiclaus (approved):

PR #2180, opened 2024-05-14
MQTT example for SWIOT1L and a couple fixes

CiprianRegus:
## Pull Request Description

- Adding an MQTT example for the SWIOT1L project. This is an alternative to the current default firmware.
- The UART parity error flag was sometimes randomly set for the Maxim platform (probably due to noise on the UART bus). Since it was not cleared in either no-OS or the msdk, the interrupt would continuously trigger.

## PR Type

## PR Checklist


CiprianRegus:
I'll have to update the documentation as well

network/lwip_raw_socket/lwip_socket.c
```diff
@@ -808,27 +808,70 @@ static int32_t lwip_socket_recvfrom(void *net, uint32_t sock_id, void *data,
...
- * @param addr - Not used.
+ * @brief Called once a connect() completes.
+ * @param net - lwip sockets layer specific descriptor.
+ * @param pcb - unused.
+ * @param err - connection result.
+ * @return -ENOSYS
+ */
+static err_t lwip_connect_callback(void *arg, struct tcp_pcb *pcb, err_t err)
+{
+	struct lwip_socket_desc *sock = arg;
```
rbolboac, line 819: seeing how these APIs are static, I am not sure if we should check for null pointers...I suppose the upper layer is taking care of it?

network/lwip_raw_socket/lwip_socket.c
```diff
@@ -808,27 +808,70 @@ static int32_t lwip_socket_recvfrom(void *net, uint32_t sock_id, void *data,
...
+
+	socket = _get_sock(desc, sock_id);
+	if (!socket)
+		return -ENOMEM;
+
+	pcb = socket->pcb;
+
+	ret = tcp_connect(pcb, &ipaddr, addr->port, lwip_connect_callback);
+	if (ret)
+		return ret;
```
rbolboac, line 861: can return directly

projects/swiot1l/src.mk
```diff
@@ -68,26 +65,51 @@ SRCS += $(DRIVERS)/digital-io/max149x6/max149x6-base.c	\
...
+INCS += $(DRIVERS)/adc-dac/ad74413r/iio_ad74413r.h
+SRCS += $(DRIVERS)/adc-dac/ad74413r/iio_ad74413r.c
+INCS += $(DRIVERS)/digital-io/max149x6/iio_max14906.h
+SRCS += $(DRIVERS)/digital-io/max149x6/iio_max14906.c
+INCS += $(DRIVERS)/temperature/adt75/iio_adt75.h
+SRCS += $(DRIVERS)/temperature/adt75/iio_adt75.c
+
+endif
+
+
```
rbolboac, line 97: remove extra blank line

projects/swiot1l/src/examples/swiot1l-mqtt/swiot1l_mqtt.c
```diff
@@ -0,0 +1,282 @@
...
+	int ret;
+
+	struct ad74413r_desc *ad74413r;
+	struct lwip_network_param lwip_ip = {
+		.platform_ops = &adin1110_lwip_ops,
+		.mac_param = &adin1110_ip,
+	};
+	struct lwip_network_desc *lwip_desc;
+	struct tcp_socket_desc *tcp_socket;
+	// struct connection *client_socket;
```
rbolboac, line 81: remove commented line

projects/swiot1l/src/examples/swiot1l-mqtt/swiot1l_mqtt.c
```diff
@@ -0,0 +1,282 @@
...
+		return ret;
+
+	struct tcp_socket_init_param tcp_ip = {
+		.net = &lwip_desc->no_os_net,
+		.max_buff_size = 0
+	};
+
+	ret = socket_init(&tcp_socket, &tcp_ip);
+	if (ret)
+		return ret;
```
rbolboac, line 175: any memory leaks we should handle?
also, in case of returning, especially if !0, maybe it would be a good idea to print some error messages which would explain what has happened?

rbolboac (commented):
some remarks from my side, also it looks like the build is failing because the max UART flags are not found

network/lwip_raw_socket/lwip_socket.c
CiprianRegus: The upper layer is not checking `arg`. I've added the check in the `lwip_connect_function`, however this pointer is shared between multiple callbacks (`lwip_accept_callback()`, `lwip_recv_callback()`) and an invalid value will result in a fatal error (meaning the lwip stack won't be able to recover from this and any action such as connect, accept, packet RX, etc will be broken from that point). So, it's not much different that a null pointer access.

CiprianRegus:
Addressed all the comments from @rbolboac.

rbolboac (approved):
lgtm

Issue #2181, opened 2024-05-14
Maxim, STM32 makefiles overwrite VS Code configuration files


trupples:
## Behaviour

The `$(PROJECT_TARGET)_configure` targets in [tools/scripts/stm32.mk](https://github.com/analogdevicesinc/no-OS/blob/main/tools/scripts/stm32.mk) and [tools/scripts/maxim.mk](https://github.com/analogdevicesinc/no-OS/blob/main/tools/scripts/maxim.mk) use the [GNU Make `file` function](https://www.gnu.org/software/make/manual/html_node/File-Function.html) to set up a default VS Code configuration for projects, with no regard to the contents of those files.

https://github.com/analogdevicesinc/no-OS/blob/3cb5f58f6815c4fa63b732bb5072d42483ad6f81/tools/scripts/maxim.mk#L139-L143

For context, the content of the files is generated in [vsc_intellisense.mk](https://github.com/analogdevicesinc/no-OS/blob/main/tools/scripts/vsc_intellisense.mk), [vsc_openocd_debug.mk](https://github.com/analogdevicesinc/no-OS/blob/main/tools/scripts/vsc_openocd_debug.mk), and [vsc_tasks.mk](https://github.com/analogdevicesinc/no-OS/blob/main/tools/scripts/vsc_tasks.mk).

## Impact

With the current makefiles, one may not change their VS Code config without it being reset to the no-OS defaults during the next build, which greatly hinders development. For instance, my WSL dev setup requires changing the gdb options in `launch.json`, but doing so is futile because the makefile resets my changes when building.

## Options

I suggest (and will shortly open a PR) adding a condition to only write to these files if they are empty or don't exist (i.e., during project setup), but keep their current content otherwise, to allow for customization of the config.

Despite some effort, I didn't manage to do it using just GNU Make syntax, but rather by changing the `file` call from `filename.json` to `filename.json.default`, and having some shell commands conditionally copy `filename.json.default` to `filename.json` if the latter doesn't exist. Finally, shell commands remove the `.default` files to keep the directory tidy.

CiprianRegus:
The commits were merged.

PR #2182, opened 2024-05-14
Don't overwrite .vscode config if it already exists

trupples:
## Pull Request Description

(Referencing issue #2181, the following content is copied from there)

### Behaviour

The `$(PROJECT_TARGET)_configure` targets in [tools/scripts/stm32.mk](https://github.com/analogdevicesinc/no-OS/blob/main/tools/scripts/stm32.mk) and [tools/scripts/maxim.mk](https://github.com/analogdevicesinc/no-OS/blob/main/tools/scripts/maxim.mk) use the [GNU Make `file` function](https://www.gnu.org/software/make/manual/html_node/File-Function.html) to set up a default VS Code configuration for projects, with no regard to the contents of those files.

https://github.com/analogdevicesinc/no-OS/blob/3cb5f58f6815c4fa63b732bb5072d42483ad6f81/tools/scripts/maxim.mk#L139-L143

For context, the content of the files is generated in [vsc_intellisense.mk](https://github.com/analogdevicesinc/no-OS/blob/main/tools/scripts/vsc_intellisense.mk), [vsc_openocd_debug.mk](https://github.com/analogdevicesinc/no-OS/blob/main/tools/scripts/vsc_openocd_debug.mk), and [vsc_tasks.mk](https://github.com/analogdevicesinc/no-OS/blob/main/tools/scripts/vsc_tasks.mk).

### Impact

With the current makefiles, one may not change their VS Code config without it being reset to the no-OS defaults during the next build, which greatly hinders development. For instance, my WSL dev setup requires changing the gdb options in `launch.json`, but doing so is futile because the makefile resets my changes when building.

### Options

I suggest (this PR) adding a condition to only write to these files if they are empty or don't exist (i.e., during project setup), but keep their current content otherwise, to allow for customization of the config.

Despite some effort, I didn't manage to do it using just GNU Make syntax, but rather by changing the `file` call from `filename.json` to `filename.json.default`, and having some shell commands conditionally copy `filename.json.default` to `filename.json` if the latter doesn't exist. Finally, shell commands remove the `.default` files to keep the directory tidy.

## PR Type

## PR Checklist

### DIFF
--- a/tools/scripts/maxim.mk
+++ b/tools/scripts/maxim.mk
@@ -137,10 +137,17 @@ $(PLATFORM)_project:
 	$(MAKE) --no-print-directory $(PROJECT_TARGET)_configure
 
 $(PROJECT_TARGET)_configure:
-	$(file > $(CPP_PROP_JSON),$(CPP_FINAL_CONTENT))
-	$(file > $(SETTINGSJSON),$(VSC_SET_CONTENT))
-	$(file > $(LAUNCHJSON),$(VSC_LAUNCH_CONTENT))
-	$(file > $(TASKSJSON),$(VSC_TASKS_CONTENT))
+	$(file > $(CPP_PROP_JSON).default,$(CPP_FINAL_CONTENT))
+	$(file > $(SETTINGSJSON).default,$(VSC_SET_CONTENT))
+	$(file > $(LAUNCHJSON).default,$(VSC_LAUNCH_CONTENT))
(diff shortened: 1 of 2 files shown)


rbolboac:
@EoghanMonahan can you please take a look at this since you might be familiar with the topic?

buha (approved):

dbogdan (approved):

PR #2183, opened 2024-05-14
drivers: adc: ad7124: Updating the Device IDs

D-Disha:
Removed comment to change device ID by user for new silicon and added macros for the same. Also updated the device ID checks in ad7124_setup function.

## Pull Request Description

## PR Type

## PR Checklist

### DIFF
--- a/drivers/adc/ad7124/ad7124.c
+++ b/drivers/adc/ad7124/ad7124.c
@@ -1037,12 +1037,29 @@ int32_t ad7124_setup(struct ad7124_dev **device,
 	ret = ad7124_read_register(dev, &dev->regs[AD7124_ID_REG]);
 	if (ret)
 		goto error_spi;
+
 	if (dev->active_device == ID_AD7124_4) {
-		if (dev->regs[AD7124_ID_REG].value != AD7124_4_ID)
+		switch (dev->regs[AD7124_ID_REG].value) {
+		case AD7124_4_STD_ID:
+		case AD7124_4_B_GRADE_ID:
+		case AD7124_4_NEW_ID:
+			break;
+
+		default:
 			goto error_spi;
-	} else if (dev->active_device == ID_AD7124_8) {
-		if (dev->regs[AD7124_ID_REG].value != AD7124_8_ID)
+		}
+	}
+
+	else if (dev->active_device == ID_AD7124_8) {
+		switch (dev->regs[AD7124_ID_REG].value) {
+		case AD7124_8_STD_ID:
+		case AD7124_8_B_W_GRADE_ID:
+		case AD7124_8_NEW_ID:
+			break;
+
+		default:
 			goto error_spi;
+		}
 	}
 
 	for (setup_index = 0; setup_index < AD7124_MAX_SETUPS; setup_index++) {
--- a/drivers/adc/ad7124/ad7124.h
+++ b/drivers/adc/ad7124/ad7124.h
@@ -60,11 +60,20 @@
 #define AD7124_MAX_SETUPS	8
 /* Maximum number of channels */
 #define AD7124_MAX_CHANNELS	16
-/* Device IDs
- * The Device IDs for REV-B of boards need to be updated as 0x07 for ad7124-4
- * and 0x17 for ad7124-8 */
-#define AD7124_4_ID         0x14
-#define AD7124_8_ID	        0x16
+
+/* AD7124-4 Standard Device ID */
+#define AD7124_4_STD_ID  0x04
+/* AD7124-4 B Grade Device ID */
+#define AD7124_4_B_GRADE_ID  0x06
+/* Device ID for the re-designed die in the AD7124-4 standard part and B-grade */
+#define AD7124_4_NEW_ID  0x07
+
+/* AD7124-8 Standard Device ID */
+#define AD7124_8_STD_ID  0x14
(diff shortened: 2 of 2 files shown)


drivers/adc/ad7124/ad7124.h
```diff
@@ -60,11 +60,17 @@
...
-/* Device IDs
- * The Device IDs for REV-B of boards need to be updated as 0x07 for ad7124-4
- * and 0x17 for ad7124-8 */
-#define AD7124_4_ID         0x14
-#define AD7124_8_ID	        0x16
+
+/* AD7124-4 Device IDs */
+#define AD7124_4_ID      0x06
+/* Chip ID for new silicon in AD7124-4 */
+#define AD7124_4_NEW_ID  0x07
```
rbolboac, line 67: ID and NEW_ID naming isn't all that suggestive... what should the user choose? based on what?
I would prefer to have a suggestive naming, related maybe to hardware revision?

drivers/adc/ad7124/ad7124.h
```diff
@@ -60,11 +60,17 @@
...
+
+/* AD7124-8 Device IDs */
+#define AD7124_8_ID1     0x14
+#define AD7124_8_ID2     0x16
```
rbolboac, line 71: what is ID1 and what is ID2 and how can the user differentiate between these? again, the naming should be more suggestive and the comment as well

drivers/adc/ad7124/ad7124.c
```diff
@@ -1037,11 +1038,14 @@ int32_t ad7124_setup(struct ad7124_dev **device,
...
+
+	id_val = dev->regs[AD7124_ID_REG].value;
 	if (dev->active_device == ID_AD7124_4) {
-		if (dev->regs[AD7124_ID_REG].value != AD7124_4_ID)
+		if (id_val != AD7124_4_ID && id_val != AD7124_4_NEW_ID)
 			goto error_spi;
 	} else if (dev->active_device == ID_AD7124_8) {
-		if (dev->regs[AD7124_ID_REG].value != AD7124_8_ID)
+		if (id_val != AD7124_8_ID1 && id_val != AD7124_8_ID2
+		    && id_val != AD7124_8_NEW_ID)
```
rbolboac, line 1048: I think a switch case would fit better here

drivers/adc/ad7124/ad7124.c
```diff
@@ -1037,11 +1038,14 @@ int32_t ad7124_setup(struct ad7124_dev **device,
 	ret = ad7124_read_register(dev, &dev->regs[AD7124_ID_REG]);
 	if (ret)
 		goto error_spi;
+
+	id_val = dev->regs[AD7124_ID_REG].value;
 	if (dev->active_device == ID_AD7124_4) {
-		if (dev->regs[AD7124_ID_REG].value != AD7124_4_ID)
+		if (id_val != AD7124_4_ID && id_val != AD7124_4_NEW_ID)
```
rbolboac, line 1044: a switch case might fit better here, especially if there will be other 'new ids' in the future or other supported devices by the same driver

rbolboac (commented):
some comment from my side

drivers/adc/ad7124/ad7124.h
D-Disha: The ID can be either 
Screenshot 2024-05-15 101849

The user per say is not choosing any ID, the intention is that if he uses a board with new silicon or the old silicon , the chip id read check should not fail

drivers/adc/ad7124/ad7124.h
D-Disha: The user per say is not choosing any ID, the intention is that if he uses a board with new silicon or the old silicon , the chip id read check should not fail.
In terms of the boards. It is only the silicon that has been revised. They will only replace the old AD7124 IC on the board with the new silicon before releasing it.  I am not sure if they will call the newer boards with a new revision number or just replace the rev B boards with new silicon and still call it rev B boards. I will confirm this.

drivers/adc/ad7124/ad7124.h
rbolboac: I was referring to the 1 and 2 part. if the 1 is related to the old silicon and the 2 is related to the new silicon maybe there is a way to differentiate between them, like  AD7124_8_ID_REV_0 and AD7124_8_ID_REV_A.
But I see there is also  AD7124_8_NEW_ID. All of these values are compared against the read `id_val` but it is unclear to me what they actually represent (from chip version point of view).

D-Disha: @rbolboac I found this information on ezone https://ez.analog.com/data_converters/precision_adcs/w/documents/15293/ad7124-4-8-standard-silicon-b-grade-and-w-grade
I would rename the ID1 and ID2 as standard and B/W_GRADE respectively

drivers/adc/ad7124/ad7124.h
D-Disha: I have checked with the Apps Engineer, the new silicon that is redesigned has no name as such and we can keep it as new_id

rbolboac (approved):

D-Disha:
Hi all, are there any more comments on this PR or does it look good to merge?

amiclaus (approved):

PR #2190, opened 2024-05-22
stm32 usb serial (and a bugfix)

buha:
## Pull Request Description

Add posibility to use USB CDC ACM instead of a normal UART for stm32 projects.

## PR Type

## PR Checklist


drivers/platform/stm32/stm32_usb_uart.c
```diff
@@ -0,0 +1,247 @@
...
+#include "usbd_cdc.h"
+
+static uint8_t UserRxBufferFS[2048];
+static uint8_t UserTxBufferFS[2048];
+extern USBD_HandleTypeDef hUsbDeviceFS;
+
+volatile int tx_pending = 0;
+
+struct lf256fifo
+	*gfifo; // TODO: remove this, I could not find a better way to do it due the way USBD_Interface function ops are designed.
```
rbolboac, line 54: I would remove the todo and simply leave the explanatory comment in place, seeing how this is the final implementation

drivers/platform/stm32/stm32_usb_uart.c
```diff
@@ -0,0 +1,247 @@
...
+	return USBD_OK;
+}
+
+void stm32_on_usb_cdc_acm_rx(uint8_t* buf, uint32_t len)
+{
+	unsigned int i = 0;
+	int ret;
+
+	while(i < len) {
+		ret = lf256fifo_write(gfifo, buf[i]); // TODO: replace this gfifo global.
```
rbolboac, line 94: remove todo

drivers/platform/stm32/stm32_usb_uart.c
```diff
@@ -0,0 +1,247 @@
...
+		ret = -ENOMEM;
+		goto err_sdesc;
+	}
+	descriptor->extra = sdesc;
+
+	ret = lf256fifo_init(&sdesc->fifo);
+	if (ret)
+		goto err_fifo;
+
+	gfifo = sdesc->fifo; // TODO: remove gfifo.
```
rbolboac, line 162: remove todo

rbolboac (commented):
only some small remarks from my side

drivers/platform/stm32/stm32_usb_uart.c
CiprianRegus: In my opinion it's fine to have these TODO comments, since we never have a "final" implementation.

drivers/platform/stm32/stm32_usb_uart.c
```diff
@@ -0,0 +1,247 @@
+/***************************************************************************//**
+ *   @file   stm32/stm32_uart.c
```
CiprianRegus, line 2: `stm32_usb_uart.c`

buha:
V2, implements all the changes proposed above by @rbolboac

drivers/platform/stm32/stm32_usb_uart.c
```diff
@@ -0,0 +1,252 @@
...
+*******************************************************************************/
+#include <errno.h>
+#include <stdlib.h>
+#include "no_os_util.h"
+#include "stm32_usb_uart.h"
+#include "no_os_lf256fifo.h"
+#include "no_os_alloc.h"
+#include "usbd_cdc.h"
+
+static uint8_t UserRxBufferFS[2048];
```
CiprianRegus, line 47: The USB driver can work with smaller buffers right? It would be nice to have some sort of configuration option in order to allow the user to set their size. Forcing everyone to use 4kB of RAM even though they might not require the full bandwidth of the USB connection is a bit too much in my opinion.

drivers/platform/stm32/stm32_usb_uart.c
```diff
@@ -0,0 +1,252 @@
...
+
+volatile int tx_pending = 0;
+
+/* This global FIFO is needed because of how the STM32 CDC API is implemented.
+ * The function ops do not allow passing context parameters and so it is 
+ * impossible to pass this fifo to CDC_Receive_FS() in any other way.
+ * The implication is that only 1 USB CDC ACM interface may be used at once
+ * but this is an acceptable compromise, since having more than one doesn't
+ * make much sense anyway. */
+struct lf256fifo *gfifo;
```
CiprianRegus, line 59: `static`

drivers/platform/stm32/stm32_usb_uart.c
```diff
@@ -0,0 +1,252 @@
...
+	CDC_Init_FS,
+	CDC_DeInit_FS,
+	CDC_Control_FS,
+	CDC_Receive_FS,
+	CDC_TransmitCplt_FS
+};
+
+static int8_t CDC_Init_FS(void)
+{
+	USBD_CDC_SetTxBuffer(&hUsbDeviceFS, UserTxBufferFS, 0);
```
CiprianRegus, line 77: I think we should use snake case for all the functions and variables which are not defined in SDKs.

drivers/platform/stm32/stm32_usb_uart.c
```diff
@@ -0,0 +1,252 @@
...
+ * @return Positive number of transmitted bytes in case of success, negative error code otherwise.
+ */
+static int32_t stm32_usb_uart_write(struct no_os_uart_desc *desc,
+				    const uint8_t *data,
+				    uint32_t bytes_number)
+{
+	unsigned int len = no_os_min(bytes_number, 2048);
+	while(tx_pending)
+		;
+	memcpy(UserTxBufferFS, data, len);
```
CiprianRegus, line 209: Since the user provided `data` is const, I think we can avoid the memcpy() and use that buffer directly.

projects/iio_demo/builds.json
```diff
@@ -20,10 +20,13 @@
   },
   "stm32": {
     "iio": {
-      "flags": "IIO_EXAMPLE=y IIO_SW_TRIGGER_EXAMPLE=n IIO_TIMER_TRIGGER_EXAMPLE=n"
+      "flags": "IIO_EXAMPLE=y IIO_SW_TRIGGER_EXAMPLE=n IIO_TIMER_TRIGGER_EXAMPLE=n HARDWARE=sdp-ck1z.ioc"
+    },
+    "iio_usb_cdc_acm": {
```
CiprianRegus, line 25: Also, we should use `RELEASE=y` for CI builds, since they won't be used for debugging. `-O2` might also reveal the use of undefined behavior.

drivers/platform/stm32/stm32_usb_uart.c
buha: The transmit part uses DMA and is nonblocking so I use a combination of a memcpy into separate buffer and the tx_complete flag to make sure the buffer is available for the DMA until TX is complete. In other words, if I don't use the memcpy, behavior is not ok, I see mixed up content if there are subsequent calls to the Tx function.

This is the only comment from your review that I won't implement, the rest I'm implementing right now.

projects/iio_demo/builds.json
buha: You are right.

I can add it, but I just figured, shouldn't the CI build all projects with RELEASE=y regardless of what's in the builds.json ? To me it would make sense.

drivers/platform/stm32/stm32_usb_uart.c
buha: good point, changed defaults to 512 and added a way for user to change this

buha:
V4:
- made the transmit function blocking, to keep in line with the no-OS API expected behavior

buha:
V5:
- added and tested RELEASE=y for the new entry in builds.json

CiprianRegus (approved):

rbolboac (approved):

PR #2192, opened 2024-05-27
ad469x: project update and multiple variant support

ahaslam2:
## Pull Request Description

This PR updates the ad469x project to the new multi platform structure while
adding support for parts with different number of channels.

## PR Type
- [ ] Bug fix (change that fixes an issue)
- [x] New feature (change that adds new functionality)
- [ ] Breaking change (has dependencies in other repos or will cause CI to fail)

## PR Checklist
- [x] I have followed the [Coding style guidelines](http://analogdevicesinc.github.io/no-OS/drivers_guide.html#coding-style)
- [x] I have performed a self-review of the changes
- [x] I have commented my code, at least hard-to-understand parts
- [x] I have build all projects affected by the changes in this PR
- [ ] I have tested in hardware affected projects, at the relevant boards
- [x] I have signed off all commits from this PR
- [x] I have updated the documentation (wiki pages, ReadMe etc), if applies

PR #2199, opened 2024-05-30
projects: ad463x: Migrate to new project structure & stm32 support 

ahaslam2:
## Pull Request Description

Migrate to the new project structrure and add support for stm32.
STM32 uses standard spi with one line. so we need to set the 4630
in interleaved mode and de-interleave the bits. 

## PR Type

## PR Checklist


amiclaus (commented):
there are some build issues that need addressing in the first place.

ahaslam2:
yes.. this is just a draft push.. 
perhaps you were referring to my other PR https://github.com/analogdevicesinc/no-OS/pull/2192?
where CI seems to be blocked

buha (dismissed):

buha:
is this still wip ?

ahaslam2:
@buha  @amiclaus 
ready for review now.

projects/ad463x_fmcz/src/examples/basic_example/basic_example.c
```diff
@@ -0,0 +1,93 @@
...
+				    dev->real_bits_precision, &pgia_gain_idx);
+	if (ret)
+		goto out;
+	/** set pgia gain */
+	ret = ad463x_set_pgia_gain(dev, pgia_gain_idx);
+	if (ret)
+		goto out;
+#endif
+
+	ret = ad463x_read_data(dev, buf, SAMPLES_PER_CHANNEL);
```
jonathanns, line 82: I am also working with this device (for Xilinx and spi-engine), so I used your branch to test my changes and so far the only thing i see missing here is the `ad463x_exit_reg_cfg_mode` before reading the data. If you don't do that, the device does not enter the streaming mode. It should be like this:

```
        /* Exit register configuration mode */
	ret = ad463x_exit_reg_cfg_mode(dev);
	if (ret)
		goto out;

	ret = ad463x_read_data(dev, buf, SAMPLES_PER_CHANNEL);
	if (ret)
		goto out;
```

ahaslam2: yes you are right. 
in fact its is mentioned on the datasheet that conversion mode is on by default
and i had a patch that would enter cfg mode  it on init, (not sure why its not done)
and enable conversion mode before exiting init. but i doped that patch and forgot to put this back
on the basic example. (iio driver part takes is doing the exit cfg mode).. will double check this.

ahaslam2: now that im testing again, i remember why i drop that patch, seems HDL plays some role
and while it works on "standard spi" when adding HDL for xilinx, there is some issue..
ill find out what is going on.

ahaslam2: after some debuging seems there currently  a buffer overflow on the iio driver when using offload.
that needs to be fixed, then i can add my patch.

ahaslam2: Ok, i was still working on it too,  i did not quite figure %100 why the read raw is timing out .
 i was able to get better results by increasing the buffer size from 2 to 4, and perhaps adding
the "aligned" property to the buffer. 
but then i got it to hang randomly with a timeout sometimes too. i thought there was some corruption
going on past the buffer, but with some experiments to print values past the buffer,  i was not able to
"see" any corruption.

ahaslam2: @jonathanns  i will just add exit_cfg mode after the init to get this merged.

(i was trying to exit cfg mode on init, and  remove the enter/exit cfg calls in iio driver)
but its not working for iio+xilinx not sure why, (hags on raw read)
perhaps the chip is entering cfg mode somehow after the init, not sure were/how. 

so ill drop this idea for now.

ahaslam2:
Changelog:

basic example:
- exit cfg mode 
- sign extention on print 
- init uart console

projects/ad463x_fmcz/src/examples/basic_example/basic_example.c
jonathanns: > @jonathanns i will just add exit_cfg mode after the init to get this merged.
> 
> (i was trying to exit cfg mode on init, and remove the enter/exit cfg calls in iio driver) but its not working for iio+xilinx not sure why, (hags on raw read) perhaps the chip is entering cfg mode somehow after the init, not sure were/how.
> 
> so ill drop this idea for now.

Looks like after every offload sequence the device enters the config mode again. So, before enabling the offload again, I had to call the exit cfg mode again. Not sure yet why this happens, but it is recommended to exit config mode before start getting data and enter config mode after that.
So far, this should be enough i will try your changes again.

ahaslam2: i think your patches sitll have size = 4. 

With buffer size 2 it *might* work sometimes.  It works better with size = 4,
but its still something going on here i dont fully understand. 

another thing that seems to work is to make move the buffer out of the stack 
and declare it as a "global". this is why i thought there was some corruption happening.

jonathanns: The last pacth has buffer size 2, actually. But is this "global" buffer only for the single read (raw attribute)? because for the buffer reading (readdev) works the way it is.
I guess i will try putting this buffer outside of the function stack.

ahaslam2: yes.. only for the raw attribute. i just did that as an experiment.
no need to do it if we dont understand why its needed.

ahaslam2:
A new full PR with these changes + @jonathanns  fixes + stm32 implementation here 

https://github.com/analogdevicesinc/no-OS/pull/2225

ahaslam2:
closing the PR as https://github.com/analogdevicesinc/no-OS/pull/2225 includes all this work.

PR #2215, opened 2024-06-17
rev_D functionality

raezt:
## Pull Request Description

Revision D firmware update.

## PR Type

## PR Checklist

### DIFF
--- a/projects/ad-acevsecrdset-sl/Makefile
+++ b/projects/ad-acevsecrdset-sl/Makefile
@@ -1,3 +1,11 @@
+REV_A = n
+REV_D = y
+ifeq (y,$(strip $(REV_A)))
+CFLAGS += -DREV_A
+else
+CFLAGS += -DREV_D
+endif
+
 include ../../tools/scripts/generic_variables.mk
 
 include src.mk
--- a/projects/ad-acevsecrdset-sl/src/inter/inter.c
+++ b/projects/ad-acevsecrdset-sl/src/inter/inter.c
@@ -1,0 +1,341 @@
+/***************************************************************************//**
+ *   @file   inter.c
+ *   @brief  Interrupt on port 2 implementation file.
+ *   @author GMois (george.mois@analog.com), REtz (radu.etz@analog.com)
+********************************************************************************
+ * Copyright (c) 2024 Analog Devices, Inc.
+ *
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *  - Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ *  - Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in
+ *    the documentation and/or other materials provided with the
+ *    distribution.
+ *  - Neither the name of Analog Devices, Inc. nor the names of its
+ *    contributors may be used to endorse or promote products derived
+ *    from this software without specific prior written permission.
+ *  - The use of this software may or may not infringe the patent rights
+ *    of one or more patent holders.  This license does not release you
+ *    from the requirement that you obtain separate licenses from these
+ *    patent holders to use this software.
+ *  - Use of the software either in source or binary form, must be run
+ *    on or directly connected to an Analog Devices Inc. component.
+ *
+ * THIS SOFTWARE IS PROVIDED BY ANALOG DEVICES "AS IS" AND ANY EXPRESS OR
+ * IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, NON-INFRINGEMENT,
+ * MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED.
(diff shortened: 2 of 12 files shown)


amiclaus (commented):
there are some build errors that need handling in the first place, prior to the review.
Also, what about REV_B, REV_C? Are those equivalent from the software point of view with REV_D? I'm asking because if that's the case, the software handling for the revisions might be simplified.

projects/ad-acevsecrdset-sl/src/inter/inter.c
```diff
@@ -0,0 +1,375 @@
...
+	}
+
+	ret = no_os_gpio_direction_input(pin3);
+	if (ret) {
+		printf("ERROR\n");
+		return ret;
+	}
+
+	/* Initialize GPIO IRQ controller in order to be able to enable GPIO IRQ interrupt */
+	struct no_os_irq_init_param irq_gpio_ip = {
```
amiclaus, line 269: move this above together with the rest of the structure initializations.

projects/ad-acevsecrdset-sl/src/inter/inter.c
```diff
@@ -0,0 +1,375 @@
...
+	ret = no_os_gpio_get(&pin, &gpio_ip);
+	if (ret) {
+		printf("ERROR\n");
+		return ret;
+	}
+
+	ret = no_os_gpio_direction_input(pin);
+	if (ret) {
+		printf("ERROR\n");
+		return ret;
```
amiclaus, line 241: where is the memory deallocated for the `pin` in case of error? same applies for the rest of the gpio/irq descriptors.

projects/ad-acevsecrdset-sl/src/inter/inter.c
```diff
@@ -0,0 +1,375 @@
...
+	}
+
+	ret = no_os_gpio_get(&pin2, &gpio_ip);
+	if (ret) {
+		printf("ERROR\n");
```
amiclaus, line 246: these code duplications can be optimized a bit with conditional jumps.

projects/ad-acevsecrdset-sl/src/self_test/self_test.c
```diff
@@ -441,6 +470,26 @@ static int self_test_pilot(struct stout *stout)
...
+{
+	pr_debug("TEST upstream PE. opto1 = %d, opto2 = %d \n",
+		 get_gpio_opto_out1_flag_state(), get_gpio_opto_out2_flag_state());
+	if (get_gpio_opto_out1_flag_state() ^ get_gpio_opto_out2_flag_state()) {
+		pr_debug("PE upstream presence: Passed\n");
+		return INTF_NO_ERR;
+	} else
+		return INTF_PE_UPSTREAM_ERR;
+
+	return 0;
```
amiclaus, line 489: is this ever reached?

projects/ad-acevsecrdset-sl/src/self_test/self_test.c
```diff
@@ -441,6 +470,26 @@ static int self_test_pilot(struct stout *stout)
 	return 0;
 }
 
+#if defined(REV_D)
+/**
+ * @brief Startup test for PE upstream presence
+ * @param stout - state  machine descriptor
+ * @return 0 in case of success, error code otherwise
+ */
+int self_test_PE_upstream(struct stout *stout)
```
amiclaus, line 479: not sure about the previous implementation of the project, but snake case convention is applied.

amiclaus (commented):
some comments on my side.

raezt:
v3
- renamed self_test_PE_upstream
- modified the self_test_PE_upstream return
- moved "struct no_os_irq_init_param irq_gpio_ip" where the other structures are initialized

raezt:
v4

- squash 2 commits (the second one was fixing the previous one)

projects/ad-acevsecrdset-sl/Makefile
```diff
@@ -1,7 +1,15 @@
+ifdef REV_A
+CFLAGS += -DREV_A=1
+else
+CFLAGS += -DREV_D=1
```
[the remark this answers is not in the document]
amiclaus: If there will be other Revisions in the future then we might want to keep both flags (REV_A and REV_D). Also drop the explicit initialization with 1 since it is redundant. To keep the approach similar to the rest of the projects, we should do something like this:
```
REV_A = n
REV_D = y
ifeq (y,$(strip $(REV_A)))
CFLAGS += -DREV_A
else
CFLAGS += -DREV_D
endif
```

raezt:
v5

- optimized the code with conditional jumps (interupt.c).
- removed the statement guards where there is only one expression.
- changed printf to pr_err.
- dropped the extra comment blocks.
- memory deallocated in case of error.

raezt:
v6

- squash 2 commits (the first one was implementing just a part of the added functionality)

raezt:
v7

- changed the makefile for revision

amiclaus (commented):
some more minor comments on my side. otherwise looks good.

raezt:
v8

- REV_A=true changed to REV_A=y in commit message.
- deleted line from eof in makefile.
- add Functions Definition header comment in inter.c.
- unrelated change in self_test.c resolved.

danmois (approved):

PR #2221, opened 2024-06-20
drivers: platform: maxim: pwm: Add Maxim PWM platform driver

RaduSabau1:
## Pull Request Description

This PR contains the commit adding the Maxim PWM driver.

 - Extra field from no_os_pwm_desc is used as the timer's prescaler used to calculate the ticks for both period and duty cycle.
 - The prescaler for both should be the same and is calculated depending on the duty cycle since it is always smaller than
    the period, therefor a good resolution of the PWM will be obtained.

## PR Type

## PR Checklist

### DIFF
--- a/drivers/platform/maxim/max32655/maxim_pwm.c
+++ b/drivers/platform/maxim/max32655/maxim_pwm.c
@@ -1,0 +1,321 @@
+/***************************************************************************//**
+ *   @file   maxim_pwm.c
+ *   @brief  Maxim platform specific implementation PWM API.
+ *   @author Dragos Bogdan (dragos.bogdan@analog.com)
+********************************************************************************
+ * Copyright 2024(c) Analog Devices, Inc.
+ *
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *  - Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ *  - Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in
+ *    the documentation and/or other materials provided with the
+ *    distribution.
+ *  - Neither the name of Analog Devices, Inc. nor the names of its
+ *    contributors may be used to endorse or promote products derived
+ *    from this software without specific prior written permission.
+ *  - The use of this software may or may not infringe the patent rights
+ *    of one or more patent holders.  This license does not release you
+ *    from the requirement that you obtain separate licenses from these
+ *    patent holders to use this software.
+ *  - Use of the software either in source or binary form, must be run
+ *    on or directly connected to an Analog Devices Inc. component.
+ *
+ * THIS SOFTWARE IS PROVIDED BY ANALOG DEVICES "AS IS" AND ANY EXPRESS OR
+ * IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, NON-INFRINGEMENT,
+ * MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED.
+ * IN NO EVENT SHALL ANALOG DEVICES BE LIABLE FOR ANY DIRECT, INDIRECT,
+ * INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, INTELLECTUAL PROPERTY RIGHTS, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+*******************************************************************************/
+
+#include "tmr.h"
+#include "no_os_alloc.h"
+#include "no_os_error.h"
+#include "no_os_util.h"
+#include "no_os_units.h"
+#include "maxim_pwm.h"
+
+/**
+ * @brief Convert the prescaler from a clock division value to
+ * an enum's value for the init function.
+ * @param div - clock division value.
+ * @param prescaler - enum value to be provided to the init function.
+ * @return 0 in case of success, -EINVAL otherwise.
+ */
+static int _get_prescaler(uint32_t div, mxc_tmr_pres_t *prescaler)
+{
+	int i;
+
+	i = 1;
+
+	while (div / MAX_PWM_GET_PRESCALER(i) > MAX_PWM_TMR_MAX_VAL)
+		i++;
+
+	*prescaler = MAX_PWM_PRESCALER_VAL(i);
+
+	return 0;
+}
+
+/**
+ * @brief Enable the PWM.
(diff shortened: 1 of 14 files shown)


RaduSabau1:
V2:

- Use mxc_tmr_cfg_t pointer variable as descriptor extra field and pass it in duty cycle and period setter/getter functions in order to use the prescaler.
- Dynamically allocate the memory for the cfg at init and also add error labels in case of failure to avoid memory leaks (same procedure goes for the remove function).

drivers/platform/maxim/common/maxim_pwm.c
```diff
@@ -0,0 +1,355 @@
...
+	uint32_t div;
+	int32_t ret;
+
+	descriptor = (struct no_os_pwm_desc *)no_os_calloc(1, sizeof(*descriptor));
+	if (!descriptor)
+		return -ENOMEM;
+
+	cfg = (mxc_tmr_cfg_t *)no_os_calloc(1, sizeof(*cfg));
+	if (!cfg) {
+		return -ENOMEM;
```
amiclaus, line 283: this return should be dropped to fix the cppcheck build. the return is handled already on next line with the conditional jump

drivers/platform/maxim/common/maxim_pwm.c
```diff
@@ -0,0 +1,355 @@
+/***************************************************************************//**
+ *   @file   maxim_pwm.c
+ *   @brief  Implementation of MAX32655 PWM driver.
```
dbogdan, line 3: Since this driver will be common to all supported Maxim MCUs, delete the mention of MAX32655.

drivers/platform/maxim/common/maxim_pwm.c
```diff
@@ -0,0 +1,355 @@
...
+		return -EINVAL;
+
+	no_os_free(desc->extra);
+	no_os_free(desc);
+
+	return 0;
+}
+
+/**
+* @brief MAX32655 specific PWM ops
```
dbogdan, line 344: MAX32655 -> maxim

drivers/platform/maxim/common/maxim_pwm.h
```diff
@@ -0,0 +1,58 @@
...
+#define MAX_PWM_PRESCALER_TRUE(n)	((n) - 1)
+#define MAX_PWM_PRESCALER_VAL(n)	((n) / 16)
+
+struct max_pwm_init_param {
+	mxc_gpio_vssel_t vssel;
+	uint32_t prescaler;
+};
+
+/**
+ * @brief MAX32655 specific PWM ops
```
dbogdan, line 54: MAX32655 -> maxim

drivers/platform/maxim/common/maxim_pwm.h
```diff
@@ -0,0 +1,58 @@
+/***************************************************************************//**
+ *   @file   maxim_pwm.h
+ *   @brief  Header file of MAX32655 PWM driver.
```
dbogdan, line 3: MAX32655 -> Maxim

drivers/platform/maxim/common/maxim_pwm.c
```diff
@@ -0,0 +1,355 @@
...
+		*prescaler = TMR_PRES_256;
+	else if (div < 1024)
+		*prescaler = TMR_PRES_512;
+	else if (div < 2048)
+		*prescaler = TMR_PRES_1024;
+	else if (div < 4096)
+		*prescaler = TMR_PRES_2048;
+	else
+		*prescaler = TMR_PRES_4096;
+
```
amiclaus, line 85: nit: extra line.

drivers/platform/maxim/common/maxim_pwm.c
```diff
@@ -0,0 +1,355 @@
...
+#include "maxim_pwm.h"
+
+/**
+ * @brief Convert the prescaler from a clock division value to
+ * an enum's value for the init function.
+ * @param div - clock division value (this should be a power of 2).
+ * @param prescaler - enum value to be provided to the init function.
+ * @return 0 in case of success, -EINVAL otherwise.
+ */
+static int _get_prescaler(uint32_t div, mxc_tmr_pres_t *prescaler)
```
amiclaus, line 54: this looks like the same as in the timer implementation. maybe we can somehow use that directly (just a thought)

drivers/platform/maxim/common/maxim_pwm.c
```diff
@@ -0,0 +1,355 @@
...
+
+	return 0;
+}
+
+/**
+ * @brief Enable the PWM.
+ * @param desc - PWM descriptor.
+ * @return 0 in case of success, negative error code otherwise.
+ */
+int32_t max_pwm_enable(struct no_os_pwm_desc *desc)
```
amiclaus, line 95: `int` type. check the entire code.

drivers/platform/maxim/common/maxim_pwm.c
```diff
@@ -0,0 +1,355 @@
...
+
+	tmr_cfg.mode = TMR_MODE_PWM;
+	tmr_cfg.bitMode = TMR_BIT_MODE_32;
+	tmr_cfg.clock = MXC_TMR_APB_CLK;
+	tmr_cfg.cmp_cnt = period_ticks;
+	tmr_cfg.pol = desc->polarity;
+
+	max_pwm_disable(desc);
+
+	ret = MXC_TMR_Init(MXC_TMR_GET_TMR(desc->id), &tmr_cfg, 1);
```
CiprianRegus, line 144: This function will overwrite the VDDIO level.

drivers/platform/maxim/common/maxim_pwm.c
RaduSabau1: I am guessing just assigning ret = -ENOMEM, so the correct error code is returned.

RaduSabau1:
V3

- Added max_pwm_extra struct and using to store pin VDDIO and timer configuration and use along PWM API.
- Changed get_prescaler static function in order to get the best possible resolution out of the timer therefore being able to have smaller duty_cycles.

drivers/platform/maxim/max32655/maxim_pwm.c
```diff
@@ -0,0 +1,322 @@
...
+ * @return 0 in case of success, -EINVAL otherwise.
+ */
+static int _get_prescaler(uint32_t div, mxc_tmr_pres_t *prescaler)
+{
+	int cnt, i;
+
+	i = MAX_PWM_TMR_RES;
+	cnt = 0;
+
+	while (div > NO_OS_BIT(i))
```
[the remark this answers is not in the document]
RaduSabau1: Indeed, "i" should be decremented as well.

drivers/platform/maxim/max32655/maxim_pwm.c
```diff
@@ -0,0 +1,322 @@
...
+
+	pwm_extra = desc->extra;
+
+	duty_ticks = MXC_TMR_GetPeriod(MXC_TMR_GET_TMR(desc->id),
+				       MXC_TMR_APB_CLK,
+				       MAX_PWM_PRESCALER_TRUE(pwm_extra->tmr_cfg.pres),
+				       NO_OS_DIV_ROUND_CLOSEST_ULL(NANO, duty_cycle_ns));
+
+	/* Set duty cycle to a minimum. */
+	if (!duty_ticks)
```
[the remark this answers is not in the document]
RaduSabau1: Thought that in case of the duty cycle that the user wants to set, a minimuk duty cycle should be set.

RaduSabau1:
V3:

- Added maxim_pwm files for MAX32650, MAX32660 and MAX32665
- Modified for each MAX uController to support all available timers.
- Modified prescaler calculation function in order to obtain the maximum count possible without risking an overflow.

RaduSabau1:
V4

- Fixed prescaler function.

RaduSabau1:
V5

- Change the way real duty cycle value is computed.

RaduSabau1:
V6

- When PWM is disabled it only stops the timer, and the timer comes to shut down at remove.

PR #2225, opened 2024-06-21
STM32: Standard SPI DMA fixes and ad463x example

ahaslam2:
## This is a working example of spi dma xfer for stm32 (standard spi)

This uses 3 pwm timers to do dma xfers using the stm32.

1 timer to toggle CNV (master)
1 timer to toggle CS (slave to CNV)
1 timer to  write to SPI data register and trigger (slave to CS) 

The xfer timer can be repeated to xfer multiple bytes. 

the ad630 that needs a 6 byte xfer of interleaved data per sample 
and spi at 20Mbps (could not get 40Mbps working with current hw setup)
and respecting 300ns between CNV and CS low 
the sample rate obtained is  ~300ksps.

This PR includes previous work to convert ad4630 to new project structure and add stm32 support
as well as a couple of fixes from @jonathanns  for ad4630 iio driver. IOW all known current work
for 4630 is in this PR. 

## PR Type

## PR Checklist


ahaslam2:
@jonathanns  FYI: Here is a branch with all the current ad463x work i know of so far.
it includes:

- stm32 and project migration
- iio fixes
- spi dma.

drivers/adc/ad463x/ad463x.h
```diff
@@ -43,9 +43,11 @@
 /***************************** Include Files **********************************/
 /******************************************************************************/
 #include <stdint.h>
-#include "spi_engine.h"
 #include "no_os_util.h"
+#if !defined(USE_STANDARD_SPI)
```
danmois, line 47: In general we have #ifndef in our code. Basically, it is the same, but maybe is more concise.

danmois (commented):
The commit message seems a bit strange for 6e3d05ecd.

drivers/platform/stm32/stm32_pwm.c
```diff
@@ -499,7 +502,6 @@ int32_t stm32_pwm_disable(struct no_os_pwm_desc *desc)
 
 	if (ret != HAL_OK)
 		return -EIO;
-
```
danmois, line 502: Unrelated change?

drivers/platform/stm32/stm32_spi.c
```diff
@@ -588,6 +587,51 @@ int32_t stm32_config_dma_and_start(struct no_os_spi_desc* desc,
...
+	CLEAR_BIT(sdesc->hspi.Instance->CR2, SPI_CR2_TXDMAEN);
+#endif
+
+#if defined (STM32H5)
+	CLEAR_BIT(sdesc->hspi.Instance->CFG1, SPI_CFG1_RXDMAEN);
+#else
+	CLEAR_BIT(sdesc->hspi.Instance->CR2, SPI_CR2_RXDMAEN);
+#endif
+
+	ret = no_os_dma_xfer_abort(sdesc->dma_desc, sdesc->txdma_ch);
```
danmois, line 625: you are not doing anything with ret here and it is overwritten.

drivers/platform/stm32/stm32_spi.c
```diff
@@ -633,7 +634,7 @@ void stm32_spi_dma_callback(struct no_os_dma_xfer_desc *old_xfer,
 }
 
 /**
- * @brief Configure and start a series of transfers using DMA. Wait for the
+ * @brief Configure and start a series of transfers using DMA. Dont Wait for the
```
danmois, line 644: Don't

drivers/platform/stm32/stm32_spi.c
```diff
@@ -656,6 +657,38 @@ int32_t stm32_spi_dma_transfer_async(struct no_os_spi_desc* desc,
...
+		no_os_mdelay(1);
+		if (sdesc->stm32_spi_dma_done)
+			break;
+	};
+
+	/* need some cleanup here? */
+	if (timeout == 0)
+		return -ETIME;
+
+
```
danmois, line 695: extra line

drivers/adc/ad463x/ad463x.c
```diff
@@ -370,11 +370,19 @@ static int32_t ad463x_init_gpio(struct ad463x_dev *dev,
 
 	no_os_mdelay(100);
 
+	ret = no_os_gpio_get_optional(&dev->gpio_cnv, init_param->gpio_cnv);
+	if (ret != 0)
+		return ret;
```
amiclaus, line 377: you want a conditional jump here too, i think.

drivers/adc/ad463x/ad463x.c
```diff
@@ -530,29 +563,68 @@ static int32_t ad463x_read_single_sample(struct ad463x_dev *dev,
...
+	if (!rx_buf)
+		return -ENOMEM;
+
+	spi_msg.tx_buff = &tx_buf,
+	spi_msg.bytes_number = samples * dev->read_bytes_no,
+	spi_msg.rx_buff = rx_buf,
+
+	ret = no_os_pwm_enable(dev->trigger_pwm_desc);
+	if (ret != 0)
+		return ret;
```
danmois, line 601: Should rx_buf be freed?

drivers/adc/ad463x/ad463x.c
```diff
@@ -530,29 +563,68 @@ static int32_t ad463x_read_single_sample(struct ad463x_dev *dev,
...
+
+	ret = no_os_spi_transfer_dma_sync(dev->spi_desc, &spi_msg, 1);
+	if (ret)
+		goto out;
+
+	ret = no_os_pwm_disable(dev->trigger_pwm_desc);
+	if (ret != 0)
+		return ret;
```
danmois, line 609: Same as above.

drivers/adc/ad463x/ad463x.c
```diff
@@ -446,7 +457,120 @@ int32_t ad463x_read_data(struct ad463x_dev *dev,
...
+	ret = no_os_spi_write_and_read(dev->spi_desc, data, dev->read_bytes_no);
+	if (ret)
+		return ret;
+
+	ad463x_pext(data[0], data[1], &ch0[0], &ch1[0]);
+	ad463x_pext(data[2], data[3], &ch0[1], &ch1[1]);
+	ad463x_pext(data[4], data[5], &ch0[2], &ch1[2]);
+	ad463x_pext(data[6], data[7], &ch0[3], &ch1[3]);
+
+	*ch0_out = ch0[3] | ch0[2] << 8 | ch0[1] << 16 | ch0[0] << 24;
```
amiclaus: `no_os_get_unaligned_be32`

drivers/platform/stm32/stm32_spi.c
```diff
@@ -227,6 +227,15 @@ int32_t stm32_spi_init(struct no_os_spi_desc **desc,
 		if (ret)
 			goto error_pwm;
 	}
+	if (sinit->tx_pwm_init) {
```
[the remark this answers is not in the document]
ahaslam2: Hi @CiprianRegus 

The general sampling procedure for an ADC is:
- toggle CNV pin
- CS goes low
- Read spi words.

When trying to read multiple samples in standard SPI
mode there is significant latency setting up each write,
then polling or waiting treating the IRQ to and xfer data
from the SPI data register. This, for each 8bit spi word.

This latency could be reduced if DMA is used to read
directly words from the spi data register without cpu intervention.

However, we cannot simply program the TX DMA and RX DMA for N number of
transfers because:

- CNV pin needs to be toggled and synchronized with SPI read.
- CS is not automatically set between words (TI mode disabled)
- even if CS is set to auto: SPI does 8bit words or 16 bit words (max)
but some chips need more bits to be read between CS toggles.

Hence some other method of synchronizing CNV, CS, and SPI data read
needs to be implemented.

The method here uses STM32 timers to achieve a coordinated DMA transfer:
with stm32 timmers we can:

- configure the timer in pwm mode.
- master>slave hierarchy.
- one shot mode with repeat capability
- dma operations to peripheral. 

The proposed solution uses 3 timers and spi rx dma:

CNV timer:
 - job is to trigger the CNV pin and start CS timer
 - master to CS timer (event sent on timmer start)
 - period is sampling rate
 - duty cycle depends on part, typically 10ns.

 
CS timer
 - job is to wait for conversion ready, toggle CS and start TX timer
 - slave to CNV timer
 - master to TX timer (event sent on duty cycle complete)
 - one shot mode (one spi read each CNV toggle)
 - one repetition (CS stays low during n-byte read)
 - period is sampling rate
 - duty cycle is max busy time between CNV and data ready.
 
 TX timer
  - job is to write to SPI data register to write single 8 bit word to SPI
  - slave to CS timer
  - DMA enabled mem to peripheral mode (write to SPI->DR)
  - one shot mode
  - repetition is the number of bytes to transfer.
  - period is speed spi xfers 8bits IOW depends on spi baudrate
  - duty cycle dont care. we just use the start event.
 
 SPI RX DMA
  - triggered by SPI hw after each TX timer wrote to SPI->DR and SPI xfer is done.
  - SPI DMA xfers data to memory from SPI->DR

drivers/adc/ad463x/ad463x.h
ahaslam2: ok, will change.

ahaslam2:
> The commit message seems a bit strange for [6e3d05e](https://github.com/analogdevicesinc/no-OS/commit/6e3d05ecde8f22fb04bf770aa4b9c5b062987997). Same for [263d47e](https://github.com/analogdevicesinc/no-OS/commit/263d47e3ddbb0a50011705dbebcf8eb0eeebcc21).
> 
I just cherry  picked those from @jonathanns 
> Also for commit messages containing the fix work, should also a Fixes tag be used?
ok, did not know fix tag was used for no-os.
what is the fixes tag used for in no-os? are there stable branches that they need
to be back ported to?

drivers/adc/ad463x/ad463x.c
```diff
@@ -434,19 +449,205 @@ int32_t ad463x_read_data(struct ad463x_dev *dev,
...
+	int shift;
+
+	memcpy(data, buf, size);
+	ad463x_pext(data[0], data[1], &ch0[0], &ch1[0]);
+	ad463x_pext(data[2], data[3], &ch0[1], &ch1[1]);
+	ad463x_pext(data[4], data[5], &ch0[2], &ch1[2]);
+	ad463x_pext(data[6], data[7], &ch0[3], &ch1[3]);
+
+	*ch0_out = ch0[3] | ch0[2] << 8 | ch0[1] << 16 | ch0[0] << 24;
+	*ch1_out = ch1[3] | ch1[2] << 8 | ch1[1] << 16 | ch1[0] << 24;
```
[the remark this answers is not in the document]
ahaslam2: i used no_os_get_unaligned_be32.

danmois:
> @danmois
> 
> > The commit message seems a bit strange for [6e3d05e](https://github.com/analogdevicesinc/no-OS/commit/6e3d05ecde8f22fb04bf770aa4b9c5b062987997). Same for [263d47e](https://github.com/analogdevicesinc/no-OS/commit/263d47e3ddbb0a50011705dbebcf8eb0eeebcc21).
> 
> I just cherry picked those from @jonathanns
> 
> > Also for commit messages containing the fix work, should also a Fixes tag be used?
> 
> ok, did not know fix tag was used for no-os. what is the fixes tag used for in no-os? are there stable branches that they need to be back ported to?

I was referring to a tag like the one following commit message: https://github.com/analogdevicesinc/no-OS/pull/2233/commits/af1fc13b010695ff04b6c4243450d96b584b4e3d

ahaslam2:
> > @danmois
> > > The commit message seems a bit strange for [6e3d05e](https://github.com/analogdevicesinc/no-OS/commit/6e3d05ecde8f22fb04bf770aa4b9c5b062987997). Same for [263d47e](https://github.com/analogdevicesinc/no-OS/commit/263d47e3ddbb0a50011705dbebcf8eb0eeebcc21).
> > 
> > 
> > I just cherry picked those from @jonathanns
> > > Also for commit messages containing the fix work, should also a Fixes tag be used?
> > 
> > 
> > ok, did not know fix tag was used for no-os. what is the fixes tag used for in no-os? are there stable branches that they need to be back ported to?
> 
> I was referring to a tag like the one following commit message: [af1fc13](https://github.com/analogdevicesinc/no-OS/commit/af1fc13b010695ff04b6c4243450d96b584b4e3d)

yes, i understood about the fix tag.
i was just wondering if there are stable branches being maintained and/or i need to do some back-porting to these branches (like in the linux kernel)?

danmois:
> > > @danmois
> > > > The commit message seems a bit strange for [6e3d05e](https://github.com/analogdevicesinc/no-OS/commit/6e3d05ecde8f22fb04bf770aa4b9c5b062987997). Same for [263d47e](https://github.com/analogdevicesinc/no-OS/commit/263d47e3ddbb0a50011705dbebcf8eb0eeebcc21).
> > > 
> > > 
> > > I just cherry picked those from @jonathanns
> > > > Also for commit messages containing the fix work, should also a Fixes tag be used?
> > > 
> > > 
> > > ok, did not know fix tag was used for no-os. what is the fixes tag used for in no-os? are there stable branches that they need to be back ported to?
> > 
> > 
> > I was referring to a tag like the one following commit message: [af1fc13](https://github.com/analogdevicesinc/no-OS/commit/af1fc13b010695ff04b6c4243450d96b584b4e3d)
> 
> yes, i understood about the fix tag. i was just wondering if there are stable branches being maintained and/or i need to do some back-porting to these branches (like in the linux kernel)?

No back-porting is required. The latest release (2022_r2) is currently being maintained.

ahaslam2:
Change log:
- change !defined to #ifndef
- drop patch to fix undersampling as this was fixes on spi_engine
- use uint8_t instead of unsigned char in pext function
- spelling fixes

drivers/adc/ad463x/ad463x.h
```diff
@@ -206,9 +206,9 @@ struct ad463x_init_param {
 	struct no_os_gpio_init_param *gpio_cnv;
 	struct no_os_gpio_init_param *gpio_pgia_a0;
 	struct no_os_gpio_init_param *gpio_pgia_a1;
-#ifndef USE_STANDARD_SPI
 	/** PWM */
 	struct no_os_pwm_init_param *trigger_pwm_init;
```
machschmitt: This sounds restricting. What if the platform doesn't have PWM?
I see that CNV (often triggered by a PWM) and CS must be synchronized for data read in CS zone 2.
But maybe better to not assume all steups will want to use a PWM to trigger CNV. It looks like one could use a GPIO to pulse CNV and CS would not need be that much synchronized with CNV (as long as the transfer is done after BUSY goes low (zone 1 capture)).
I'd check the init params to decide what to use for CNV instead of assuming a PWM is there.

projects/ad463x_fmcz/src/common/common_data.h
```diff
@@ -1,9 +1,9 @@
 /***************************************************************************//**
- *   @file   parameters.h
- *   @brief  Parameters Definitions.
- *   @author Antoniu Miclaus (antoniu.miclaus@analog.com)
```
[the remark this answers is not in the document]
ahaslam2: yes its a big change, but sure ill keep the original authors in these files.

projects/ad463x_fmcz/src/examples/basic_example/basic_example.c
```diff
@@ -0,0 +1,104 @@
...
+	/* Estimating a gain of 1/3, calculate the closest accepted pgia gain value */
+	int32_t gain_int = 0;
+	int32_t gain_fract = 1464843;
+	enum ad463x_pgia_gain pgia_gain_idx;
+	ret = ad463x_calc_pgia_gain(gain_int, gain_fract, dev->vref,
+				    dev->real_bits_precision, &pgia_gain_idx);
+	if (ret)
+		goto out;
+	/** set pgia gain */
+	ret = ad463x_set_pgia_gain(dev, pgia_gain_idx);
```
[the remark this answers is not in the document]
ahaslam2: this is actually the only part of code i brought in "as-is" from the old project file.

drivers/adc/ad463x/ad463x.h
ahaslam2: What you mention is actually how it is working:

 the pwm is needed in two cases:
- spi engine offload is enabled
- using standard spi with dma to toggle the CNV without cpu intervention.

otherwise in the "standard spi" mode without dma a gpio is used to toggle
cnv and antother gpio to toggle cs. 

the driver checks for these two conditions on init to enable the pwm:

see: https://github.com/analogdevicesinc/no-OS/pull/2225/files/06d1222f7f7d125869f526cd69e074ed5a1a5a10..cf315b0fa0fd565db9b3c7c4de314b49b2079a43#diff-f169717926e4e3f8e673fc4f56466d92742b7e39635579b62d80259fc2e8de60R848
```
        if (dev->spi_dma_enable || dev->offload_enable) {
                ret = no_os_pwm_init(&dev->trigger_pwm_desc,
                                     init_param->trigger_pwm_init);
                if (ret != 0)
                        goto error_spi;
        }

```

the ad463x_read_data function will:
 - use offload when USE_STANDARD_SPI is not defined.
 - use "ad463x_read_data_dma" when spi_dma_enable is set
 - use "ad463x_read_single_sample" when spi_dma_enable is not set


```
static int32_t ad463x_read_single_sample(struct ad463x_dev *dev,
                uint32_t *ch0_out,  uint32_t *ch1_out)
{
        uint8_t data[8] = {0};
        int ret;

        if (!dev)
                return -EINVAL;

        ret = no_os_gpio_set_value(dev->gpio_cnv, NO_OS_GPIO_HIGH);
        if (ret)
                return ret;

        ret = no_os_gpio_set_value(dev->gpio_cnv, NO_OS_GPIO_LOW);
        if (ret)
                return ret;

        ret = no_os_spi_write_and_read(dev->spi_desc, data, dev->read_bytes_no);
        if (ret)
                return ret;

        ad463x_pext_sample(dev, data, dev->read_bytes_no, ch0_out, ch1_out);

        return 0;
}

```

projects/ad463x_fmcz/src/platform/stm32/parameters.h
```diff
@@ -0,0 +1,98 @@
...
+
+/******************************************************************************/
+/********************** Macros and Constants Definitions **********************/
+/******************************************************************************/
+extern UART_HandleTypeDef huart5;
+#define SAMPLES_PER_CHANNEL_PLATFORM 2000
+#define MAX_SIZE_BASE_ADDR     (SAMPLES_PER_CHANNEL_PLATFORM * 2 * sizeof(uint32_t))
+extern uint8_t in_buff[];
+
+#define ADC_DDR_BASEADDR	in_buff
```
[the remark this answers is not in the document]
ahaslam2: one solution is to disable the mmu for xilinx:

Xil_DisableMMU

this way we can have buffers larger than 16k which is not too much just
about 2000 samples:
16384/ (4*2) = 2048 samples (2 ch - 4byte)

projects/ad463x_fmcz/src/common/common_data.h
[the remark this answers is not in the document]
ahaslam2: i dont even know why github compares these two files.. but ok.. ill add both

ahaslam2:
Change log:
- added "Antoniu Miclaus" as dual author
- set 1/3 gain directly for AD4224
- avoid calloc on iio read
- fix number of samples for basic example
- RFC to add big buffer for xilinx and disable mmu (last patch)

projects/ad463x_fmcz/src/platform/stm32/parameters.h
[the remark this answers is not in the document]
ahaslam2: i pushed an rfc patch for this.

drivers/adc/ad463x/ad463x.h
machschmitt: Oh, I see now. Okay, that looks good.
I think I got misled by the commit message.

Instead of
> Enable the CNV pwm for standard spi and  Use the dma sync function
> to do sample reads.

I would say something more along the lines of:

Enable DMA transfers with standard SPI when the SPI controller is DMA capable. In that case, request and use a PWM to trigger conversions on the CNV pin.

drivers/adc/ad463x/ad463x.c
```diff
@@ -868,11 +946,11 @@ int32_t ad463x_remove(struct ad463x_dev *dev)
 
 	if (!dev)
 		return -1;
-#ifndef USE_STANDARD_SPI
+
 	ret = no_os_pwm_remove(dev->trigger_pwm_desc);
 	if (ret != 0)
 		return ret;
```
machschmitt: Now we need to check a PWM was initialized before removing it, no?
```c
if (dev->spi_dma_enable || dev->offload_enable) {
        ret = no_os_pwm_remove(dev->trigger_pwm_desc);
        ...
```

drivers/adc/ad463x/ad463x.h
ahaslam2: ok.. ill re-word it.

drivers/adc/ad463x/ad463x.c
ahaslam2: actually i had done some thing like this on a different PR in the past,
but after discussions with the reviewer, it seems that in noOs the duality is
not required as its expected that remove can be called without init being called first
there should be checks for null in every remove.

conclusion was that i should not check if init was called in the remove/error paths.

ahaslam2:
Change log
- modify commit message
-fix astyle error

ahaslam2:
Change Log:
- use Xil_SetTlbAttributes instead of disabling MMU.

ahaslam2:
Seems the large buffer issue was due to offload still writing to buffer when the function returns
and cache invalidate is called. By adding some delay (not sure how to choose it) the glitch on large
buffer is gone. 

Change log
- dropped disable mmu patch
- added delay after offload xfer

ahaslam2:
Change log:

Dropped delay patch as its not longer needed after offload xfer  is fixed with: https://github.com/analogdevicesinc/no-OS/pull/2258

projects/ad463x_fmcz/src/platform/stm32/parameters.h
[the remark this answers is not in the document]
ahaslam2: @CiprianRegus 

i finally traced the buffer issue to offload not waiting for the xfer to complete.
and the buffer to be randomly corrupted with old data.
this is fixed with:
https://github.com/analogdevicesinc/no-OS/pull/2258

ill see try to do the common buffer.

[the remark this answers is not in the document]
ahaslam2: common buffer was added on the latest push.

drivers/adc/ad463x/ad463x.c
```diff
@@ -271,14 +271,14 @@ int32_t ad463x_exit_reg_cfg_mode(struct ad463x_dev *dev)
 				   AD463X_EXIT_CFG_MODE);
 	if (ret != 0)
 		return ret;
-
+#ifndef USE_STANDARD_SPI
```
buha, line 276: i think you could do something like has been done here to avoid these macros (see my first comment and the discussion that followed)
https://github.com/analogdevicesinc/no-OS/pull/2248

drivers/adc/ad463x/iio_ad463x.c
```diff
@@ -328,7 +323,6 @@ static int32_t _iio_ad463x_read_dev(struct iio_ad463x *desc, uint32_t *buff,
 			buff[j++] = data[i];
 		}
 
-	ad463x_enter_config_mode(desc->ad463x_desc);
```
buha, line 331: is this change related ?

drivers/adc/ad463x/ad463x.c
ahaslam2: sure, i prefer not to use macros too. 
It just means that the spi engine stuff will be built and not used
when using standard SPI for all platforms (including stm32).. 

ok with that?

drivers/adc/ad463x/iio_ad463x.c
ahaslam2: this is from the patch by @jonathanns. ill answer from what i understood:

there is no iio fuction that actually needs config mode set, so we dont
set cfg mode after reading the raw data bytes. 

however at start of each function to read data, exit_cofig mode is called
just to make sure that the chip is in a good state to return data, and that
cfg mode was not set by any noise on the MOSI line.

drivers/adc/ad463x/ad463x.c
buha: yes, i'm ok with that

drivers/platform/stm32/stm32_spi.c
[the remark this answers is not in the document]
CiprianRegus: Thank you for explaining this, I understand it now. Also, the timer synchronization mechanism sounds good in my opinion.

drivers/adc/ad463x/ad463x.c
ahaslam2: not so trivial to do. im starting to run into  issues when compiling for STM32:

* spi engine is using  #include "xilinx_spi.h"
so i have to add a hacky  (DRIVERS)/platform/xilinx/xilinx_spi.h"
to the src.mk. not big deal

but then:
* spi_engine is using usleep that is not available in STM32:
/drivers/axi_core/spi_engine/spi_engine.c :
#include <sleep.h>

need to start fixing these to see if i get to compile ;(

ahaslam2: seems to be the only error, so just switching to no_os_delay, in spi_engine looks like enough.

ahaslam2: but then, we get into missing all the axi_ funtions, so i guess we would have to provide
spi_engine stubs for other platforms?


```
/opt/st/stm32cubeide_1.14.0/plugins/com.st.stm32cube.ide.mcu.externaltools.gnu-tools-for-stm32.11.3.rel1.linux64_1.1.100.202309141235/tools/bin/../lib/gcc/arm-none-eabi/11.3.1/../../../../arm-none-eabi/bin/ld: /home/axelh/gits/no-OS2/projects/ad463x_fmcz/build/objs/ad463x_fmcz/build/app/Core/noos/drivers/axi_core/axi_dmac/axi_dmac.o: in function `axi_dmac_read':
/home/axelh/gits/no-OS2/projects/ad463x_fmcz/build/app/Core/noos/drivers/axi_core/axi_dmac/axi_dmac.c:236: undefined reference to `no_os_axi_io_read'
/opt/st/stm32cubeide_1.14.0/plugins/com.st.stm32cube.ide.mcu.externaltools.gnu-tools-for-stm32.11.3.rel1.linux64_1.1.100.202309141235/tools/bin/../lib/gcc/arm-none-eabi/11.3.1/../../../../arm-none-eabi/bin/ld: /home/axelh/gits/no-OS2/projects/ad463x_fmcz/build/objs/ad463x_fmcz/build/app/Core/noos/drivers/axi_core/axi_dmac/axi_dmac.o: in function `axi_dmac_write':
/home/axelh/gits/no-OS2/projects/ad463x_fmcz/build/app/Core/noos/drivers/axi_core/axi_dmac/axi_dmac.c:254: undefined reference to `no_os_axi_io_write'
/opt/st/stm32cubeide_1.14.0/plugins/com.st.stm32cube.ide.mcu.externaltools.gnu-tools-for-stm32.11.3.rel1.linux64_1.1.100.202309141235/tools/bin/../lib/gcc/arm-none-eabi/11.3.1/../../../../arm-none-eabi/bin/ld: /home/axelh/gits/no-OS2/projects/ad463x_fmcz/build/objs/ad463x_fmcz/build/app/Core/noos/drivers/axi_core/clk_axi_clkgen/clk_axi_clkgen.o: in function `axi_clkgen_write':
/home/axelh/gits/no-OS2/projects/ad463x_fmcz/build/app/Core/noos/drivers/axi_core/clk_axi_clkgen/clk_axi_clkgen.c:165: undefined reference to `no_os_axi_io_write'
/opt/st/stm32cubeide_1.14.0/plugins/com.st.stm32cube.ide.mcu.externaltools.gnu-tools-for-stm32.11.3.rel1.linux64_1.1.100.202309141235/tools/bin/../lib/gcc/arm-none-eabi/11.3.1/../../../../arm-none-eabi/bin/ld: /home/axelh/gits/no-OS2/projects/ad463x_fmcz/build/objs/ad463x_fmcz/build/app/Core/noos/drivers/axi_core/clk_axi_clkgen/clk_axi_clkgen.o: in function `axi_clkgen_read':
/home/axelh/gits/no-OS2/projects/ad463x_fmcz/build/app/Core/noos/drivers/axi_core/clk_axi_clkgen/clk_axi_clkgen.c:177: undefined reference to `no_os_axi_io_read'
/opt/st/stm32cubeide_1.14.0/plugins/com.st.stm32cube.ide.mcu.externaltools.gnu-tools-for-stm32.11.3.rel1.linux64_1.1.100.202309141235/tools/bin/../lib/gcc/arm-none-eabi/11.3.1/../../../../arm-none-eabi/bin/ld: /home/axelh/gits/no-OS2/projects/ad463x_fmcz/build/objs/ad463x_fmcz/build/app/Core/noos/drivers/axi_core/spi_engine/spi_engine.o: in function `spi_engine_write':
/home/axelh/gits/no-OS2/projects/ad463x_fmcz/build/app/Core/noos/drivers/axi_core/spi_engine/spi_engine.c:98: undefined reference to `no_os_axi_io_write'
/opt/st/stm32cubeide_1.14.0/plugins/com.st.stm32cube.ide.mcu.externaltools.gnu-tools-for-stm32.11.3.rel1.linux64_1.1.100.202309141235/tools/bin/../lib/gcc/arm-none-eabi/11.3.1/../../../../arm-none-eabi/bin/ld: /home/axelh/gits/no-OS2/projects/ad463x_fmcz/build/objs/ad463x_fmcz/build/app/Core/noos/drivers/axi_core/spi_engine/spi_engine.o: in function `spi_engine_read':
/home/axelh/gits/no-OS2/projects/ad463x_fmcz/build/app/Core/noos/drivers/axi_core/spi_engine/spi_engine.c:116: undefined reference to `no_os_axi_io_read'
```

CiprianRegus (approved):
Approving the STM32 DMA/SPI related commits: b3f90418a8d84345a021402c5a813933e4e084cf ... 5956c3f0efc55b170fea2e7479a5068d0c6f9687

drivers/adc/ad463x/ad463x.c
buha: the stubs approach looks good, would marking them with weak linkage be of any help ?
like these for example: https://github.com/analogdevicesinc/no-OS/blob/main/util/no_os_mutex.c#L47

ahaslam2: @buha 

i could probably make it work with stubs too, but we might have the additional burden in the 
projects to conditionally compile the .c file. the way its right now, the project does not need to do anything special the implementation selection is taken into account by the STANDARD_SPI flag in spi_engine.c

i think weak is for defining multiple implementations of a function, and allow for future unknown implementations too. 
in this case  we only have two known implementations: do-nothing and do-something.

buha: you're right

buha (approved):

PR #2232, opened 2024-06-26
fixes for no_os_dma and stm32_dma

ahaslam2:
## random  dma related fixes for no_os_dma and stm32_dma

Some fixes that jumped to my attention while looking 

## PR Type

## PR Checklist

### DIFF
--- a/drivers/api/no_os_dma.c
+++ b/drivers/api/no_os_dma.c
@@ -251,28 +251,13 @@ int no_os_dma_config_xfer(struct no_os_dma_desc *desc,
 {
 	uint32_t i;
 	int ret;
+	void *discard;
 	struct no_os_callback_desc *sg_callback;
-
 
 	if (!desc || !xfer || !len || !ch)
 		return -EINVAL;
 
 	no_os_mutex_lock(ch->mutex);
-	sg_callback = &ch->cb_desc;
-	sg_callback->peripheral = xfer[0].periph;
-
-	switch (xfer[0].xfer_type) {
-	case MEM_TO_DEV:
-	case MEM_TO_MEM:
-		sg_callback->event = NO_OS_EVT_DMA_TX_COMPLETE;
-		break;
-	case DEV_TO_MEM:
-		sg_callback->event = NO_OS_EVT_DMA_RX_COMPLETE;
-		break;
-	default:
-		ret = -EINVAL;
-		goto unlock;
-	}
 
 	/*
 	 * Add the transfers to the channel's SG list. It's safe to do so, since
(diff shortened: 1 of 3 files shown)


drivers/api/no_os_dma.c
```diff
@@ -303,6 +303,7 @@ int no_os_dma_config_xfer(struct no_os_dma_desc *desc,
 		no_os_irq_set_priority(desc->irq_ctrl, ch->irq_num, xfer[0].irq_priority);
 	}
 
+	no_os_mutex_unlock(ch->mutex);
```
amiclaus: this isn't enough. you need to call it even if an error occurs.

drivers/api/no_os_dma.c
```diff
@@ -248,46 +248,38 @@ int no_os_dma_config_xfer(struct no_os_dma_desc *desc,
 			  uint32_t len, struct no_os_dma_ch *ch)
 {
 	uint32_t i;
-	int ret;
+	int ret = 0;
```
amiclaus, line 251: this explicit initialization is not needed i think.

drivers/api/no_os_dma.c
ahaslam2: yes, and we need to clear the list too.

drivers/api/no_os_dma.c
```diff
@@ -76,8 +76,11 @@ static void default_sg_callback(void *context)
 		old_xfer->xfer_complete_cb(old_xfer, next_xfer,
 					   old_xfer->xfer_complete_ctx);
 
-	if (!list_size) {
+	if (!list_size && data->desc->irq_ctrl) {
 		no_os_irq_disable(data->desc->irq_ctrl, data->channel->irq_num);
+		no_os_irq_unregister_callback(data->desc->irq_ctrl,
```
CiprianRegus, line 81: This isn't really required, since calling `no_os_irq_register_callback()` for the same event/handler (the DMA channel's specific interrupt number in this case) is supposed to just update the already registered IRQ handler, without allocating memory again. So I have not unregistered the callback in this case as an optimization (since the whole calloc/free sequence will take a bit longer than just disabling the interrupt).

I guess the best place to call `no_os_irq_unregister_callback()` would be in `no_os_dma_remove()`.

ahaslam2: Ok,  so stm32_irq needs to be fixed to not allocate then.

ahaslam2: looking at https://github.com/analogdevicesinc/no-OS/blob/df71b3f243e1ae83ba9702e2db548761801ec048/drivers/platform/stm32/stm32_irq.c#L287 it seems we would need to do a lot of book-keeping and tracking to do to  "just update the already registered IRQ handler" in stm32_irq_register_callback if we do it for all the peripheral types.
Looks like what would need to happen  is an unregister and register even if its behind the scene. 
do you see some other way?

do we really need to optimize here? this is a one time thing whenthe xfers are already completed right?

drivers/api/no_os_dma.c
ahaslam2: But the unlock is allready called on error right? check line 311 below:

drivers/api/no_os_dma.c
CiprianRegus: Ok, you're right. I knew this was the case on Maxim and pico platforms and assumed STM32 to be the same. So we can keep you version then.

ahaslam2: looking at it more..and the examples in Maxim and pico.. perhaps we can factor out the allocation in stm32
ill give it a try.

ahaslam2: @CiprianRegus  the new push should address  your original comment.

drivers/api/no_os_dma.c
ahaslam2: i modified the patch and added code to empty the list we filled in case of error.
so i think we need the error path handling with this.

CiprianRegus (approved):

amiclaus (approved):

PR #2234, opened 2024-06-28
util: no_os_clk_fixed.c: add fixed rate clock implementation

jonathanns:
Add a Fixed-Rate Clock implementation using no_os_clk framework.

## Pull Request Description

## PR Type

## PR Checklist

### DIFF
--- a/util/no_os_clk_fixed.c
+++ b/util/no_os_clk_fixed.c
@@ -1,0 +1,83 @@
+/***************************************************************************//**
+ *   @file   no_os_clk_fixed.c
+ *   @brief  Implementation of Fixed Rate Clock Driver based on no_os_clk
+ *      framework.
+ *   @author Jonathan Santos (jonathan.santos@analog.com)
+********************************************************************************
+ * Copyright 2024(c) Analog Devices, Inc.
+ *
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *  - Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ *  - Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in
+ *    the documentation and/or other materials provided with the
+ *    distribution.
+ *  - Neither the name of Analog Devices, Inc. nor the names of its
+ *    contributors may be used to endorse or promote products derived
+ *    from this software without specific prior written permission.
+ *  - The use of this software may or may not infringe the patent rights
+ *    of one or more patent holders.  This license does not release you
+ *    from the requirement that you obtain separate licenses from these
+ *    patent holders to use this software.
+ *  - Use of the software either in source or binary form, must be run
+ *    on or directly connected to an Analog Devices Inc. component.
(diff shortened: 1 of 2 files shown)


util/no_os_clk_fixed.c
```diff
@@ -0,0 +1,83 @@
...
+static int32_t fixed_clk_recalc_rate(struct no_os_clk_desc *hw, uint64_t *rate)
+{
+	struct no_os_clk_fixed_param *dev = hw->dev_desc;
+
+	*rate = dev->clk_frequency;
+
+	return 0;
+}
+
+static const struct no_os_clk_platform_ops fixed_clk_ops = {
```
amiclaus, line 62: I am supposing u are using `recalc_rate` just because u have no other option to get the rate. wondering if a `no_os_clk_set_rate` function in the api makes more sense .  https://github.com/torvalds/linux/blob/22a40d14b572deb80c0648557f4bd502d7e83826/include/linux/clk.h#L1041
But I guess this approach works too.

jonathanns: Actually, according to the `no_os_clk_platform_ops`, the `clk_recalc_rate ` is used to get the current frequency of the CLK poiinter. The fixed clock does not allow setting new frequency.

amiclaus (approved):

util/no_os_clk_fixed.c
```diff
@@ -0,0 +1,83 @@
...
+#include "no_os_alloc.h"
+#include "no_os_error.h"
+#include "no_os_clk.h"
+#include "no_os_clk_fixed.h"
+
+/******************************************************************************/
+/************************** Functions Implementation **************************/
+/******************************************************************************/
+
+static int32_t fixed_clk_recalc_rate(struct no_os_clk_desc *hw, uint64_t *rate)
```
buha, line 53: I'd have implemented this differently because I don't see the need to introduce another init param solely for this purpose (`struct no_os_clk_fixed_param`).

One could populate a `struct no_os_clk_init_param`, provide fixed_clk_ops to it and subsequently do a `clk_set_rate` which is written by you in such a way that only 1 call is allowed (with static boolean in the function).

This way, from the point of view of the user, the only different thing they do when initializing the fixed clock (vs variable) is passing the fixed_clk_ops.

jonathanns: I understand your point, but where would the clock rate information be stored if it's not saved in a struct? Also, is it okay to add a static boolean for this purpose? Would that approach still work if we have more than one fixed clock?

I chose this method to avoid changing the no_os_clk framework, similar to how they handle it in Linux. By the way, I just realized that this `init_param ` does not work weel if we don't have a descriptor to allocate, since the init_param might be deallocated sometime.

buha: okay, what about not using a static boolean in the function (which indeed won't work if you have multiple of these) to a boolean in the `struct no_os_clk` called `fixed` which defaults to false if uninitialized, so it's backward compatible ?

so we keep the requirement that the only thing the API user needs to do is to initialized with 'fixed_clk_ops`

i don't understand the last sentence

jonathanns:
Closing PR because we have a easier way to do so only using the no_os_clk with same effort.

PR #2238, opened 2024-07-02
Staging/max32670

dbogdan:
## Pull Request Description
MAX32670 VDDIO selection support for SPI and UART.

## PR Type

## PR Checklist

### DIFF
--- a/drivers/platform/maxim/max32670/maxim_spi.c
+++ b/drivers/platform/maxim/max32670/maxim_spi.c
@@ -57,6 +57,38 @@
 /************************ Functions Definitions *******************************/
 /******************************************************************************/
 
+/**
+ * @brief Configure the VDDIO level for a SPI interface
+ * @param desc - the SPI descriptor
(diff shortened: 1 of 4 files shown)


drivers/platform/maxim/max32670/maxim_spi.c
```diff
@@ -57,6 +57,38 @@
...
+		break;
+	default:
+		return -EINVAL;
+	}
+
+	spi_pins.vssel = eparam->vssel;
+	MXC_GPIO_Config(&spi_pins);
+
+	return 0;
+}
```
CiprianRegus, line 90: May be worth mentioning in the commit description that MAX32670 only has 1 CS per SPI instance (which is already included in the `gpio_cfg_spi` pin config structs), so we don't need to configure it independently, as for the other Maxim drivers.

amiclaus (commented):
looks good to me overall, just some of the commits are lacking a body.

PR #2241, opened 2024-07-04
drivers: spi_engine: Dont hardcode rx dma size

ahaslam2:
The rx offload size can be 2, 4 or more bytes, it should use the size read from the IP and not be hardcoded to 4.

## PR Type

## PR Checklist


ahaslam2:
@dbogdan Here is the fix for the hardcoded rx size.

amiclaus (approved):

PR #2243, opened 2024-07-05
drivers: More driver fixes for offload_transfer size

ahaslam2:
There seems to have been some mixup between what "sample" means when using offload_transfer function. Drivers sometimes sending number of bytes, and sometimes sending number of chip samples to collect, when in reality The offload_transfer funtion parameters define the "sample" argument as "Number of time the messages will be transferred"

A bug were the spi_engine would trasfer less bytes than num_samples required, may have added to the mixup, making drivers ask for more messages to fill (and sometimes overrun) their data buffer.

Also, some chips collect one channel per offload transaction (ad400x, ad469x) while others collect multiple channels per offload transaction (ad4630x), perhaps ad713x?

ad400x (single channel, 32bit DMA) one message for single channel:
	total_offload_mesages = num_sampels

ad469x (16 channels, 32bit DMA) one message per channel:
	total_offload_mesages = num_sampels * num_channel_active.

ad4630x (2 channel, 64bit DMA) one message for 2 channels:
	total_offload_mesages = num_sampels;

ad713x (4 channel, 128bit DMA) one message for 4 channels:
	total_offload_mesages = num_sampels; (not tested in real hw!)

## PR Type

## PR Checklist

### DIFF
--- a/drivers/adc/ad463x/ad463x.c
+++ b/drivers/adc/ad463x/ad463x.c
@@ -434,15 +434,15 @@ int32_t ad463x_read_data(struct ad463x_dev *dev,
 	msg.commands_data = commands_data;
 
 	if (dev->dcache_invalidate_range)
-		dev->dcache_invalidate_range(msg.rx_addr, samples * 2);
+		dev->dcache_invalidate_range(msg.rx_addr, samples * 2 * sizeof(buf[0]));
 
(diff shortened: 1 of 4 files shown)


ahaslam2:
if someone could test this with  ad713x it would be nice, since i dont have the HW.

PR #2246, opened 2024-07-08
spi-engine: fix data_width read from HDL

commodo:
## Pull Request Description

It's only the lower 16 bits that are actually the data-width in register 0xC.
Bits 23:16 are NUM_SDI.
```
      8'h03: up_rdata_ff <= {8'b0, NUM_OF_SDI, DATA_WIDTH};
```

Link:
  https://github.com/analogdevicesinc/hdl/blob/d95528a533cefc6d82b9fdebd30ee85b3e10c880/library/spi_engine/axi_spi_engine/axi_spi_engine.v#L348

## PR Type

## PR Checklist

### DIFF
--- a/drivers/axi_core/spi_engine/spi_engine.c
+++ b/drivers/axi_core/spi_engine/spi_engine.c
@@ -666,7 +666,9 @@ int32_t spi_engine_init(struct no_os_spi_desc **desc,
 
 	/* Get current data width */
 	spi_engine_read(eng_desc, SPI_ENGINE_REG_DATA_WIDTH, &data_width);
-	eng_desc->max_data_width = data_width;
+	/* Only the lower 16 bits are relevant for the actual data-width */
+	eng_desc->max_data_width = no_os_field_get(SPI_ENGINE_REG_DATA_WIDTH_MSK,
+				   data_width);
 
 	spi_engine_set_transfer_width(*desc, spi_engine_init->data_width);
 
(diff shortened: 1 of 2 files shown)


drivers/axi_core/spi_engine/spi_engine.c
```diff
@@ -666,7 +666,8 @@ int32_t spi_engine_init(struct no_os_spi_desc **desc,
 
 	/* Get current data width */
 	spi_engine_read(eng_desc, SPI_ENGINE_REG_DATA_WIDTH, &data_width);
-	eng_desc->max_data_width = data_width;
+	/* Only lower 16 bits are relevant for the actual data-width */
+	eng_desc->max_data_width = data_width & 0xFFFF;
```
amiclaus, line 670: not mandatory, but i guess u can use:
https://github.com/analogdevicesinc/no-OS/blob/7add1a7756f9f08c62cda688102408d5f8753b41/include/no_os_util.h#L86
and
https://github.com/analogdevicesinc/no-OS/blob/7add1a7756f9f08c62cda688102408d5f8753b41/include/no_os_util.h#L144

amiclaus (commented):
looks good.
Since this is a fix, then a `Fixes:` tag would be nice :). Also, I'd drop that long link in the commit body.

drivers/axi_core/spi_engine/spi_engine.c
commodo: oh; i thought about using something like that, but i didn't find these on a quick search;
i was probably expecting something closer to the kernel naming

amiclaus (approved):

PR #2248, opened 2024-07-11
ad7606: add PWM and AXI clock-gen support for driver

commodo:
## Pull Request Description

This is just a preliminary change that can go in before others, to add AXI/SPI-engine support to the driver.

## PR Type

## PR Checklist

### DIFF
--- a/drivers/adc/ad7606/ad7606.c
+++ b/drivers/adc/ad7606/ad7606.c
@@ -228,6 +228,8 @@ static const uint16_t tconv_max[] = {
 struct ad7606_axi_dev {
 	/* Clock gen for hdl design structure */
 	struct axi_clkgen *clkgen;
+	/* Trigger conversion PWM generator descriptor */
+	struct no_os_pwm_desc *trigger_pwm_desc;
 };
 
 /**
@@ -1197,6 +1199,10 @@ static int32_t ad7606_axi_init(struct ad7606_dev *device,
 	if (ret != 0)
 		goto error;
 
+	ret = no_os_pwm_init(&axi->trigger_pwm_desc, axi_init->trigger_pwm_init);
+	if (ret != 0)
+		goto error;
+
 	/* Note: more validation will be added later */
 error:
 	return ret;
@@ -1368,6 +1374,7 @@ void ad7606_axi_remove(struct ad7606_dev *dev)
 
 	axi = &dev->axi_dev;
 
+	no_os_pwm_remove(axi->trigger_pwm_desc);
 	axi_clkgen_remove(axi->clkgen);
 }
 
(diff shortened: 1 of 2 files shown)


amiclaus (approved):

drivers/adc/ad7606/ad7606.h
```diff
@@ -318,12 +327,18 @@ struct ad7606_dev {
...
 
 /**
- * @struct ad7606_dev
+ * @struct ad7606_init_param
  * @brief Device driver initialization parameters
  */
 struct ad7606_init_param {
 	/** SPI initialization parameters */
 	struct no_os_spi_init_param spi_init;
+#ifdef XILINX_PLATFORM
```
buha, line 338: i think we should not pollute the device driver code with #ifdefs that distinguish hardware platforms, device driver shouldn't care much about what hardware it runs on

in this case it just needs to know that it should or should not use an axi_clkgen (which could be intel, could be xilinx etc.). i suggest to remove all the ifdefs and operate on the fact that clkgen_init is populated or not, this way you move the #ifdef logic to the app level code, which indeed should care about what exact hw platform we're running on

commodo: we can do it like that as well;
this was copied from `ad7616.c`, which doesn't mean that we should use that model;

i'll probably need to re-consider how to group them;
the only benefit these `#ifdef` guards have, is that there is a clear separation between "simple" mode for this driver and FPGA-mode;
the initial version of this driver, seems to have been developed as directly connected to an SoC;

commodo:
v1 -> v2:
* moved `struct ad7606_dev` struct into ad7606.c driver file; this makes it private now;
* split `struct ad7606_axi_dev`, `struct ad7606_axi_init_param ` and functions  `ad7606_axi_init()` & `ad7606_axi_remove()` ; these should kick-in, if there's an AXI init param struct attached to `struct ad7606_init_param`

drivers/adc/ad7606/ad7606.c
```diff
@@ -1197,6 +1199,10 @@ static int32_t ad7606_axi_init(struct ad7606_dev *device,
 	if (ret != 0)
 		goto error;
 
+	ret = no_os_pwm_init(&axi->trigger_pwm_desc, axi_init->trigger_pwm_init);
```
buha, line 1202: i think this makes the PWM usage mandatory while at the same time there is a convst gpio in the init_param / device driver desc 

it's very good that you added the pwm method but there should be a proper way for the user to use either method without being forced to use a particular one

buha: it makes sense then

buha (approved):

drivers/adc/ad7606/ad7606.c
commodo: > also, i think (but i'd need to check with the HDL guys) in the HDL design, the CONVST pin is wired to the CONVST pin to drive the conversion;

typo when writing here; apologies;
CONVST pin is wired to the PWM/HDL pin to drive the conversion;

PR #2260, opened 2024-07-22
ad7606: minor cleanups

commodo:
## Pull Request Description

Just a bunch of cleanups before coming in with some more changes

## PR Type

## PR Checklist

### DIFF
--- a/drivers/adc/ad7606/ad7606.c
+++ b/drivers/adc/ad7606/ad7606.c
@@ -1078,10 +1078,6 @@ int32_t ad7606_set_config(struct ad7606_dev *dev,
 	if (dev->sw_mode) {
 
 		val |= no_os_field_prep(AD7606_CONFIG_OPERATION_MODE_MSK, config.op_mode);
-		/* This driver currently supports only normal SPI with 1 DOUT line.
-		 * TODO: remove this check when implementing multi-line DOUT. */
-		if ((uint8_t)config.dout_format > AD7606_1_DOUT)
-			return -EINVAL;
 		if ((uint8_t)config.dout_format > (uint8_t)dev->max_dout_lines)
 			return -EINVAL;
 		val |= no_os_field_prep(AD7606_CONFIG_DOUT_FORMAT_MSK, config.dout_format);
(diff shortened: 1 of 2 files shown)


drivers/adc/ad7606/ad7606.c
```diff
@@ -649,6 +650,38 @@ int32_t ad7606_read(struct ad7606_dev *dev, uint32_t * data)
...
 
+/***************************************************************************//**
+ * @brief Blocking conversion start and data read (multiple samples).
+ *
+ * This function performs a conversion start and then proceeds to reading
+ * the conversion data.
+ *
+ * @param dev        - The device structure.
+ * @param data       - Pointer to location of buffer where to store the data.
+ *
```
amiclaus, line 661: missing samples.

drivers/adc/ad7606/ad7606.c
```diff
@@ -1069,26 +1069,26 @@ int32_t ad7606_set_ch_gain(struct ad7606_dev *dev, uint8_t ch,
  *                  0 - No errors encountered.
 *******************************************************************************/
 int32_t ad7606_set_config(struct ad7606_dev *dev,
-			  struct ad7606_config config)
+			  struct ad7606_config *config)
```
buha, line 1105: i prefer the pass by value instead of pass by reference here because the structure is very small and also this is a function that's called at the initialization stage of the program, not in some loops/recursion, where repetitive stack copying could be a problem

pointers only make sense when structures are large and/or when structure could be optional (in this case it cannot be optional)

commodo: fine by me: will drop;

drivers/adc/ad7606/ad7606.c
```diff
@@ -649,6 +650,38 @@ int32_t ad7606_read(struct ad7606_dev *dev, uint32_t * data)
...
+int32_t ad7606_read_samples(struct ad7606_dev *dev, uint32_t * data,
+			    uint32_t samples)
+{
+	int32_t ret;
+	uint32_t i;
+
+	for (i = 0; i < samples; i++) {
+		ret = ad7606_read_one_sample(dev, data);
+		if (ret) {
+			printf("Got error at sample %lu: %ld\n", i, ret);
```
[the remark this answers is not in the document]
commodo: `pr_err() ` would be a first here;
i tried to use it initially, but it gave me some build error (linker or something);
maybe i'll just remove this for now;

buha (approved):

amiclaus (approved):

PR #2262, opened 2024-07-23
drivers: axi_core: Add spi engine stubs

ahaslam2:
Currently divers are using STANDARD_SPI macro to wrap all spi engine calls when the platform does not support spi_engine.

Add stub functions to spi_engine and clkgen so that drivers can avoid the use of STANDARD_SPI macro and use a runtime a flag instead.

## PR Type

## PR Checklist


ahaslam2:
abandon this as its part of the ad4630 series.

PR #2266, opened 2024-07-30
ad400x: Fix SDO line idle, and rename to pulsar-adc

ahaslam2:
The hdl binary i had been using to test these parts has an option enabled by default, to set SDO high in idle.
I tested compiling my own hdl, while doing offload reads, i noticed  SDO was going low on idle. 

Turns out the latest HDL needs needs to explicitly enable a bit in cfg register to set the SDO line high in idle,
this is needed for the 400x series of parts. (just writing FF's while reading does not work)

Also ad400x hdl is used for a broader set of parts like the ad798x series.
This is called the "pulsar_adc" in the HDL project.

Follow the same naming convention in No-Os driver/project.

## Pull Request Description

## PR Type

## PR Checklist


buha (dismissed):
ok from me, why is it still a draft ?

ahaslam2:
@buha 

i added to the rename a fix to an issue i found when using the latest HDL i built myself instead of the shared binary.
ad400x parts need sdo line high on idle and this needs a bit on spi engine to be enabled. 
ad7985/86  worked with SDO high or low on idle ,  so we dont need to do it per-part.

PR #2276, opened 2024-08-13
Staging/no os device

dbogdan:
## Pull Request Description

Add no_os_device.h

To be included in all no-OS drivers.

## PR Type

## PR Checklist


buha:
next step would be to add no_os_device.h in src.mk files

buha:
and the next next step would be to actually use these compatible strings in device driver initialization, currently they are unused globals

buha:
and a question: should these also be marked with static ?

PR #2280, opened 2024-08-20
Add max42500 driver and basic project example

JManiti:
## Pull Request Description

The MAX42500 is a SoC power-system monitor with up to seven voltage monitor inputs that will be SIL 3-Certified. Each input has programmable OV/UV thresholds of between 2.5% and 10% with ±1.3% accuracy over the full temperature range. Two of the inputs have a separate remote ground-sense input and support DVS through the integrated I2C interface.
The MAX42500 contains a programmable flexible power sequence recorder (FPSR). This recorder stores power-up and power-down timestamps separately, and supports on/off and sleep/standby power sequences. The MAX42500 also contains a programmable challenge/response watchdog, which is accessible through the I2C interface, along with a configurable RESET output.
The MAX42500 improves reliability while significantly reducing system size and component count as compared to separate ICs or discrete components. The MAX42500 is suitable for use in safety functions up to SIL 3. The device is designed to operate over the ambient temperature range of -40°C to +125°C.

## PR Type

## PR Checklist


drivers/power/max42500/max42500.h
```diff
@@ -0,0 +1,280 @@
...
+
+/* WDLOCK bit masks */
+#define MAX42500_WDLOCK_LOCK_MASK   NO_OS_BIT(0)
+
+/* RSTCTRL bit masks */
+#define MAX42500_RSTCTRL_MR1_MASK   NO_OS_BIT(2)
+#define MAX42500_RSTCTRL_RHLD_MASK  (0x3)
+
+/* MAX42500 registers list */
+enum max42500_registers {
```
danmois, line 84: Declared also in the header.

drivers/power/max42500/max42500.c
```diff
@@ -0,0 +1,904 @@
...
+		en0 = NO_OS_GPIO_HIGH;
+		en1 = NO_OS_GPIO_HIGH;
+		break;
+	}
+
+	ret = no_os_gpio_set_value(desc->en0, en0);
+	if (ret)
+		return ret;
+
+	ret = no_os_gpio_set_value(desc->en1, en1);
```
danmois, line 136: You can directly return here.
return no_os_gpio_set_value();

drivers/power/max42500/max42500.c
```diff
@@ -0,0 +1,904 @@
...
+ * @param desc - The device structure.
+ * @param mbst - The MBST bit value.
+ * @return 0 in case of success, error code otherwise
+ */
+int max42500_set_mbst(struct max42500_dev *desc, bool mbst)
+{
+	if (!desc)
+		return -EINVAL;
+
+	int ret;
```
danmois, line 268: Should be declared at start of function.

drivers/power/max42500/max42500.c
```diff
@@ -0,0 +1,904 @@
...
+
+/**
+ * @brief Enable watchdog
+ * @param desc - The device structure.
+ * @param wd_enable - Enable/disable MAX42500 watchdog.
+ * @return 0 in case of success, error code otherwise
+ */
+int max42500_set_watchdog_enable(struct max42500_dev *desc, bool wd_enable)
+{
+	if (!desc)
```
danmois, line 537: Remove one occurrence.

drivers/power/max42500/max42500.c
```diff
@@ -0,0 +1,904 @@
...
+
+/**
+ * @brief Set watchdog lock
+ * @param desc - The device structure.
+ * @param wd_lock - Enable/disable MAX42500 watchdog register access.
+ * @return 0 in case of success, error code otherwise
+ */
+int max42500_set_watchdog_lock(struct max42500_dev *desc, bool wd_lock)
+{
+	if (!desc)
```
danmois, line 626: Remove one occurrence.

drivers/power/max42500/max42500.c
```diff
@@ -0,0 +1,904 @@
...
+	for (uint8_t idx = 0; idx < len; idx++) {
+		crc ^= data[idx];
+
+		for (uint8_t loop=0; loop < 8; loop++) {
+			msb_set = ((crc & 0x80) == 0x80);
+			crc <<= 1;
+
+			/* Divide by PEC if bit 8 was set */
+			if (msb_set)
+				crc ^= CRC8_PEC;
```
buha, line 79: is this PEC incompatible with [this implementation](https://github.com/analogdevicesinc/no-OS/blob/main/util/no_os_crc8.c) ?

drivers/power/max42500/max42500.c
```diff
@@ -0,0 +1,904 @@
...
+}
+
+/**
+ * @brief 8-bit watchdog key computation
+ * @param curr_wd_key - Current watchdog key value.
+ * @return New 8-bit watchdog key value
+ */
+static uint8_t max42500_new_watchdog_key(uint8_t curr_wd_key)
+{
+	return curr_wd_key;
```
buha, line 93: is this what this function is supposed to be doing ?

drivers/power/max42500/max42500.c
```diff
@@ -0,0 +1,904 @@
...
+	else
+		reg.value &= ~MAX42500_CONFIG1_RR_MASK;
+
+	/* Write register */
+	ret = max42500_reg_write(desc, reg);
+	if (ret)
+		return ret;
+
+	/* Update register value */
+	desc->regs[MAX42500_CONFIG1].value = reg.value;
```
buha, line 312: I find this `regs` register table where you cache their values makes the code more verbose and also more inefficient and you can also run into issues like `regs` cached version straying from reality.

But in this review I care more about verbosity. For example this register value update can be moved in the `reg_write()` function, this way you simplify all these "feature" r/w routines.

Also, because of how it's designed, you need to kind of write a function for every interesting bit or bitmask in every register resulting again in verbose code. Take [this example](https://github.com/analogdevicesinc/no-OS/blob/main/drivers/adc/ad7606/ad7606.c#L967) where we wrote a `ad7606_spi_write_mask` routine that performs read&write with a mask to any register bit or field, so we don't need a function for every bit and every bit field. Most of our device drivers have such a `write with mask` function.

drivers/power/max42500/max42500.h
```diff
@@ -0,0 +1,280 @@
...
+#define MAX42500_VMON_IN_MASK(bit)  NO_OS_BIT(bit)
+#define MAX42500_VMON_VMPD_MASK     NO_OS_BIT(7)
+
+/* RSTMAP bit masks */
+#define MAX42500_RSTMAP_IN_MASK(bit)    NO_OS_BIT(bit)
+#define MAX42500_RSTMAP_PARM_MASK       NO_OS_BIT(7)
+
+/* WDCDIV bit masks */
+#define MAX42500_WDCDIV_SWW_MASK    NO_OS_BIT(6)
+#define MAX42500_WDCDIV_WDIC_MASK   (0x3F)
```
buha, line 70: use `NO_OS_GENMASK`

drivers/power/max42500/max42500.h
```diff
@@ -0,0 +1,280 @@
...
+#define MAX42500_I2C_WR_FRAME_SIZE  (4)
+#define MAX42500_I2C_RD_FRAME_SIZE  (5)
+
+/* CONFIG1 bit masks */
+#define MAX42500_CONFIG1_PECE_MASK  NO_OS_BIT(0)
+#define MAX42500_CONFIG1_MBST_MASK  NO_OS_BIT(1)
+#define MAX42500_CONFIG1_RR_MASK    NO_OS_BIT(2)
+
+/* VMON bit masks */
+#define MAX42500_VMON_IN_MASK(bit)  NO_OS_BIT(bit)
```
buha, line 61: This is just a rename and is not necessary, you can use NO_OS_BIT directly where you configure this VMON_IN

drivers/power/max42500/max42500.c
JManiti: This is noted.

drivers/power/max42500/max42500.h
JManiti: This is noted.

drivers/power/max42500/max42500.h
```diff
@@ -0,0 +1,280 @@
...
+
+/* Device register info */
+struct max42500_reg_st {
+	const enum max42500_registers addr;
+	uint8_t value;
+};
+
+/* MAX42500 watchdog mode */
+enum max42500_state {
+	MAX42500_STATE_OFF = 0,
```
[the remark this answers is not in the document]
JManiti: This is noted.

drivers/power/max42500/max42500.c
JManiti: Sorry, will update this function.

drivers/power/max42500/max42500.c
JManiti: will update the code using that implementation

JManiti:
Update v1: 

- Fixed nits
- use the no_os_crc8
- create write with a mask function and remove some of the functions
- remove duplicated codes
- use NO_OS_GENMASK

drivers/power/max42500/max42500.c
```diff
@@ -0,0 +1,473 @@
...
+ * @return 0 in case of success, error code otherwise.
+ */
+int max42500_reg_write2(struct max42500_dev *desc,
+			uint32_t reg,
+			uint32_t writeval)
+{
+	desc->regs[reg].value = writeval;
+
+	return max42500_reg_write(desc, desc->regs[reg]);
+}
```
CiprianRegus, line 206: There is no need to have 2 of these "register write" functions. In order to keep compatibility with the rest of the no os drivers, you should only implement this one.

drivers/power/max42500/max42500.c
```diff
@@ -0,0 +1,473 @@
...
+		return ret;
+
+	ret = no_os_gpio_remove(desc->en0);
+	if (ret)
+		return ret;
+
+	no_os_free(desc);
+
+	return 0;
+}
```
CiprianRegus, line 473: The timer and interrupt controller also have to be removed

drivers/power/max42500/max42500.c
```diff
@@ -0,0 +1,473 @@
...
+
+	/* Initialize timer for watchdog */
+	ret = no_os_timer_init(&descriptor->timer_desc, &init_param->timer_param);
+	if (ret)
+		goto free_i2c;
+
+	/* Initialize IRQ controller */
+	ret = no_os_irq_ctrl_init(&descriptor->irq_desc, &init_param->irq_param);
+	if (ret)
+		goto free_timer;
```
CiprianRegus, line 409: This watchdog functionality seems to only be half implemented. I assume you want to reset the watchdog based on a timer interrupt? Anyway, I would suggest to let the user decide whether or not they need this (by checking the `timer_param` field for `NULL` for example).

Also, currently, the user of the driver has to register a callback for the timer and enable it in order for this to be useful. If you choose to keep this functionality, I would suggest to add a few parameters or functions in order to do the full interrupt setup.

projects/max42500/src/platform/parameters.h
```diff
@@ -0,0 +1,75 @@
+/***************************************************************************//**
```
CiprianRegus, line 1: This header file should be for macros and variables specific to a target platform (e.g Maxim, STM32, etc). Which one is it?

drivers/power/max42500/max42500_regs.h
```diff
@@ -0,0 +1,48 @@
...
+* SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+*******************************************************************************/
+
+#ifndef __MAX42500_REG_H__
+#define __MAX42500_REG_H__
+
+#include "max42500.h"
+
+/*! Array holding the info for the MAX42500 registers - address, initial value */
+extern struct max42500_reg_st max42500_regs[MAX42500_REG_NO];
```
CiprianRegus, line 46: The user doesn't need to access this register array, so you can only define it in the source file. Or even better, you can delete it and just initialize all the registers to 0 (apart from the first one) in a loop during init.

projects/max42500/src/common/common_data.c
```diff
@@ -0,0 +1,121 @@
...
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+*******************************************************************************/
+
+#include "common_data.h"
+#include "maxim_gpio.h"
+#include "maxim_i2c.h"
+#include "maxim_irq.h"
+#include "maxim_timer.h"
+#include "maxim_uart.h"
+
```
CiprianRegus, line 46: Extra empty line

drivers/power/max42500/max42500.c
JManiti: This is noted.

drivers/power/max42500/max42500.c
JManiti: I will remove the watchdog functionality for now.

drivers/power/max42500/max42500_regs.h
JManiti: Is it okay to keep this one?

projects/max42500/src/platform/parameters.h
JManiti: I now added a folder and additional files to address this one.

JManiti:
The files are updated. Thank you.

JManiti:
Files are updated. Thank you

amiclaus (commented):
it looks like the links for the documentation are missing. please also rebase the code with the latest main branch.

JManiti:
Update:
-The register files are now removed.
-The driver is updated with newly added functions.

drivers/power/max42500/README.rst
```diff
@@ -0,0 +1,90 @@
...
+MAX42500 Device Configuration
+-----------------------------
+
+Driver Initialization
+---------------------
+In order to be able to use the device, you will have to provide the support for 
+the communication protocol (I2C) alongside other GPIO pins if needed in the 
+specific application (depends on the way the device is used).
+
+The first API to be called is max42500_init. Make sure that it returns 0, which
```
buha, line 40: add a few more examples here besides the init function, it doesn't need to cover the whole device driver API but the basic stuff like data reading

JManiti:
Update:

- Added more API documentation in the driver documentation
- Edited the function documentation in the driver
- Edited the JSON file to hopefully fix the unsuccessful checks

JManiti:
Update:

-Added a title in the project documentation

amiclaus (approved):

PR #2299, opened 2024-09-17
Add driver for LT7170/LT7170-1, LT7171/LT7171-1

csarip:
## Pull Request Description

The LT7170/LT7170-1 are monolithic DC/DC synchronous step-down regulators that 
deliver up to 20 A of continuous output current. The LT7170-1 option has two 
switching phases that are connected to two inductors to drive a 
single-regulated output supply.

Similarly, the LT7171/LT7171-1 are monolithic PolyPhase DC/DC synchronous 
step-down regulators that also deliver up to 20A of continuous output current. 
The LT7171-1 option has two switching phases that are connected to two 
inductors to drive a single-regulated output supply.

Both series offer quick, clean, low-overshoot switching edges deliver high 
efficiency while minimizing electromagnetic interference (EMI) emissions.

An I2C-based PMBus 1.3-compliant serial interface allows control of device 
functions while providing telemetry information for system monitoring. The 
LT7170/LT7170-1 and LT7171/LT7171-1 are supported by the LTpowerPlay graphical 
user interface (GUI) tool.

## PR Type

## PR Checklist

### DIFF
--- a/drivers/power/lt7170/README.rst
+++ b/drivers/power/lt7170/README.rst
@@ -1,0 +1,287 @@
+LT7170 no-OS driver
+====================
+
+Supported Devices
+-----------------
+
+* `LT7170/LT7170-1 <https://www.analog.com/LT7170>`_
+* `LT7171/LT7171-1 <https://www.analog.com/LT7171>`_
+
+Overview
+--------
+
+The LT7170/LT7170-1 are monolithic DC/DC synchronous step-down regulators that 
+deliver up to 20 A of continuous output current. The LT7170-1 option has two 
+switching phases that are connected to two inductors to drive a 
+single-regulated output supply.
+
+Similarly, the LT7171/LT7171-1 are monolithic PolyPhase DC/DC synchronous 
+step-down regulators that also deliver up to 20A of continuous output current. 
+The LT7171-1 option has two switching phases that are connected to two 
+inductors to drive a single-regulated output supply.
+
+Both series offer quick, clean, low-overshoot switching edges deliver high 
+efficiency while minimizing electromagnetic interference (EMI) emissions.
+
+An I2C-based PMBus 1.3-compliant serial interface allows control of device 
+functions while providing telemetry information for system monitoring. The 
+LT7170/LT7170-1 and LT7171/LT7171-1 are supported by the LTpowerPlay graphical 
+user interface (GUI) tool.
+
+Applications
+------------
+
+LT7170
+-------
+
+* Communications, Storage, and Industrial Systems
+* Data Center and Solid-State Drives
+
+LT7170 Device Configuration
+----------------------------
+
+Driver Initialization
+---------------------
+
+In order to be able to use the device, you will have to provide the support
+for the communication protocol (I2C) alongside other GPIO pins if needed in the
+specific application (depends on the way the device is used).
+
+The first API to be called is **lt7170_init**. Make sure that it return 0,
+which means that the driver was initialized correctly.
+
+The initialization API uses the device descriptor and an initialization
+parameter. The initialization parameter contains the **chip_id** of the device 
+and the optional GPIO
+parameters for controlling POWER GOOD, RUN, FAULT and ALERT pins. Use of packet
+error checking and the external clock can also be configured in the parameter.
+These are defined in the header file of the driver.
+
+VOUT Configuration
+------------------
+
+The VOUT_COMMAND sets the output voltage, while the VOUT_MAX sets the upper
+limit on the commanded voltage. The output voltage and limit can be set using 
+the **lt7170_vout_value** API.
+
+The output voltage transition rate can be configured using the 
+**lt7170_vout_tr** API.
+
+Output voltage lower and higher margin can also be configured using the
+**lt7170_vout_margin** API.
+
+The OPERATION command via the **lt7170_set_operation** can be used to command
+the output voltage to be regulated at the high margin, low margin, or at the
+nominal VOUT_COMMAND. This can also be used to set a channel off.
+
+VIN Configuration
+-----------------
+
+VIN_ON and VIN_OFF command values sets the input voltage window at which power
+conversion will proceed. Both of which can be configured through the
+**lt7170_set_vin** API.
+
+Status Configuration
+--------------------
+
+Statuses bytes/words of the device can be read using **lt7170_read_status**
+API. Faults asserted by these statuses can be cleared using the
+**lt7170_clear_faults** API.
+
+Telemetry
+---------
+
+Measurements can be read using the **lt7170_read_value** API. Some telemetry 
+values includes input/output voltage,
+input/output current, external VCC, die temperature, and output power.
+
+The frequency at which telemetry values are converted, as well as enabling debug
+telemetries such as external VCC and ITH values, can be configured using the
+**lt7170_adc_control** API.
+
+Timing Configuration
+--------------------
+
+The TON_DELAY command sets the time from when a start condition is received
+until the output voltage starts to rise. The TON_RISE sets the time from the
+from the time the output starts to rise to the time the output enters the
+regulation band. The TOFF_DELAY command sets the time from when a stop condition
+is received until the output voltage starts to fall. The TOFF_FALL command sets
+the time from the end of the turn-off delay time until the output voltage is
+commanded to zero. These commands can be set using the **lt7170_set_timing**
+API.
+
+Overvalue and Undervalue Limits Configuration
+---------------------------------------------
+
+Overvalue and undervalue limits sets the threshold at which the device voltage,
+current, temperature, and timing must meet. When these measurements cross the
+limits, a status bit may be asserted. These limits can be configured using the
+**lt7170_set_limit** API.
+
+PWM Configuration
+-----------------
+
+The PWM can be set on forced continuous mode or on
+pulse skip mode. This can be set using the **lt7170_pwm_mode** API.
+
+The phase offset of the PWM can also be set using the
+**lt7170_pwm_phase** API.
+
+Clock Configuration
+-------------------
+
+When using the device in parallel with others of the same device, the PWM clock
+of all devices can be synchronized. Using an external clock or enabling the
+clock input can be configured using the **lt7170_sync_config** API.
+
+NVM commands
+------------
+
+The device has an internal EEPROM that saves the settings configured to the
+device even when power is removed. The **lt7170_nvm_cmd** API allows the user
+to lock and unlock EEPROM, and store, compare and restore user settings.
+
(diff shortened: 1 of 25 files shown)


dbogdan (approved):

amiclaus (approved):

drivers/power/lt7170/lt7170.c
```diff
@@ -0,0 +1,1074 @@
...
+
+	/* Convert to sign, 5 bit exponent, 10 bit mantissa */
+	*reg = no_os_field_prep(LT7170_IEEE754_SIGN_BIT, sign) |
+	       no_os_field_prep(LT7170_IEEE754_MANTISSA_MSK, mantissa) |
+	       no_os_field_prep(LT7170_IEEE754_EXPONENT_MSK, exponent);
+
+	return 0;
+}
+/**
+ * @brief Converts raw IEEE754 register data to its actual value
```
kister-jimenez, line 134: nit: add a newline between functions.

drivers/power/lt7170/lt7170.c
```diff
@@ -0,0 +1,1074 @@
...
+	if (!block)
+		return -ENOMEM;
+
+	dev = (struct lt7170_dev *)no_os_calloc(1, sizeof(struct lt7170_dev));
+	if (!dev) {
+		ret = -ENOMEM;
+		goto free_block;
+	}
+
+	/* Initialize I2C */
```
kister-jimenez, line 321: You can probably remove this comment as this is obvious.

drivers/power/lt7170/lt7170.c
```diff
@@ -0,0 +1,1074 @@
...
+	ret = lt7170_read_word(dev, LT7170_MFR_SPECIAL_ID, &word);
+	if (word != LT7170_SPECIAL_ID_VALUE) {
+		ret = -EIO;
+		goto dev_err;
+	}
+
+	ret = lt7170_read_block_data(dev, LT7170_IC_DEVICE_ID, block,
+				     lt7170_info[init_param->chip_id].name_size);
+	if (ret)
+		goto dev_err;
```
kister-jimenez, line 336: nit: adding a newline after the goto line makes this easier to read.

drivers/power/lt7170/lt7170.c
```diff
@@ -0,0 +1,1074 @@
...
+}
+
+/**
+ * @brief Set switching frequency
+ * @param dev - Device structure
+ * @param freq - Frequency to set in Hz. Value should be between 400 kHz and
+ * 		 4000 kHz.
+ * @return 0 in case of success, negative error code otherwise
+ */
+int lt7170_switch_freq(struct lt7170_dev *dev, int freq) // freq in kHz
```
kister-jimenez, line 881: No need for the inline comment. freq is describe in the @param.

drivers/power/lt7170/lt7170.c
```diff
@@ -0,0 +1,1074 @@
...
+		      enum lt7170_timing_type timing_type, int time) // in us
+{
+	return lt7170_write_word_data(dev, timing_type, time);
+}
+
+/**
+ * @brief Set switching frequency
+ * @param dev - Device structure
+ * @param freq - Frequency to set in Hz. Value should be between 400 kHz and
+ * 		 4000 kHz.
```
kister-jimenez, line 878: Confusing parameter. Is it Hz or kHz?

drivers/power/lt7170/iio_lt7170.c
```diff
@@ -0,0 +1,1039 @@
...
+	},
+	{
+		.name = "pulse_skipping_available",
+		.show = lt7170_iio_read_enable_available,
+		.shared = IIO_SHARED_BY_ALL
+	},
+	END_ATTRIBUTES_ARRAY
+};
+
+
```
kister-jimenez, line 885: nit: for consistency, just remove this extra new line after the function.

drivers/power/lt7170/README.rst
```diff
@@ -0,0 +1,287 @@
...
+
+Software Reset Configuration
+----------------------------
+
+Software Reset operation is available through **lt7170_software_reset** API.
+
+LT7170 Driver Initialization Example
+-------------------------------------
+
+.. code-block:: bash
```
kister-jimenez, line 153: Should this be c instead of bash?

drivers/power/lt7170/README.rst
```diff
@@ -0,0 +1,287 @@
...
+* ``status_input - INPUT status byte value of channel 0``
+* ``status_mfr_specific - MFR_SPECIFIC status byte value of channel 0``
+* ``status_word - Status word value of the channel 0``
+* ``status_temperature - TEMPERATURE status byte value of the device``
+* ``status_cml - CML status byte value of the device``
+
+LT7170 IIO Driver Initialization Example
+-----------------------------------------
+
+.. code-block:: bash
```
kister-jimenez, line 255: ditto

projects/lt7170/src/common/common_data.c
```diff
@@ -0,0 +1,71 @@
...
+	.i2c_init = &lt7170_i2c_ip,
+	.pg_param = NULL,
+	.run_param = NULL,
+	.alert_param = NULL,
+	.fault_param = NULL,
+	.fault_cfg = LT7170_FAULT_PIN_OUTPUT,
+	.chip_id = ID_LT7170,
+	.external_clk_en = false,
+	.crc_en = false,
+};
```
kister-jimenez, line 71: Add a newline at the end of the file.

kister-jimenez (changes requested):
Please use the 50/72 rule for the commit subject/body

drivers/power/lt7170/lt7170.c
csarip: It is set in Hz. Would it be better if I avoid using 'k' and display the zeros here instead?

csarip: It is set in Hz. Would it be better if I avoid using 'k' and display the zeros here instead?

drivers/power/lt7170/lt7170.c
```diff
@@ -0,0 +1,1074 @@
...
+	ret = no_os_i2c_init(&dev->i2c_desc, init_param->i2c_init);
+	if (ret)
+		goto i2c_err;
+
+	/* Identify device */
+	ret = lt7170_read_word(dev, LT7170_MFR_SPECIAL_ID, &word);
+	if (word != LT7170_SPECIAL_ID_VALUE) {
+		ret = -EIO;
+		goto dev_err;
+	}
```
CiprianRegus, line 331: Do we need to check this? If the MCU can't communicate with the device, the next check would fail anyway, right?

drivers/power/lt7170/lt7170.c
```diff
@@ -0,0 +1,1074 @@
...
+		ret = no_os_i2c_read(dev->i2c_desc, rx_buf, 2, 1);
+		if (ret)
+			return ret;
+
+		if (lt7170_pec(dev, cmd, rx_buf, 1, 1) != rx_buf[1])
+			return -EBADMSG;
+
+		*data = rx_buf[0];
+		return ret;
+	} else
```
CiprianRegus, line 523: add `{` for the `else` branch as well if the first case required it.

drivers/power/lt7170/iio_lt7170.c
```diff
@@ -0,0 +1,1039 @@
...
+	case LT7170_REVISION:
+	case LT7170_MFR_FAULT_RESPONSE:
+	case LT7170_MFR_ADC_CONTROL:
+	case LT7170_MFR_COMMON:
+	case LT7170_MFR_CHANNEL_STATE:
+	case LT7170_MFR_SYNC_CONFIG:
+	case LT7170_MFR_RAIL_ADDRESS:
+	case LT7170_MFR_DISABLE_OUTPUT:
+		return lt7170_read_byte(lt7170, (uint8_t)reg,
+					(uint8_t *)readval);
```
CiprianRegus, line 119: You will have to use a separate `uint8_t` variable for this. Currently, the result will be different on big endian architectures

drivers/power/lt7170/iio_lt7170.c
```diff
@@ -0,0 +1,1039 @@
...
+		.show = lt7170_iio_read_enable,
+		.store = lt7170_iio_write_enable,
+		.priv = LT7170_IIO_PULSE_ENABLE
+	},
+	{
+		.name = "pulse_skipping_available",
+		.show = lt7170_iio_read_enable_available,
+		.shared = IIO_SHARED_BY_ALL
+	},
+	END_ATTRIBUTES_ARRAY
```
CiprianRegus, line 882: Most of these are not IIO attributes defined in the IIO ABI (https://www.kernel.org/doc/Documentation/ABI/testing/sysfs-bus-iio). Thus, they should be debug attributes instead.

drivers/power/lt7170/iio_lt7170.c
csarip: Hi, I apologize I'm a bit confused. Could you kindly elaborate on this? I referred to this implementation on that line of code:  [LT7182S_IIO](https://github.com/analogdevicesinc/no-OS/blob/main/drivers/power/lt7182s/iio_lt7182s.c#L123)

CiprianRegus: The function call should be done as following:

```
uint8_t reg_val;
lt7170_read_byte(lt7170, reg, &readval);
*read_val = reg_val;
```

Since your `lt7170_read_byte()` function takes a parameter that's a pointer to an 1 byte unsigned int, it will end up setting the first byte at the address `readval`. Depending on the microcontroller endianess, the actual result will be interpreted as `0xFE000000` or `0x000000FE`, so the values will be different.

Another issue is that by setting only 1 byte, you don't clear any non 0 values in that memory region, so the result will be wrong. For example:

uint32_t reg_val; /* Not initialized, so its value is set to whatever was on the stack */
lt7170_iio_reg_read(..., &reg_val);

At this point, only the first byte of `reg_val` was set, so your result will be wrong.

csarip:
V2:

- Corrected formatting as per review comments.
- Replaced dynamically allocated array with stack-allocated array.
- Removed unnecessary checking of device.
- Used a separate uint8_t variable to ensure correct result regardless of the microcontroller's endianness.
- Moved attributes that belong to debug attributes.

drivers/power/lt7170/README.rst
csarip: Used Bash for consistency, as other drivers use Bash instead of C.

CiprianRegus (approved):

buha (approved):

kister-jimenez (approved):

PR #2301, opened 2024-09-18
drivers:platform:stm32:Added support for STM32H5 in the DMA driver

Shetty-Anush:
## Pull Request Description

The DMA drivers specific to STM32 did not have support for STM32H5 which lead to build errors in certain projects.

In this Pull Request, we can find changes made according to the DMA handler of STM32H5 for proper initialization in the driver.

## PR Type

## PR Checklist

### DIFF
--- a/drivers/platform/stm32/stm32_dma.c
+++ b/drivers/platform/stm32/stm32_dma.c
@@ -70,11 +70,55 @@ int stm32_dma_config_xfer(struct no_os_dma_ch *channel,
 
 	/* Note: Channel number is assigned via the Instance for MCU
 	 * families other than STM32F2, STM32F4 and STM32F7 */
-#if defined (STM32F2) || defined (STM32F4) || defined (STM32F7)
-	sdma_ch->hdma->Init.Channel = sdma_ch->ch_num;
-#else
+#if defined (STM32H5)
 	sdma_ch->hdma->Instance = sdma_ch->ch_num;
+#else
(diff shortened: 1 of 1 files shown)


drivers/platform/stm32/stm32_dma.c
```diff
@@ -118,6 +119,50 @@ int stm32_dma_config_xfer(struct no_os_dma_ch *channel,
 	default:
 		return -EINVAL;
 	}
+#else
```
buha, line 122: I think the best play here is to wrap H5 specific code into an #if defined (STM32F5) directive, and put the existing implementation into the #else branch.

buha: Even better, it looks like the fudamental difference is between the DMA Cores so maybe you can find a macro that you can use to distinguish instead of distinguishing by chip family because we are not ST and we don't want to care which family has which type of DMA Core.

buha:
Also, click on the CLA button to sign the agreement.

drivers/platform/stm32/stm32_dma.c
Shetty-Anush: Wrapping the H5 specific code into #if directive can be done; but even though it is true that F2, F4 and F7 have a different DMA type than H5, there does not seem to be any specific macro in the STM32 drivers that help us distinguish between them.

buha (approved):

amiclaus (approved):

PR #2318, opened 2024-09-30
XSPI (OctoSPI/Quad SPI/Dual SPI) support

NagaHimanshu:
## Pull Request Description

- Added no_os_xspi to drivers/api
- Added support for xspi on stm32 platform
- Tested on Nucleo-H563ZI board

## PR Type

## PR Checklist

### DIFF
--- a/drivers/platform/stm32/stm32_xspi.c
+++ b/drivers/platform/stm32/stm32_xspi.c
@@ -1,0 +1,721 @@
+/***************************************************************************//**
+ *   @file   stm32/stm32_xspi.c
+ *   @brief  Implementation of stm32 xspi driver.
+ *   @author Naga Himanshu Indraganti (naga.indraganti@analog.com)
+********************************************************************************
+ * Copyright 2024(c) Analog Devices, Inc.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice,
+ *    this list of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * 3. Neither the name of Analog Devices, Inc. nor the names of its
+ *    contributors may be used to endorse or promote products derived from this
+ *    software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY ANALOG DEVICES, INC. “AS IS” AND ANY EXPRESS OR
+ * IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF
+ * MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO
+ * EVENT SHALL ANALOG DEVICES, INC. BE LIABLE FOR ANY DIRECT, INDIRECT,
+ * INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA,
+ * OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF
+ * LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING
+ * NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE,
+ * EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+*******************************************************************************/
+
+#include <stdlib.h>
+#include <string.h>
+#include <errno.h>
+#include "no_os_irq.h"
+#include "no_os_delay.h"
+#include "no_os_alloc.h"
+#include "no_os_mutex.h"
+#include "no_os_dma.h"
+#include "stm32_gpio.h"
+#include "stm32_dma.h"
+#include "stm32_xspi.h"
+
+/*
+ * @brief Initialize the XSPI peripheral.
+ * @param desc - The XSPI descriptor
+ * @param param - The structure that contains the XSPI parameters.
+ * @return 0 in case of success, negative code otherwise.
+ */
+static int32_t stm32_xspi_peripheral_init(struct no_os_spi_desc *desc,
+		const struct no_os_spi_init_param *param)
+{
+	int32_t ret;
+	uint32_t prescaler = 0u;
+	struct stm32_xspi_desc *sdesc;
+	XSPI_TypeDef *base = NULL;
+
+	if (!desc || !param)
+		return -EINVAL;
+
+	sdesc = desc->extra;
+
+	/* Calculate prescaler based on given clock frequency */
+	if (desc->max_speed_hz != 0u) {
+		uint32_t div = sdesc->input_clock / desc->max_speed_hz;
+
+		/* If exact division is possible, use div - 1 (since the peripheral does
+		 * clock frequency as 2 ^ (prescaler + 1) */
+		if (!(sdesc->input_clock % desc->max_speed_hz)) {
+			prescaler = div - 1;
+		} else {
+			prescaler = div;
+		}
+	} else {
+		prescaler = 24;
+	}
+
+	switch (desc->device_id) {
+	case 1:
+#if defined(OCTOSPI1)
+		base = OCTOSPI1;
+		break;
+#endif
+#if defined(OCTOSPI2)
+	case 2:
+		base = OCTOSPI2;
+		break;
+#endif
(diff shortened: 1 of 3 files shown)


include/no_os_xspi.h
```diff
@@ -0,0 +1,255 @@
+/***************************************************************************//**
```
buha, line 1: I don't think we should add no_os_xpi.h/.c

We can just as well use no_os_spi.h module for the same purpose.
I've diffed your xspi.h to spi.h and it seems the only fundamental change is that you added a specifier for the number of lanes (single, bi, quad, octo). This could be easily done by adding a `lanes` member to the spi init_param and desc, which by default/uninitialized would mean single lane to be backwards compatible.

Then you'd proceed to implement your own implementation of the no_os_spi API in stm32_octospi for example, and populate the platform ops. Keep in mind that the same no_os_spi API could be implemented as xilinx_octalspi, so no_os_spi needs to be as generic as possible.

drivers/platform/stm32/stm32_xspi.c
```diff
@@ -0,0 +1,718 @@
+/***************************************************************************//**
```
buha, line 1: so this is good in principle, I'd rename it and the function ops to stm32_octospi to reflect the true peripheral being used

and as I said in the other comment, make it use generic (and slightly modified) no_os_spi instead of a new no_os_xspi

include/no_os_xspi.h
NagaHimanshu: Sure, this is a great idea and can be done. 

In STM HAL files, they have referred it to as XSPI driver and I was using the same convention as _xspi in this case. Will it be better to use the same or shall we proceed with _octospi?

buha: Maybe STM has very similar implementations for dual/quad/octo SPI's with very few register map differences and behavior, in which case it's better to make one stm32_xspi implementation that handles all of them, yes. But if they have wildly different implementations then it's better to keep them separate. I'm not very familiar with their offer.

drivers/platform/stm32/stm32_xspi.c
```diff
@@ -0,0 +1,718 @@
...
+	timeout = msgs->bytes_number;
+	while(timeout--) {
+		no_os_mdelay(1);
+		if (sdesc->stm32_xspi_dma_done)
+			break;
+	};
+
+	/* need some cleanup here? */
+	if (timeout == 0)
+		return -ETIME;
```
CiprianRegus, line 702: You should also call `no_os_dma_abort()` if you reach this point.

drivers/platform/stm32/stm32_xspi.h
```diff
@@ -0,0 +1,137 @@
...
+#include <stdint.h>
+#include <stdbool.h>
+#include "no_os_xspi.h"
+#include "stm32_hal.h"
+
+/**
+ * @struct stm32_xspi_command
+ * @brief Structure holding the stm32 platform specific XSPI command parameters.
+ */
+typedef struct stm32_xspi_command {
```
CiprianRegus, line 51: We don't allow `typedef`. Just use `struct stm32_xspi_command` when declaring a variable of this type (as you're already doing). Moreover, it doesn't look like you want to expose this type to the user. In this case, please move it to the source file.

drivers/platform/stm32/stm32_xspi.h
```diff
@@ -0,0 +1,137 @@
...
+	/** Get peripheral source clock function */
+	uint32_t (*get_input_clock)(void);
+	/** FIFO threshold */
+	uint32_t fifo_threshold;
+	/** Sample Shifting */
+	bool sample_shifting;
+	/** CS high time (in number of clock cycles) */
+	uint32_t cs_high_time;
+	/** DMA Initialization Parameters */
+	struct no_os_dma_init_param* dma_init;
```
CiprianRegus, line 95: `struct no_os_dma_init_param *dma_init;`

drivers/platform/stm32/stm32_xspi.c
```diff
@@ -0,0 +1,718 @@
...
+			sdesc->dma_ch->irq_num = sinit->irq_num;
+		}
+	}
+
+	*desc = xspi_desc;
+
+	return 0;
+
+error:
+	no_os_dma_remove(sdesc->dma_desc);
```
CiprianRegus, line 323: `sdesc` may not be allocated at this point if `xspi_desc` alloc fails. Also, please initialize all the pointers you're calling `free()` on here to `NULL`. It's undefined behavior to free a random pointer (which doesn't point to a memory region previously allocated).

drivers/platform/stm32/stm32_xspi.c
```diff
@@ -0,0 +1,718 @@
...
+	return 0;
+}
+
+/**
+ * @brief Configure XSPI with new parameters.
+ * @param desc - The XSPI descriptor
+ * @param params - The structure that contains the XSPI parameters.
+ * @return 0 in case of success, negative code otherwise.
+ */
+int32_t stm32_xspi_configure(struct no_os_xspi_desc *desc,
```
CiprianRegus, line 152: All the functions in this file should be `static`

include/no_os_spi.h
```diff
@@ -84,6 +84,21 @@ enum no_os_spi_bit_order {
 	NO_OS_SPI_BIT_ORDER_LSB_FIRST = 1,
 };
 
+/**
+ * @enum no_os_spi_lanes
+ * @brief SPI configuration for number of lanes.
+ */
+enum no_os_spi_lanes {
+	/** Single Lane */
+	NO_OS_SPI_SINGLE_LANE = 0,
```
buha, line 93: no need to specify their values

include/no_os_spi.h
```diff
@@ -187,7 +208,11 @@ struct no_os_spi_desc {
 	enum no_os_spi_mode	mode;
 	/** SPI bit order */
 	enum no_os_spi_bit_order	bit_order;
+	/** SPI Lanes */
+	enum no_os_spi_lanes   lanes;
+	/** SPI bus platform ops */
```
buha, line 213: these documentation strings should go in a separate commit, something like "document all no_os_spi structure members"

drivers/platform/stm32/stm32_xspi.c
```diff
@@ -5,36 +5,30 @@
 ********************************************************************************
  * Copyright 2024(c) Analog Devices, Inc.
  *
- * All rights reserved.
```
buha: please squash this commit into the commit that adds stm32_xspi implementation

since you add that implementation, add it with the proper license (3 clause BSD) from the start

buha (dismissed):

include/no_os_spi.h
```diff
@@ -233,6 +253,10 @@ int32_t no_os_spi_init(struct no_os_spi_desc **desc,
 /* Free the resources allocated by no_os_spi_init(). */
 int32_t no_os_spi_remove(struct no_os_spi_desc *desc);
 
+/* Configure the SPI peripheral */
+int32_t no_os_spi_config(struct no_os_spi_desc *desc,
+			 const struct no_os_spi_init_param *param);
+
```
CiprianRegus, line 266: Can you add more details about what this new function does and why is it needed? "Configure the SPI peripheral" is too vague and it doesn't look like you can actually reconfigure the fields defined in `struct no_os_spi_desc` (only those for xspi).

CiprianRegus (changes requested):

drivers/platform/stm32/stm32_xspi.c
```diff
@@ -0,0 +1,722 @@
...
+	struct stm32_xspi_desc *sdesc;
+	XSPI_RegularCmdTypeDef *cmd;
+
+	if (!desc || !desc->extra || !msgs)
+		return -EINVAL;
+
+	sdesc = desc->extra;
+	cmd = sdesc->command;
+
+	for (uint32_t i = 0; i < len; i++) {
```
CiprianRegus, line 380: For the standard SPI peripheral we're configuring the hardware registers before the transfer (based on the fields in the no os and STM32 specific SPI descriptors). I think we could do the same in this case and avoid the need to have the new `configure` API.

NagaHimanshu: The original intent for adding configure API is to reconfigure the vendor specific params without reinitializing the peripheral and creating descriptors again. If this is the case, instead of `no_os_spi_init_params` as argument, is it good to have `void *params` just to reconfigure vendor specific params?

If configuration need to be done, then user will use `no_os_spi_init` with new set of init params.

buha: I think you could do what @CiprianRegus says for now and drop it. A configure function would be very useful but not necessarily how it's implemented. The way I see it is that for each peripheral driver we could have:
- init() - allocate memory, calls config() internally
- config() - configure the peripheral
- remove() - deallocate memory (and put the hardware in a good state, if needed)

The user could then reconfigure the peripheral with a new set of init_param structure without deallocating/reallocating memory. 

But this is quite a big change and I don't think this PR should do such big changes, just drop the config function for now and we'll see about it in the future.

buha: PS: the idea was good, we even have an issue with the UART where we'd like to reconfigure the baudrate without reallocating it, like you proposed for this spi, it's just that we need to have a proper discussion about this internally

buha (approved):
Ciprian's request for changes was made, LGTM so I'm merging

CiprianRegus (approved):

PR #2320, opened 2024-09-30
Dummy read to clear any pending read on SPI

Shetty-Anush:
## Pull Request Description

Dummy read in spi dma callback function clears appropriate data register

## PR Type

## PR Checklist

### DIFF
--- a/drivers/platform/stm32/stm32_spi.c
+++ b/drivers/platform/stm32/stm32_spi.c
@@ -633,7 +633,12 @@ void stm32_spi_dma_callback(struct no_os_dma_xfer_desc *old_xfer,
 	sdesc->stm32_spi_dma_done = true;
 
 	/* Dummy read to clear any pending read on SPI */
+#ifndef SPI_SR_RXNE
+	*(volatile uint8_t *)&SPIx->RXDR;
+#else
 	*(volatile uint8_t *)&SPIx->DR;
+#endif
+
 	if (sdesc->stm32_spi_dma_user_cb)
 		sdesc->stm32_spi_dma_user_cb(sdesc->stm32_spi_dma_user_ctx);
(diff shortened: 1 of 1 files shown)


drivers/platform/stm32/stm32_spi.c
```diff
@@ -633,7 +633,12 @@ void stm32_spi_dma_callback(struct no_os_dma_xfer_desc *old_xfer,
 	sdesc->stm32_spi_dma_done = true;
 
 	/* Dummy read to clear any pending read on SPI */
+#ifndef SPI_SR_RXNE
```
buha, line 636: i see that the newer chips don't have this bit in the SR register so this is your method of distinguishing SPI peripherals for newer chips like H5 from the older chips like F4

could you add a comment on this line explaining this ?

buha (commented):
Also, drop the `1) ` from the commit description, a list with 1 item is not a list...

drivers/platform/stm32/stm32_spi.c
Shetty-Anush: The file already utilizes the SPI_SR_RXNE and SPI_SR_TXE bits to distinguish between the chips. Given this consistent usage, is the comment needed?

buha: i just saw what Ramona wrote here:
https://github.com/analogdevicesinc/no-OS/blame/57499efc27ba9980f3856ceee1dd629062a1c4d5/drivers/platform/stm32/stm32_spi.c#L387-L390

no need anymore :D

buha (approved):

buha:
Xilinx and STM32 CI builds fail, but for unrelated reasons so I'm merging.

PR #2326, opened 2024-10-03
STM32 DMA driver issue fix

Shetty-Anush:
## Pull Request Description

Channel number is assigned via the Instance for MCU families other than STM32F2, STM32F4 and STM32F7.
In this Pull Request, we can find changes made to support the DMA handlers of all MCUs.

## PR Type

## PR Checklist

### DIFF
--- a/drivers/platform/stm32/stm32_dma.c
+++ b/drivers/platform/stm32/stm32_dma.c
@@ -70,10 +70,10 @@ int stm32_dma_config_xfer(struct no_os_dma_ch *channel,
 
 	/* Note: Channel number is assigned via the Instance for MCU
 	 * families other than STM32F2, STM32F4 and STM32F7 */
-#if defined (STM32H5)
-	sdma_ch->hdma->Instance = sdma_ch->ch_num;
-#else
+#if defined (STM32F2) || defined (STM32F4) || defined (STM32F7)
 	sdma_ch->hdma->Init.Channel = sdma_ch->ch_num;
+#else
+	sdma_ch->hdma->Instance = sdma_ch->ch_num;
 #endif
 #if defined (STM32H5)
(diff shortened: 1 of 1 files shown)


drivers/platform/stm32/stm32_dma.c
```diff
@@ -70,10 +70,10 @@ int stm32_dma_config_xfer(struct no_os_dma_ch *channel,
 
 	/* Note: Channel number is assigned via the Instance for MCU
 	 * families other than STM32F2, STM32F4 and STM32F7 */
-#if defined (STM32H5)
-	sdma_ch->hdma->Instance = sdma_ch->ch_num;
-#else
+#if defined (STM32F2) || defined (STM32F4) || defined (STM32F7)
 	sdma_ch->hdma->Init.Channel = sdma_ch->ch_num;
+#else
+	sdma_ch->hdma->Instance = sdma_ch->ch_num;
```
buha, line 76: please move all this code in the #ifdef/#else just below it, there is no reason to have 2 separate #ifdefs for the same condition

Shetty-Anush: Channel number is assigned via the Instance for MCU families other than STM32F2, STM32F4 and STM32F7. This includes chips like H5 and L4 (the chip because of which the test failed previously). The code from line 79 to 119 is specific to H5 only and all the other MCUs will use the code of the #else directive.

` sdma_ch->hdma->Instance = sdma_ch->ch_num;`
If we include the above line with if defined (STM32H5) the channel number will not be assigned to chips like L4.

buha: ok fair enough

buha (approved):

PR #2329, opened 2024-10-04
iio: iio_app: don't overwrite status-code in iio_app_init() on the error path

commodo:
## Pull Request Description

Two changes done during the development of the `ad7606x-fmc` project

## PR Type

## PR Checklist

### DIFF
--- a/iio/iio_app/iio_app.c
+++ b/iio/iio_app/iio_app.c
@@ -410,8 +410,7 @@ error:
 
 	no_os_free(application);
 
-	status = print_uart_error_message(&uart_desc,
-					  &app_init_param.uart_init_params, status);
+	print_uart_error_message(&uart_desc, &app_init_param.uart_init_params, status);
 	no_os_uart_remove(uart_desc);
 
 	return status;


iio/iio_app/iio_app.c
```diff
@@ -79,31 +79,11 @@
...
-
-	status = no_os_uart_write(*uart_desc, (uint8_t *)message, msglen);
-	if (status < 0)
-		return status;
-
-	delay_ms = _calc_uart_xfer_time(msglen, UART_BAUDRATE_DEFAULT);
-	no_os_mdelay(delay_ms);
-
-	/** Reinitialize UART with parameters required by the IIO application */
-	no_os_uart_remove(*uart_desc);
```
buha, line 105: IIRC this was needed and is still needed because there is no API to change baudrate once the UART was initialized. And sometimes the iiod runs at higher baudrate than the iio greeting message. I think adding an UART baudrate change API would be the proper way to address this mess.

And because there was no API, we had to "needlessly" _remove() and _init() with a delay in between because on fast processors or slow UARTs or both, you'd see a partial iio greeting message.

commodo: so, if that's the case then i'd need to re-visit this

because the `no_os_uart_remove()` + `no_os_uart_init()` doesn't seem to initialize properly;
some init_data gets lost

amiclaus (approved):

PR #2335, opened 2024-10-11
drivers: dac: max22017 : Change driver name to max22017

RaduSabau1:
## Pull Request Description
Old driver name was max2201x, and now it has to be changed to max22017 since it will only support
the specific chip.

## PR Type

## PR Checklist


RaduSabau1:
V2:
- Fix broken dependencies

buha:
Screenshot from 2024-10-11 15-25-47

Why do the .h files don't show up as `oldfile -> newfile` ? also, the changes count is different (282 vs 289) so this commit does more than it says in the description.

I recommend using `git mv` command to rename files (or rename + move to a different directory)

buha:
Also, I don't see it necessary to have so many commits for this, I'd make 1 commit that renames everything. I find it easier when browsing git log to see only 1 rename commit instead of 5.

RaduSabau1:
Regarding the difference in code lines, that appears because I have removed the enum with the chip IDs, should I mention this in the commit body. Also should all the commits be squashed then?

RaduSabau1:
V3:
- Squashed all commits

amiclaus (approved):

buha:
```$ ag 2201X
drivers/adc/ad405x/README.rst
70:MAX2201X Driver Initialization Example
```

There is another occurrence in another device driver documentation. Could you please fix that too in a separate commit ?

buha:
actually i'll merge and do it myself in the github editor

PR #2344, opened 2024-10-23
 add spi read data correction feature for  sign extend and split status 

mi-zeng:
add a new function ad7606_data_correction_serial to correct spi read serial raw data for
1. sign extend for hardware/bipolar mode in case of negative value
2. split data and status information into 2 arrays.

## PR Type
- [x ] New feature (change that adds new functionality)

## PR Checklist
- [x ] I have followed the [Coding style guidelines](http://analogdevicesinc.github.io/no-OS/drivers_guide.html#coding-style)
- [x ] I have performed a self-review of the changes
- [x ] I have commented my code, at least hard-to-understand parts
- [x ] I have tested in hardware affected projects, at the relevant boards
- [x ] I have signed off all commits from this PR

### DIFF
--- a/drivers/adc/ad7606/ad7606.h
+++ b/drivers/adc/ad7606/ad7606.h
@@ -247,9 +247,9 @@ struct ad7606_range {
  */
 struct ad7606_oversampling {
 	/** Oversampling padding */
-	uint8_t os_pad:4;
+	uint8_t os_pad: 4;
 	/** Oversampling ratio */
-	enum ad7606_osr os_ratio:4;
+	enum ad7606_osr os_ratio : 4;
 };
 
 /**
@@ -400,5 +400,7 @@ int32_t ad7606_set_digital_diag(struct ad7606_dev *dev,
 				struct ad7606_digital_diag diag);
 int32_t ad7606_init(struct ad7606_dev **device,
 		    struct ad7606_init_param *init_param);
+int32_t ad7606_data_correction_serial(struct ad7606_dev *dev,
+				      uint32_t *buf, int32_t *data, uint8_t *status);
 int32_t ad7606_remove(struct ad7606_dev *dev);
 #endif /* AD7606_H_ */
--- a/drivers/adc/ad7606/ad7606.c
+++ b/drivers/adc/ad7606/ad7606.c
@@ -616,16 +616,20 @@ static int32_t cpy18b32b(uint8_t *psrc, uint32_t srcsz, uint32_t *pdst)
 	if (srcsz % 9)
 		return -EINVAL;
 
-	for(i = 0; i < srcsz; i += 9) {
+	for (i = 0; i < srcsz; i += 9) {
 		j = 4 * (i / 9);
-		pdst[j+0] = ((uint32_t)(psrc[i+0] & 0xff) << 10) | ((uint32_t)psrc[i+1] << 2)
-			    | ((uint32_t)psrc[i+2] >> 6);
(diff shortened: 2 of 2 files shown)


buha:
the astyle will sure fail, please run `ci/astyle.sh` script and amend your commit with what it proposes

also, make sure to put a whitespace after `//` like:
```
// comment
```
not
```
//comment
```

drivers/adc/ad7606/ad7606.h
```diff
@@ -391,5 +391,7 @@ int32_t ad7606_set_digital_diag(struct ad7606_dev *dev,
 				struct ad7606_digital_diag diag);
 int32_t ad7606_init(struct ad7606_dev **device,
 		    struct ad7606_init_param *init_param);
+int32_t ad7606_data_correction_serial(struct ad7606_dev *dev,
```
buha, line 394: i assume this new function is called after getting the samples with ad7606_read_samples(), i'd suggest to document this sequence somewhere

another alternative would be to make this new function private and propose a function that gets the data and performs the splitting into data and status buffers internally, perhaps called ad7606_read_data_status()

mi-zeng: hi,
  I prefer former solution because even creating new function, a new pointer for status buffer is still requires, this will change the parameter list of original data read function `ad7606_read_samples()`, which affects all exist project. I feel this is redundent for call 2 functions in a sequence but cannot figure out a better way. I believe this happens a lot during a Build-and-Fix develop mode.
 so where can I do the documentation? I find there is no readme or anything else for single device SDK, perhaps just point out in comments that could be called after read functions?

buha: So what I meant by the 2nd proposal is that you'd have both these functions:
```
ad7606_read_data() // existing, no changes
ad7606_read_data_status() // new

// private/static, inaccessible to the device driver user
ad7606_data_correction_serial()

// the new function makes data reading into data/status buffers "atomic" in the sense that it's only one operation and the user doesn't need to check for intermediate status codes or handle intermediate buffers
ad7606_read_data_status()
     call ad7606_read_data()
     calls ad7606_data_correction_serial() with the output of ad7606_read_data()
```

I hope this is more clear, let me know which approach you choose.

mi-zeng: I chose the 1nd suggestion and have added a note in the funciton description part, thx
image

buha:
So there are some code style changes that we did not enforce so far but will enforce from now on.
https://github.com/analogdevicesinc/no-OS/pull/2352

If you run the ci/astyle.sh from this pull request on your patch, you'll see that it will propose several code style changes. Please do so and resubmit, even though the new astyle.sh is not on main branch yet.

mi-zeng:
Hi,
 
  I have resubmited with new code style changes and the astyle config file.

best

buha:
Good, i see you applied the astyle config changes and ran it over your code, that's good.

But you included the astyle_config changes, they should not be in the commit, just locally on your machine (until we merge the other pull request).

buha:
please don't put the changelog in commit description, put it in a comment in the PR discussion
```
2024.11.11: resubmit with new code style changes without astyle_config change
```

also, what's with the ad3552r changes ?

mi-zeng:
I have made some mistakes with the commit, and trying to fix it right now.

mi-zeng:
11.11 resubmit with #2352 code style

buha (approved):

buha:
looks good now, and sorry for the inconvenience with astyle, from now on the new astyle config is on the main branch and you will not need to do these extra steps

PR #2355, opened 2024-11-07
add script to convert Windows to Unix paths

buha:
## Pull Request Description

When sourcing an environment script such as the Xilinx's settings64.sh, PATH entries need to be converted to Unix compatible paths. This script can be used to perform the conversion like this:

source tools/scripts/git-bash-paths.sh /c/Xilinx/Vitis/2022.2/settings64.sh

It's not necessarily specific to Xilinx platform but we're currently using it only for this purpose with possibility to extend it for other platforms.

Tested with Vitis 2022.2 on various projects for various targets such as zynq7000, microblaze, ultrascale.

## PR Type

## PR Checklist

### DIFF
--- a/tools/scripts/git-bash-paths.sh
+++ b/tools/scripts/git-bash-paths.sh
@@ -1,0 +1,54 @@
+#!/bin/bash
+# SPDX-License-Identifier: GPL-2.0+
+#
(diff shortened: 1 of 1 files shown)


buha:
rebase on newsest main because it contains some relevant changes

amiclaus (approved):

PR #2358, opened 2024-11-07
tools: generic.mk: re-generate object files if flags.txt files have changed

commodo:
## Pull Request Description

    In case CFLAGS,CPPFLAGS,ASFLAGS and OBJS lists change, their respective
    files will also change. This should re-trigger a build.
    
    The 'pre_build' rule is run via the 'update' build rule, which is now
    getting called everything we build 'all'.
    When the 'pre_build' gets run, we re-generate all flags.txt files, and
    overwrite them if they changed.
    When they change, the object file will rebuild.

## PR Type

## PR Checklist

### DIFF
--- a/tools/scripts/generic.mk
+++ b/tools/scripts/generic.mk
@@ -301,32 +301,41 @@ LSCRIPT_FLAG = -T$(LSCRIPT)
 endif
 
 define generate_cflags_func
-echo $(subst \",\\\",$(1)) >> $(CFLAGS_FILE)
+echo $(subst \",\\\",$(1)) >> $(CFLAGS_FILE).tmp
 endef
 
 define generate_cppflags_func
-echo $(subst \",\\\",$(1)) >> $(CPPFLAGS_FILE)
+echo $(subst \",\\\",$(1)) >> $(CPPFLAGS_FILE).tmp
 endef
 
 define generate_asflags_func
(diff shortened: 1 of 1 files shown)


buha:
Yeah this is good.

The most obvious improvement is when content creators keep adding files in src.mk to build up a project from scratch, they can now simply `make` instead of `make update; make`.

buha (dismissed):

buha:
i think we need testing on Windows tho, WIndows builds are not covered by CI and I just wanna make sure it doesn't break it

buha:
> i'm relatively confident that Make behaves similarly between Windows & Linux, but yeah: definitely worth testing;

me too but the whole thing about CFLAGS in files (instead of at command line) was because of this stupidly low windows command character limit and you don't know about it until you find about it the hard way :D

commodo:
pushed an update for fixing the AD7768-1FMCZ project.

looks like it's this HDL change that caused the no-OS project to fail:  https://github.com/analogdevicesinc/hdl/commit/136844ade13efc174571662cd360d5f0d523a2e1

i can drop it, if needed;
let's see if the CI passes at least;
then it can be discussed what to do with it;

commodo:
nevermind;
looks like a fix was pushed before me;

i've rebased;

Changelog v1 -> v2:
* rebased on newer main

amiclaus (approved):

PR #2352, opened 2024-11-04
ci: astyle_config: add padding options

amiclaus:
## Pull Request Description

Add padding options in order to ease the review process. The astyle build checks the following:

- add spaces between 'if', 'for', 'while' keywords and parens
- remove spaces inside parentheses
- Add space after commas

## PR Type
- [ ] Bug fix (change that fixes an issue)
- [x] New feature (change that adds new functionality)
- [ ] Breaking change (has dependencies in other repos or will cause CI to fail)

## PR Checklist
- [ ] I have followed the [Coding style guidelines](http://analogdevicesinc.github.io/no-OS/drivers_guide.html#coding-style)
- [ ] I have performed a self-review of the changes
- [ ] I have commented my code, at least hard-to-understand parts
- [ ] I have build all projects affected by the changes in this PR
- [x] I have tested in hardware affected projects, at the relevant boards
- [x] I have signed off all commits from this PR
- [ ] I have updated the documentation (wiki pages, ReadMe etc), if applies

PR #2351, opened 2024-11-04
Add support for LTC4162L

mrcsosa:
## Pull Request Description

Please replace this with a detailed description and motivation of the changes. 
You can tick the checkboxes below with an 'x' between square brackets or just check them after publishing the PR. 
If this PR contains a breaking change, list dependent PRs and try to push all related PRs at the same time.

## PR Type
- [ ] Bug fix (change that fixes an issue)
- [x] New feature (change that adds new functionality)
- [ ] Breaking change (has dependencies in other repos or will cause CI to fail)

## PR Checklist
- [x] I have followed the [Coding style guidelines](http://analogdevicesinc.github.io/no-OS/drivers_guide.html#coding-style)
- [x] I have performed a self-review of the changes
- [x] I have commented my code, at least hard-to-understand parts
- [ ] I have build all projects affected by the changes in this PR
- [x] I have tested in hardware affected projects, at the relevant boards
- [x] I have signed off all commits from this PR
- [ ] I have updated the documentation (wiki pages, ReadMe etc), if applies

PR #2361, opened 2024-11-09
add ad5293 drivers 

mi-zeng:
## Pull Request Description
drivers includes
1. no-os initilization
2. calibration, wiper r/w ,hard/soft reset and other operation
3. daisy-chain multiple chip support

has formatted according to #2352 code style with astyle.sh

## PR Type

## PR Checklist


amiclaus (commented):
there are some build issues that need addressing in the first place. have a look at https://github.com/analogdevicesinc/no-OS/pull/2351 for example on how documentation should be integrated.

mi-zeng:
Hi,
  I have ref to #2351, modified related sphinx source file and a `make html` is build successful, please take a look.
best

drivers/potentiometer/ad5293/ad5293.c
```diff
@@ -0,0 +1,531 @@
...
+	// spi initialization
+	init_param->spi_init.mode = NO_OS_SPI_MODE_1; //CPOL=0, CPHA=1
+	ret = no_os_spi_init(& dev->spi_desc, & init_param->spi_init);
+	if (ret < 0) {
+		ret = - EIO;
+		goto error;
+	}
+
+	*device = dev;
+	printf("Initializing succeed.\r\n ");
```
buha, line 123: no need to put these messages in the device driver

drivers/potentiometer/ad5293/ad5293.c
```diff
@@ -0,0 +1,531 @@
...
+#include "no_os_delay.h"
+
+/***************************************************************************//**
+ * @brief Reset ad5293 chip info
+ *
+ * Reset ad5293 chip info to default situation after hard/soft reset
+ *
+ * @param dev - The device structure.
+ *
+ * @return NAN
```
buha, line 46: void

drivers/potentiometer/ad5293/ad5293.c
```diff
@@ -0,0 +1,531 @@
...
+
+	dev->chip = chp;
+
+	// reset all chips
+	ret = ad5293_hard_reset(dev);
+	if (ret < 0)
+		goto error;
+
+	// spi initialization
+	init_param->spi_init.mode = NO_OS_SPI_MODE_1; //CPOL=0, CPHA=1
```
buha, line 115: the comment is redundant, but as you wish

drivers/potentiometer/ad5293/ad5293.c
```diff
@@ -0,0 +1,531 @@
...
+			goto error;
+		}
+	}
+
+	// chip individual
+	dev->chip_num = init_param->chip_num;
+	chp = (struct ad5293_chip_info*)no_os_malloc(dev->chip_num * sizeof(* chp));
+	if (! chp) {
+		ret = - ENOMEM;
+		goto error;
```
buha, line 104: if you jump from here to `error` label, execution will attempt to free the unallocated `chp` pointer which is undefined behavior

drivers/potentiometer/ad5293/ad5293.c
```diff
@@ -0,0 +1,531 @@
...
+	int32_t ret = 0;
+
+	if ((! device) || (! init_param))
+		return - EINVAL;
+
+	dev = (struct ad5293_dev*)no_os_malloc(sizeof(* dev));
+	if (! dev)
+		return - ENOMEM;
+
+	printf("Initializing ad5293, number of chips %d.\r\n", init_param->chip_num);
```
buha, line 87: remove all printf statements from the device driver, they only belong to the application

you may use debug pr_* statements from no_os_print_log.h, but not printf directly

drivers/potentiometer/ad5293/ad5293.c
```diff
@@ -0,0 +1,531 @@
...
+		buf[2 * i] = (uint8_t)(CMD_SDO_Z >> 8);
+		buf[2 * i + 1] = (uint8_t)CMD_SDO_Z;
+	}
+
+	ret = no_os_spi_write_and_read(dev->spi_desc, & buf[0], sz);
+	if (ret < 0)
+		return ret;
+
+	memset(& buf, 0, sz);
+	ret = no_os_spi_write_and_read(dev->spi_desc, & buf[0], sz);
```
buha, line 254: you can return on the same line here

drivers/potentiometer/ad5293/ad5293.c
```diff
@@ -0,0 +1,531 @@
...
+	uint8_t i;
+	uint16_t sz = 2 * dev->chip_num;
+	uint8_t buf[sz];
+
+	for (i = 0; i < dev->chip_num; i++) {
+		buf[2 * i] = (uint8_t)(SHUTDOWN_FRAME(sd) >> 8);
+		buf[2 * i + 1] = (uint8_t)SHUTDOWN_FRAME(sd);
+	}
+
+	ret = no_os_spi_write_and_read(dev->spi_desc, & buf[0], sz);
```
buha, line 284: same here, return on the same line, no need for `ret` variable, please apply to all if there are other instances

drivers/potentiometer/ad5293/ad5293.h
```diff
@@ -0,0 +1,149 @@
...
+
+struct ad5293_dev {
+	/* SPI */
+	struct no_os_spi_desc	*spi_desc;
+	/* GPIO */
+	struct no_os_gpio_desc	*gpio_reset;
+	/* number of chips */
+	uint16_t chip_num;
+	/* pointer of chip struct  */
+	struct ad5293_chip_info* chip; //point to chip 0
```
buha, line 105: please document better the fact that this is an array and which instance it's used (daisy chain?)

drivers/potentiometer/ad5293/ad5293.h
```diff
@@ -0,0 +1,149 @@
...
+};
+
+struct ad5293_init_param {
+	/* SPI */
+	struct no_os_spi_init_param spi_init;
+	/* GPIO */
+	struct no_os_gpio_init_param* gpio_reset;
+	/* number of chips */
```
buha, line 113: same as above, is this the number of chips in daisy chain config ? 

i'd prefer it if the default uninitialized value of 0 would imply 1 chip

drivers/potentiometer/ad5293/ad5293.c
```diff
@@ -0,0 +1,531 @@
...
+ * @param dev - The device structure.
+ * @param cali - calibration mode selection.
+ * @param num - chip indication.
+ *
+ * @return nan
+ *******************************************************************************/
+void ad5293_update_cali(struct ad5293_dev* dev, enum calibration_mode_t cali,
+			uint8_t num)
+{
+	(dev->chip + num)->cali = cali;
```
buha, line 303: test num value, it should not be greatir than dev->chip_num, otherwise the API user can start overwriting memory with junk

drivers/potentiometer/ad5293/ad5293.h
```diff
@@ -0,0 +1,149 @@
...
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE.
+*******************************************************************************/
+
+#ifndef _ad5293_H_
```
buha, line 29: lower/upper case mix

drivers/potentiometer/ad5293/ad5293.h
```diff
@@ -0,0 +1,149 @@
...
+#define _ad5293_H_
+/*****************************************************************************/
```
buha, line 31: delete these comment delimiters, they are completely useless

i know we have a lot of them in our code but we should get rid of them

drivers/potentiometer/ad5293/ad5293.h
```diff
@@ -0,0 +1,149 @@
...
+};
+
+/* operation mode */
+enum operation_mode_t {
+	BUS,
+	DAISY_CHAIN
+};
+
+/* SDO mode */
+enum SDO_mode_t {
```
buha, line 78: lower/uppser case mix

drivers/potentiometer/ad5293/README.rst
```diff
@@ -0,0 +1,159 @@
...
+- Low resolution DAC replacements
+- Sensor calibration
+
+Device Configuration
+--------------------
+
+1.  Serial peripheral interface (SPI) is necessary to access device
+2.  Daisy-chain mode is required for multiple chip use, in this configuration 
+    chip select line remains low untill all bytes are shifted out
+3.  High impedence mode of SDO pin is not recommanded,
```
buha, line 28: this sentence needs rephrasing

drivers/potentiometer/ad5293/README.rst
```diff
@@ -0,0 +1,159 @@
...
+Device Configuration
+--------------------
+
+1.  Serial peripheral interface (SPI) is necessary to access device
+2.  Daisy-chain mode is required for multiple chip use, in this configuration 
+    chip select line remains low untill all bytes are shifted out
+3.  High impedence mode of SDO pin is not recommanded, 
+    as tested somehow this could cause communications 
+    unfunctional
+
```
buha, line 31: also i think more details are needed in the documentation

i just searched a random device driver .rst to use as template:
https://github.com/analogdevicesinc/no-OS/blob/main/drivers/adc/adm1177/README.rst

drivers/potentiometer/ad5293/ad5293.c
buha: this applies to many places where you do this pointer arithmetic

drivers/potentiometer/ad5293/ad5293.c
```diff
@@ -0,0 +1,531 @@
...
+ *                  -EIO - SPI communication error.
+ *                  -EIO - GPIO initialization error.
+ *                  0 - No errors encountered.
+ *******************************************************************************/
+int32_t ad5293_init(struct ad5293_dev **device,
+		    struct ad5293_init_param* init_param)
+{
+	struct ad5293_dev* dev;
+	struct ad5293_chip_info* chp;
+	int32_t ret = 0;
```
amiclaus, line 78: no need for explicit initialization.

drivers/potentiometer/ad5293/ad5293.c
```diff
@@ -0,0 +1,531 @@
...
+
+	dev = (struct ad5293_dev*)no_os_malloc(sizeof(* dev));
+	if (! dev)
+		return - ENOMEM;
+
+	printf("Initializing ad5293, number of chips %d.\r\n", init_param->chip_num);
+
+	// hardware reset pin initialization
+	if (init_param->gpio_reset) {
+		ret |= no_os_gpio_get_optional(& dev->gpio_reset, init_param->gpio_reset);
```
amiclaus, line 91: check ret after each assignment. avoid propagating errors with `|=`

drivers/potentiometer/ad5293/ad5293.c
```diff
@@ -0,0 +1,531 @@
...
+		ret |= no_os_gpio_set_value(dev->gpio_reset, NO_OS_GPIO_HIGH);
+		if (ret < 0) {
```
amiclaus, line 93: `if (ret)` check all occurences.

drivers/potentiometer/ad5293/ad5293.h
```diff
@@ -0,0 +1,149 @@
...
+/*****************************************************************************/
+/****************************** Include Files ********************************/
+/*****************************************************************************/
+#include <stdint.h>
+#include "no_os_gpio.h"
+#include "no_os_spi.h"
+#include "no_os_util.h"
+
+/* SPI Read/Write commands */
+#define CMD_NOP        0b0000
```
amiclaus, line 40: go for hex values here entirely.

drivers/potentiometer/ad5293/ad5293.h
```diff
@@ -0,0 +1,149 @@
...
+	struct no_os_spi_init_param spi_init;
+	/* GPIO */
+	struct no_os_gpio_init_param* gpio_reset;
+	/* number of chips */
+	uint16_t chip_num;
+};
+
+/*******************************functions prototype*******************************/
+/* Initialize the ad5293 device structure. */
+int32_t ad5293_init(struct ad5293_dev **device,
```
amiclaus, line 119: all functions should return `int`

drivers/potentiometer/ad5293/ad5293.c
mi-zeng: will add explicity initilazation to pointer dev and chp as NULL, then it won't affect even if chp is not allocated a address

drivers/potentiometer/ad5293/ad5293.c
mi-zeng: will remove `= 0`

drivers/potentiometer/ad5293/ad5293.c
mi-zeng: will check all function

drivers/potentiometer/ad5293/ad5293.c
mi-zeng: will add parameter num check in case of point to unallocated memory

drivers/potentiometer/ad5293/ad5293.h
mi-zeng: I would say value 0 imply 1 is fine for array, but 0 imply for 1 chip is kinda anti-intuitive. I would chose do a better documentation that explains this well. Do you insist on that?
Daisy chain is only supported for multiple chips, because I cant figure a clearly/comprehensive way to switch CS line configuration on spi_desc( especially when port changed).
In the way I wrote single and daisy chain mode are compatible, only difference is several more bytes needs to be set for SPI tranfer buffer, I think this is better.

drivers/potentiometer/ad5293/ad5293.h
mi-zeng: Do I have to return an error code? 
I returned data in last 2 get operation functions

mi-zeng:
11.17 
several change according to review
1. reorgnize document README.rst, add more description
2. fix void function return description
3. turn `ret` initialization as non-explicit
4. remove printf information
5. check `ret` for each ret assignment, simplify `if(ret<0)` as `if(ret)`
6. `chp` init as `NULL` to prevent undefined memory release behavior
7. simplify return and function return in same line
8. fix lower/upper case mix
9. remove redundant delimiters
10. marco define turn form binary to hex

mi-zeng:
any build error information from LinuxBuilds cppcheck?

buha:
here's the log if you dont have access to it:
```
drivers/potentiometer/ad5293/ad5293.c:149:6: error: Uninitialized variable: ret [uninitvar]
 if (ret)
     ^
drivers/potentiometer/ad5293/ad5293.c:313:9: error: Uninitialized variable: ret [uninitvar]
 return ret;
        ^
drivers/potentiometer/ad5293/ad5293.c:338:9: error: Uninitialized variable: ret [uninitvar]
 return ret;
        ^
```

drivers/potentiometer/ad5293/ad5293.c
buha: that works too

drivers/potentiometer/ad5293/ad5293.h
buha: My reasoning is one may use this device driver with 1-chip only, no daisy chain configuration needed, so they shouldn't bother to populate the ad5293_init_param.chip_num. So chip_num could be interpreted at 2 levels:
1. `if (chip_num)`, easy test that you will likely do in a few places in the device driver
2. chip_num value should mean the number of devices in daisy chain (document this as you proposed!): 0 - don't bother because of the test, 1 - you could treat it as invalid in the init() as it doesn't make sense, >=2 - use them to compute how many bytes need to be transferred

mi-zeng: Oh, I guess we have a misunderstanding here, since you are worrying 1 chip use. Let me explain clearer, 
1. if chip num is 1, hardware schamatic is normal SPI device topology. In init function 1 chip info struct space is allocated  for chip info. In drivers functions, sz indicate size of tranfer buffer size, so 2 is set. For communications 2 bytes will be shifted out.

memory allocate
>     chp = (struct ad5293_chip_info*)no_os_malloc(dev->chip_num * sizeof(* chp));
buffer array declare and size compute
> 	uint16_t sz = 2 * dev->chip_num;
> 	uint8_t buf[sz];
> 
> 	for (i = 0; i < dev->chip_num; i++) {
> 		buf[2 * i] = (uint8_t)(AD5293_R_WIPER_FRAME >> 8);
> 		buf[2 * i + 1] = (uint8_t)(AD5293_R_WIPER_FRAME);
> 	}
2. if chip num greater than 1,  hardware schamatic is required to be daisy chain. In init function `num` chip info space will be allocated for multiple chip. Driver functions will calculate proper size for data buffer, which is `2*num`, and there is no need to change anything because `no_os_spi_write_and_read` api keep CS line low until all `2*num` bytes are shifted.

The thing needs to change is the `chip_num` parameter for init struct and proper hardware topology, all programs work for both 1 chip or multiple chips set without necessary testing in operation functions. This is what meant by compatible. To avoid confusing I will add more description on README.

mi-zeng:
11.18
1. fix build error related code
2. add chip num check in init function
3. add struct member memory free in initilazation error part
4. add single chip setup description

mi-zeng:
BTY I found the access to check logs, thanks :>

amiclaus (approved):

PR #2363, opened 2024-11-11
Add IIO support for AD4170

machschmitt:
## Pull Request Description

The reported issue is the following: "The AD4170 drivers don't do a sign extension when reading the ADC registers, as the ADC is 24 bits, but the int variable is 32-bit, it fills the 4th bit with zeroes even if bit 23 is 1."
While that could be fixed in the base driver by checking the bipolar field of AFE register in `ad4170_get_ch_data()`, I believe it's better to support the bipolar/unipodar data configuration through IIO so truly raw data access is preserved through the base driver channel read function.

Add support for basic IIO support that will output `_raw` data that is properly signed according to the bipolar/unipolar AFE configuration.

## PR Type

## PR Checklist


machschmitt:
The fix for the data sign extension was implemented in the precision-converters-firmware repo (https://github.com/analogdevicesinc/precision-converters-firmware).
PR: https://github.com/analogdevicesinc/precision-converters-firmware/pull/105

PR #2368, opened 2024-11-18
Staging/adf4368

skucukar:
The ADF4368 is a high performance, ultra-low jitter, integer-N and fractional-N phase-locked loop (PLL) with integrated VCO ideally suited for frequency conversion applications.

The high performance PLL has a figure of merit of −239 dBc/Hz, very low 1/f noise of normalized −287 dBc/Hz and high PFD frequency that can achieve ultra-low in-band noise and integrated jitter. The ADF4368 can generate any frequency from 800 MHz to 12.8 GHz without an internal doubler, which eliminates the need for sub-harmonic filters. The Σ-Δ modulator includes a 25-bit fixed modulus that allows hertz frequency resolution and an additional 17-bit variable modulus, which allows even finer resolution and flexibility for frequency planning. The 9 dBm output power at 12.8 GHz in single-ended configuration with 16 step power adjust feature makes it very useful for any application.

For multiple frequency conversion applications, such as phase array radar or massive MIMO systems, the outputs of multiple ADF4368 can be aligned by using the SYNC input or EZSync™. The EZSync method is used when it is difficult to distribute the SYNC signal to all devices precisely. For applications that require deterministic delay or delay adjustment capability, a programmable reference to output delay with <1 ps resolution is provided. The reference to output delay is guaranteed across multiple devices and temperature, allowing for predictable and precise multichip alignment.

The simplicity of the ADF4368 block diagram eases development time with a simplified serial-peripheral interface (SPI) register map, external SYNC input, and repeatable multichip phase alignment both in integer mode and fractional mode.

## Pull Request Description

## PR Type

## PR Checklist

### DIFF
--- a/drivers/frequency/adf4368/adf4368.c
+++ b/drivers/frequency/adf4368/adf4368.c
@@ -1,0 +1,1992 @@
+/***************************************************************************//**
+ *   @file   adf4368.c
+ *   @brief  Implementation of ADF4368 Driver.
+ *   @author Sirac Kucukarabacioglu (sirac.kucukarabacioglu@analog.com)
+********************************************************************************
+ * Copyright 2024(c) Analog Devices, Inc.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice,
+ *    this list of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * 3. Neither the name of Analog Devices, Inc. nor the names of its
+ *    contributors may be used to endorse or promote products derived from this
+ *    software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY ANALOG DEVICES, INC. “AS IS” AND ANY EXPRESS OR
+ * IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF
+ * MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO
+ * EVENT SHALL ANALOG DEVICES, INC. BE LIABLE FOR ANY DIRECT, INDIRECT,
+ * INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA,
+ * OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF
+ * LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING
+ * NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE,
+ * EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+*******************************************************************************/
+
+#include "adf4368.h"
+#include "no_os_alloc.h"
+#include "no_os_delay.h"
+#include "no_os_error.h"
+#include "no_os_print_log.h"
+
+/**
+ * @struct adf4368_reg_sequence
+ * @brief ADF4368 register format structure for default values
+ */
+static struct adf4368_reg_sequence {
+	uint16_t reg;
+	uint8_t val;
+};
+
+/**
+ * @struct adf4368_reg_defaults
+ * @brief ADF4368 register initialization
+ */
+static const struct adf4368_reg_sequence adf4368_reg_defaults[] = {
+
+	{ 0x054, 0x00 },
+	{ 0x053, 0x25 },
+	{ 0x052, 0x00 },
+	{ 0x051, 0x00 },
+	{ 0x050, 0x00 },
+	{ 0x04f, 0x00 },
+	{ 0x04e, 0x10 },
+	{ 0x04d, 0x00 },
+	{ 0x04c, 0x2B },
+	{ 0x04b, 0x5D },
+	{ 0x04a, 0x00 },
+	{ 0x048, 0x00 },
+	{ 0x047, 0x00 },
+	{ 0x046, 0x00 },
+	{ 0x045, 0x08 },
+	{ 0x044, 0x18 },
+	{ 0x043, 0x09 },
+	{ 0x042, 0x09 },
+	{ 0x041, 0x00 },
+	{ 0x040, 0x00 },
+	{ 0x03f, 0x83 },
+	{ 0x03e, 0x26 },
+	{ 0x03d, 0xC0 },
+	{ 0x03c, 0x00 },
+	{ 0x03b, 0x8C },
+	{ 0x03a, 0x00 },
+	{ 0x039, 0xB0 },
+	{ 0x038, 0x00 },
+	{ 0x037, 0x3E },
+	{ 0x036, 0xD6 },
+	{ 0x035, 0x04 },
+	{ 0x034, 0x99 },
+	{ 0x033, 0x32 },
+	{ 0x032, 0xD3 },
+	{ 0x031, 0x69 },
+	{ 0x030, 0x1B },
+	{ 0x02f, 0xA7 },
+	{ 0x02e, 0x00 },
+	{ 0x02d, 0xF1 },
+	{ 0x02c, 0x4C },
+	{ 0x02b, 0x03 },
+	{ 0x02a, 0x10 },
+	{ 0x029, 0xDD },
+	{ 0x028, 0x20 },
+	{ 0x027, 0x28 },
+	{ 0x026, 0x80 },
+	{ 0x025, 0x00 },
+	{ 0x024, 0x00 },
+	{ 0x023, 0x80 },
+	{ 0x022, 0x00 },
+	{ 0x021, 0x09 },
+	{ 0x020, 0xC1 },
+	{ 0x01f, 0x1E },
+	{ 0x01e, 0x28 },
+	{ 0x01d, 0xDD },
+	{ 0x01c, 0xFF },
+	{ 0x01b, 0xFF },
+	{ 0x01a, 0xFF },
+	{ 0x019, 0x55 },
+	{ 0x018, 0x55 },
+	{ 0x017, 0x55 },
+	{ 0x016, 0x00 },
+	{ 0x015, 0x00 },
+	{ 0x014, 0x43 },
+	{ 0x013, 0x55 },
+	{ 0x012, 0x55 },
+	{ 0x011, 0x00 },
+	{ 0x010, 0x2B },
+};
+
+/* Charge pump current values expressed in uA */
+static const int adf4368_ci_ua[] = {
+	790,
+	990,
+	1190,
+	1380,
+	1590,
+	1980,
+	2390,
+	2790,
+	3180,
+	3970,
+	4770,
+	5570,
+	6330,
+	7910,
+	9510,
+	11100
+};
+
+/**
+ * @brief Writes data to ADF4368 over SPI.
+ * @param dev	   - The device structure.
+ * @param reg_addr - The register address.
+ * @param data 	   - Data value to write.
+ * @return 	   - 0 in case of success or negative error code otherwise.
+ */
+int adf4368_spi_write(struct adf4368_dev *dev, uint16_t reg_addr, uint8_t data)
+{
+	uint8_t buff[ADF4368_BUFF_SIZE_BYTES];
+	uint16_t cmd;
+
+	if (!dev)
+		return -EINVAL;
+
+	cmd = ADF4368_SPI_WRITE_CMD | reg_addr;
+	if (dev->spi_desc->bit_order) {
+		buff[0] = no_os_bit_swap_constant_8(cmd & 0xFF);
+		buff[1] = no_os_bit_swap_constant_8(cmd >> 8);
+		buff[2] = no_os_bit_swap_constant_8(data);
+	} else {
+		buff[0] = cmd >> 8;
+		buff[1] = cmd & 0xFF;
+		buff[2] = data;
+	}
+
+	return no_os_spi_write_and_read(dev->spi_desc, buff,
+					ADF4368_BUFF_SIZE_BYTES);
+}
+
+/**
+ * @brief Reads data from ADF4368 over SPI.
+ * @param dev 	   - The device structure.
+ * @param reg_addr - The register address.
+ * @param data	   - Data read from the device.
+ * @return	   - 0 in case of success or negative error code otherwise.
+ */
+int adf4368_spi_read(struct adf4368_dev *dev, uint16_t reg_addr, uint8_t *data)
+{
+	uint8_t buff[ADF4368_BUFF_SIZE_BYTES];
+	uint16_t cmd;
+	int ret;
+
+	if (!dev)
+		return -EINVAL;
+
+	cmd = ADF4368_SPI_READ_CMD | reg_addr;
+	if (dev->spi_desc->bit_order) {
+		buff[0] = no_os_bit_swap_constant_8(cmd & 0xFF);
+		buff[1] = no_os_bit_swap_constant_8(cmd >> 8);
+		buff[2] = no_os_bit_swap_constant_8(ADF4368_SPI_DUMMY_DATA);
+	} else {
+		buff[0] = cmd >> 8;
+		buff[1] = cmd & 0xFF;
+		buff[2] = ADF4368_SPI_DUMMY_DATA;
+	}
+
+	ret = no_os_spi_write_and_read(dev->spi_desc, buff,
+				       ADF4368_BUFF_SIZE_BYTES);
+	if (ret)
+		return ret;
+
+	if (dev->spi_desc->bit_order)
+		*data = no_os_bit_swap_constant_8(buff[2]);
+	else
+		*data = buff[2];
+
+	return 0;
+}
+
+/**
+ * @brief Updates the values of the ADF4368 register.
+ * @param dev 	   - The device structure.
+ * @param reg_addr - The register address.
+ * @param mask 	   - Bits to be updated.
+ * @param data 	   - Update value for the mask.
+ * @return	   - 0 in case of success or negative error code otherwise.
+ */
+int adf4368_spi_update_bits(struct adf4368_dev *dev, uint16_t reg_addr,
+			    uint8_t mask, uint8_t data)
+{
+	uint8_t tmp, orig;
+	int ret;
+
+	ret = adf4368_spi_read(dev, reg_addr, &orig);
+	if (ret)
+		return ret;
+
+	tmp = orig & ~mask;
+	tmp |= data & mask;
+
+	if (tmp != orig)
+		return adf4368_spi_write(dev, reg_addr, tmp);
+
+	return 0;
+}
+
+/**
+ * @brief Will output on the terminal the values of all the ADF4368 registers.
+ * @param dev 	- The device structure.
+ * @return 	- 0 in case of success or negative error code.
+ */
+int adf4368_reg_dump(struct adf4368_dev *dev)
+{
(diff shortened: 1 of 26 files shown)


projects/adf4368/src/platform/platform_includes.h
```diff
@@ -0,0 +1,51 @@
...
+ * NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE,
+ * EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+*******************************************************************************/
+#ifndef __PLATFORM_INCLUDES_H__
+#define __PLATFORM_INCLUDES_H__
+
+/******************************************************************************/
+/***************************** Include Files **********************************/
+/******************************************************************************/
+#ifdef STM32_PLATFORM
```
amiclaus, line 39: does the project have stm32 support? if so include it in the builds.json file too.

skucukar: It doesn't support the STM32 platform.

amiclaus: If it doesn't support the STM32 platform then remove all the related additions within code and also the .ioc file.

drivers/frequency/adf4368/adf4368.c
```diff
@@ -0,0 +1,1949 @@
...
+	cmd = ADF4368_SPI_WRITE_CMD | reg_addr;
+	if (dev->spi_desc->bit_order) {
+		buff[0] = no_os_bit_swap_constant_8(cmd & 0xFF);
+		buff[1] = no_os_bit_swap_constant_8(cmd >> 8);
+		buff[2] = no_os_bit_swap_constant_8(data);
+	} else {
+		buff[0] = cmd >> 8;
+		buff[1] = cmd & 0xFF;
+		buff[2] = data;
+	}
```
CiprianRegus, line 84: I'm a bit confused by this. If the user of the driver is setting the SPI bit order why wouldn't they only use MSB (or what is specified in the datasheet)? In that case you only need to handle one bit order setting.

drivers/frequency/adf4368/adf4368.h
```diff
@@ -0,0 +1,661 @@
...
+#define MHZ					MEGA
+#define S_TO_NS					NANO
+#define NS_TO_PS				KHZ_PER_MHZ
+#define PS_TO_NS				KHZ_PER_MHZ
+#define PS_TO_US				MICRO
+#define US_TO_MS				MILLI
+#define MS_TO_S					MILLI
+#define PS_TO_S					PICO
+#define US_TO_FS				NANO
+#define NS_TO_S					NANO
```
CiprianRegus, line 416: These names are generic enough to create future collision with other macros. If you really need to use them, please add the `ADF4368_` prefix.

drivers/frequency/adf4368/adf4368.h
```diff
@@ -0,0 +1,661 @@
...
+	uint64_t			freq_min;
+	uint64_t			ref_freq_hz;
+	uint64_t			freq;
+};
+
+/**
+ * @struct reg_sequence
+ * @brief ADF4368 register format structure for default values
+ */
+struct reg_sequence {
```
CiprianRegus, line 465: You can make this static. Better yet to move it and the default values to the driver source file. Also, aren't these register values automatically loaded after a reset.

drivers/frequency/adf4368/adf4368.c
skucukar: Hi Ciprian, 

I checked this feature. This code snippet is correct but there are other bugs. 
In the "mbed_spi_init" and "no_os_spi_init" functions there is no place to set spi desc->bit_order. Its always 0.
Also in the driver while I am reading I should also swap the bits. I will fix the driver bug.

drivers/frequency/adf4368/adf4368.h
skucukar: This values are the recommended register values for given initial state. After a reset chip's default register values are different.

Struct definition is not static but struct array itself is static in the header file.

I wanted to make the driver similar to adf4382 driver but if its better in the source file I can . 
You mean static struct reg_sequence in the header file and  move the struct array to source file, right?

drivers/frequency/adf4368/adf4368.h
skucukar: Thank you! Fixed

projects/adf4368/src/platform/platform_includes.h
skucukar: Thank you! Fixed

drivers/frequency/adf4368/adf4368.c
skucukar: Fixed

skucukar:
V2

- added  "ADF4368_" prefix to calculations 
- modified struct reg_sequence to static struct reg_sequence
- Removed STM32_PLATFORM related additions and .ioc file
- Fix Driver Read - LSB mode

drivers/frequency/adf4368/adf4368.c
```diff
@@ -0,0 +1,1953 @@
...
+
+#include "adf4368.h"
+#include "no_os_alloc.h"
+#include "no_os_delay.h"
+#include "no_os_error.h"
+#include "no_os_print_log.h"
+
+/* Charge pump current values expressed in uA */
+static const int adf4368_ci_ua[] = {
+	790,
```
amiclaus, line 42: you can shrink this in 2 lines or so.

drivers/frequency/adf4368/adf4368.c
```diff
@@ -0,0 +1,1953 @@
...
+		buff[1] = cmd & 0xFF;
+		buff[2] = ADF4368_SPI_DUMMY_DATA;
+	}
+
+	ret = no_os_spi_write_and_read(dev->spi_desc, buff,
+				       ADF4368_BUFF_SIZE_BYTES);
+	if (ret)
+		return ret;
+
+	if (dev->spi_desc->bit_order) {
```
amiclaus, line 123: drop brackets.

drivers/frequency/adf4368/adf4368.c
```diff
@@ -0,0 +1,1953 @@
...
+		if (mod2_tmp > mod2_max) {
+			channel_spacing *= 5;
+		} else {
+			mod2_wd = mod2_tmp;
+			break;
+		}
+
+	} while (channel_spacing < ADF4368_CHANNEL_SPACING_MAX);
+
+	if (!en_phase_resync) {
```
amiclaus, line 392: drop brackets.

drivers/frequency/adf4368/adf4368.c
```diff
@@ -0,0 +1,1953 @@
...
+ * @param dev 	     - The device structure.
+ * @return 	     - 0 in case of success or negative error code.
+ */
+static int adf4368_optimize_ldwin(struct adf4368_dev *dev)
+{
+	uint64_t pfd_freq;
+	uint8_t ldwin_pw;
+	uint8_t tmp;
+	int ret;
+
```
amiclaus, line 480: drop extra line

drivers/frequency/adf4368/adf4368.c
```diff
@@ -0,0 +1,1953 @@
...
+int adf4368_get_ref_clk(struct adf4368_dev *dev, uint64_t *val)
+{
+	if (!dev)
+		return -EINVAL;
+
+	*val = dev->ref_freq_hz;
+
+	return 0;
+}
+
```
amiclaus, line 657: drop extra line

drivers/frequency/adf4368/adf4368.c
```diff
@@ -0,0 +1,1953 @@
...
+	ret = adf4368_spi_read(dev, 0x20, &tmp);
+	if (ret)
+		return ret;
+
+	dev->ref_div = no_os_field_get(ADF4368_R_DIV_MSK, tmp);
+	*div = dev->ref_div;
+
+	return 0;
+}
+
```
amiclaus, line 745: ditto.

drivers/frequency/adf4368/adf4368.c
```diff
@@ -0,0 +1,1953 @@
...
+		} else {
+			ret = adf4368_optimize_bleed_word(dev);
+			if (ret)
+				return ret;
+		}
+	}
+
+	// Need to set N_INT last to write Double Buffered Registers
+	val = dev->n_int & ADF4368_N_INT_LSB_MSK;
+	ret = adf4368_spi_write(dev, 0x10, val);
```
amiclaus, line 803: return directly.

drivers/frequency/adf4368/adf4368.c
```diff
@@ -0,0 +1,1953 @@
...
+			return ret;
+
+		ret = adf4368_optimize_ldwin(dev);
+		if (ret)
+			return ret;
+	}
+
+	// Need to set N_INT last to write Double Buffered Registers
+	val = dev->n_int & ADF4368_N_INT_LSB_MSK;
+	ret = adf4368_spi_write(dev, 0x10, val);
```
amiclaus, line 891: return directly.

drivers/frequency/adf4368/adf4368.c
```diff
@@ -0,0 +1,1953 @@
...
+		      no_os_field_prep(ADF4368_DNCLK_DEL_MSK, delay);
+		ret = adf4368_spi_update_bits(dev, 0x30,
+					      ADF4368_DRCLK_DEL_MSK | ADF4368_DNCLK_DEL_MSK, val);
+
+		if (ret)
+			return ret;
+
+	} else {
+		/* Power Down SYNC */
+		ret = adf4368_spi_update_bits(dev, 0x2A, ADF4368_PD_SYNC, 0xff);
```
amiclaus, line 1122: this return value is lost.

drivers/frequency/adf4368/adf4368.c
```diff
@@ -0,0 +1,1953 @@
...
+		return -EINVAL;
+
+	ret = adf4368_spi_read(dev, 0x1F, &tmp);
+	if (ret)
+		return ret;
+	*sw_sync = no_os_field_get(tmp, ADF4368_SW_SYNC_MSK);
+
+	return 0;
+}
+
```
amiclaus, line 1192: drop line

drivers/frequency/adf4368/adf4368.c
```diff
@@ -0,0 +1,1953 @@
...
+	if (int_mode) {
+		ret = adf4368_set_phase_bleedi(dev, phase_fs, phase_pol);
+		if (ret)
+			return ret;
+	} else {
+		// Only Positive Phase
+		if (phase_pol)
+			return -EINVAL;
+		ret = adf4368_set_phase_sdm(dev, phase_fs);
```
amiclaus, line 1469: return directly

amiclaus (commented):
looks good overall. small comments on my side.

drivers/frequency/adf4368/adf4368.c
skucukar: Bracket after "if" is forced with /ci/astyle.sh

drivers/frequency/adf4368/adf4368.c
skucukar: Bracket after "if" is forced with /ci/astyle.sh

drivers/frequency/adf4368/adf4368.c
skucukar: Fixed

drivers/frequency/adf4368/adf4368.c
skucukar: Fixed

drivers/frequency/adf4368/adf4368.c
skucukar: Fixed

drivers/frequency/adf4368/adf4368.c
skucukar: Fixed

drivers/frequency/adf4368/adf4368.c
skucukar: Fixed

drivers/frequency/adf4368/adf4368.c
skucukar: Fixed

drivers/frequency/adf4368/adf4368.c
skucukar: Fixed

drivers/frequency/adf4368/adf4368.c
skucukar: Fixed

drivers/frequency/adf4368/adf4368.c
skucukar: Fixed

drivers/frequency/adf4368/adf4368.c
skucukar: Not changed for readability

skucukar:
V3

- removed extra lines
- added direct returns 
- removed brackets for single line if statements

drivers/frequency/adf4368/iio_adf4368.c
```diff
@@ -0,0 +1,1199 @@
...
+
+	if (!iio_adf4368)
+		return -EINVAL;
+
+	adf4368 = iio_adf4368->adf4368_dev;
+
+	if (!adf4368)
+		return -EINVAL;
+
+	iio_parse_value(buf, IIO_VAL_INT, &val, NULL);
```
buha, line 379: i know this is not going to fail as it is now, but please check the return code of this function everywhere you use it, somebody may add failure points to it in the future and it can silently fail if it's not properly checked

drivers/frequency/adf4368/iio_adf4368.c
```diff
@@ -0,0 +1,1199 @@
...
+	END_ATTRIBUTES_ARRAY
+};
+
+
+
+static struct iio_attribute adf4368_iio_ch_attrs[] = {
+	{
+		.name = "en",
+		.shared = IIO_SEPARATE,
+		.show = adf4368_iio_read_en,
```
buha, line 1100: instead of doing 4 different read and 4 different write functions for these attributes you could've done 1 single read and 1 write function

please see how it was done here:
https://github.com/analogdevicesinc/no-OS/blob/cc73a3d70d056d143b783bd846489e2d2819ba7c/drivers/rf-transceiver/hmc630x/iio_hmc630x.c#L326-L366

i know there are lots of places that define a function for an attribute (i've done it too) but we should not be doing it like that anymore, it just takes 200 lines of code to write something that could be done in 50

projects/adf4368/src.mk
```diff
@@ -0,0 +1,45 @@
...
+	$(DRIVERS)/api/no_os_timer.c	\
+        $(DRIVERS)/api/no_os_dma.c      \
+        $(NO-OS)/util/no_os_fifo.c      \
+	$(NO-OS)/util/no_os_mutex.c     \
+	$(NO-OS)/util/no_os_lf256fifo.c \
+        $(NO-OS)/util/no_os_list.c      \
+        $(NO-OS)/util/no_os_util.c      \
+        $(NO-OS)/util/no_os_alloc.c
+
+INCS += $(INCLUDE)/no_os_delay.h     \
```
buha, line 27: you could use `NO_OS_INCS_DIR += $(INCLUDE)` to replace everything from line 27 to 42

buha (commented):
solid work, just please see about my other comments and i think it'll be good to merge

projects/adf4368/src.mk
skucukar: Fixed

drivers/frequency/adf4368/iio_adf4368.c
skucukar: Fixed and tested

drivers/frequency/adf4368/iio_adf4368.c
skucukar: Fixed

skucukar:
V4

- Added iio channel attr read-write: IIO Channel attributes combined in to one function 
- Added missing error checks

drivers/frequency/adf4368/iio_adf4368.c
```diff
@@ -0,0 +1,1087 @@
...
+	};
+
+	if (ret)
+		return ret;
+
+	return 0;
+}
+
+static struct iio_attribute adf4368_iio_attrs[] = {
+	{
```
buha, line 919: could you please to the same for device attributes like you did for channel attributes ? 2 functions that manage all these device attributes, it will result in much more compact code.

adf4368_iio_read_device_attr()
adf4368_iio_write_device_attr()

sorry i didn't see this last pass...

buha: and just for completeness, it's not even mandatory to use single function for all, but they just seem so similar that in this case i think it justifies a single function

you could even have a scenario where 1 attribute is so different than the others that you may wanna process it in its own function and have the others be treated in the main function with the big switch case, that's fine too...

projects/adf4368/src.mk
```diff
@@ -0,0 +1,47 @@
...
+	$(DRIVERS)/api/no_os_timer.c	\
+        $(DRIVERS)/api/no_os_dma.c      \
+        $(NO-OS)/util/no_os_fifo.c      \
+	$(NO-OS)/util/no_os_mutex.c     \
+	$(NO-OS)/util/no_os_lf256fifo.c \
+        $(NO-OS)/util/no_os_list.c      \
+        $(NO-OS)/util/no_os_util.c      \
+        $(NO-OS)/util/no_os_alloc.c
+
+INCS += $(NO_OS_INCS_DIR)/no_os_delay.h     \
```
buha, line 29: you did this above:
```
NO_OS_INCS_DIR += $(INCLUDE)
```

so delete all thise section, line 29 to 44

projects/adf4368/Makefile
```diff
@@ -0,0 +1,12 @@
+# Uncomment to use the desired platform
+PLATFORM = mbed
+
+# Select the example you want to enable by chossing y for enabling and n for disabling
+BASIC_EXAMPLE = n
```
buha, line 5: (commenting something that i commented on another PR that is being reviewed right now)

i don't think editing Makefiles is good to set the example to build even though we have existing projects done this way... let's strive to make the user interaction easy so that they run one of these commands to choose the example to build:

```
make EXAMPLE=basic_example
make EXAMPLE=iio_example
```

Basically you could default to basic like this:
```
EXAMPLE ?= basic_example
```

And take action based on it:
```
ifeq ($(INTERFACE), iio_example)
...
endif
```

projects/adf4368/README.rst
```diff
@@ -0,0 +1,428 @@
...
+
+In order to build the basic example make sure you have the following
+configuration in the Makefile
+`Makefile <https://github.com/analogdevicesinc/no-OS/tree/master/projects/adf4368/Makefile>`_
+
+.. code-block:: bash
+
+	# Select the example you want to enable by choosing y for enabling and n for disabling
+	BASIC_EXAMPLE = y
+	IIO_EXAMPLE = n
```
buha, line 371: this will need to change when you switch to the `make EXAMPLE=basic_example` approach

drivers/frequency/adf4368/iio_adf4368.c
skucukar: or I can make separated snprintf for every case

projects/adf4368/src.mk
skucukar: when I delete them it does not compile

projects/adf4368/Makefile
skucukar: If its possible I would like to continue with traditional approach. Our very similar product ADF4382 uses the traditional approach. So we want them to be really similar for the customer experience.

skucukar: Fixed

drivers/frequency/adf4368/iio_adf4368.c
skucukar: Fixed

drivers/frequency/adf4368/adf4368.h
skucukar: structure and array relocated to source file.

skucukar:
V5

- Add Device Attributes into one function
- Changed Makefile as describe 
- Bug Fix mbed_spi missing bit_order parameter
- Fix Project README
- Relocated default regs to adf4368.c from adf4368.h

projects/adf4368/src.mk
buha: we discussed in private this is not applicable here because there is a potentially a bug about NO_OS_INC_DIRS on mbed platform

buha (approved):

buha:
to be merged when CI passes

PR #2370, opened 2024-11-19
projects: eval-pqmon: add serial and t1l impl as iio interface

rbudai98:
By choosing the proper defines the user can choose between different implementation for connections.

## Pull Request Description

Allowing user to choose between the desired interface type for iio connection. Possible options are:
* USB
* SERIAL (not complete, documented)
* T1L 

## PR Type

## PR Checklist


buha (approved):

buha:
I was thinking we need a common approach to this in the future because there are now several projects which can do both usb/serial/ethernet and an easier way to select them would be to literally specify with something like:

```
export INTERFACE=serial
export INTERFACE=usb
export INTERFACE=ethernet
```

rbudai98:
> I was thinking we need a common approach to this in the future because there are now several projects which can do both usb/serial/ethernet and an easier way to select them would be to literally specify with something like:
> 
> ```
> export INTERFACE=serial
> export INTERFACE=usb
> export INTERFACE=ethernet
> ```

Should we add this variable to `generic.mk` ?

rbudai98:
v2:
* added INTERFACE as general variable that defines the connection for iio

buha (approved):

PR #2373, opened 2024-11-22
add polling to tcp and ssl layer

jeanADI:
## Pull Request Description

Building on this commit: https://github.com/analogdevicesinc/no-OS/commit/8ca2a15b8b7cc7a53985bf5a773bdd99f4bc63bc#diff-f9781ad4653f9192f0521846b4116a5b34ae3c2abd2c4256ee0a813e3f947aa4R158

Currently, all devices (ADIN1110 and ADIN2111) which may be used with
the LWIP sockets layer require polling for new frames.

This polling was already added before socket_recv() in the mqtt layer, now adding it to the tcp and ssl layers. 

## PR Type

## PR Checklist

### DIFF
--- a/network/tcp_socket.c
+++ b/network/tcp_socket.c
@@ -41,6 +41,7 @@
 #include "tcp_socket.h"
 #include "no_os_util.h"
 #include "no_os_alloc.h"
+#include "no_os_delay.h"
 
 #ifndef DISABLE_SECURE_SOCKET
 #include "noos_mbedtls_config.h"
@@ -104,6 +105,26 @@ static int tls_net_recv(struct tcp_socket_desc *sock, unsigned char *buff,
 {
 	int32_t ret;
 
+#ifdef NO_OS_LWIP_NETWORKING
+	/*
+	 * Currently, the LWIP networking layer doesn't implement packet RX
+	 * using interrupts, so we have to poll.
+	 *
+	 * Timeout defaults to 500ms
+	 */
+	int i = 0;
+	do {
+		ret = no_os_lwip_step(sock->net->net, NULL);
+
+		if (ret == 0) {
+			break;
+		}
+
+		no_os_mdelay(1);
+		i++;
+	} while (i < LWIP_POLLING_TIMEOUT_MS);
+
+#endif /* NO_OS_LWIP_NETWORKING */
 	ret = sock->net->socket_recv(sock->net->net, sock->id, buff, len);
 	if (ret == -EAGAIN)
 		return MBEDTLS_ERR_SSL_WANT_READ;
--- a/network/lwip_raw_socket/lwip_socket.c
+++ b/network/lwip_raw_socket/lwip_socket.c
@@ -873,7 +873,27 @@ static int32_t lwip_socket_connect(void *net, uint32_t sock_id,
 
 	pcb = socket->pcb;
 
-	return tcp_connect(pcb, &ipaddr, addr->port, lwip_connect_callback);
+	ret =  tcp_connect(pcb, &ipaddr, addr->port, lwip_connect_callback);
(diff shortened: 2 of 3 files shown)


network/tcp_socket.c
```diff
@@ -339,7 +353,21 @@ int32_t socket_connect(struct tcp_socket_desc *desc,
...
+		 * Adding this as a temporary workaround, based off this commit: https://github.com/analogdevicesinc/no-OS/commit/8ca2a15b8b7cc7a53985bf5a773bdd99f4bc63bc#diff-f9781ad4653f9192f0521846b4116a5b34ae3c2abd2c4256ee0a813e3f947aa4R158
+		 * 
+		 * Previously this polling was only applied to the mqtt layer
+		 */
+	int i = 500;
+	while(i>0){
+		no_os_lwip_step(desc->net->net, NULL);
+		i--;
+	}
+#endif /*NO_OS_LWIP_NETWORKING*/
```
CiprianRegus, line 370: Instead of doing this, I'd say we move this sequence after: https://github.com/analogdevicesinc/no-OS/blob/main/network/lwip_raw_socket/lwip_socket.c#L876

Also, you'll need to add a delay (1ms for example) between `no_os_lwip_step()` calls. This way we could configure the timeout.

CiprianRegus:
Hi @jeanADI

A few things which you'll have to change:
- Sign the CLA posted above by @CLAassistant
- Add a commit message in which you explain what is wrong and how your solution works. We prefer to have this in the commit message instead of the PR description because it's easier to check the reason for a commit that way.
- Sign the commit (see other commit messages in this repository, e.g https://github.com/analogdevicesinc/no-OS/commit/054757a19bfbdca5b40a86b04e7816c3bf7ded07 ).
- Code indentation looks off, please fix it.

network/tcp_socket.c
jeanADI: Thank you, I've now used the no_os_delay method and moved this sequence to the lwip_socket_connect function. My project still builds and runs correctly with the change

CiprianRegus:
Code changes due to review suggestions should be applied to the original commit, instead of creating a new one. So you should squash the 2 commits and keep the description of the second one.

Also, `no_os_lwip_step()` should be called until the connection is established or when the timeout (500 ms in your case) expires. I would also define a macro for the timeout, so it can be configured by the user.

kister-jimenez:
Hi @jeanADI , please format your commit message similar to commit [054757](https://github.com/analogdevicesinc/no-OS/commit/054757a19bfbdca5b40a86b04e7816c3bf7ded07) suggested in the previous comment.

It's easier to know what will be affected based on the subject if you format it like "network: ...". 

If possible, use 50 characters for the subject and wrap the commit body lines every 72 characters. You may also want to move some of the comments in the code you added to this commit message body.

CiprianRegus (approved):

CiprianRegus:
Please run `./ci/astyle.sh` locally to fix the CI issue.

jeanADI:
> Please run `./ci/astyle.sh` locally to fix the CI issue.

Hi Ciprian, I've run astyle yesterday and I believe reformatted my new code as a result. However, when I did it also edited much of the `tcp_socket.c` and  `lwip_socket.c` files, I didn't add these changes. I'm not sure if this is why the tests or failing or if it is the new code - I cannot find detailed test output to see where astyle is failing.

CiprianRegus:
This commit https://github.com/analogdevicesinc/no-OS/commit/19084c7abea7e9fc644e8a57acad9a7e1c382cfa was merged to main. Can you please rebase your branch and push again?

buha:
closing for now, feel free to reopen when you have a fix

PR #2374, opened 2024-11-23
tools: generic.mk: fix race-y builds with 'flags.txt' files generation

commodo:
## Pull Request Description

In PR https://github.com/analogdevicesinc/no-OS/pull/2358 a functionality was added to regenerate the 'flags.txt' files. But under some parallel builds, this seems to have a few race-conditions, where some files aren't generated when they're needed.

This changeset tries to fix this.

## PR Type

## PR Checklist


buha:
i think it'd be best to squash them into 1 good commit

buha (approved):

PR #2375, opened 2024-11-23
tools: generic.mk: add NO_OS_INC_DIRS build var for pulling header files tools

commodo:
## Pull Request Description

This is inspired by the SRC_DIRS build variable, but only focuses on
include files. Using the SRC_DIRS can be problematic, as it pulls
every .c & .h file from a directory.
    
But in some cases, we just want to pull in the header files. These header
files will only get compiled if a .c file pulls them in.
    
If this build variable is unspecified, it will be empty, so nothing will be
done (to potentially affect current projects).

## PR Type

## PR Checklist

### DIFF
--- a/projects/ad7606x-fmc/src.mk
+++ b/projects/ad7606x-fmc/src.mk
@@ -1,16 +1,25 @@
 # Project includes + source code
-INCS := $(PROJECT)/src/parameters.h      \
-        $(PROJECT)/src/common_data.h
 SRCS := $(PROJECT)/src/common_data.c
+
+NO_OS_INC_DIRS := \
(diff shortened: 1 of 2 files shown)


commodo:
Changelog v1 -> v2:
* renamed INC_DIRS -> NO_OS_INC_DIRS ;  Mbed seems to have an INC_DIRS variable and this seems to collide with it in some builds

amiclaus (approved):

commodo:
Changelog v2 - v3:
- rebased on newer main
- squashed commits into one

buha (approved):

PR #2385, opened 2024-12-02
Add uart_map in MAX32665 uart extras.

cencarna:
## Pull Request Description

This adds max_uart_map in the init_param of the MAX32665 UART driver.
By default, the console UART for the MAX32665FTHR uses MAP_B, while the
EvKit_V1's uses MAP_A based on their corresponding BSPs [here](https://github.com/analogdevicesinc/msdk/blob/main/Libraries/Boards/MAX32665/FTHR/Source/board.c#L98) and [here](https://github.com/analogdevicesinc/msdk/blob/main/Libraries/Boards/MAX32665/EvKit_V1/Source/board.c#L248).

The sys_map_t is undocumented in the MSDK docs but it appears it sets
specific GPIO configs which varies with different board form factors according
to [here](https://github.com/analogdevicesinc/msdk/blob/main/Libraries/PeriphDrivers/Source/UART/uart_me14.c#L64). This change will allow users to access EvKit_V1 UART at the application
level without touching the platform drivers.

Furthermore, this introduces the new enum max_uart_map that sets MAP_B
as default to avoid breaking changes with adjacent projects that uses the
MAX32665FTHR board.

## PR Type

## PR Checklist

### DIFF
--- a/drivers/platform/maxim/max32660/maxim_uart.c
+++ b/drivers/platform/maxim/max32660/maxim_uart.c
@@ -244,6 +244,7 @@ static int32_t max_uart_init(struct no_os_uart_desc **desc,
 {
 	int32_t ret;
 	int32_t stop, size, flow, parity;
+	sys_map_t map;
 	mxc_uart_regs_t *uart_regs;
 	struct max_uart_init_param *eparam;
 	struct no_os_uart_desc *descriptor;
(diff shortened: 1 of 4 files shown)


cencarna:
Astyle fails but this is already outside my changes. Let me know if this should be fixed in this pull request.

buha:
regarding the astyle issue, yeah we know where it's coming from and we'll come up with a PR to fix it

you will then have to rebase on main once the fix is merged so that the error disappears

buha:
can you please rebase now on `main` branch ? the astyle error should disappear

buha (dismissed):

buha:
does this need to be added to the other maxim targets as well ? or is it a feature of max32665 specifically ?

cencarna:
changelog v2:

- Dropped UART_MAP_DEFAULT, first enum member (which maps to zero) is now set as default
- Added similar changes for MAX32660 as well. [This](https://github.com/analogdevicesinc/msdk/blob/main/Libraries/PeriphDrivers/Source/UART/uart_me11.c#L83) suggests the device supports three maps.

buha (approved):

PR #2388, opened 2024-12-03
Dev/ade7880

raezt:
## Pull Request Description

ADE7880 Basic driver with SPI communication and example for maxim platform. 

## PR Type

## PR Checklist

### DIFF
--- a/drivers/meter/ade7880/ade7880.c
+++ b/drivers/meter/ade7880/ade7880.c
@@ -1,0 +1,478 @@
+/***************************************************************************//**
+ *   @file   ade7880.c
+ *   @brief  Implementation of ADE7880 Driver.
+ *   @author REtz (radu.etz@analog.com)
+********************************************************************************
+ * Copyright 2024(c) Analog Devices, Inc.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice,
+ *    this list of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * 3. Neither the name of Analog Devices, Inc. nor the names of its
+ *    contributors may be used to endorse or promote products derived from this
+ *    software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY ANALOG DEVICES, INC. “AS IS” AND ANY EXPRESS OR
+ * IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF
+ * MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO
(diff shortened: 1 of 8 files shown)


drivers/meter/ade7880/ade7880.c
```diff
@@ -0,0 +1,487 @@
...
+			    uint8_t *status)
+{
+	int ret;
+	/* register value read */
+	uint32_t reg_val;
+
+	if (!dev)
+		return -ENODEV;
+	if (!status)
+		return -EINVAL;
```
CiprianRegus, line 244: You could skip these checks, as they are already done in the `ade7880_read()` function.

drivers/meter/ade7880/ade7880.c
```diff
@@ -0,0 +1,487 @@
...
+
+	// Lock the SPI port as active
+	ret = ade7880_update_bits(dev, ADE7880_REG_CONFIG2, ADE7880_I2C_LOCK, ENABLE);
+
+	/* Use a valid register with default value different from 0 */
+	ret = ade7880_read(dev, ADE7880_REG_CFMODE, &chip_id);
+	if (ret)
+		goto error_spi;
+
+
```
CiprianRegus, line 396: Remove one empty line

drivers/meter/ade7880/ade7880.c
```diff
@@ -0,0 +1,487 @@
...
+		goto error_dev;
+
+	/* Hard reset the device */
+	if (!init_param.reset_desc)
+		goto error_dev;
+	dev->reset_desc = init_param.reset_desc;
+	ret = no_os_gpio_set_value(dev->reset_desc,
+				   NO_OS_GPIO_LOW);
+	if (ret)
+		return ret;
```
CiprianRegus, line 357: Not likely to fail, but you'll have to free `dev` if `ret != 0`.

drivers/meter/ade7880/ade7880.c
```diff
@@ -0,0 +1,487 @@
...
+		goto error_dev;
+	dev->ss_desc = init_param.ss_desc;
+	for (int i = 0; i < 3; i++) {
+		ret = no_os_gpio_set_value(dev->ss_desc,
+					   NO_OS_GPIO_LOW);
+		no_os_mdelay(1);
+		ret = no_os_gpio_set_value(dev->ss_desc,
+					   NO_OS_GPIO_HIGH);
+		no_os_mdelay(1);
+	}
```
CiprianRegus, line 381: The datasheet specifies two ways of selecting the SPI as MCU interface:

```
First, use the SS/HSA pin of the master device (that is, the
microcontroller) as a regular I/O pin and toggle it three times.
Second, execute three SPI write operations to a location in the
address space that is not allocated to a specific ADE7880 register
(for example 0xEBFF, where eight bit writes can be executed).
These writes allow the SS/HSA pin to toggle three times.
```
As far as I understand, you could also issue 3 SPI transactions instead of controlling the SS pin as a GPIO, which would simplify things a bit, since the user doesn't need configure SS a GPIO prior to calling `ade7880_init()`.

raezt: True. I changed the SPI select method to the one using 3 consecutive write operations.

raezt:
v2 change log:

- all comments were addressed.
- additionally changed the comment at l83 (l85 v1) in platform.h from "Port and Pin for PSM1" to "Port and Pin for Reset"

amiclaus (approved):

amiclaus (commented):
we should have though separate commits for driver/project documentation.

PR #2389, opened 2024-12-05
Change MAX32670 UART to Initialize with APB CLK

Brandon-Hurst:
## Pull Request Description

MSDK board support packages and all other MAX32 No-OS drivers use APB CLK to initialize the UART, while MAX32670 uses IBRO. This caused issues printing to stdio/serial port on a project for me, so I'm submitting a fix to change this to APB CLK to match other implementations. This fix solved my problems with printing.

## PR Type

## PR Checklist

### DIFF
--- a/drivers/platform/maxim/max32670/maxim_uart.c
+++ b/drivers/platform/maxim/max32670/maxim_uart.c
@@ -307,7 +307,7 @@ static int32_t max_uart_init(struct no_os_uart_desc **desc,
 		goto error;
 	}
 
-	ret = MXC_UART_Init(uart_regs, descriptor->baud_rate, MXC_UART_IBRO_CLK);
+	ret = MXC_UART_Init(uart_regs, descriptor->baud_rate, MXC_UART_APB_CLK);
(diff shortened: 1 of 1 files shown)


CiprianRegus:
Please sign your commit and wrap the message to 75 characters. Other than that, it looks good.

buha:
it bugs me that i don't have enough context to understand why is it like this

that IBRO clock should give better error rate and reduce to near zero the possibility of junk characters and weird UART issues, so why changing it to a regular 1000 multiple clock works better is beyond me

anyway, for the sake of homogenous code i'm merging it as is because it definitely works for the other platforms with the APB clock

buha (approved):

Brandon-Hurst:
Thanks, I agree. There's some things to speculate it _could_ be. UART has some strict jitter requirements for the receiver to recover the clock -- that could be part of it. Could also be a SW issue of initialization or perhaps the voltage supply rail for the IBR0 has an issue. I'm also bugged by it, but don't have much time to look into it either right now.

PR #2390, opened 2024-12-05
Fix AD5421 driver bugs

Brandon-Hurst:
## Pull Request Description

AD5421 driver had a bug when integrating into a custom example:
- SPI R/W status checks check against an expected length, rather than 0/-1 which is what the No-OS API comments indicate it should return.
- Very minor: I added the "No-Op" code from AD5421 datasheet (0x09). Currently the driver sends 0x00 which should suffice for reading data back, but I added 0x09 to match what the AD5421 GUI sends.

## PR Type

## PR Checklist


CiprianRegus (approved):

CiprianRegus:
Please sign your commit and wrap the message to 75 characters. Other than that, it looks good.

amiclaus (approved):

PR #2398, opened 2024-12-12
Staging/adf4382 dev

Jude-Osems:
## Pull Request Description
This is to publish new features to the existing ADF4382 No-OS driver. The features include:
- EZSync and Timed Sync.
- Fast Calibration.
- Change frequency for lock time measurement.
- Update to README, describing the latest features implemented.

## PR Type

## PR Checklist

### DIFF
--- a/drivers/frequency/adf4382/adf4382.c
+++ b/drivers/frequency/adf4382/adf4382.c
@@ -506,41 +506,6 @@ int adf4382_get_en_chan(struct adf4382_dev *dev, uint8_t ch, bool *en)
 }
 
 /**
- * @brief Set the sync to enable or disable based on the passed parameter. If
- * the parameter is different then 0 it will set the doubler to enable.
- * @param dev 		- The device structure.
- * @param en	 	- The enable or disable sync.
- * @return    		- Result of the writing procedure, error code otherwise.
- */
-int adf4382_set_en_sync(struct adf4382_dev *dev, bool en)
-{
-	uint8_t enable;
-
-	enable = no_os_field_prep(ADF4382_PD_SYNC_MSK, en);
-	return adf4382_spi_update_bits(dev, 0x2A, ADF4382_PD_SYNC_MSK, enable);
-}
-
-/**
- * @brief Gets the value the sync if it is enabled or disable.
- * @param dev 		- The device structure.
- * @param en	 	- The read status of the sync enable.
- * @return    		- 0 in case of success or negative error code.
- */
-int adf4382_get_en_sync(struct adf4382_dev *dev, bool *en)
-{
-	uint8_t tmp;
-	int ret;
-
-	ret = adf4382_spi_read(dev, 0x2A, &tmp);
-	if (ret)
-		return ret;
-
-	*en = no_os_field_get(tmp, ADF4382_PD_SYNC_MSK);
-	return 0;
-
-}
-
-/**
  * @brief Set the desired output frequency and reset everything over to maximum
  * supported value of 22GHz (21GHz for ADF4382A) to the max. value and
  * everything under the minimum supported value of 687.5MHz (2.875GHz for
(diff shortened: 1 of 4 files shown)


drivers/frequency/adf4382/adf4382.h
```diff
@@ -588,7 +622,7 @@ static const struct reg_sequence adf4382_reg_defaults[] = {
 	{ 0x03c, 0x00 },
 	{ 0x03b, 0x00 },
 	{ 0x03a, 0xFA },
-	{ 0x039, 0x00 },
+	{ 0x039, 0x80 },
```
buha, line 625: This defaults are not really defaults at reset, they seem to be some custom all-purpose config, more like a template. In this case O_VCO_DB is set which is 0 at reset. Dont change anything for now, my comment is more about the way things were done before, the whole adf4382_reg_defaults thing is not okay in my opinion, all features should be configurable via config APIs, not like this.

drivers/frequency/adf4382/adf4382.c
```diff
@@ -506,38 +506,499 @@ int adf4382_get_en_chan(struct adf4382_dev *dev, uint8_t ch, bool *en)
...
+ * @param en_fast_cal	- Enables the fast calibration routine.
+ * @return		- N_INT value corresponding to minimum VCO frequency for
+ * 		      fast calibration LUT generation.
+ */
+int adf4382_set_en_fast_calibration(struct adf4382_dev *dev, bool en_fast_cal)
+{
+	if (!dev)
+		return -EINVAL;
+	if (en_fast_cal == 1) {
+		uint64_t min_vco_frequency;
```
buha, line 563: move these declarations at the top of the function

drivers/frequency/adf4382/adf4382.c
```diff
@@ -506,38 +506,499 @@ int adf4382_get_en_chan(struct adf4382_dev *dev, uint8_t ch, bool *en)
...
+ * @param dev 	     - The device structure.
+ * @param en_fast_cal	- Enables the fast calibration routine.
+ * @return		- N_INT value corresponding to minimum VCO frequency for
+ * 		      fast calibration LUT generation.
+ */
+int adf4382_set_en_fast_calibration(struct adf4382_dev *dev, bool en_fast_cal)
+{
+	if (!dev)
+		return -EINVAL;
+	if (en_fast_cal == 1) {
```
buha, line 562: i would suggest to 
```
if (en_fast_cal != 1)
    return /* whatever necessary */;

// place the rest of code at this indentation
```

drivers/frequency/adf4382/adf4382.c
```diff
@@ -506,38 +506,499 @@ int adf4382_get_en_chan(struct adf4382_dev *dev, uint8_t ch, bool *en)
...
 
-	enable = no_os_field_prep(ADF4382_PD_SYNC_MSK, en);
-	return adf4382_spi_update_bits(dev, 0x2A, ADF4382_PD_SYNC_MSK, enable);
+		val = no_os_field_prep(ADF4382_LUT_WR_ADDR_MSK, 31) |
+		      no_os_field_prep(ADF4382_O_VCO_LUT_MSK, 1);
+		ret = adf4382_spi_write(dev, 0x200, val);
+		if (ret)
+			return ret;
+		dev->en_lut_cal = en_fast_cal;
+		adf4382_spi_update_bits(dev, 0x36, ADF4382_EN_LUT_CAL_MSK,
```
buha, line 918: status code is not checked

drivers/frequency/adf4382/iio_adf4382.c
```diff
@@ -870,10 +1188,46 @@ static struct iio_attribute adf4382_iio_attrs[] = {
 		.store = adf4382_iio_write_divider,
 	},
 	{
-		.name = "sync_en",
+		.name = "sw_sync",
+		.shared = IIO_SHARED_BY_ALL,
+		.show = adf4382_iio_read_sw_sync,
```
buha, line 1193: By creating a dedicated function for each .show of each attribute you create a lot of code duplication. Sure, the code will work as-is, but it will be a maintenance hell with so much duplication.

See the discussion here with suggestions on how to change the code.
https://github.com/analogdevicesinc/no-OS/pull/2368#discussion_r1883928847

drivers/frequency/adf4382/adf4382.h
Jude-Osems: Does this mean we should create a separate configurable API for setting default register values and expose it to IIO? Is there a usecase why this should be? I thought we initialize the registers when we call the adf5611_init function on power up?

CiprianRegus:
Please sign your commits and provide a description for each one (in the commit message). This should include what is done and the reason for the change. Also the commit titles are a bit misleading, since the driver already exists and this PR seems to add new functionality instead.

Jude-Osems:
> Please sign your commits and provide a description for each one (in the commit message). This should include what is done and the reason for the change. Also the commit titles are a bit misleading, since the driver already exists and this PR seems to add new functionality instead.

Jude-Osems:
> Please sign your commits and provide a description for each one (in the commit message). This should include what is done and the reason for the change. Also the commit titles are a bit misleading, since the driver already exists and this PR seems to add new functionality instead.

@CiprianRegus could you provide an instance of the accepted commit message? eg "drivers: frequency: adf4382: Update driver support for ADF4382" or "drivers: frequency: adf4382: Add Fast calibration feature". Is the commits feature based in this case or divided as is but with "update" as the key word? Please clarify, thanks.

drivers/frequency/adf4382/iio_adf4382.c
Jude-Osems: I see. this is carried forward based on an old implementation. I only expanded on the feature. Let me know if this has to change now, or we can reserve the task of changing the IIO driver structure for later, when this pull request is merged.

drivers/frequency/adf4382/adf4382.c
Jude-Osems: Here is how I understand your suggestion. Am I correct?

```
uint8_t val;
uint8_t tmp;
int ret;

if (!dev)
           return -EINVAL;
if (!en_fast_cal)
           return 0;  //IDEA: Function should do nothing or stay disabled if the value is not 1

if (en_fast_cal == 1) {
           n_value = NO_OS_DIV_ROUND_CLOSEST(fclk_max_hz, dev->ref_freq_hz);

           ref_freq_khz = NO_OS_DIV_U64(dev->ref_freq_hz, KHZ_PER_MHZ);
```

buha: yes, correct

the idea is to avoid code with too many indentation levels if possible

drivers/frequency/adf4382/adf4382.h
buha: No. What I mean is that the device driver init sequence should be like this:
```
driver_api_init()
    driver_api_hw_reset(); // this alone will set all the registers to default values, no need to do it manually

    // config sequence
    driver_api_set_clock(); 
    driver_api_set_mode();
    ... // various initializations that put the chip in the desired state...
```

Making a large hard to read table that achieves the same thing as the config sequence above is a lazy way to do it. It works, but it's unmaintainable because one has to decipher the code with a datasheet. The config sequence above is much more clear about what it does.

But again, no need to change anything in this PR because the table was already there, somebody might tackle it separately.

drivers/frequency/adf4382/iio_adf4382.c
buha: please change this now, it's pretty easy to do and we don't want to carry any more bad old implementations, i know there are quite a few already in the codebase...

Jude-Osems:
This branch and pull request will be deleted to propagate the new changes

PR #2408, opened 2025-01-06
drivers: power: lt3074: add driver for LT3074

cencarna:
## Pull Request Description

The LT3074 is a low voltage, ultralow noise, and ultrafast transient
response linear regulator with a PMBus serial interface. The device
supplies up to 3A with a typical dropout voltage of 45mV.

This pull request adds the no-OS and IIO driver, as well as examples
and documentations, to quickly evaluate the LT3074.

## PR Type

## PR Checklist

### DIFF
--- a/drivers/power/lt3074/iio_lt3074.c
+++ b/drivers/power/lt3074/iio_lt3074.c
@@ -1,0 +1,446 @@
+/***************************************************************************//**
+*   @file   iio_lt3074.c
+*   @brief  Source file for the LT3074 IIO Driver
+*   @author Cedric Justine Encarnacion (cedricjustine.encarnacion@analog.com)
+********************************************************************************
+* Copyright 2025(c) Analog Devices, Inc.
+*
+* Redistribution and use in source and binary forms, with or without
+* modification, are permitted provided that the following conditions are met:
+*
+* 1. Redistributions of source code must retain the above copyright notice,
+*    this list of conditions and the following disclaimer.
+*
+* 2. Redistributions in binary form must reproduce the above copyright notice,
+*    this list of conditions and the following disclaimer in the documentation
+*    and/or other materials provided with the distribution.
+*
+* 3. Neither the name of Analog Devices, Inc. nor the names of its
+*    contributors may be used to endorse or promote products derived from this
+*    software without specific prior written permission.
+*
+* THIS SOFTWARE IS PROVIDED BY ANALOG DEVICES, INC. “AS IS” AND ANY EXPRESS OR
+* IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF
+* MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO
+* EVENT SHALL ANALOG DEVICES, INC. BE LIABLE FOR ANY DIRECT, INDIRECT,
+* INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+* LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA,
+* OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF
+* LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING
+* NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE,
+* EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+*******************************************************************************/
+#include <stdlib.h>
+#include <stdio.h>
+#include <string.h>
+#include "no_os_alloc.h"
+#include "no_os_error.h"
+#include "no_os_units.h"
+#include "no_os_util.h"
+
+#include "lt3074.h"
+#include "iio_lt3074.h"
+
+static const int32_t lt3074_margin_avail[] = {0, 1, 3, 5, 10, 15, 20, 25, 30};
+
+enum lt3074_iio_chan_type {
+	LT3074_IIO_VIN_CHAN,
+	LT3074_IIO_VOUT_CHAN,
+	LT3074_IIO_IOUT_CHAN,
+	LT3074_IIO_TEMP_CHAN,
+	LT3074_IIO_VBIAS_CHAN,
+};
+
+enum lt3074_iio_attr_id {
+	LT3074_IIO_CHAN_ATTR_RAW,
+	LT3074_IIO_CHAN_ATTR_SCALE,
+	LT3074_IIO_GLOBAL_ATTR_VOUT_MARGIN_LOW,
+	LT3074_IIO_GLOBAL_ATTR_VOUT_MARGIN_HIGH,
+	LT3074_IIO_GLOBAL_ATTR_VOUT_MARGIN_AVAILABLE,
+	LT3074_IIO_GLOBAL_ATTR_VOUT_OV_WARN_LIMIT = LT3074_VOUT_OV_WARN_LIMIT_TYPE,
+	LT3074_IIO_GLOBAL_ATTR_VOUT_UV_WARN_LIMIT = LT3074_VOUT_UV_WARN_LIMIT_TYPE,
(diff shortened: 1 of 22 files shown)


amiclaus (approved):

drivers/power/lt3074/iio_lt3074.c
```diff
@@ -0,0 +1,706 @@
...
+
+/**
+ * @brief Read register value.
+ * @param dev     - The iio device structure.
+ * @param reg	  - Register to read.
+ * @param readval - Read value.
+ * @return ret    - Result of the reading procedure.
+ * 		    In case of success, the size of the read data is returned.
+*/
+static int32_t lt3074_iio_reg_read(void *dev, uint32_t reg, uint32_t *readval)
```
amiclaus, line 74: actually there is one comment: the return type of the functions should be `int`.

cencarna: These are mapped to `.debug_reg_read` and `.debug_reg_write` which return `int32_t`. Compiler also gives warnings about incompatible pointers. Should I still proceed with `int`?

amiclaus: It's fine like this, i did not take into account that aspect.

drivers/power/lt3074/iio_lt3074.c
```diff
@@ -0,0 +1,706 @@
...
+	case LT3074_MFR_BIAS_UV_WARN_LIMIT:
+	case LT3074_MFR_IOUT_MIN_WARN_LIMIT:
+	case LT3074_READ_VIN:
+	case LT3074_READ_VOUT:
+	case LT3074_READ_IOUT:
+	case LT3074_READ_TEMPERATURE_1:
+	case LT3074_IC_DEVICE_REV:
+	case LT3074_MFR_SPECIAL_ID:
+		return lt3074_read_word(lt3074, (uint8_t)reg,
+					(uint16_t *)readval);
```
CiprianRegus, line 120: Don't cast pointers. The result will be different based on CPU endianess

cencarna:
v2:
- Removed pointer casts. Used intermediate variables instead to handle type differences.

buha (changes requested):
Please also update the project to the new project infrastructure.

Here's the PR with the changes with 3 modified projects as example: https://github.com/analogdevicesinc/no-OS/pull/2412

And here's a pending PR where I gave feedback with pretty much that needs to be changed, same applies to your PR.
https://github.com/analogdevicesinc/no-OS/pull/2411

drivers/power/lt3074/lt3074.c
```diff
@@ -0,0 +1,778 @@
...
+
+	/* If value too low, increase mantissa. */
+	while (data < LT3074_LIN11_MANTISSA_MIN &&
+	       exp > LT3074_LIN11_EXPONENT_MIN) {
+		exp--;
+		data <<= 1;
+	}
+
+	mant = no_os_clamp(NO_OS_DIV_ROUND_CLOSEST_ULL(data, MILLI),
+			   0, 0x3FF);
```
CiprianRegus, line 105: `NO_OS_GENMASK()`

drivers/power/lt3074/iio_lt3074.c
```diff
@@ -0,0 +1,727 @@
...
+ * @brief Free resources allocated by the init function.
+ * @param iio_desc - The iio device descriptor.
+ * @return 0 in case of success, an error code otherwise.
+ */
+int lt3074_iio_remove(struct lt3074_iio_desc *iio_desc)
+{
+	if (!iio_desc)
+		return -ENODEV;
+
+	no_os_free(iio_desc->iio_dev->channels);
```
CiprianRegus, line 535: The `lt3074_channels` array is statically allocated.

drivers/power/lt3074/iio_lt3074.c
```diff
@@ -0,0 +1,727 @@
...
+
+	if (!iio_lt3074->lt3074_dev)
+		return -EINVAL;
+
+	ret = lt3074_read_word_data(lt3074, priv, &word);
+	if (ret)
+		return ret;
+
+	vals[0] = (int32_t)(word / MILLI);
+	vals[1] = (int32_t)((word * MILLI) % MICRO);
```
CiprianRegus, line 458: These are implicit casts, so they don't have any effect.

drivers/power/lt3074/iio_lt3074.c
```diff
@@ -0,0 +1,727 @@
...
+	case LT3074_STATUS_CML:
+	case LT3074_STATUS_MFR_SPECIFIC:
+	case LT3074_MFR_MARGIN:
+	case LT3074_MFR_RAIL_ADDRESS:
+	case LT3074_CAPABILITY:
+	case LT3074_VOUT_MODE:
+	case LT3074_IOUT_OC_FAULT_RESPONSE:
+	case LT3074_REVISION:
+	case LT3074_MFR_DEFAULT_CONFIG:
+		ret = lt3074_read_byte(lt3074, (uint8_t)reg, &byte);
```
buha, line 102: this will definitely work but i expect this kind of register level access size to be handled in the device driver so at the lower level, not here

buha: this is how we prefer to do it:
https://github.com/analogdevicesinc/no-OS/blob/2d0e5d5ed9244bf91e4b47870702a696406aab71/drivers/adc/ad9250/ad9250.h#L54

each register has size information coded into the register "address" macro so the lowest level spi/i2c/whatever read/write, knows how many bytes to read/write

drivers/power/lt3074/iio_lt3074.c
```diff
@@ -0,0 +1,727 @@
...
+		.name = "scale",
+		.show = lt3074_iio_read_scale,
+	},
+	END_ATTRIBUTES_ARRAY
+};
+
+static struct iio_attribute lt3074_global_attrs[] = {
+	{
+		.name = "vout_margin_low",
+		.show = lt3074_iio_read_vout_margin,
```
buha, line 557: please use a single function for all of these if they are similar enough to avoid code duplication

projects/lt3074/Makefile
```diff
@@ -0,0 +1,9 @@
+# Select the example you want to enable by choosing y for enabling and n for disabling
+BASIC_EXAMPLE = n
```
buha, line 2: and i see you didn't yet update the project structure, please update it

cencarna:
v3:

- Remove `free` for statically allocated `lt3074_channels` array
- Used `NO_OS_GENMASK` for mantissa computation
- Removed redundant typecasts for variable assignments that are already implicitly cast
- Moved all register-level access to device driver. 
- Macros now have a prefix for size information
- Generic reg read and reg write for size-encoded macros
- Used single function for all attribute reads and writes
- Updated project structure using examples.mk

drivers/power/lt3074/lt3074.h
```diff
@@ -0,0 +1,275 @@
...
+/* LINEAR data format params */
+#define LT3074_LIN11_MANTISSA_MAX		1023000L
+#define LT3074_LIN11_MANTISSA_MIN		511000L
+#define LT3074_LIN11_EXPONENT_MAX		15
+#define LT3074_LIN11_EXPONENT_MIN		-15
+#define LT3074_LIN11_MANTISSA_MSK		NO_OS_GENMASK(10,0)
+#define LT3074_LIN11_EXPONENT_MSK		NO_OS_GENMASK(15,11)
+#define LT3074_LIN11_EXPONENT(x)		((int16_t)(x) >> 11)
+#define LT3074_LIN11_MANTISSA(x)		(((int16_t)((x & 0x7FF) << 5)) >> 5)
+#define LT3074_LIN16_EXPONENT			-13
```
CiprianRegus, line 123: These could be moved to the source file, since they're not meant to be used directly

drivers/power/lt3074/iio_lt3074.c
```diff
@@ -0,0 +1,456 @@
...
+	0,
+	1,
+	3,
+	5,
+	10,
+	15,
+	20,
+	25,
+	30,
+};
```
CiprianRegus, line 54: Define these in one line.

CiprianRegus (approved):
Some minor remarks, looks good otherwise

cencarna:
v4:
- Moved LINEAR reg2data macros to source file
- Define `lt3074_margin_avail` in one line

kister-jimenez:
lgtm

CiprianRegus (approved):

amiclaus (approved):

buha (approved):

PR #2414, opened 2025-01-16
drivers:adc-dac:ad5592r: Driver Enhancement

D-Disha:
1. Added driver apis for adc range, dac range, power down,set internal reference and set adc buffer to the base driver.
2. Added API to enable busy indicatir on AD5592r
3. Removed static for spi nop API for visibilty.
4. Moved some macros from ad5593.c to header file for visibility.

## Pull Request Description

## PR Type
- [] Bug fix (change that fixes an issue)

## PR Checklist

### DIFF
--- a/drivers/adc-dac/ad5592r/ad5592r-base.c
+++ b/drivers/adc-dac/ad5592r/ad5592r-base.c
@@ -3,7 +3,7 @@
  *   @brief  Implementation of AD5592R Base Driver.
  *   @author Mircea Caprioru (mircea.caprioru@analog.com)
 ********************************************************************************
- * Copyright 2018, 2020(c) Analog Devices, Inc.
+ * Copyright 2018, 2020, 2025(c) Analog Devices, Inc.
  *
  * Redistribution and use in source and binary forms, with or without
  * modification, are permitted provided that the following conditions are met:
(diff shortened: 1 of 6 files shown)


drivers/adc-dac/ad5592r/ad5593r.h
```diff
@@ -36,6 +36,17 @@
...
 #include "ad5592r-base.h"
 
+#define AD5593R_MODE_CONF		(0 << 4)
+#define AD5593R_MODE_DAC_WRITE		(1 << 4)
+#define AD5593R_MODE_ADC_READBACK	(4 << 4)
+#define AD5593R_MODE_DAC_READBACK	(5 << 4)
+#define AD5593R_MODE_GPIO_READBACK	(6 << 4)
+#define AD5593R_MODE_REG_READBACK	(7 << 4)
+
+#define STOP_BIT	1
```
buha, line 46: tbh i would drop this STOP_BIT, it's pretty clear what it does, no need to have a macro for it

alternatively, prefix it, because at some point there is going to be a name clash

drivers/adc-dac/ad5592r/ad5593r.h
```diff
@@ -36,6 +36,17 @@
...
+#define RESTART_BIT	0
```
buha, line 47: same for RESTART_BIT

D-Disha:
Change log: Added power down configuration to init sequence as well. Initialized dev->num_of_channels and ldac mode as it is used by the driver.

drivers/adc-dac/ad5592r/ad5592r-base.c
```diff
@@ -320,3 +320,165 @@ int32_t ad5592r_reset_channel_modes(struct ad5592r_dev *dev)
...
+ */
+int32_t ad5592r_power_down(struct ad5592r_dev *dev, uint8_t chan, bool status)
+{
+	int ret;
+	uint16_t temp_reg_val;
+
+	if (!dev)
+		return -1;
+
+	ret = ad5592r_base_reg_read(dev, AD5592R_REG_PD, &temp_reg_val);
```
buha, line 408: instead of this read - process - write sequence that you repeat in all these functions you could have a function named `ad5592r_base_reg_update` which does the sequence for you

D-Disha: Updated with these changes now

drivers/adc-dac/ad5592r/ad5592r-base.c
```diff
@@ -320,3 +320,184 @@ int32_t ad5592r_reset_channel_modes(struct ad5592r_dev *dev)
...
+	int32_t ret;
+	uint16_t status;
+
+	if (!dev)
+		return -1;
+
+	if (adc_range)
+		status |= AD5592R_REG_CTRL_ADC_RANGE;
+	else
+		status &= ~AD5592R_REG_CTRL_ADC_RANGE;
```
buha, line 376: you can rewrite this whole part more compact as:
```
uint16_t status = adc_range ? AD5592R_REG_CTRL_ADC_RANGE : 0;
```

or even ditch the status local variable and put it all inline in the function call, although that's maybe a bit too much

drivers/adc-dac/ad5592r/ad5592r-base.c
```diff
@@ -320,3 +320,184 @@ int32_t ad5592r_reset_channel_modes(struct ad5592r_dev *dev)
...
+	dev->int_ref = status;
+
+	return 0;
+}
+
+/**
+ * Set ADC Buffer for the device
+ *
+ * @param dev - The device structure.
+ * @param status - Status to enable/disable adc buffer.
```
buha, line 481: i wouldn't name it status, rename it to `enable`, then it's even more clear what it does

buha:
also, check out cppcheck and astyle CI results

D-Disha:
> also, check out cppcheck and astyle CI results

Fixed all review comments and the checks are passing now

drivers/adc-dac/ad5592r/ad5592r.c
```diff
@@ -126,6 +134,34 @@ int32_t ad5592r_read_adc(struct ad5592r_dev *dev, uint8_t chan,
...
+
+	dev->channel_modes[7] = CH_MODE_GPO;
+	ad5592r_set_channel_modes(dev);
+
+	if (!dev)
+		return -1;
+
+	temp_reg_val = enable ? AD5592R_REG_GPIO_OUT_EN_ADC_NOT_BUSY : 0;
+
+	ret = ad5592r_base_reg_update(dev, AD5592R_REG_GPIO_OUT_EN, temp_reg_val,
```
CiprianRegus, line 157: Return directly

drivers/adc-dac/ad5592r/ad5592r.c
```diff
@@ -19,7 +19,7 @@
  *    contributors may be used to endorse or promote products derived from this
  *    software without specific prior written permission.
  *
- * THIS SOFTWARE IS PROVIDED BY ANALOG DEVICES, INC. “AS IS” AND ANY EXPRESS OR
+ * THIS SOFTWARE IS PROVIDED BY ANALOG DEVICES, INC. �AS IS� AND ANY EXPRESS OR
```
CiprianRegus, line 22: Unrelated change (may have added a white space by mistake).

drivers/adc-dac/ad5592r/ad5592r-base.c
```diff
@@ -320,3 +320,174 @@ int32_t ad5592r_reset_channel_modes(struct ad5592r_dev *dev)
...
+		return -1;
+
+	status = adc_range ? AD5592R_REG_CTRL_ADC_RANGE : 0;
+
+	ret = ad5592r_base_reg_update(dev, AD5592R_REG_CTRL, status,
+				      AD5592R_REG_CTRL_ADC_RANGE);
+	if (ret < 0)
+		return ret;
+
+	dev->adc_range = adc_range;
```
CiprianRegus, line 380: We usually avoid caching register values at the driver level, since the value could be changed by a direct register access. There are some cases where you may want to do that to avoid recurrent register reads, but it doesn't look like it applies here. Is there a reason for which you may want these values to be stored?

drivers/adc-dac/ad5592r/ad5593r.c
```diff
@@ -258,37 +254,70 @@ int32_t ad5593r_gpio_read(struct ad5592r_dev *dev, uint8_t *value)
...
-int32_t ad5593r_init(struct ad5592r_dev *dev,
+int32_t ad5593r_init(struct ad5592r_dev **device,
 		     struct ad5592r_init_param *init_param)
 {
 	int32_t ret;
-	uint16_t temp_reg_val;
+	uint8_t i;
+	struct ad5592r_dev *dev;
 
+	dev = (struct ad5592r_dev *)no_os_malloc(sizeof(*dev));
```
CiprianRegus, line 268: `no_os_calloc()`

drivers/adc-dac/ad5592r/ad5592r-base.h
```diff
@@ -108,8 +111,22 @@ struct ad5592r_rw_ops {
...
 struct ad5592r_init_param {
 	bool int_ref;
+	struct no_os_spi_init_param *spi_init;
+	struct no_os_i2c_init_param *i2c_init;
+	uint8_t channel_modes[8];
+	uint8_t channel_offstate[8];
+	enum ad559xr_range adc_range;
+	enum ad559xr_range dac_range;
+	bool adc_buf;
+	uint16_t cached_dac[8];
```
CiprianRegus, line 128: I'd say having an initial values for the DACs output is a bit awkward. There is already `ad5592r_write_dac()`, so you could use that in the application after `ad5592r_init()`.

CiprianRegus (commented):
Some comments from my side. Also, for future contributions like this one, it would be worth splitting your work into multiple commits, since the changes are not directly related to each other.

drivers/adc-dac/ad5592r/ad5592r-base.c
D-Disha: The adc and dac range are runtime attributes in the IIO firmware where a user can change the values on the fly. This in turn updates the scale value of the device. 

Hence, I considered it best to have it as a device member and init_param member as well so the user can easily update the ranges compile time as well if needed.

drivers/adc-dac/ad5592r/ad5592r-base.h
D-Disha: cached_dac was present as a device member in the existing no-os drivers already. Having it as a part of init_param as well will help user change the dac output values runtime through our user configuration file.

drivers/adc-dac/ad5592r/ad5592r.c
D-Disha: I copy pasted those lines again from the preexisting files. Not sure what is wrong here. It still shows up in the diff

D-Disha:
> As @CiprianRegus said, please split work into multiple commits in the future. The reason is, if one is bad and gets merged, we can revert it and keep the other good ones for users to benefit. So if you feel like listing 5 items in the commit message, you probably have to split the commit into 5 commits.
@buha @CiprianRegus Sure, I understand. Will take the suggestion into account from next time.

D-Disha:
Hi all, are there any more comments on this PR?

drivers/adc-dac/ad5592r/ad5592r-base.h
buha: i also don't see the reason for cached_dac

yes, it was there in the desc, but if you search for it in the prior code version you'll see it's unused, probably a bad copy paste from linux from previous author

i'd remove these and any code related to them:
```
uint16_t cached_dac[8];
uint16_t cached_gp_ctrl;
```

buha: what user configuration file ? how do you pass that info into the init param ? what's the disadvantage of performing a ad5592r_write_dac() sequence ?

CiprianRegus: The users already have to open an IIO client to configure output values for other DACs, so I don't really see why this one should be different. It seems like an application specific thing to me. So, I would say:

1. Remove `cached_dac` from the init parameter.
2. Let the user configure the DAC output values using the same array, but define it at the application level.
3. Call `ad5592r_write_dac()` for the 8 channels after you initialize the IIO device

CiprianRegus (approved):

PR #2427, opened 2025-01-29
drivers:platform:freeRTOS:Added mem functions

landas:
Add freeRTOS memory functions for no-OS implementation.

## Pull Request Description

Added memory allocation functions for use in freeRTOS platform.

## PR Type

## PR Checklist

### DIFF
--- a/drivers/platform/freeRTOS/freertos_alloc.c
+++ b/drivers/platform/freeRTOS/freertos_alloc.c
@@ -1,0 +1,78 @@
+/*******************************************************************************
+ *   @file   freeRTOS/freertos_alloc.c
+ *   @brief  Implementation of freertos allocation functions.
+ *   @author Lars Andre Landås (landas@gmail.com)
+********************************************************************************
+ * Copyright 2025(c) Analog Devices, Inc.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice,
+ *    this list of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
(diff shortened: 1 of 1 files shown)


buha:
Looks good to me

As a side note, I think freertos.mk should contain something like, perhaps as a separate commit:
```
SRCS += $(NO-OS)/drivers/platform/freeRTOS/freertos_mutex.c \
        $(NO-OS)/drivers/platform/freeRTOS/freertos_semaphore.c \
        $(NO-OS)/drivers/platform/freeRTOS/freertos_delay.c \
        $(NO-OS)/drivers/platform/freeRTOS/freertos_alloc.c
```

This would avoid having to specify these in the project .mk files.

landas:
I was a bit hasty with my pull request. The last commit is compiling without error or warnings with STM32CubeIDE.

drivers/platform/freeRTOS/freertos_alloc.c
```diff
@@ -0,0 +1,76 @@
...
+#include "no_os_alloc.h"
+#include "portable.h"
+#include <string.h>
+
+/**
+ * @brief Allocate memory and return a pointer to it.
+ * @param size - Size of the memory block, in bytes.
+ * @return Pointer to the allocated memory, or NULL if the request fails.
+ */
+__attribute__((weak)) void *no_os_malloc(size_t size)
```
buha, line 45: is the weak linkage specified needed here ? i would argue that this file is THE implementation of no_os_alloc for freertos platform so it should replace the 3 alloc API functions there (which are specified as weak) with non-weak versions

buha: btw, is the weak linkage what changed between your first and 2nd version ?

landas: I agree that the weak linkage should be removed, and I have done that in my last commit now.

buha (approved):

PR #2445, opened 2025-02-11
drivers: spi: Add SPI Abort API

NagaHimanshu:
## Pull Request Description

Add SPI Abort API to abort the SPI transactions.

This is useful when user has to terminate the SPI transactions in the application, especially when circular DMA is used.

## PR Type

## PR Checklist

### DIFF
--- a/drivers/platform/stm32/stm32_spi.c
+++ b/drivers/platform/stm32/stm32_spi.c
@@ -615,22 +615,10 @@ void stm32_spi_dma_callback(struct no_os_dma_xfer_desc *old_xfer,
 		no_os_pwm_disable(sdesc->tx_pwm_desc);
 #endif
 
-#if defined (STM32H5)
-	CLEAR_BIT(sdesc->hspi.Instance->CFG1, SPI_CFG1_TXDMAEN);
-#else
-	CLEAR_BIT(sdesc->hspi.Instance->CR2, SPI_CR2_TXDMAEN);
-#endif
+	/* Perform abort SPI transfers */
+	no_os_spi_transfer_abort(desc);
 
-#if defined (STM32H5)
-	CLEAR_BIT(sdesc->hspi.Instance->CFG1, SPI_CFG1_RXDMAEN);
-#else
-	CLEAR_BIT(sdesc->hspi.Instance->CR2, SPI_CR2_RXDMAEN);
-#endif
-
-	no_os_dma_xfer_abort(sdesc->dma_desc, sdesc->txdma_ch);
-
-	no_os_dma_xfer_abort(sdesc->dma_desc, sdesc->rxdma_ch);
-
+	/* Free the allocated memory for tx and rx transfers */
 	no_os_free(sdesc->tx_ch_xfer);
 	no_os_free(sdesc->rx_ch_xfer);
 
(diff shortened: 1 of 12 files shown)


drivers/platform/stm32/stm32_spi.c
```diff
@@ -705,6 +686,58 @@ int32_t stm32_spi_dma_transfer_sync(struct no_os_spi_desc* desc,
...
+	sdesc = desc->extra;
+	SPI_TypeDef *SPIx = sdesc->hspi.Instance;
+
+	if (sdesc->rxdma_ch) {
+		ret = no_os_dma_xfer_abort(sdesc->dma_desc, sdesc->rxdma_ch);
+		if (ret) {
+			return ret;
+		}
+
+#if defined (STM32H5)
```
buha, line 711: experiment with #ifdef HAL_GPDMA_MODULE_ENABLED to replace all these STM32H5 based conditions

NagaHimanshu: Umm... I guess that macro can't help here because this is related to SPIx registers - CFG1 for H5 & CR2 for F4. So probably we may need to use processor macros itself?

include/no_os_spi.h
```diff
@@ -276,6 +278,9 @@ int32_t no_os_spi_transfer_dma_async(struct no_os_spi_desc *desc,
 				     void (*callback)(void *),
 				     void *ctx);
 
+/* Abort SPI transfers. */
```
buha, line 281: make it clear that this function aborts transfers started by `no_os_spi_transfer_dma_async` by documenting this

buha:
Can you also make a separate commit that renames `no_os_spi_transfer_dma_sync` to `no_os_spi_transfer_dma` ? Synchronous behavior is kind of implied. There are also 2 usages of it across no-OS repo that need changing.

include/no_os_spi.h
NagaHimanshu: Also, shouldn't that be useful for sync SPI transfers where we may abort the transfer in ISR? However right now it is aborting for DMA transfers, but in future it should be applicable for aborting other kinds of transfers as well. Isn't it?

CiprianRegus:
That's not really  required, since it's only internal naming, but if you really want to do so:

`find drivers/platform -type f -exec sed -i 's/dma_transfer_sync/dma_transfer/g' {} +`

drivers/platform/stm32/stm32_spi.c
NagaHimanshu: I was looking through the reference manual of STM32H743/53, and observed that it uses BDMA but the SPI registers contain CFG1 for TXDMAEN & RXDMAEN. So it looks like we can't use the DMA macro here.

image

include/no_os_spi.h
buha: you're right, i think it's better to keep it broad for now, i thought you had a focus only on transfer_dma only but it can potentially apply to async interrupt based transfer, not necessarily dma

drivers/platform/stm32/stm32_spi.c
```diff
@@ -714,5 +747,6 @@ const struct no_os_spi_platform_ops stm32_spi_ops = {
 	.remove = &stm32_spi_remove,
 	.transfer = &stm32_spi_transfer,
 	.dma_transfer_async = &stm32_spi_dma_transfer_async,
-	.dma_transfer_sync = &stm32_spi_dma_transfer_sync
+	.dma_transfer_sync = &stm32_spi_dma_transfer_sync,
```
buha, line 750: Please make a separate commit where you rename this line (and all the implications thereof) to:
`.transfer_dma_sync = &stm32_spi_transfer_dma_sync,`

This will break some projects, but you can fix them in the same commit. I want to keep the naming consistent like:

```
stm32_spi_transfer_dma
^     ^       ^    ^--- method (could also be interrupt based or other)
|     |       |
|     |       |--- action
|     |      
|     |--- interface
|
|--- platform (or top level no_os)

buha: it's semantics but i think it's important

buha (dismissed):

buha (approved):

PR #2446, opened 2025-02-11
prefix admv1013 and admv1014 enum members

### LINKED ISSUE #2444: ADMV1013 and 1014 cannot coexist
Looks like enums for the two parts are conflicting. Recommend action is to prepend enums with the part number.

buha:
Some enum members have identical names in the admv1013 and admv1014 device drivers which renders them unable to coexist. Fix this by prefixing the enum member names.

Closes #2444

## PR Type

## PR Checklist


CiprianRegus (dismissed):

drivers/frequency/admv1013/admv1013.h
```diff
@@ -99,9 +99,6 @@
 #define ADMV1013_SPI_READ_CMD			NO_OS_BIT(7)
 #define ADMV1013_SPI_WRITE_CMD			(0 << 7)
 
-#define MIXER_GATE_0_to_1_8_V(x)		((2389 * x/ 1000000 + 8100) / 100)
```
amiclaus, line 102: i'd add a prefix for these also. so we keep them in the header file.

drivers/frequency/admv1013/admv1013.c
```diff
@@ -39,6 +39,9 @@
 #include "no_os_error.h"
 #include "no_os_alloc.h"
 
+#define MIXER_GATE_0_to_1_8_V(x)		((2389 * x/ 1000000 + 8100) / 100)
```
amiclaus, line 42: if doing the change, i'd also uppercase the "_to" -> "_TO_"

buha: will do

drivers/frequency/admv1013/admv1013.h
buha: are they useful in the header ? after all they're used by some update() function so the user doesn't need these macros

buha:
V2:
- to -> TO

drivers/frequency/admv1013/admv1013.h
amiclaus: Not really, just to keep it consistent with other drivers where we hold all macros in the header file.

amiclaus (approved):

PR #2447, opened 2025-02-12
drivers: stm32: pwm: Fix PWM set period

NagaHimanshu:
## Pull Request Description

Fix PWM set period function as reinitialization of timer with new extra params as local variable demolishing other properties such as trigger, polarity etc.

## PR Type

## PR Checklist

### DIFF
--- a/drivers/platform/stm32/stm32_pwm.c
+++ b/drivers/platform/stm32/stm32_pwm.c
@@ -1,5 +1,5 @@
 /***************************************************************************//**
- *   @file   stm32/stm32_pwm.h
+ *   @file   stm32/stm32_pwm.c
  *   @brief  Implementation of stm32 pwm driver.
 ********************************************************************************
  * Copyright 2023(c) Analog Devices, Inc.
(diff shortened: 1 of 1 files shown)


drivers/platform/stm32/stm32_pwm.c
```diff
@@ -599,32 +599,31 @@ int32_t stm32_pwm_disable(struct no_os_pwm_desc *desc)
...
-	if (!desc || !desc->extra)
+	if (!sdesc)
 		return -EINVAL;
 
-	ret = stm32_init_timer(desc->extra, &param);
-	if (ret != HAL_OK)
-		return -EIO;
+	if (sdesc->get_timer_clock) {
+		timer_frequency_hz = sdesc->get_timer_clock();
+		timer_frequency_hz *= sdesc->clock_divider;
```
buha, line 611: question1: why not divide by clock_divider ? you're multiplying instead if dividing here

drivers/platform/stm32/stm32_pwm.c
```diff
@@ -599,32 +599,31 @@ int32_t stm32_pwm_disable(struct no_os_pwm_desc *desc)
...
+		timer_frequency_hz /= (sdesc->prescaler + 1);
+		period = _compute_period_ticks(sdesc, timer_frequency_hz, period_ns);
+	} else {
+		period = PWM_DEFAULT_PERIOD - 1;
+	}
+
+	sdesc->htimer.Init.AutoReloadPreload = sdesc->timer_autoreload ?
+					       TIM_AUTORELOAD_PRELOAD_ENABLE : TIM_MASTERSLAVEMODE_DISABLE;
+	sdesc->htimer.Init.ClockDivision = TIM_CLOCKDIVISION_DIV1;
```
buha, line 620: question 2: why not use the clock_divider value here ? you're hardcoding this to DIV1

buha:
the previous implementation was indeed bad, it was creating an sparam from scratch instead of using the one from desc, this is the correct approach

drivers/platform/stm32/stm32_pwm.c
NagaHimanshu: I took the same piece of code from existing driver which is during stm32_init_timer.
https://github.com/analogdevicesinc/no-OS/blob/03b6f82d777db799ee72796126ee3b19257fa9a4/drivers/platform/stm32/stm32_pwm.c#L106

drivers/platform/stm32/stm32_pwm.c
NagaHimanshu: Applies to here as well. I didnt want to do any extra modifications in the existing implementation.
https://github.com/analogdevicesinc/no-OS/blob/03b6f82d777db799ee72796126ee3b19257fa9a4/drivers/platform/stm32/stm32_pwm.c#L184

drivers/platform/stm32/stm32_pwm.c
buha: I think the computation is wrong but for the sake of keeping things separate, let's say we don't care about it being right or wrong for now.

You should create an internal/static function that performs the computation and use it both in stm32_init_timer and here. You don't want this kind of code duplication.

NagaHimanshu: Yeah that makes sense, will do that.

amiclaus (approved):

buha (approved):

PR #2457, opened 2025-02-17
drivers: stm32: i2c: Add I2C GPIOs init

NagaHimanshu:
Initialize I2C GPIOs (SCL, SDA) using respective init params.

When I2C is selected in a fresh project instance, the gpio configuration is not accurate and requires additional effort to configure them in ioc file. Hence the idea is to consolidate the gpio configuration in driver itself so as to eliminate separate configuration in ioc file. This will be useful in migration of projects among various stm32 series.

## Pull Request Description

Add I2C GPIOs (SCL, SDA) initialization as a part of driver instead of relying on auto generated files from ioc file.

## PR Type

## PR Checklist

### DIFF
--- a/drivers/platform/stm32/stm32_i2c.c
+++ b/drivers/platform/stm32/stm32_i2c.c
@@ -50,6 +50,7 @@ int32_t stm32_i2c_init(struct no_os_i2c_desc **desc,
 	struct no_os_i2c_desc *descriptor;
 	struct stm32_i2c_desc *xdesc;
 	struct stm32_i2c_init_param *i2cinit;
+	struct no_os_gpio_desc *gpio_desc;
 	I2C_TypeDef *base = NULL;
 
 	if (!desc || !param)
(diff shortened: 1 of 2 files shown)


CiprianRegus:
Can you please add the reason for this change in the commit description? Also, I'm confused why is this necessary? Shouldn't the STM32 HAL I2C driver already handle setting the alternate functions of those pins?

NagaHimanshu:
When I2C is selected in a fresh project instance, the gpio configuration is not accurate and this requires additional effort to set the pull-up/down in ioc file. This effort has to be carried when same project is ported to multiple platforms (in our case F4 & H5).

Hence we'd like to consolidate all necessary configurations within application files and there is no need for users to adjust settings in ioc files.

image

buha:
I don't understand the reason for this change. The selection of pin functions for stm32 platform was left up to the .ioc writer by design because they have app knowledge. If your project can run on 2 micros (F4 and H5 let's say) why not just make 2 .ioc files, one for each board ? What's the advantage of moving this info in-code and putting some #ifdefs around them ? #ifdefs will 100% be required, the port/pin numbers certainly won't match when you change micros.

NagaHimanshu:
I agree there will be separate ioc files for separate micros. But if an user needs to migrate the project to another micro, he can simply enable the i2c in ioc file, change the pins in FW and everything will be good. Or else all the configuration from one micro should be migrated to another micro in ioc file. Ultimately the objective is the code should take care of everything rather configurations in bits & pieces. This was the idea actually so to enhance the user experience while migration.

NagaHimanshu:
Closing this PR as this would make the code bulky and replicative from auto-generated files.

PR #2458, opened 2025-02-17
drivers:platform:stm32:dma Circular DMA and Trigger Support

SwaroopPKADI:
## Pull Request Description

Added support for circular DMA using nodes and linked lists, enabling continuous data transfer without manual intervention, which enhances efficiency for repetitive tasks. Additionally, implemented triggering support for DMA, allowing precise control over data transfer initiation based on specific events or conditions

## PR Type

## PR Checklist

### DIFF
--- a/drivers/platform/stm32/stm32_dma.c
+++ b/drivers/platform/stm32/stm32_dma.c
@@ -65,50 +65,6 @@ int stm32_dma_config_xfer(struct no_os_dma_ch *channel,
 #else
 	sdma_ch->hdma->Instance = sdma_ch->ch_num;
 #endif
-#if defined (STM32H5)
-	sdma_ch->hdma->Init.DestInc = sdma_ch->mem_increment ? DMA_DINC_INCREMENTED :
-				      DMA_DINC_FIXED;
-	sdma_ch->hdma->Init.SrcInc = sdma_ch->per_increment ? DMA_SINC_INCREMENTED :
-				     DMA_SINC_FIXED;
-
-	switch (sdma_ch->mem_data_alignment) {
-	case DATA_ALIGN_BYTE:
-		sdma_ch->hdma->Init.DestDataWidth = DMA_DEST_DATAWIDTH_BYTE;
-		break;
-	case DATA_ALIGN_HALF_WORD:
-		sdma_ch->hdma->Init.DestDataWidth = DMA_DEST_DATAWIDTH_HALFWORD;
-		break;
-	case DATA_ALIGN_WORD:
-		sdma_ch->hdma->Init.DestDataWidth = DMA_DEST_DATAWIDTH_WORD;
-		break;
-	default:
-		return -EINVAL;
-	}
-
-	switch (sdma_ch->per_data_alignment) {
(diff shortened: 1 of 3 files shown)


SwaroopPKADI:
### PR Rev2

- Fixed the astyle issues in stm32_dma.c

drivers/platform/stm32/stm32_dma.c
```diff
@@ -40,6 +40,13 @@
 /************************ Functions Definitions *******************************/
 /******************************************************************************/
 
+#if defined (STM32H5)
```
buha, line 43: (I've had this comment in previous contributions to this file regarding the H5 DMA vs other devices' DMA.)

I don't like keeping track of which stm device has which kind of dma and I was assured in the past that there is no way to tell them apart other than by their part name.

But I did some digging and found that ST has 2 types of DMA called Basic DMA or BDMA found in most devices and GPDMA or General Purpose DMA, which is used on more advanced devices and is more capable (notably has these features that you're using here, nodes and list).

I think we could get rid of those #ifdef STM32H5 statements by runtime checking if GPDMA is availabe
#if defined(HAL_GPDMA_MODULE_ENABLED) and modify code accordingly

SwaroopPKADI: I verified the HAL as well as the generated files, there is no differentiation between BDMA and GPDMA. Both are enabled using ```HAL_DMA_MODULE_ENABLED```. Even the function name and the handle structure name (```DMA_HandleTypeDef```) are the same.

Following is my proposal for this:

- Define a macro for GPDMA initially based on the processor macro.
- Use the defined macro to separate functions for GPDMA/BDMA.
- Configure will be divided into 2 parts to configure BDMA and GPDMA separately.

buha:
Looks good

Moving stuff around in a new file (stm32_dma_common.h) might break builds that use DMA already, but let's see the CI output if any src.mk needs to be changed.

SwaroopPKADI:
### Rev3

- Modified the DMA to remove dependency on the MCU macros.
- The DMA has been split into (B)DMA and GPDMA.
- Both DMA and GPDMA will use a common include file containing all the necessary definitions.
- Modified the necessary projects to include the stm32_dma_common.h in their respective src.mk files.
- Modified SPI and XSPI code to use stm32_dma_common.h instead of stm32_dma.h to remove dependency of SPI on type of DMA.

buha:
I am pretty sure it's related:
```
/no-OS_agent/_work/1/s/projects/eval-adxl355-pmdz/build/app/Core/noos/drivers/platform/stm32/stm32_dma.c: In function 'stm32_dma_config_xfer':
/no-OS_agent/_work/1/s/projects/eval-adxl355-pmdz/build/app/Core/noos/drivers/platform/stm32/stm32_dma.c:61:21: error: 'DMA_InitTypeDef' {aka 'struct <anonymous>'} has no member named 'Channel'
   61 |  sdma_ch->hdma->Init.Channel = sdma_ch->ch_num;
      |                     ^
      ```

buha (approved):

PR #2464, opened 2025-02-19
Lp timer pwm support

SwaroopPKADI:
## Pull Request Description

Added PWM support using the Low Power Timer (LPTimer) and implemented an interrupt for PWM completion. 

- The timers or LP Timer related are put under macros such that it gets enabled only when Timer/LPTimer module has been enabled on HAL. 
- The instance of Timer handle in PWM descriptor is changed to a pointer and an enum is included to provide support for multiple types of timers and it also allows future expansion.
- The projects ad738x_fmcz and ad463x_fmcz are affected by the changes and is fixed in the same PR.

## PR Type

## PR Checklist


SwaroopPKADI:
### PR Rev2

- Fixed the AStyle of stm32_pwm.c and stm32_irq.c

amiclaus (commented):
there are some build issues that need addressing in the first place (see the cppcheck build)

SwaroopPKADI:
> there are some build issues that need addressing in the first place (see the cppcheck build)

image

The error mentions an uninitialized variable but I've verified the code and every path the code can take, the return variable has been set a value. The same case with all the other errors mentioned as well.

To resolve this, can I write a redundant line assigning it to a default value?

SwaroopPKADI:
### Rev3

- Fixed the CPP Check issues.
- Pulled the main and fixed the merge conflicts

drivers/platform/stm32/stm32_pwm.h
```diff
@@ -39,7 +39,10 @@
 #include "stm32_gpio.h"
 #include "stm32_hal.h"
 
-#include <stdbool.h>
```
buha, line 42: this shouldnt be remove, i see bool being used below

SwaroopPKADI: line 36 has the same include.

drivers/platform/stm32/stm32_pwm.h
```diff
@@ -39,7 +39,10 @@
...
+enum stm32_pwm_timer {
+	PWM_TIM_TIMER = 0,
```
buha, line 43: this needs to be prefixed, I would suggest:
```
enum stm32_pwm_timer {
	STM32_PWM_TIMER_TIM = 0,
	STM32_PWM_TIMER_LPTIM,
}
```

buha: Normally I'd use a boolean but it seems there's a 3rd type available in STM32 HAL, HRTIM_HandleTypeDef, that someone may wanna add in the future so enum is ok

PR #2468, opened 2025-02-21
remove useless code delimiters

buha:
These code delimiters are useless and we've decided not to use them in future code. However, many contributors use existing files as templates and they keep creeping up in new code. So the only solution is to remove them completely.

Here's the command and python script I used to do this: 
```
$ find . -name "*.c" -o -name "*.h" -print0 | xargs -0 python code-delimiters.py 
$ cat code-delimiters.py
import sys
import re
import os
import chardet

def remove_include_comments(filenames):
    # Regex for comments with leading/trailing blank lines (replace with single newline)
    comment_pattern_with_blanks = re.compile(r"\n/\*+\/\n(?:/\*+[^\n]*\n){1}/\*+\/\n\s*\n", re.DOTALL)

    # Regex for comments without leading/trailing blank lines (replace with nothing)
    comment_pattern_no_blanks = re.compile(r"/\*+\/\n(?:/\*+[^\n]*\n){1}/\*+\/\n", re.DOTALL)

    for filename in filenames:
        try:
            with open(filename, 'rb') as f:  # Open in binary mode first
                raw_content = f.read()

            encoding_result = chardet.detect(raw_content)  # Detect the encoding
            encoding = encoding_result['encoding']

            if encoding is None:
                print(f"Could not detect encoding for {filename}. Skipping.")
                continue # Skip the file if encoding cannot be detected

            try:
                content = raw_content.decode(encoding) # Decode with detected encoding
            except UnicodeDecodeError:
                print(f"Decoding error with {filename} using {encoding}. Skipping.")
                continue # Skip the file if decoding fails

            # First pass: Remove comments with leading/trailing blanks
            new_content = comment_pattern_with_blanks.sub("\n", content)

            # Second pass: Remove comments without leading/trailing blanks
            new_content = comment_pattern_no_blanks.sub("", new_content)  # Operate on the modified content

            if new_content != content:
                with open(filename, 'w', newline='') as f:
                    f.write(new_content)
                print(f"Modified: {filename}")
            else:
                print(f"No changes: {filename}")

        except FileNotFoundError:
            print(f"File not found: {filename}")
        except Exception as e:
            print(f"Error processing {filename}: {e}")

if __name__ == "__main__":
    if len(sys.argv) < 2:
        print("Usage: python script.py <file1> <file2> ...")
        sys.exit(1)

    filenames = sys.argv[1:]
    remove_include_comments(filenames)
```
## Pull Request Description

## PR Type

## PR Checklist


buha:
V2:
- added a 2nd commit with astyle fixes (I did it as separate commit because it has nothing to with 1st commit changes)

CiprianRegus:
It looks like the same type of comments are common in mbed-os, and the script modified the contents of the submodule:
```
git show HEAD~1 --stat | grep libraries 
 libraries/mbed/mbed-os                               |  2 +-
```

I think you can exclude that directory as such:

`find . -not \( -path "./libraries/mbed*" -type d -prune \) -name "*.c" -o -name "*.h" -print0 | xargs -0 python code-delimiters.py`

buha:
good catch

i usedd `git add -u` after excluding some stuff manually, but didnt see the libraries change

buha:
V3:
- removed `libraries/mbed/mbed-os` submodule change

CiprianRegus (approved):

PR #2481, opened 2025-02-28
filter: admv8818: force initialization of SDO

amiclaus:
## Pull Request Description

When a weak pull-up is present on the SDO line, spi_update_bits fails to write both the SOFTRESET and SDOACTIVE bits because it incorrectly reads them as already set.

Since the soft reset disables the SDO line, performing a read-modify-write operation on ADI_SPI_CONFIG_A to enable the SDO line doesn't make sense. This change directly writes to the register instead of using spi_update_bits.

Fixes: 3755205246724f658a89eae985266bcc0185ae48 ("drivers:filter: add admv8818 support")

## PR Type

## PR Checklist

### DIFF
--- a/drivers/filter/admv8818/admv8818.c
+++ b/drivers/filter/admv8818/admv8818.c
@@ -346,19 +346,15 @@ int admv8818_init(struct admv8818_dev **device,
 	if (ret)
 		goto error_dev;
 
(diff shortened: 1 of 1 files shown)


amiclaus (commented):
v2:
- add enough characters of the commit SHA such as it becomes interactive.

PR #2489, opened 2025-03-05
projects: eval-pqmon: add fallback path

romandariana:
Add fallback for new Maxim library path.

## Pull Request Description

The path for libarm_cortexM4l_math.a is now MaximSDK/Libraries/CMSIS/5.9.0/DSP/1.16.2/Lib. Add fallback for this path. Keep old path for compatibility (?).

## PR Type

## PR Checklist

### DIFF
--- a/projects/eval-pqmon/src/platform/maxim/platform_src.mk
+++ b/projects/eval-pqmon/src/platform/maxim/platform_src.mk
@@ -36,4 +36,7 @@ INCS +=	$(MAXIM_LIBRARIES)/MAXUSB/include/core/usb.h		\
 	
 SRCS +=	$(MAXIM_LIBRARIES)/MAXUSB/src/core/usb_event.c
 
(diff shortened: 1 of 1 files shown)


CiprianRegus:
We had a similar situation with the gcc compiler distributed in the Maxim SDK (path changed from `GNUTools/gcc` to `GNUTools/10.1/gcc`): https://github.com/analogdevicesinc/no-OS/blob/main/tools/scripts/maxim.mk#L18

In your case, I don't know how likely it is for the path to change again in the future, but if it doesn't slow down the compilation too much, I'd say a similar approach would save the need for an update down the line.

romandariana:
@CiprianRegus  thank you for the suggestion! I applied something similar in v2.

v1 -> v2:
    - instead of adding fallback, search for the needed library in $(MAXIM_LIBRARIES)/CMSIS/*/DSP/*

CiprianRegus (approved):

PR #2501, opened 2025-03-17
rf-transceiver: ad9361: ad9361.c: Return calib err

danmois:
## Pull Request Description

Return error in case TX_QUAD_CAL or RFDC_CAL fail.

## PR Type

## PR Checklist

### DIFF
--- a/drivers/rf-transceiver/ad9361/ad9361.c
+++ b/drivers/rf-transceiver/ad9361/ad9361.c
@@ -5671,7 +5671,7 @@ int32_t ad9361_setup(struct ad9361_rf_phy *phy)
 int32_t ad9361_do_calib_run(struct ad9361_rf_phy *phy, uint32_t cal,
 			    int32_t arg)
 {
-	int32_t ret;
+	int32_t ret, ret2;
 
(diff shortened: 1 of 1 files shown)


drivers/rf-transceiver/ad9361/ad9361.c
```diff
@@ -5695,11 +5695,11 @@ int32_t ad9361_do_calib_run(struct ad9361_rf_phy *phy, uint32_t cal,
 		break;
 	}
 
-	ret = ad9361_tracking_control(phy, phy->bbdc_track_en,
-				      phy->rfdc_track_en, phy->quad_track_en);
+	ret2 = ad9361_tracking_control(phy, phy->bbdc_track_en,
```
amiclaus, line 5698: why not just check `ret` after the switch..case scenario?

danmois: If the cal fails then we need to restore state.

amiclaus: hmm, conditional jump can works also. but i guess it is fine like this too.

PR #2505, opened 2025-03-20
drivers: digital-io : Fix reg_write function for spi_transfer usage

RaduSabau1:
## Pull Request Description

Some platforms have an spi_transfer function in their platform ops, and some don't.
In case of those who don't, noOS calls no_os_spi_write_then_read for each message,
and one condition is for the rx_buff to be equal to tx_buff, which for these specific
drivers was not true as for the reg_write functions.

This PR fixes this problem by allocating the buffer pointer to both rx_buff and tx_buff,
although for a reg_write the rx_buff, doesn't really matter.

## PR Type

## PR Checklist


amiclaus (commented):
commits that fix things deserve `Fixes:` tag.

PR #2512, opened 2025-03-26
ad9656_fmc: src: app: ad9656_fmc.c: Add delay

danmois:
## Pull Request Description

Add a 10ms delay so that messages have time to be displayed before starting the iio server.

## PR Type

## PR Checklist

### DIFF
--- a/projects/ad9656_fmc/src/app/ad9656_fmc.c
+++ b/projects/ad9656_fmc/src/app/ad9656_fmc.c
@@ -282,6 +282,8 @@ int main(void)
 
 	printf("ad9656: setup, configuration and test program is done\n");
 
+	no_os_mdelay(10);
+
 #ifdef IIO_SUPPORT
 	struct xil_uart_init_param platform_uart_init_par = {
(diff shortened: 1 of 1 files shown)


CiprianRegus:
While the delay would solve the issue, I think the underlying problem is the Xilinx `xil_uart_write()` returns before the UART FIFO is empty. There should be a check if the FIFO is empty somewhere after this loop https://github.com/analogdevicesinc/no-OS/blob/main/drivers/platform/xilinx/xilinx_uart.c#L186-L192

CiprianRegus (approved):
The Xilinx platform doesn't use our UART driver for printf(), so I think this is a fine workaround for now.

buha:
@CiprianRegus i think what you said applies here as well: https://github.com/analogdevicesinc/no-OS/blob/main/iio/iio_app/iio_app.c#L101-L102

buha:
I made an issue, #2514 as a follow up. I'm merging this because this is a separate issue, it seems xilinx's own printf behaves in a nonblocking way and we expect it to be blocking.

PR #2549, opened 2025-04-11
drivers: platform: ftd2xx : Fix GPIO driver

RaduSabau1:
## Pull Request Description

Fix same variable name for different variables in GPIO driver.
Fix syntax error in gpio function.

## PR Type

## PR Checklist

### DIFF
--- a/drivers/platform/ftd2xx/mpsse/ftd2xx_gpio.c
+++ b/drivers/platform/ftd2xx/mpsse/ftd2xx_gpio.c
@@ -259,20 +259,20 @@ int32_t ftd2xx_gpio_get_value(struct no_os_gpio_desc *desc, uint8_t *value)
 {
 	struct ftd2xx_gpio_desc *extra_desc;
 	FT_STATUS status;
-	UCHAR value = 0;
+	UCHAR val = 0;
 	int32_t ret;
 
 	extra_desc = desc->extra;
 	if (no_os_field_get(NO_OS_BIT(desc->number),
 			    extra_desc->pins_dir) == NO_OS_GPIO_IN) {
-		status = FT_ReadGPIO(extra_desc->ftHandle, &value);
+		status = FT_ReadGPIO(extra_desc->ftHandle, &val);
 		if (status != FT_OK) {
 			ret = status;
 			return ret;
 		}
 		extra_desc->pins_val &= ~NO_OS_BIT(desc->number);
 		extra_desc->pins_val |= NO_OS_BIT(desc->number) & no_os_field_prep(NO_OS_BIT(
-						desc->number), no_os_field_get(NO_OS_BIT(desc->number), value));
+						desc->number), no_os_field_get(NO_OS_BIT(desc->number), val));
 	}
 
 	*value = no_os_field_get(NO_OS_BIT(desc->number), extra_desc->pins_val);
--- a/drivers/platform/ftd2xx/mpsse/ftd2xx_gpio.c
+++ b/drivers/platform/ftd2xx/mpsse/ftd2xx_gpio.c
@@ -108,7 +108,7 @@ error:
 int32_t ftd2xx_gpio_get_optional(struct no_os_gpio_desc **desc,
 				 const struct no_os_gpio_init_param *param)
 {
-	if (param = NULL) {
+	if (!param) {
 		*desc = NULL;
 		return 0;
 	}


amiclaus (commented):
these are not fixes, rather code refactoring. so i'd change the commit body/title. otherwise, lgtm.

drivers/platform/ftd2xx/mpsse/ftd2xx_gpio.c
```diff
@@ -259,20 +259,20 @@ int32_t ftd2xx_gpio_get_value(struct no_os_gpio_desc *desc, uint8_t *value)
 {
 	struct ftd2xx_gpio_desc *extra_desc;
 	FT_STATUS status;
-	UCHAR value = 0;
+	UCHAR val = 0;
```
buha, line 262: i dont see the need for this change, and anyway it shouldnt be mixed with the NULL change

RaduSabau1:
> these are not fixes, rather code refactoring. so i'd change the commit body/title. otherwise, lgtm.

First change causes warnings, and I think causes the function not to work as it should.
Second change causes an error at build.
Wouldn't this be marked as a fix in this case?

drivers/platform/ftd2xx/mpsse/ftd2xx_gpio.c
```diff
@@ -108,7 +108,7 @@ int32_t ftd2xx_gpio_get(struct no_os_gpio_desc **desc,
 int32_t ftd2xx_gpio_get_optional(struct no_os_gpio_desc **desc,
 				 const struct no_os_gpio_init_param *param)
 {
-	if (param = NULL) {
+	if (!param) {
```
buha, line 111: i can see a need for this one, probably stdlib.h is not included, but should be separate

drivers/platform/ftd2xx/mpsse/ftd2xx_gpio.c
RaduSabau1: in this case if the variable name isn't changed, it would have the same variable name as the parameter of the function, though they are different variables. Makes the build return an error.

drivers/platform/ftd2xx/mpsse/ftd2xx_gpio.c
RaduSabau1: Making a different commit for this one, in this case.

amiclaus:
> > these are not fixes, rather code refactoring. so i'd change the commit body/title. otherwise, lgtm.
> 
> First change causes warnings, and I also think causes the function not to work as it should. Second change causes an error at build. Wouldn't this be marked as a fix in this case?

That's strange. Probably the issue is somewhere else. if param is a pointer then if (!param) is equivalent to if (param == NULL).

RaduSabau1:
> > > these are not fixes, rather code refactoring. so i'd change the commit body/title. otherwise, lgtm.
> > 
> > 
> > First change causes warnings, and I also think causes the function not to work as it should. Second change causes an error at build. Wouldn't this be marked as a fix in this case?
> 
> That's strange. Probably the issue is somewhere else. if param is a pointer then if (!param) is equivalent to if (param == NULL).

Thought so too, then I saw the line is actually `param = NULL` instead of `param == NULL`

amiclaus:
> > > > these are not fixes, rather code refactoring. so i'd change the commit body/title. otherwise, lgtm.
> > > 
> > > 
> > > First change causes warnings, and I also think causes the function not to work as it should. Second change causes an error at build. Wouldn't this be marked as a fix in this case?
> > 
> > 
> > That's strange. Probably the issue is somewhere else. if param is a pointer then if (!param) is equivalent to if (param == NULL).
> 
> Thought so too, then I saw the line is actually `param = NULL` instead of `param == NULL`

true that. minor stuff, barely noticed by reviewers too.

RaduSabau1:
V2:
- Split commits

drivers/platform/ftd2xx/mpsse/ftd2xx_gpio.c
buha: oh yeah didnt see that, but how come was it up on github then ? would've failed all CI checks

buha (approved):

drivers/platform/ftd2xx/mpsse/ftd2xx_gpio.c
RaduSabau1: I guess, since ftd2xx at this is moment is seen as a platform, the CI didn't run the build for it since there was no CI script implemented for the platform?

amiclaus (approved):

PR #2550, opened 2025-04-11
Add ethernet support to eval-pqmon

romandariana:
## Pull Request Description

This PR adds support for the W5500 Ethernet controller chip, providing standard Ethernet connectivity for no-os projects. The implementation includes:

- Core W5500 driver with SPI interface and socket management
- Network interface adapter to integrate with the no-os network_interface API
- IIO subsystem support for W5500-based network communications
- Integration with the eval-pqmon project, adding a new "ethernet" build option

## PR Type

## PR Checklist

### DIFF
--- a/drivers/net/w5500/w5500.c
+++ b/drivers/net/w5500/w5500.c
@@ -1,0 +1,1118 @@
+/***************************************************************************//**
+*   @file   w5500.c
+*   @brief  Implementation of W5500 Driver
+*   @author Alisa-Dariana Roman (alisa.roman@analog.com)
+********************************************************************************
+* Copyright 2025(c) Analog Devices, Inc.
+*
+* Redistribution and use in source and binary forms, with or without
+* modification, are permitted provided that the following conditions are met:
+*
+* 1. Redistributions of source code must retain the above copyright notice,
+*    this list of conditions and the following disclaimer.
+*
+* 2. Redistributions in binary form must reproduce the above copyright notice,
+*    this list of conditions and the following disclaimer in the documentation
+*    and/or other materials provided with the distribution.
+*
+* 3. Neither the name of Analog Devices, Inc. nor the names of its
+*    contributors may be used to endorse or promote products derived from this
+*    software without specific prior written permission.
+*
+* THIS SOFTWARE IS PROVIDED BY ANALOG DEVICES, INC. “AS IS” AND ANY EXPRESS OR
+* IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF
+* MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO
+* EVENT SHALL ANALOG DEVICES, INC. BE LIABLE FOR ANY DIRECT, INDIRECT,
+* INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+* LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA,
+* OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF
+* LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING
+* NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE,
+* EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+*******************************************************************************/
+
+#include "w5500.h"
+
+/***************************************************************************//**
+ * @brief Write data to a W5500 register
+ *
+ * @param dev   - The device descriptor
+ * @param block - The block select bits (common registers, socket registers, etc.)
+ * @param addr  - The register address
+ * @param data  - The data buffer to write
+ * @param len   - The data length
+ *
+ * @return 0 in case of success, negative error code otherwise
+*******************************************************************************/
+int w5500_reg_write(struct w5500_dev *dev, uint8_t block,
+		    uint16_t addr, const uint8_t *data, uint16_t len)
+{
+	uint8_t spi_tx[3 + len];
+
+	spi_tx[0] = W5500_BYTE_HIGH(addr);
+	spi_tx[1] = W5500_BYTE_LOW(addr);
+
+	spi_tx[2] = W5500_BSB(block) | W5500_RWB_WRITE | W5500_OM_VDM;
+
+	memcpy(&spi_tx[3], data, len);
+
+	struct no_os_spi_msg xfer = {
+		.tx_buff = spi_tx,
+		.rx_buff = NULL,
+		.bytes_number = 3 + len,
+		.cs_change = 1,
+	};
+
+	return no_os_spi_transfer(dev->spi, &xfer, 1);
+}
+
+/***************************************************************************//**
+ * @brief Read data from a W5500 register
+ *
+ * @param dev   - The device descriptor
+ * @param block - The block select bits (common registers, socket registers, etc.)
+ * @param addr  - The register address
+ * @param data  - The buffer to store read data
+ * @param len   - The data length to read
+ *
+ * @return 0 in case of success, negative error code otherwise
+*******************************************************************************/
+int w5500_reg_read(struct w5500_dev *dev, uint8_t block,
+		   uint16_t addr, uint8_t *data, uint16_t len)
+{
+	uint8_t spi_tx[3];
+	uint8_t spi_rx[3 + len];
+	int ret;
+
+	spi_tx[0] = W5500_BYTE_HIGH(addr);
+	spi_tx[1] = W5500_BYTE_LOW(addr);
+
+	spi_tx[2] = W5500_BSB(block) | W5500_RWB_READ | W5500_OM_VDM;
+
+	struct no_os_spi_msg xfer = {
+		.tx_buff = spi_tx,
+		.rx_buff = spi_rx,
+		.bytes_number = 3 + len,
+		.cs_change = 1,
+	};
+
+	ret = no_os_spi_transfer(dev->spi, &xfer, 1);
+	if (ret)
+		return ret;
+
+	memcpy(data, &spi_rx[3], len);
+
+	return 0;
+}
+
+/***************************************************************************//**
+ * @brief Read a 16-bit register reliably by reading until consecutive values match
+ *
+ * @param dev   - The device descriptor
+ * @param block - The block select bits (common registers, socket registers, etc.)
+ * @param addr  - The register address
+ * @param value - Pointer to store the 16-bit value
+ *
+ * @return 0 in case of success, negative error code otherwise
+*******************************************************************************/
+int w5500_read_16bit_reg(struct w5500_dev *dev, uint8_t block,
+			 uint16_t addr, uint16_t *value)
+{
+	int ret;
+	uint8_t data[2];
+	uint16_t val1, val2;
+	int attempts = 0;
+	const int max_attempts = 5;
+
+	do {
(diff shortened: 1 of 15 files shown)


drivers/net/w5500/w5500.h
```diff
@@ -0,0 +1,352 @@
...
+* EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+*******************************************************************************/
+
+#ifndef __W5500_H__
+#define __W5500_H__
+
+#include "network_interface.h"
+
+#include <errno.h>
+#include <string.h>
```
buha, line 40: i think you can move errno.h and string.h in the .c file, possibly also delay.h and alloc.h, to clean things up a bit

buha (commented):
add a builds.json

i see this project doesn't have one, but i would add it at this stage to ensure we don't break this project or the w5500 networking infrastructure

buha:
and make sure to review the #ifdefs you added, they seem to break some existing projects that use networking

drivers/net/w5500/w5500.c
```diff
@@ -0,0 +1,1118 @@
...
+			return ret;
+		val1 = (data[0] << 8) | data[1];
+
+		ret = w5500_reg_read(dev, block, addr, data, 2);
+		if (ret)
+			return ret;
+		val2 = (data[0] << 8) | data[1];
+
+		attempts++;
+	} while (val1 != val2 && attempts < max_attempts);
```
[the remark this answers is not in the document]
romandariana: The documentation recommends reading it twice. Using a loop seems a bit too much indeed. Should I keep the double reading and return an error if the 2 values are different, without the loop?

Regarding the builds.json, I think it was skipped because an external library is provided at build time. For example:

    "pqmon_ethernet": {
      "flags": "INTERFACE=ethernet TARGET=max32650 BASIC_EXAMPLE=y IIOD=y HEAP_SIZE=0x80000 STACK_SIZE=0x20000 PQLIB_PATH=/path/to/pqlib"
    },
    
How should I handle the pqlib?

romandariana:
I pushed a v2 with some fixes for now:

v1 -> v2
 - use /** instead of /*! doxygen syntax in header files
 - moved errno.h and string.h in the .c file, same for delay.h and alloc.h
 - instead of using network_interface.h, i refactored the driver to use driver specific structure for socket address to see if we get rid of some of the build errors this way. a low-level driver would better now depend on higher level network interface. the mapping between the two is done in w5500_network

drivers/net/w5500/w5500.c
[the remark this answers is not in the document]
buha: 1st question, ok, if the datasheet recommends that (very weird!) keep a 2 retry counter then

2nd question, i think we have the infrastructure for this but it has to be discussed with the CI team.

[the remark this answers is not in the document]
buha: can you give me a page number where that retry is documented ? i searched and couldn't find anything in the datasheet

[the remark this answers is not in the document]
romandariana: "Therefore, it is recommended that you read all 16-bits twice or more until getting 
the same value."
I found this on page 54  /  67

romandariana:
v2 -> v3

 - add #ifdef NO_OS_W5500_NETWORKING to w5500_network files to avoid getting it built in random projects

drivers/net/w5500/w5500.c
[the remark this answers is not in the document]
buha: that's crazy, it even says twice or more :D ok, let's leave it as-is.

buha:
ok, i'm dropping the requirement to add builds.json at this point, that's a separate discussion because any build of this project requires pqlib and i am not sure we have the infrastructure now, that can be a separate pr

network/w5500_network/w5500_network.c
```diff
@@ -0,0 +1,643 @@
...
+
+	w5500_protocol = w5500_net_translate_protocol(proto);
+
+	ret = w5500_socket_open(net_dev->mac_dev, physical_id, w5500_protocol, buff_kb);
+	if (ret)
+		return ret;
+
+	socket = &net_dev->sockets[physical_id];
+
+	socket->role = (proto == PROTOCOL_TCP) ? W5500_ROLE_CLIENT : W5500_ROLE_UNUSED;
```
CiprianRegus, line 228: I don't think you can know if the socket will be used by a client or server at this point

projects/eval-pqmon/src/examples/basic/basic_example.c
```diff
@@ -225,6 +225,23 @@ int basic_pqm_firmware()
 	app_init_param.lwip_param.extra = NULL;
 #endif
 
+#if defined(PQM_CONN_ETH)
+	struct w5500_dev *mac_dev;
+	struct w5500_network_dev *net_dev;
+
+	status = w5500_init(&mac_dev, &w5500_ip);
```
CiprianRegus, line 232: I think you can move this part in the `w5500_network_init()` function in order to simplify the application code.

projects/eval-pqmon/src/common/common_data.c
```diff
@@ -33,6 +33,47 @@
...
+	.chip_select = 0,
+	.extra = WIZ_SPI_EXTRA,
+};
+
+struct w5500_init_param w5500_ip = {
+	.spi_init = w5500_spi_init_params,
+	.gpio_reset = w5500_rst_gpio_ip,
+	.gpio_int = w5500_int_gpio_ip,
+	.mac_addr = {0x00, 0x08, 0xDC, 0x01, 0x02, 0x03},
+	.ip_addr = {192, 168, 1, 110},
```
CiprianRegus, line 69: Is there a way to support DHCP besides the static IP assignment? Also, it would be nice if you could define a macro for the IP address and netmask (similarly to this approach: https://github.com/analogdevicesinc/no-OS/blob/main/projects/apard32690/Makefile#L13-L16 , however you can set it directly into the init param). I think that would provide a simple way of building the project with different ip addresses for the server.

network/w5500_network/w5500_network.c
```diff
@@ -0,0 +1,643 @@
...
+		return ret;
+
+	new_socket = &net_dev->sockets[new_physical_id];
+	new_socket->sock_id = sock_id;
+	new_socket->local_port = server_socket->local_port;
+	new_socket->in_use = 1;
+	new_socket->role = W5500_ROLE_SERVER;
+
+	server_socket->sock_id = net_dev->next_virtual_id++;
+	server_socket->role = W5500_ROLE_CLIENT;
```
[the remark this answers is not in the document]
romandariana: If I understood correctly the iio module where accept is called expects a new client socket to be created and the server socket keeps listening for new connections. But for W5500, the socket that is listening changes status and becomes a "client" socket. To bypass this, in w5500 accept function I created a new socket that will be the server socket, thus the 3 function callings above, then I switch the virtual ids. Should I extract this switch in another static function to make things clearer?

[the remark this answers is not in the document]
CiprianRegus: I'd say just adding a comment explaining this behavior would be good enough.

drivers/net/w5500/w5500.c
```diff
@@ -0,0 +1,1099 @@
...
+ * @param data  - The buffer to store read data
+ * @param len   - The data length to read
+ *
+ * @return 0 in case of success, negative error code otherwise
+*******************************************************************************/
+int w5500_reg_read(struct w5500_dev *dev, uint8_t block,
+		   uint16_t addr, uint8_t *data, uint16_t len)
+{
+	uint8_t spi_tx[3];
+	uint8_t spi_rx[3 + len];
```
CiprianRegus, line 89: You can use just one buffer.

drivers/net/w5500/w5500.c
```diff
@@ -0,0 +1,1099 @@
...
+
+	ret = w5500_reg_write(dev, W5500_COMMON_REG, W5500_RTR, rtr_buf, 2);
+	if (ret)
+		return ret;
+
+	ret = w5500_reg_write(dev, W5500_COMMON_REG, W5500_RCR, &retry_count, 1);
+	if (ret)
+		return ret;
+
+	return 0;
```
CiprianRegus, line 305: You can return directly

drivers/net/w5500/w5500.c
```diff
@@ -0,0 +1,1099 @@
...
+	ret = w5500_socket_reg_read(dev, sock_id, W5500_Sn_SR, &status, 1);
+	if (ret)
+		return ret;
+
+	if (status != W5500_Sn_SR_ESTABLISHED &&
+	    status != W5500_Sn_SR_UDP &&
+	    status != W5500_Sn_SR_MACRAW)
+		return -ENOTCONN;
+
+	while (total_sent < len) {
```
CiprianRegus, line 613: Would it be possible to remain stuck in this loop (e.g the other side disconnects while the buffer is full)? If so, I'd add a timeout.

drivers/net/w5500/w5500.c
```diff
@@ -0,0 +1,1099 @@
...
+ * @brief Receive data from a socket
+ *
+ * @param dev     - Device descriptor
+ * @param sock_id - Socket number (0-7)
+ * @param buf     - Buffer to store received data
+ * @param len     - Maximum length to receive
+ *
+ * @return Number of bytes received on success, negative error code otherwise
+*******************************************************************************/
+int w5500_socket_recv(struct w5500_dev *dev, uint8_t sock_id, void *buf,
```
CiprianRegus, line 666: By default, `recv()` is a blocking function. I think you should implement it the same way you did with `send()` above.

drivers/net/w5500/w5500.c
```diff
@@ -0,0 +1,1099 @@
...
+ *
+ * @param dev     - Device descriptor
+ * @param sock_id - Socket number (0-7)
+ * @param buf     - Data buffer to send
+ * @param len     - Data length
+ * @param to      - Destination address (IP and port)
+ *
+ * @return 0 on success, negative error code otherwise
+*******************************************************************************/
+int w5500_socket_sendto(struct w5500_dev *dev, uint8_t sock_id, const void *buf,
```
CiprianRegus, line 739: This should block as well.

drivers/net/w5500/w5500.c
```diff
@@ -0,0 +1,1099 @@
...
+	if (ret)
+		goto free_gpio_int;
+
+	memcpy(dev->mac_addr, init_param->mac_addr, 6);
+	memcpy(dev->ip_addr, init_param->ip_addr, 4);
+	memcpy(dev->subnet_mask, init_param->subnet_mask, 4);
+	memcpy(dev->gateway, init_param->gateway, 4);
+
+	memcpy(&dev->retry_time, &init_param->retry_time, 1);
+	memcpy(&dev->retry_count, &init_param->retry_count, 1);
```
CiprianRegus, line 1057: The W5500 seems to have some default values for these that look reasonable. I think these 2 fields in the init parameter should be "optional" in the sense that if they are 0, you don't actually write the register. This way you don't force the users to check the datasheet.

drivers/net/w5500/w5500.c
```diff
@@ -0,0 +1,1099 @@
...
+ *
+ * @return 0 in case of success, negative error code otherwise
+*******************************************************************************/
+int w5500_init(struct w5500_dev **device,
+	       struct w5500_init_param *init_param)
+{
+	struct w5500_dev *dev;
+	int ret;
+
+	dev = (struct w5500_dev *)malloc(sizeof(*dev));
```
CiprianRegus, line 1035: Use `no_os_calloc()` instead to zero out the memory.

CiprianRegus (dismissed):

romandariana:
v5 -> v6
    - solve the dev leak in w5500_network_init()

romandariana:
v6 -> v7:
   - new commit with builds.json

romandariana:
v7 -> v8 -> v9:
    - add mac_dev clean up in w5500_network_init()
    
Note: If the maxim projects builds fails again, I am pretty sure it is not because of these commits. I saw another pr builds fail because of the same project error.

buha (approved):

PR #2601, opened 2025-05-15
drivers: platform: ftd2xx : Fix FTD2XX library GPIO driver.

RaduSabau1:
## Pull Request Description
If another pin from the same port is being used, the pins direction won't be remembered, therefore fix this issue by having a global variable for it.

Also add global variable for values, since it encounters the same problem, especially when setting a new value for another pin, which can break the other pin values.

Fix : 5a9583a ("drivers: platform: ftd2xx : Add GPIO support for FTD2XX platform:)

## PR Type

## PR Checklist


drivers/platform/ftd2xx/mpsse/ftd2xx_gpio.h
```diff
@@ -38,7 +38,13 @@
 #include "no_os_gpio.h"
 #include "ftd2xx.h"
 
+#define FTD2XX_GPIO_PIN(x)		NO_OS_BIT(x)
+
 #define FTD2XX_MAX_PIN_NB		7
+#define FTD2XX_MAX_PORT_NB		4
+
+extern uint8_t ftd2xx_gpio_pins_dir[FTD2XX_MAX_PORT_NB];
+extern uint8_t ftd2xx_gpio_pins_val[FTD2XX_MAX_PORT_NB];
```
CiprianRegus, line 47: Why do you want to make these available in the header? Looks like an implementation detail that should only be accessible to the driver. Also, you will only allow one instance of the driver by having a single set of arrays.

drivers/platform/ftd2xx/mpsse/ftd2xx_gpio.h
```diff
@@ -50,8 +56,10 @@ extern const struct no_os_gpio_platform_ops ftd2xx_gpio_ops;
  * @brief Structure holding the initialization parameters for ftd2xx platform
  */
 struct ftd2xx_gpio_init {
-	/* Initial direction of the port pins at initialization time. */
-	uint8_t pins_dir;
+	/* Initial direction of pin at initialization time. */
+	uint8_t pin_dir;
+	/* Initial value of pin at initialization time.*/
+	uint8_t pin_val;
```
CiprianRegus, line 62: I don't think having this struct is a good idea. The direction and value of a GPIO should be set using the `no_os_gpio_` API.

RaduSabau1: True, I will remove this structure.

PR #2702, opened 2025-08-20
drivers:platform:stm32:dma: Fix DMA Abort API

NagaHimanshu:
The DMA_Abort_IT disables the DMA and rest of the flags are cleared in the ISR. At this point, the execution do not reach ISR since the IRQ is disabled in no_os_dma_abort platform's abort is called. Hence abort needs to be done without ISR branch.

HAL returns an error when application tries to abort an already completed DMA transaction.

## Pull Request Description

## PR Type

## PR Checklist

### DIFF
--- a/drivers/platform/stm32/stm32_dma.c
+++ b/drivers/platform/stm32/stm32_dma.c
@@ -247,7 +247,10 @@ int stm32_dma_xfer_abort(struct no_os_dma_desc* desc,
 
 	ret = HAL_DMA_Abort(sdma_chan->hdma);
 
-	if (ret != HAL_OK)
+	/* If transfer already completed, the state changes to READY
+	 * and hence aborting again would lead to error from HAL */
+	if ((ret != HAL_OK) &&
+	    (sdma_chan->hdma->ErrorCode != HAL_DMA_ERROR_NO_XFER))
 		return -EINVAL;
 
 	return 0;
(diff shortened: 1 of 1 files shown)


drivers/platform/stm32/stm32_dma.c
```diff
@@ -245,12 +245,12 @@ int stm32_dma_xfer_abort(struct no_os_dma_desc* desc,
 
 	sdma_chan = chan->extra;
 
-	if (chan->irq_num)
-		ret = HAL_DMA_Abort_IT(sdma_chan->hdma);
-	else
-		ret = HAL_DMA_Abort(sdma_chan->hdma);
+	ret = HAL_DMA_Abort(sdma_chan->hdma);
```
buha, line 248: There are 2 changes here:
1. the abort type change
2. the hal error on aborting 2 times in a row

I think you should split them in 2 commits.
1st commit, I'm not sure I understand the problem.
2nd commit, ok, seems like a legit fix.

NagaHimanshu: Sure, I split into two separate commits.

The first one is like this - 
in no_os_dma.c we will be disabling the IRQ before the stm32_dma_abort is being called.
When the DMA is disabled in DMA_Abort_IT, it raises an interrupt, which branches to an ISR and clear the flags and put the DMA into ready state.

But since we already disabled IRQ, it will never reach the DMA handler and dma state wont be ready and still flags are not cleared.

buha (approved):

PR #2725, opened 2025-09-17
drivers:adc:ad7768-1: Fix minor issues

Shetty-Anush:
1. Dynamically size buf array in ad77681_compute_xor() according to data_size argument
2. Fix switch logic in ad77681_update_sample_rate() to handle SINC5 and FIR filters separately
3. Remove unused data_frame_byte field from ad77681_init_param and ad77681_setup()

## Pull Request Description

## PR Type

## PR Checklist

### DIFF
--- a/drivers/adc/ad7768-1/ad77681.c
+++ b/drivers/adc/ad7768-1/ad77681.c
@@ -448,7 +448,8 @@ int32_t ad77681_update_sample_rate(struct ad77681_dev *dev)
 
 	/* Finding out decimation ratio */
 	switch (dev->filter) {
-	case (AD77681_SINC5 | AD77681_FIR):
+	case AD77681_SINC5:
+	case AD77681_FIR:
 		/* Decimation ratio of FIR or SINC5 (x32 to x1024) */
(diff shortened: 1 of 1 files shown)


drivers/adc/ad7768-1/ad77681.c
```diff
@@ -80,7 +80,7 @@ uint8_t ad77681_compute_xor(uint8_t *data,
 			    uint8_t init_val)
 {
 	uint8_t crc = init_val;
-	uint8_t buf[3];
+	uint8_t buf[data_size];
```
buha: if you can break the description of the patch into 1) 2) 3) bullet points, you can also break the commits into individual commits, please break it into separate commits

buha: now regarding this change specifically, I don't really like VLA's (variable length array), they have issues like compiler compatibility, C standard compatibility, possible stack overflow etc. but since it's bound to max 256 size due to the uint8_t, i think it's acceptable to use it here, it does keep things simple

PR #2772, opened 2025-10-06
build_projects.yml: skip projects build if it's a documentation PR

AAndrisa:
## Pull Request Description

If the documentation label is set to the opened PR then we skip the projects building jobs.
The condition on the jobs is: if the build reason is different from PR build anyway, and if it's a PR, build just if the documentation label is not found.

## PR Type

## PR Checklist

### DIFF
--- a/build_projects.yml
+++ b/build_projects.yml
@@ -25,7 +25,24 @@ variables:
     value: $(Build.SourceBranchName)
 
 jobs:
+- job: CheckLabel
+  pool:
+    vmImage: ubuntu-latest
+  steps:
+  - bash: |
+      sudo apt install curl
+      if curl -s "https://api.github.com/repos/$BUILD_REPOSITORY_ID/issues/$SYSTEM_PULLREQUEST_PULLREQUESTNUMBER/labels" | grep '"name": "documentation"'
+      then
+        echo "##vso[task.setvariable variable=prHasLabel;isOutput=true]true"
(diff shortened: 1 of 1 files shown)


amiclaus (commented):
I am not sure how useful this approach is, taking into consideration that people (especially the external contributers) don't add labels to the PRs (or add the labels after the PR is opened).

I'd rather check if the PR contains only changes in the `doc` folder or `README.rst` files and skip the unnecessary builds.

gastmaier (commented):
If you use jq, know the base and head commit and has the gh-token
, you can do the diff api call, jq to get changed files list, then grep inverse
```
base_sha=c57ad676601eeaaae079ba957226aac4946cf578
head_sha=ed74585994569c068c25f23ccdf45173ffce731e
curl -s -L  \
  -H "Accept: application/vnd.github+json" \
  -H "Authorization: Bearer $gh_token"  \
  -H "X-GitHub-Api-Version: 2022-11-28"  \
      "https://api.github.com/repos/analogdevicesinc/no-OS/compare/$base_sha...$head_sha" | \
   jq '.files[].filename' -r | \
   grep -v '^docs/' && \ 
  echo "others changed" || \
  echo "only docs changed"
```

I do this api call for the checkout that unrolls the merge commit at
https://github.com/analogdevicesinc/doctools/blob/action/checkout/action.yml#L79-L84
and wondered a few times how to do what we are discussing here, so decide to chime in as well and try out other solution

PR #2963, opened 2026-03-05
Xilinx no os hello world

stefpopa:
## Pull Request Description

projects: hello_world: Add Hello World example project
    
Add a minimal no-OS example project targeting Xilinx Zynq-7000 and ZynqMP platforms. The project demonstrates the core no-OS APIs (UART init, pr_info logging, no_os_mdelay delay) and is structured following the iio_demo multi-platform pattern so that support for additional platforms can be added by dropping in a new src/platform/<platform>/ directory.

## PR Type

## PR Checklist

### DIFF
--- a/projects/hello_world/README.rst
+++ b/projects/hello_world/README.rst
@@ -1,0 +1,239 @@
+Hello World no-OS Example Project
+==================================
+
+Overview
+--------
+
+This is the simplest possible no-OS project: it initialises the board's UART
+peripheral and prints a "Hello World" counter message over the serial port once
+per second.  It is intended as a **quick-start template** for engineers who are
+new to no-OS and want to understand how to build, flash, and run a bare-metal
+embedded application on a Xilinx board.
+
+What it demonstrates:
+
+- How a no-OS project is structured (``Makefile``, ``src.mk``, platform tree)
+- The no-OS peripheral descriptor pattern (``no_os_uart_init`` / ``no_os_uart_remove``)
+- The no-OS logging macros (``pr_info``, ``pr_err``, …)
+- The no-OS delay API (``no_os_mdelay``)
+- Proper error handling (checking return values, cleanup on failure)
+
+Supported Platforms
+-------------------
+
+Xilinx
+^^^^^^
+
+The project is generic — no board-specific code is required.  The same source
+builds correctly for:
+
+- **Zynq-7000**: Cora Z7, ZED, ZC706, and any other Zynq-7000 carrier
+- **ZynqMP (Zynq UltraScale+)**: ZCU102, and any other ZynqMP carrier
+
+The target architecture (Cortex-A9 for Zynq-7000, Cortex-A53 for ZynqMP) is
+auto-detected from the XSA file you provide — see `Hardware Requirements`_
+below.
+
+Project Layout
+--------------
+
+::
+
+   no-OS/projects/hello_world/
+   ├── Makefile
+   ├── README.rst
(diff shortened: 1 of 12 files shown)


projects/hello_world/src/platform/platform_includes.h
```diff
@@ -0,0 +1,50 @@
...
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA,
+ * OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF
+ * LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING
+ * NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE,
+ * EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+*******************************************************************************/
+#ifndef __PLATFORM_INCLUDES_H__
+#define __PLATFORM_INCLUDES_H__
+
+#ifdef STM32_PLATFORM
```
buha: i recommend using the examples infrastructure, you can simply do `#include "parameters.h"` due to this line here: https://github.com/analogdevicesinc/no-OS/blob/main/tools/scripts/examples.mk#L6, you just need to have a specific project directory tree

projects/hello_world/README.rst
```diff
@@ -0,0 +1,239 @@
...
+
+   # 3. Build
+   make
+
+Useful build variants:
+
+.. code-block:: bash
+
+   # Build with full debug symbols and -O0 optimisation (recommended for stepping through code)
+   make DEBUG=1
```
buha, line 112: also btw, RELEASE=y is indeed documented in a few places, so maybe just stick to it:
https://github.com/search?q=repo%3Aanalogdevicesinc%2Fno-OS%20RELEASE%3Dy&type=code

stefpopa: The RELEASE=y indeed adds -O2 optimization, but does not remove the debug symbols. However, by default we have -O0, with no debug symbols, the problem here is that we get slow code, but can't even attach a debugger usefully. 

The DEBUG=1 does more:

- Adds -g3 (full debug symbols) + -fdebug-prefix-map=...
- Forces -O0 (strips any other -O flags in the pre_build step)
- Explicitly disables RELEASE

What we can do is to add -g unconditionally:
```

# Default: always build with debug symbols
CFLAGS += -g

ifeq (y,$(strip $(RELEASE)))
CFLAGS += -O2
CFLAGS := $(filter-out -g -g1 -g2 -g3,$(CFLAGS))
endif

```
Then, we can remove DEBUG=1. The only issue is that -g will increase the binary size on MCUs, for Xilinx this is irrelevant, but it might affect other platforms. But I think other platforms such as stm32, ADuCM, or Pico, already hardcode their own -g variants. Please contradict me if I am wrong

buha: fair point, i think that's a good argument to have both a release flag and a debug flag, we needed the release flag exactly for size constraints so adding -g unconditionally is not a good idea

projects/hello_world/src/platform/platform_includes.h
stefpopa: done

amiclaus (dismissed):

buha (approved):

PR #3007, opened 2026-03-19
projects: adalm-lsmspg: update builds.json

nielacuna:
1. Synchronize naming of builds.json with a relative project adalm-lsmspg.

## Pull Request Description

## PR Type

## PR Checklist
- [X ] I have followed the [Coding style guidelines](http://analogdevicesinc.github.io/no-OS/drivers_guide.html#coding-style)
- [X] I have complied with the [Submission Checklist](http://analogdevicesinc.github.io/no-OS/contributing.html#submission-checklist)
- [X] I have build all projects affected by the changes in this PR
- [X] I have tested in hardware affected projects, at the relevant boards
- [X] I have updated the documentation (wiki pages, ReadMe etc), if applies

### DIFF
--- a/projects/adalm-mmsc/builds.json
+++ b/projects/adalm-mmsc/builds.json
@@ -1,7 +1,7 @@
 {
-	"maxim": {
-		"iio_example_max32665": {
-			"flags" : "TARGET=max32665"
-		}
-	}
+  "maxim": {
+    "iio":  {
+      "flags" : "TARGET=max32665"
+    }
+  }
 }


amiclaus (approved):

amiclaus (commented):
@nielacuna is this ready for review?

nielacuna:
@amiclaus , thank you for pinging me about this. @thorenscientific , Hi mark! Would like to get your go/no-go for this PR.

thorenscientific:
Hi @nacuna-adi , @amiclaus ,
Wow, I got confused there for a second - the title of the PR is adalm-lsmspg, but we're working on the mmsc builds.json.

It actually looks like your original is more in line with my favorite example:
https://github.com/analogdevicesinc/no-OS/blob/main/projects/eval-adxl355-pmdz/builds.json

top-level object specifies family (maxim:), nested object is named such that describes what is generated, and flags tell what key/value to use to generate said object. We don't need to append ad4080 since we've only got a single part. It IS conceivable that we may some day generate some dummy examples that dump CSV to the terminal, not unlike the lsmspg.

Toni what do you think? Should we leave as-is?

And extending this idea to the lsmspg - it looks like maybe we should add the processor name to that one. But let's hold off, it's likely that we'll be porting that to other processors, just not sure which at the moment. We can add at that time.

amiclaus (commented):
> Hi @nacuna-adi , @amiclaus , Wow, I got confused there for a second - the title of the PR is adalm-lsmspg, but we're working on the mmsc builds.json.
> 
> It actually looks like your original is more in line with my favorite example: https://github.com/analogdevicesinc/no-OS/blob/main/projects/eval-adxl355-pmdz/builds.json
> 
> top-level object specifies family (maxim:), nested object is named such that describes what is generated, and flags tell what key/value to use to generate said object. We don't need to append ad4080 since we've only got a single part. It IS conceivable that we may some day generate some dummy examples that dump CSV to the terminal, not unlike the lsmspg.
> 
> Toni what do you think? Should we leave as-is?
> 
> And extending this idea to the lsmspg - it looks like maybe we should add the processor name to that one. But let's hold off, it's likely that we'll be porting that to other processors, just not sure which at the moment. We can add at that time.

 Agree, let's leave the mmsc builds.json as-is for now. The current naming already follows the common pattern and there's no reason to add processor suffixes until we actually have multiple targets. We can revisit both mmsc and lsmspg naming when we start porting to other processors.

PR #3059, opened 2026-04-07
driver: Add network support for Xilinx boards

DeaconescuDaniel:
## Pull Request Description

Added ethernet support for xilinx boards using the XEMacPs driver, following the [AMD Programming Guide](https://docs.amd.com/r/en-US/ug585-zynq-7000-SoC-TRM/Programming-Guide?tocId=dp5v~GxO_gy4dRHC6sYxjQ).

## Components:

- Standalone XEmacPs MAC driver(drivers/net/xemacps/): Includes PHY registration, autonegotiation, and init/remove, send/recv/poll functionality. Note: Zynq-7000 processors have a [hardware bug](https://adaptivesupport.amd.com/s/article/52028?language=en_US) that causes the RX data path to lock up under heavy traffic, requiring resets. 
- LwIP integration layer(network/lwip_raw_socket/netdevs/xemacps/): Integration layer for LwIP that calls the driver code, similar to the adin1110 driver
- LwIP raw socket updates (network/lwip_raw_socket/lwip_socket.c): removed dependency on adin1110 driver, updated heap allocations, removed unused variables.
- Project examples (projects/xilinx_lwip/) : an LwIP TCP echo server and an IIO server over ethernet.
- README file detailing build commands, pre-requisites and usages

## PR Type

## PR Checklist

### DIFF
--- a/drivers/net/xemacps/no_os_xemacps.c
+++ b/drivers/net/xemacps/no_os_xemacps.c
@@ -441,10 +441,14 @@ int xemacps_init(struct xemacps_desc **desc, struct xemacps_init_param *param)
 #endif
 #endif
 
-	cfg = XEmacPs_LookupConfig(param->device_id);
+#ifdef SDT
+	cfg = XEmacPs_LookupConfig((UINTPTR)param->device_id);
+#else
+	cfg = XEmacPs_LookupConfig((uint16_t)param->device_id);
+#endif
 	if (!cfg) {
-		printf("xemacps: LookupConfig failed for device %u\n",
-		       param->device_id);
+		printf("xemacps: LookupConfig failed for identifier 0x%lx\n",
+		       (unsigned long)param->device_id);
 		ret = -ENODEV;
 		goto free_desc;
 	}
--- a/drivers/net/xemacps/no_os_xemacps.c
+++ b/drivers/net/xemacps/no_os_xemacps.c
@@ -1,0 +1,814 @@
+/***************************************************************************//**
+ *   @file   no_os_xemacps.c
+ *   @brief  Standalone MAC driver for the Xilinx PS GEM (XEmacPs).
+ *
+ *   Provides init/remove/send/recv/poll for the Zynq-7000 / ZynqMP PS
+ *   Gigabit Ethernet MAC. Uses polling (no IRQs) and copy-mode DMA.
+ *   No network stack dependency - the upper layer (e.g. lwIP glue)
+ *   calls xemacps_send/recv with raw Ethernet frame buffers.
+ *
+ *   BD (buffer descriptor) memory is placed in a dedicated uncached region
+ *   following the Xilinx reference implementation (xemacpsif_dma.c).
+ *   Xil_SetTlbAttributes() only works with L1 page entries (1MB),
+ *   so a 1 MB static array is reserved and marked DEVICE_MEMORY.
+ *
+ *   This driver is limited to a single instance due to the static buffer
+ *   arrays (bd_space, tx_buf, rx_buf).
+ *
+ *   @author Nicolae-Daniel Deaconescu (Nicolae-daniel.Deaconescu@analog.com)
+********************************************************************************
+ * Copyright 2025(c) Analog Devices, Inc.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice,
+ *    this list of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * 3. Neither the name of Analog Devices, Inc. nor the names of its
+ *    contributors may be used to endorse or promote products derived from this
+ *    software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY ANALOG DEVICES, INC. "AS IS" AND ANY EXPRESS OR
+ * IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF
+ * MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO
+ * EVENT SHALL ANALOG DEVICES, INC. BE LIABLE FOR ANY DIRECT, INDIRECT,
+ * INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA,
+ * OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF
+ * LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING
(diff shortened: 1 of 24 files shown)


drivers/net/xemacps/no_os_xemacps.c
```diff
@@ -0,0 +1,546 @@
...
+
+#include "no_os_xemacps.h"
+#include "no_os_delay.h"
+#include "no_os_alloc.h"
+
+/* Maximum time to wait for PHY link-up (ms) */
+#define XGEM_PHY_LINK_TIMEOUT_MS	5000
+
+/* Uncached memory region for DMA buffer descriptors */
+static uint8_t bd_space[0x100000] __attribute__((aligned(0x100000)));
```
CiprianRegus, line 63: Do we require 1 MB of DMA buffer memory to reach the full bandwidth of the MAC? There may be cases in which we can poll and receive data more often, so then we can decrease the size of this buffer. I think we can define a macro like this, that also allows the user to configure the buffer size at compile time. 

#ifndef XEMACPS_BD_SPACE_SIZE
#define XEMACPS_BD_SPACE_SIZE    0x100000
#endif

static uint8_t bd_space[XEMACPS_BD_SPACE_SIZE] ...

drivers/net/xemacps/no_os_xemacps.h
```diff
@@ -0,0 +1,104 @@
...
+#define XGEM_CACHE_LINE_SIZE	64
+#else
+#define XGEM_CACHE_LINE_SIZE	32
+#endif
+
+/*
+ * Frame buffer size, rounded up to the next cache-line multiple.
+ */
+#define XGEM_BUFF_SIZE	(((XEMACPS_MAX_FRAME_SIZE) + XGEM_CACHE_LINE_SIZE - 1u) \
+							& ~(XGEM_CACHE_LINE_SIZE - 1u))
```
CiprianRegus: You can use the `no_os_round_up()` macro instead of manually aligning: https://github.com/analogdevicesinc/no-OS/blob/main/include/no_os_util.h#L78

drivers/net/xemacps/no_os_xemacps.h
```diff
@@ -0,0 +1,104 @@
...
+
+/**
+ * @struct xemacps_init_param
+ * @brief  Initialization parameters for the XEmacPs MAC driver.
+ */
+struct xemacps_init_param {
+	/** XEmacPs device ID from xparameters.h (e.g. XPAR_XEMACPS_0_DEVICE_ID) */
+	uint16_t device_id;
+	/** MDIO address of the on-board PHY */
+	uint32_t phy_addr;
```
CiprianRegus: Nitpick, but the type can be `uint8_t`, since you can only have 32 PHYs on an MDIO bus.

drivers/net/xemacps/no_os_xemacps.c
```diff
@@ -0,0 +1,546 @@
...
+		return -EIO;
+
+	bmcr |= PHY_REG_BMCR_ANEGEN | PHY_REG_BMCR_RSTANEG;
+	ret = XEmacPs_PhyWrite(emac, phy_addr, PHY_REG_BMCR, bmcr);
+	if (ret != XST_SUCCESS)
+		return -EIO;
+
+	do {
+		XEmacPs_PhyRead(emac, phy_addr, PHY_REG_BMSR, &bmsr);
+		ret = XEmacPs_PhyRead(emac, phy_addr, PHY_REG_BMSR, &bmsr);
```
CiprianRegus: Do you need to read the status register twice?

drivers/net/xemacps/no_os_xemacps.c
```diff
@@ -0,0 +1,546 @@
...
+	ret = XEmacPs_CfgInitialize(&d->emac, cfg, cfg->BaseAddress);
+	if (ret != XST_SUCCESS) {
+		ret = -EIO;
+		goto free_desc;
+	}
+
+	XEmacPs_SetMacAddress(&d->emac, param->hwaddr, 1);
+	XEmacPs_SetMdioDivisor(&d->emac, MDC_DIV_224);
+
+	speed = xgem_phy_setup(&d->emac, param->phy_addr);
```
CiprianRegus: This driver will fail if the link is not up during the `xemacps_init()` call, right? Having the PHY link up should't be a requirement for a MAC driver to register properly. What can be done instead is to periodically verify the status of the link (e.g Linux does this once per second) and store it somewhere in `struct xemacps_desc`. Since there is a poll function, I think we can verify the PHY link there. Then, frame RX/TX operation should only be done if the link state is up.

drivers/net/xemacps/no_os_xemacps.c
DeaconescuDaniel: The cortex-a9 limits the Xil_SetTlbAttributes function to only work with 1MB sections, it would be a good idea for portability, if the bsp supports higher granularity, but we are forced to use 1MB for this processor.
Source: [Cortex A9 BSP](https://github.com/Xilinx/embeddedsw/blob/master/lib/bsp/standalone/src/arm/cortexa9/xil_mmu.c#L71)
```C
/************************** Constant Definitions *****************************/
#define     ARM_AR_MEM_TTB_SECT_SIZE               1024*1024
#define     ARM_AR_MEM_TTB_SECT_SIZE_MASK          (~(ARM_AR_MEM_TTB_SECT_SIZE-1UL))
/************************** Variable Definitions *****************************/

extern u32 MMUTable;

/************************** Function Prototypes ******************************/

/*****************************************************************************/
/**
* @brief	This function sets the memory attributes for a section covering 1MB
*			of memory in the translation table.
*
* @param	Addr: 32-bit address for which memory attributes need to be set.
* @param	attrib: Attribute for the given memory region. xil_mmu.h contains
*			definitions of commonly used memory attributes which can be
*			utilized for this function.
*
*
* @return	None.
*
* @note		The MMU or D-cache does not need to be disabled before changing a
*			translation table entry.
*
******************************************************************************/
void Xil_SetTlbAttributes(INTPTR Addr, u32 attrib)
{
	u32 *ptr;
	u32 section;

	section = Addr / 0x100000U;
```

drivers/net/xemacps/no_os_xemacps.h
DeaconescuDaniel: Fixed, i changed it to no_os_align, it was the same macro

drivers/net/xemacps/no_os_xemacps.h
DeaconescuDaniel: Fixed

drivers/net/xemacps/no_os_xemacps.c
DeaconescuDaniel: Usually the bmsr is a latch low register, which means we need to read twice to get the updated value, but in a loop it doesn't matter since we read again on the next iteration anyway(this section was removed in the latest commit)

drivers/net/xemacps/no_os_xemacps.c
DeaconescuDaniel: You're right, i moved the phy polling to the poll function, this way it supports connects and disconnects and we don't busy wait for the connection

CiprianRegus:
I think this pull request looks good. One thing we can maybe change is to convert the sample project from being specific to the Cora Z7 to an example of using the Ethernet driver for the Xilinx platform. This way, we can add support for other Xilinx boards as well.

DeaconescuDaniel:
You're right, i moved the projects to the xilinx_lwip folder, and added support for defining the mdio address, so that for a different board we can just change the address and have it work. (I have not yet tested on other boards, but i think that's all we have to do for compatibility).
I also found a bug while doing this, if we use a non existent mdio address, the bmcr returns 0xffff, which sets the bit for 1000mbps, making the driver think there is a link when there isn't one, it worked by coincidence, but now it exits correctly.

PR #3087, opened 2026-05-04
capi: add capi phy driver implementation

DeaconescuDaniel:
## Pull Request Description

Added the [common api](https://github.com/adi-innersource/core-sw-common-hal/pull/153/changes) phy interface files along with a sample implementation for the Marvell 88E1510.
This branch includes the previous xemacps driver for testing purposes. The driver was refactored to work with the new phy implementation, although the xemacps part can be removed if the phy driver is approved to merge only the relevant part of the pull request..

## PR Type

## PR Checklist


drivers/net/phy/mrvl_88e1510/capi_marvell_88e1510.c
```diff
@@ -0,0 +1,524 @@
...
+	if (!dev)
+		return -EINVAL;
+
+	dev->interface = interface;
+
+	if (interface == CAPI_ETH_INTERFACE_RGMII) {
+		ret = mrvl_config_rgmii_delays(dev);
+		if (ret)
+			return ret;
+		return mrvl_soft_reset(dev);
```
CiprianRegus, line 319: Do we have to reset this PHY after configuring the delays? Wouldn't that overwrite the delay settings?

drivers/net/phy/mrvl_88e1510/capi_marvell_88e1510.c
```diff
@@ -0,0 +1,524 @@
...
+static void mrvl_isr(void *handle)
+{
+	struct mrvl_88e1510_handle *dev = (struct mrvl_88e1510_handle *)handle;
+	uint16_t irq_status, pss;
+	uint32_t events = 0;
+
+	if (!dev || !dev->callback)
+		return;
+
+	if (mrvl_read(dev, MRVL_REG_IRQ_STATUS, &irq_status))
```
CiprianRegus, line 492: This should be a point of discussion for the capi_phy interface. If `mrvl_isr()` is called from interrupt context, directly calling MDIO functions is wrong. In case a MDIO transfer was already in progress, this will likely fail in some way. There could be two cases in which we can get the information here:

1. Periodic polling from the application (or lwip).
2. Periodic polling on an interrupt flag, set in the `mrvl_isr`. Interrupt -> setting a "IRQ triggered" flag.

CiprianRegus (commented):
Apart from the interrupt issue, the 88e1510 driver looks good to me. Other than that, capy_phy should expose functions for c22 and c45 MDIO transfers. The current model (calling fn_read/write directly) is a bit weird and doesn't follow the other CAPI drivers.

drivers/net/phy/mrvl_88e1510/capi_marvell_88e1510.c
DeaconescuDaniel: Changing the interface mode requires a software reset, the bits are retained but don't take effect until a reset is performed

Image

drivers/net/phy/mrvl_88e1510/capi_marvell_88e1510.c
DeaconescuDaniel: Would it be possible to serialize the operations by using a mutex to make sure we don't break other mdio transfers?

Issue #2339, opened 2024-10-16
peripheral drivers should have a config() API


buha:
There is an issue with UART being needlessly reinitialized by calling remove() and init() API's only to change the baudrate. This kind of problem could be solved by adding a mandatory config() API to the peripheral drivers so that the minimal API set would look like this:

    init() - allocate memory, calls config() internally
    config() - configure the peripheral
    remove() - deallocate memory (and put the hardware in a good state, if needed)

The user could then reconfigure the peripheral with a new set of init_param structure without deallocating/reallocating memory.

buha:
closing this because no-OS API will be obsoleted anyway, and the issue seems to have a partial solution with capi uart

PR #3090, opened 2026-05-05
Add STM32 CAPI peripherals implementation

cristina-suteu:
## Pull Request Description

This PR adds CAPI based implementations to several STM32 platform drivers: GPIO, DMA, IRQ, I2C, SPI, UART and TIMER.
## PR Type

## PR Checklist

### DIFF
--- a/capi/inc/capi_alloc.h
+++ b/capi/inc/capi_alloc.h
@@ -1,0 +1,75 @@
+/*
+ * Copyright (c) 2025 Analog Devices, Inc.
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+/**
+ * @file
+ * @brief Common API Memory Allocation
+ *
+ * Overrideable allocation suite for CAPI drivers.
+ * Provides malloc/free/calloc/realloc functions that can be overridden
+ * at link time by providing strong implementations.
+ */
+
+#pragma once
+#include <stddef.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/**
+ * @brief Allocate memory block
+ *
+ * Weak implementation uses standard malloc(). Can be overridden by
+ * providing a strong implementation of capi_malloc_impl().
+ *
+ * @param size Number of bytes to allocate
+ * @return Pointer to allocated memory, or NULL on failure
+ */
+void *capi_malloc(size_t size);
+
+/**
+ * @brief Free memory block
+ *
+ * Weak implementation uses standard free(). Can be overridden by
+ * providing a strong implementation of capi_free_impl().
+ *
+ * @param ptr Pointer to memory block to free
+ */
+void capi_free(void *ptr);
+
+/**
+ * @brief Allocate and zero-initialize memory block
+ *
+ * Weak implementation uses standard calloc(). Can be overridden by
+ * providing a strong implementation of capi_calloc_impl().
+ *
+ * @param num Number of elements
+ * @param size Size of each element in bytes
+ * @return Pointer to allocated and zero-initialized memory, or NULL on failure
+ */
+void *capi_calloc(size_t num, size_t size);
+
+/**
+ * @brief Reallocate memory block
+ *
+ * Weak implementation uses standard realloc(). Can be overridden by
+ * providing a strong implementation of capi_realloc_impl().
+ *
+ * @param ptr Pointer to existing memory block (or NULL)
+ * @param size New size in bytes
+ * @return Pointer to reallocated memory, or NULL on failure
+ */
+void *capi_realloc(void *ptr, size_t size);
+
+/* Weak hook symbols to be overridden by the platform/app (don't call directly). */
+void *capi_malloc_impl(size_t size);
+void capi_free_impl(void *ptr);
+void *capi_calloc_impl(size_t num, size_t size);
+void *capi_realloc_impl(void *ptr, size_t size);
+
+#ifdef __cplusplus
+}
+#endif
--- a/capi/inc/capi_irq.h
+++ b/capi/inc/capi_irq.h
@@ -1,0 +1,159 @@
+/*
+ * Copyright (c) 2024-2025 Analog Devices, Inc.
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+/**
+ * @file
+ * @brief Common HAL Interrupt Controller.
+ */
+
+#ifndef _CAPI_IRQ_H_
+#define _CAPI_IRQ_H_
+
+#include <stdbool.h>
+#include <stdint.h>
+
+#if defined(__cplusplus)
+extern "C" {
+#endif /* __cplusplus */
+
+/**
+ * @brief Enumerated type that defines the trigger type.
+ */
+enum capi_irq_trig_level {
+	CAPI_IRQ_LEVEL_LOW,    /**< Trigger when signal is low. */
+	CAPI_IRQ_LEVEL_HIGH,   /**< Trigger when signal is high. */
+	CAPI_IRQ_EDGE_FALLING, /**< Trigger on falling edge. */
+	CAPI_IRQ_EDGE_RISING,  /**< Trigger on rising edge. */
+	CAPI_IRQ_EDGE_BOTH     /**< Trigger on rising and falling edge. */
+};
+
+/**
+ * @brief Interrupt Service Routine.
+ *
+ * @param [in] arg Points to application defined argument.
+ */
+typedef void (*capi_isr_callback_t)(void *arg);
+
+/**
+ * @struct capi_irq_config
+ * @brief Structure that holds the IRQ Controller configuration.
+ */
+struct capi_irq_config {
+	/** Interrupt request controller ID. */
+	uint32_t irq_ctrl_id;
+	/**
+	 * This is intended to store irq controller specific configurations,
+	 * it should not be a reference to any peripheral descriptor.
+	 */
+	void *extra;
+};
+
+/**
+ * @brief Initialize an instance of the IRQ controller.
+ *
+ * @param [in] config Points to the configuration for the IRQ controller.
+ * @return int 0 for success or error code.
+ */
+int capi_irq_init(struct capi_irq_config *config);
+
+/**
+ * @brief Deinitialize the IRQ controller, disable, and bring to default settings.
+ *
+ * @return int 0 for success or error code.
+ */
+int capi_irq_deinit(void);
+
+/**
+ * @brief Enable the interrupt controller.
+ *
+ * @return int 0 for success or error code.
+ */
+int capi_irq_global_enable(void);
+
+/**
+ * @brief Disable the interrupt controller.
+ *
+ * @return int 0 for success or error code.
+ */
+int capi_irq_global_disable(void);
+
+/**
+ * @brief Enable specific irq.
+ *
+ * @param [in] irq Defines the IRQ number.
+ * @return int 0 for success or error code.
+ */
+int capi_irq_enable(uint32_t irq);
+
+/**
+ * @brief Disable specific irq.
+ *
+ * @param [in] irq Defines the IRQ number.
+ * @return int 0 for success or error code.
+ */
+int capi_irq_disable(uint32_t irq);
+
+/**
+ * @brief Set isr for the irq
+ *
+ * @param [in] irq Defines the IRQ number.
+ * @param [in] isr function pointer to isr that needs to be registered.
+ * @param [in] arg A pointer to application defined argument returned when ISR is called.
+ * @return int 0 for success or error code.
+ */
+int capi_irq_connect(uint32_t irq, capi_isr_callback_t isr, void *arg);
+
+/**
+ * @brief Clear a specific pending irq.
+ *
+ * @param [in] irq Defines the IRQ number.
+ * @return int 0 for success or error code.
+ */
+int capi_irq_clear_pending(uint32_t irq);
+
+/**
+ * @brief Get status of a specific irq.
+ *
+ * @param [in] irq Defines the IRQ number.
+ * @param [out] pactive Pointer to irq status update.
+ * @return int 0 for success or error code.
+ */
+int capi_irq_get_status(uint32_t irq, uint32_t *pactive);
+
+/**
+ * @brief Set irq priority.
+ *
+ * @param [in] irq Defines the IRQ number.
+ * @param [in] priority  Priority to set.
+ * @return int 0 for success or error code.
+ */
+int capi_irq_set_priority(uint32_t irq, uint32_t priority);
+
+/**
+ * @brief Get irq priority.
+ *
+ * @param [in] irq Defines the IRQ number.
+ * @param [out] priority irq priority retrieved.
+ * @return int 0 for success or error code.
+ */
+int capi_irq_get_priority(uint32_t irq, uint32_t *priority);
+
+/**
+ * @brief Set the IRQ to fire when defined level or edge is detected based on the given
+ * configuration.
+ *
+ * @param [in] irq Defines the IRQ number.
+ * @param [in] trigger The defined state for IRQ trigger which can be level or edge.
+ * @return int 0 for success or error code.
+ */
+int capi_irq_set_level_edge_trigger(uint32_t irq,
+				    enum capi_irq_trig_level trigger);
+
+#if defined(__cplusplus)
+}
+#endif
+
+#endif // CAPI_IRQ_H_
--- a/capi/platform/stm32/stm32_capi_dma.c
+++ b/capi/platform/stm32/stm32_capi_dma.c
@@ -1,0 +1,511 @@
+/*
+ * Copyright (c) 2025-2026 Analog Devices, Inc.
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+#include "stm32_capi_dma_priv.h"
+#include "capi_alloc.h"
+#include <errno.h>
+#include <string.h>
+
+#define MAX_DMA_CHANNELS 16
+
+static struct capi_dma_handle *dma_handle_singleton = NULL;
+static struct stm32_dma_chan_priv *chan_priv_map[MAX_DMA_CHANNELS];
+
+/**
+ * @brief Find channel private data by HAL DMA handle.
+ * @param hdma - Pointer to the HAL DMA handle.
+ * @return Pointer to the matching channel private data, or NULL if not found.
+ */
+static struct stm32_dma_chan_priv *find_chan_priv_by_hdma(
+	DMA_HandleTypeDef *hdma)
+{
+	uint32_t i;
+
+	for (i = 0; i < MAX_DMA_CHANNELS; i++) {
+		if (chan_priv_map[i] && &chan_priv_map[i]->hdma == hdma)
+			return chan_priv_map[i];
+	}
+
+	return NULL;
+}
+
+/**
+ * @brief HAL DMA transfer complete callback.
+ * @param hdma - Pointer to the HAL DMA handle that completed.
+ */
+static void stm32_capi_dma_xfer_cplt_callback(DMA_HandleTypeDef *hdma)
+{
+	struct stm32_dma_chan_priv *chan_priv;
+
+	chan_priv = find_chan_priv_by_hdma(hdma);
+	if (!chan_priv || !chan_priv->capi_chan)
+		return;
+
+	chan_priv->completed = true;
+
+	if (chan_priv->capi_chan->xfer_complete_cb)
+		chan_priv->capi_chan->xfer_complete_cb(0,
+						       chan_priv->capi_chan->xfer_complete_ctx);
+}
+
+/**
+ * @brief HAL DMA transfer error callback.
+ * @param hdma - Pointer to the HAL DMA handle that errored.
+ */
+static void stm32_capi_dma_xfer_error_callback(DMA_HandleTypeDef *hdma)
+{
+	struct stm32_dma_chan_priv *chan_priv;
+
+	chan_priv = find_chan_priv_by_hdma(hdma);
+	if (!chan_priv)
+		return;
+
+	chan_priv->completed = true;
+}
+
+/**
+ * @brief Initialize the STM32 DMA controller.
+ * @param handle - Pointer to the DMA handle pointer to be allocated.
+ * @param config - Pointer to the DMA configuration structure.
+ * @return 0 on success, negative error code otherwise.
+ */
+static int stm32_capi_dma_init(struct capi_dma_handle **handle,
+			       const struct capi_dma_config *config)
+{
+	struct capi_dma_handle *dma_handle;
+	struct stm32_dma_priv_handle *dma_priv;
+
+	if (!handle || !config)
+		return -EINVAL;
+
+	if (dma_handle_singleton) {
+		*handle = dma_handle_singleton;
+		return 0;
+	}
+
+	if (*handle == NULL) {
+		dma_handle = (struct capi_dma_handle *)capi_calloc(1, sizeof(*dma_handle));
+		if (!dma_handle)
+			return -ENOMEM;
+
+		dma_priv = (struct stm32_dma_priv_handle *)capi_calloc(1, sizeof(*dma_priv));
+		if (!dma_priv) {
+			capi_free(dma_handle);
+			return -ENOMEM;
+		}
+
+		dma_handle->priv = dma_priv;
+		dma_handle->init_allocated = true;
+	} else {
+		dma_handle = *handle;
+		dma_handle->init_allocated = false;
+
+		if (!dma_handle->priv)
+			return -EINVAL;
+
+		dma_priv = dma_handle->priv;
+	}
+
+	dma_priv->num_chans = config->num_chans;
+	dma_priv->chan_privs = (struct stm32_dma_chan_priv **)capi_calloc(
+				       config->num_chans, sizeof(struct stm32_dma_chan_priv *));
+	if (!dma_priv->chan_privs) {
+		if (dma_handle->init_allocated) {
+			capi_free(dma_priv);
+			capi_free(dma_handle);
+		}
+		return -ENOMEM;
+	}
+
+	dma_handle->ops = config->ops;
+
+	dma_handle_singleton = dma_handle;
+	*handle = dma_handle;
+
+	return 0;
+}
+
+/**
+ * @brief Deinitialize the STM32 DMA controller.
+ * @param handle - Pointer to the DMA handle to deinitialize.
+ * @return 0 on success, negative error code otherwise.
+ */
+static int stm32_capi_dma_deinit(struct capi_dma_handle *handle)
+{
+	struct stm32_dma_priv_handle *dma_priv;
+	uint32_t i;
+
+	if (!handle)
+		return -EINVAL;
+
+	dma_priv = handle->priv;
+
+	if (dma_priv) {
+		for (i = 0; i < dma_priv->num_chans; i++) {
+			if (dma_priv->chan_privs[i]) {
+				HAL_DMA_DeInit(&dma_priv->chan_privs[i]->hdma);
+				if (handle->init_allocated)
+					capi_free(dma_priv->chan_privs[i]);
+			}
+		}
+		if (handle->init_allocated)
+			capi_free(dma_priv->chan_privs);
+	}
+
+	memset(chan_priv_map, 0, sizeof(chan_priv_map));
+
+	if (handle->init_allocated) {
+		capi_free(dma_priv);
+		capi_free(handle);
+	}
+
+	dma_handle_singleton = NULL;
+
+	return 0;
+}
+
+/**
+ * @brief Initialize a DMA channel.
+ * @param handle - Pointer to the DMA handle.
+ * @param chan_ptr - Pointer to the channel pointer to be allocated.
+ * @param id - Channel identifier.
+ * @return 0 on success, negative error code otherwise.
+ */
+static int stm32_capi_dma_init_chan(struct capi_dma_handle *handle,
+				    struct capi_dma_chan **chan_ptr,
+				    uint32_t id)
+{
+	struct capi_dma_chan *chan;
+	struct stm32_dma_chan_priv *chan_priv;
+	struct stm32_dma_priv_handle *dma_priv;
+
+	if (!handle || !chan_ptr)
+		return -EINVAL;
+
+	dma_priv = handle->priv;
+	if (!dma_priv)
+		return -EINVAL;
+
+	if (id >= dma_priv->num_chans)
+		return -EINVAL;
+
+	if (dma_priv->chan_privs[id])
+		return -EBUSY;
+
+	chan = (struct capi_dma_chan *)capi_calloc(1, sizeof(*chan));
+	if (!chan)
+		return -ENOMEM;
+
+	chan_priv = (struct stm32_dma_chan_priv *)capi_calloc(1, sizeof(*chan_priv));
+	if (!chan_priv) {
+		capi_free(chan);
+		return -ENOMEM;
+	}
+
+	chan->owned_by_app = false;
+	chan->handle = handle;
+	chan->id = id;
+	chan->extra = chan_priv;
+
+	chan_priv->capi_chan = chan;
+	chan_priv->completed = true;
+
+	dma_priv->chan_privs[id] = chan_priv;
+
+	if (id < MAX_DMA_CHANNELS)
+		chan_priv_map[id] = chan_priv;
+
+	*chan_ptr = chan;
+
+	return 0;
+}
+
+/**
+ * @brief Deinitialize a DMA channel.
+ * @param chan - Pointer to the DMA channel to deinitialize.
+ * @return 0 on success, negative error code otherwise.
+ */
+static int stm32_capi_dma_deinit_chan(struct capi_dma_chan *chan)
+{
+	struct stm32_dma_chan_priv *chan_priv;
+	struct stm32_dma_priv_handle *dma_priv;
+
+	if (!chan || !chan->handle)
+		return -EINVAL;
+
+	chan_priv = chan->extra;
+	dma_priv = chan->handle->priv;
+
+	if (chan_priv) {
+		HAL_DMA_DeInit(&chan_priv->hdma);
+
+		if (chan->id < MAX_DMA_CHANNELS)
+			chan_priv_map[chan->id] = NULL;
+
+		if (dma_priv && chan->id < dma_priv->num_chans)
+			dma_priv->chan_privs[chan->id] = NULL;
+
+		capi_free(chan_priv);
+	}
+
+	if (!chan->owned_by_app)
+		capi_free(chan);
+
+	return 0;
+}
+
+/**
+ * @brief Configure a DMA transfer.
+ * @param chan - Pointer to the DMA channel.
+ * @param xfer - Pointer to the DMA transfer descriptor.
+ * @return 0 on success, negative error code otherwise.
+ */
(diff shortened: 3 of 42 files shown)


drivers/platform/stm32/stm32_capi_gpio.c
```diff
@@ -0,0 +1,616 @@
...
+
+/**
+ * @brief Get STM32 GPIO port from identifier
+ * @param identifier - Port identifier (0=GPIOA, 1=GPIOB, etc. or base address)
+ * @return GPIO_TypeDef pointer or NULL if invalid
+ */
+static GPIO_TypeDef *get_gpio_port_from_identifier(uint64_t identifier)
+{
+	/* If identifier looks like a base address, use it directly */
+	if (identifier >= 0x40000000)
```
buha: yeah this is likely correct on what you tested, but will fail on other targets, so I think we need to decide on a more reasonable number for the comparison here, it has to be smaller like 256 or 1024 or something like that

cristina-suteu:
v2: 
stm32_capi_gpio: use 256 as comparision value for gpio port identifier

buha:
please fix astyle and drivers checks

cristina-suteu:
v3: 
- dropped commits with example - used only for testing purposes, no need to be merged into main
- added license header to newly added files
- added include/capi/ to the Makefile in drivers
code style issues were coming from the example

drivers/platform/stm32/stm32_capi_uart.c
```diff
@@ -0,0 +1,272 @@
...
+	case CAPI_UART_DATA_BITS_7:
+		uart_priv_handle->huart->Init.WordLength = UART_WORDLENGTH_7B;
+		break;
+	case CAPI_UART_DATA_BITS_8:
+		uart_priv_handle->huart->Init.WordLength = UART_WORDLENGTH_8B;
+		break;
+// TODO: add 9 bits option to CAPI
+//	case CAPI_UART_DATA_BITS_9:
+//		uart_priv_handle->huart->Init.WordLength = UART_WORDLENGTH_9B;
+//		break;
```
rbolboac: what is the status here? will it be supported in this PR pr later?

capi/platform/stm32/stm32_capi_uart.c
```diff
@@ -0,0 +1,272 @@
...
+
+	if (!len)
+		return 0;
+
+	uart_priv_handle = handle->priv;
+
+	ret = HAL_UART_Receive(uart_priv_handle->huart, buf, len, HAL_MAX_DELAY);
+	switch (ret) {
+	case HAL_OK:
+		break;
```
rbolboac, line 400: why not return directly? if you decide to do so, see all instanced where this applies

drivers/platform/stm32/stm32_capi_gpio.c
```diff
@@ -0,0 +1,622 @@
+/*
+ * Copyright (c) 2024-2025 Analog Devices, Inc.
```
rbolboac: I guess this should be updated to 2025, see everywhere it applies

drivers/platform/stm32/stm32_capi_gpio.c
```diff
@@ -0,0 +1,622 @@
...
+
+/**
+ * @brief Enable GPIO port clock in RCC
+ * @param identifier - Port identifier
+ * @return 0 on success, negative error code otherwise
+ */
+static int enable_gpio_port_clock(uint64_t identifier)
+{
+	/* Handle base address case */
+	if (identifier >= 0x40000000) {
```
rbolboac: does this work for all supported stm32 chip families?

drivers/platform/stm32/stm32_capi_uart.c
```diff
@@ -0,0 +1,272 @@
...
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+#include "capi_uart.h"
+#include "stm32_capi_uart.h"
+#include "no_os_alloc.h"
+#include "no_os_error.h"
+#include <sys/stat.h>
+
+int stm32_capi_uart_init(struct capi_uart_handle **handle,
```
rbolboac: APIs used in ops should be static, right?

drivers/platform/stm32/stm32_capi_gpio.c
```diff
@@ -0,0 +1,622 @@
...
+{
+	if (flags & CAPI_GPIO_ACTIVE_LOW)
+		return logical_value ? CAPI_GPIO_LOW : CAPI_GPIO_HIGH;
+	return logical_value;
+}
+
+/**
+ * @brief Initialize a GPIO port
+ */
+int stm32_capi_gpio_port_init(struct capi_gpio_port_handle **handle,
```
rbolboac: APIs used in ops should be static, right? or is there a reason we want these exposed in the header file?

drivers/platform/stm32/stm32_capi_gpio.c
```diff
@@ -0,0 +1,622 @@
...
+		return -EINVAL;
+
+	priv = handle->priv;
+
+	/* Calculate which pins to set and reset */
+	set_pins = (uint16_t)value_bitmask;
+	reset_pins = (uint16_t)(~value_bitmask);
+
+	/* Use BSRR register for atomic set/reset */
+	priv->port->BSRR = ((uint32_t)reset_pins << 16) | set_pins;
```
rbolboac: does this work for all supported stm32 chip families?

capi/platform/stm32/stm32_capi_gpio.c
```diff
@@ -0,0 +1,622 @@
...
+static inline uint8_t to_raw_value(uint8_t logical_value, uint32_t flags)
+{
+	if (flags & CAPI_GPIO_ACTIVE_LOW)
+		return logical_value ? CAPI_GPIO_LOW : CAPI_GPIO_HIGH;
+	return logical_value;
+}
+
+/**
+ * @brief Initialize a GPIO port
+ */
```
rbolboac, line 265: some APIs also contain the param and return description, I think Claude can handle doing this for all instances where it is missing

drivers/platform/stm32/stm32_capi_gpio.c
```diff
@@ -0,0 +1,622 @@
...
+{
+	struct stm32_capi_gpio_port_priv *priv;
+
+	if (!handle || !handle->priv || !value_bitmask)
+		return -EINVAL;
+
+	priv = handle->priv;
+
+	/* Read input data register */
+	*value_bitmask = priv->port->IDR & 0xFFFF;
```
rbolboac: does this work for all supported stm32 chip families? I would avoid direct register access from HAL because we had surprises with some registers missing or being renamed otherwise, maybe a generic HAL from STM can be used to obtain this value

rbolboac (commented):
my remarks regarding stm32_capi_uart and stm32_capi_gpio. I still have tot take a look at the other commits

capi/platform/stm32/stm32_capi_dma.c
```diff
@@ -0,0 +1,417 @@
...
+		return;
+
+	chan_priv->completed = true;
+
+	if (chan_priv->capi_chan->xfer_complete_cb && chan_priv->capi_chan->xfer)
+		chan_priv->capi_chan->xfer_complete_cb(chan_priv->capi_chan->xfer,
+						       chan_priv->capi_chan->xfer->user_data);
+}
+
+static void stm32_capi_dma_xfer_error_callback(DMA_HandleTypeDef *hdma)
```
rbolboac, line 58: consider using the same naming convention for all APIs, in gpio, for example, the static functions do not contain the stm32_gpio part

capi/platform/stm32/stm32_capi_dma.c
```diff
@@ -0,0 +1,417 @@
...
+	struct stm32_dma_chan_priv *chan_priv;
+
+	chan_priv = find_chan_priv_by_hdma(hdma);
+	if (!chan_priv)
+		return;
+
+	chan_priv->completed = true;
+}
+
+static int stm32_capi_dma_init(struct capi_dma_handle **handle,
```
rbolboac, line 75: add missing APIs comments, see everywhere where it is missing

drivers/platform/stm32/stm32_capi_spi.c
```diff
@@ -0,0 +1,1421 @@
...
+#include "no_os_delay.h"
+#include "capi/capi_dma.h"
+#ifdef HAL_TIM_MODULE_ENABLED
+#include "no_os_pwm.h"
+#endif
+
+/* Forward declarations */
+static int stm32_capi_spi_dma_abort(struct capi_spi_device *device);
+
+//lookup table for mapping HAL handles to our private handles
```
rbolboac: consider using the same commenting format, sometimes you are using // sometimes /* */

drivers/platform/stm32/stm32_capi_spi.c
```diff
@@ -0,0 +1,1421 @@
...
+	return 0;
+}
+
+/**
+ * @brief Get SPI base address from identifier
+ */
+static SPI_TypeDef *get_spi_base_from_identifier(uint64_t identifier)
+{
+	// If identifier is a base address, use it directly
+	if (identifier >= 0x40000000) {
```
rbolboac: is this valid for all stm32 chip families? would be nice to use a macro instead, I see PERIPH_BASE is present in stm32 HAL

capi/platform/stm32/stm32_capi_spi.c
```diff
@@ -0,0 +1,1421 @@
...
+
+	priv_handle->async_in_progress = false;
+	priv_handle->current_transfer = NULL;
+
+	if (priv_handle->callback) {
+		priv_handle->callback(CAPI_SPI_EVENT_ERROR,
+				      priv_handle->callback_arg,
+				      -ECANCELED);
+	}
+}
```
rbolboac, line 1633: missing new line at EOF

capi/platform/stm32/stm32_capi_spi.h
```diff
@@ -0,0 +1,205 @@
...
+/**
+ * @brief Async transfer multiple messages using DMA
+ * @param device - The SPI device descriptor
+ * @param transfers - Array of transfer descriptors
+ * @param transfer_count - Number of transfers
+ * @param callback - Completion callback
+ * @param callback_arg - Callback argument
+ * @return 0 in case of success, error codes otherwise
+ */
+int stm32_capi_spi_transfer_multiple_dma_async(struct capi_spi_device *device,
```
rbolboac, line 135: why some APIs are exported in the header but also present in operations?

drivers/platform/stm32/stm32_capi_i2c.c
```diff
@@ -0,0 +1,655 @@
...
+		}
+	}
+}
+
+/**
+ * @brief Get I2C base address from identifier
+ */
+static I2C_TypeDef *get_i2c_base_from_identifier(uint64_t identifier)
+{
+	if (identifier >= 0x40000000)
```
rbolboac: same remark as before

drivers/platform/stm32/stm32_capi_i2c.c
```diff
@@ -0,0 +1,655 @@
...
+	uint32_t clock_speed;
+
+	if (!handle || !handle->priv)
+		return -EINVAL;
+
+	priv_handle = handle->priv;
+
+	switch (speed) {
+	case CAPI_I2C_SPEED_STANDARD:
+		clock_speed = 100000;
```
rbolboac: macros would be nice for numeric values

drivers/platform/stm32/stm32_capi_irq.h
```diff
@@ -0,0 +1,201 @@
...
+#ifndef STM32_CAPI_IRQ_MAX_IRQS
+#define STM32_CAPI_IRQ_MAX_IRQS		128
+#endif
+
+/**
+ * @brief Maximum number of EXTI lines supported.
+ *
+ * STM32 devices typically have 16 GPIO EXTI lines (0-15).
+ */
+#define STM32_CAPI_EXTI_MAX_LINES	16
```
rbolboac: if macros are only used internally, I would only define them in the .c file, no reason to expose them in the header

capi/inc/capi_irq.h
```diff
@@ -71,7 +71,7 @@ int capi_irq_deinit(void);
  *
  * @return int 0 for success or error code.
  */
-int capi_irq_global_enalbe(void);
+int capi_irq_global_enable(void);
```
rbolboac, line 74: this change is not related to this commit

capi/src/capi_irq.c
```diff
@@ -0,0 +1,80 @@
+/*
+ * Copyright (c) 2024-2025 Analog Devices, Inc.
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+/**
+ * @file capi_irq.c
+ * @brief This file contains a set of small wrapper functions that can be used to access
+ *        the IRQ driver via its ops pointer structure. Note, it does not
```
rbolboac, line 10: the description talks about ops, however no ops are defined. is this the intended usage? if so, please adapt the brief

capi/src/capi_irq.c
```diff
@@ -0,0 +1,80 @@
...
+{
+	return capi_irq_stub_get_priority(irq, priority);
+}
+
+int capi_irq_set_level_edge_trigger(uint32_t irq, enum capi_irq_trig_level trigger)
+{
+	return capi_irq_stub_set_level_edge_trigger(irq, trigger);
+}
+
+#endif /* CAPI_STUB_IMPLEMENTATION */
```
rbolboac: it looks like many files from this commit are missing a new line at the EOF

drivers/platform/stm32/stm32_capi_timer.c
```diff
@@ -0,0 +1,983 @@
...
+	if (priv->input_clock_hz < priv->output_freq_hz) {
+		ret = -EINVAL;
+		goto error;
+	}
+
+	prescaler = (priv->input_clock_hz / priv->output_freq_hz) - 1;
+
+	priv->htim.Init.Prescaler = prescaler;
+	priv->htim.Init.CounterMode = TIM_COUNTERMODE_UP;
+	priv->htim.Init.Period = 0xFFFFFFFF;
```
rbolboac: use macro instead

capi/platform/stm32/stm32_capi_timer.c
```diff
@@ -0,0 +1,983 @@
...
+	if (sr & TIM_SR_UIF) {
+		__HAL_TIM_CLEAR_FLAG(&priv->htim, TIM_FLAG_UPDATE);
+		if (priv->event_callback) {
+			priv->event_callback(CAPI_TIMER_GLOBAL_EVENT_COUNTER_OVERFLOW,
+					     priv->event_callback_arg, 0);
+		}
+	}
+
+	for (i = 0; i < STM32_TIMER_MAX_CHANNELS; i++) {
+		uint32_t cc_flag = TIM_SR_CC1IF << i;
```
rbolboac, line 1028: does this work for all supported stm32 chip families?

rbolboac (changes requested):
most of my remarks should be checked against all stm32 specific added files.
minimal functionality changes/checks are required, specifically for the usage of STM HAL register usage
the rest are related to the form of the code: should have the same API description structure, same comment format, check if all the things defined in the header file are actually meant to be exposed, use macros instead of numerals, etc.

buha:
also can you please update CAPI to the latest ?

capi/platform/stm32/stm32_capi_uart.c
```diff
@@ -0,0 +1,272 @@
...
+	uart_extra_config = config->extra;
+
+	uart_priv_handle->huart = uart_extra_config->huart;
+
+	uart_priv_handle->huart->Init.BaudRate = config->line_config->baudrate;
+
+	switch (config->line_config->size) {
+	case CAPI_UART_DATA_BITS_7:
+		uart_priv_handle->huart->Init.WordLength = UART_WORDLENGTH_7B;
+		break;
```
rbolboac, line 111: I think I understand why you did comment case CAPI_UART_DATA_BITS_9 out, some HAL do not have these macros defines, this needs to be reworked with ifdef I guess (I am testing this with stm32f4 and UART_WORDLENGTH_7B is not defined there so the build fails).

drivers/platform/stm32/stm32_capi_dma.c
```diff
@@ -0,0 +1,417 @@
+/*
+ * Copyright (c) 2024-2025 Analog Devices, Inc.
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+#include "stm32_capi_dma.h"
+#include "capi/capi_dma.h"
+#include "no_os_alloc.h"
```
buha: as per our discussion, remove all no_os_alloc.h dependencies and replace it with capi_alloc, you'll need to also add the implementation of the capi_alloc API

drivers/platform/stm32/stm32_capi_dma.c
```diff
@@ -0,0 +1,417 @@
...
+#include "no_os_error.h"
```
buha: In my opinion no_os_error.h is useless, we can do without. Directly include <errno.h> instead.

drivers/platform/stm32/stm32_capi_gpio.c
```diff
@@ -0,0 +1,622 @@
+/*
+ * Copyright (c) 2024-2025 Analog Devices, Inc.
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+#include <stdlib.h>
+#include <errno.h>
+#include "no_os_util.h"
```
buha: I think that in order to keep these CAPI backends (implementations) independent from the rest of no-OS, so as to be able to use them into other projects, it is acceptable to not use the helper BIT(x) manipulation macros and simply perform 1 << (x) in-place.

I searched all the file and the only thing that is being used from no_os_util.h is `NO_OS_BIT()` macro so that makes things easy, just replace all those instances with an in place bit shift and you can get rid of the dependency.

drivers/api/capi/capi_adc.c
```diff
@@ -0,0 +1,216 @@
+/*
```
buha: pleae move the capi files like this.

sources:
```
no-OS/capi/src/capi_spi.c
```
headers
```
no-OS/capi/inc/capi_spi.h
```

drivers/platform/stm32/stm32_capi_spi.c
```diff
@@ -0,0 +1,1421 @@
...
+#include <stdlib.h>
+#include <errno.h>
+#include "capi/capi_spi.h"
+#include "capi/capi_gpio.h"
+#include "stm32_capi_spi.h"
+#include "stm32_capi_gpio.h"
+#include "no_os_alloc.h"
+#include "no_os_error.h"
+#include "no_os_util.h"
+#include "no_os_delay.h"
```
buha: Use capi_time instead

drivers/platform/stm32/stm32_capi_uart.c
cristina-suteu: in latest capi  uart implementation, there is still no 9 bits option. when it will be available, will update capi_uart files, and uncomment this section.

cristina-suteu:
v4 : 
- added latest capi files, only the ones needed by this PR
- moved capi files to correct location
- make sure Makefile in drivers points to new location of the capi files
-  address UART issues in PR review, replaced calls to no_os_alloc with calls to capi_calloc and replaced no_os_error with errno.h  
To do: 
- address remaining issues in PR review, per module added. 
- fix style issues introduced by latest revision of UART implementation

capi/inc/capi_alloc.h
```diff
@@ -0,0 +1,75 @@
+/*
```
buha, line 1: minor nitpick, but remove the - from c-api, it will be important when searching git log ;)

buha:
add capi_time implementation, you can basically copy paste the implementation here, with a few adjustments for the uptime function https://github.com/analogdevicesinc/no-OS/blob/main/drivers/platform/stm32/stm32_delay.c

currently, capi_time is a dependency of stm32_spi implementation, and the reason your build doesn't error out is because capi_time has a weak implementation for its functions, but if you really do a spi timing test, it's not going to behave like you'd expect, so please add it

buha:
astyle still errors out, what i do is `git rebase -i HEAD~10` and mark every relevant commit as `edit` or `e` in short, run astyle on each, stage the changes, `git commit --amend`, `git rebase --continue`

rbolboac:
@buha maybe a discussion on this topic will be useful: https://github.com/analogdevicesinc/no-OS/issues/2339
currently, I don't see a config API in CAPI

buha:
@rbolboac there is at least this one for uart: capi_uart_set_line_config

There is this one for i2c: https://github.com/analogdevicesinc/no-OS/pull/3090/changes/3826ef959241e2592df373bc7a360f6fe2274aee#diff-c5eb0379572f34544e961896cd2d9ad1588f462f85f934d19d64f3e256f0feebR214


The biggest issue was UART until now. But I can see an i2c master switching speeds after initialization phase to match slave max speeds in between talking to them. We might get away not making a config function mandatory in the CAPI implementation due to the fact that there are some line config functions, but yeah, it's worth discussing.

rbolboac:
> @rbolboac there is at least this one for uart: capi_uart_set_line_config
> 
> There is this one for i2c: [3826ef9#diff-c5eb0379572f34544e961896cd2d9ad1588f462f85f934d19d64f3e256f0feebR214](https://github.com/analogdevicesinc/no-OS/commit/3826ef959241e2592df373bc7a360f6fe2274aee#diff-c5eb0379572f34544e961896cd2d9ad1588f462f85f934d19d64f3e256f0feebR214)
> 
> The biggest issue was UART until now. But I can see an i2c master switching speeds after initialization phase to match slave max speeds in between talking to them. We might get away not making a config function mandatory in the CAPI implementation due to the fact that there are some line config functions, but yeah, it's worth discussing.

I missed the CAPI update with line configs, I think that if we implement capi_uart_set_line_config for each platform then the issue is solved. I think only iio implementation is using deinit/init flow but using capi_uart_set_line_config is definitely better.

cristina-suteu:
v5: 
- address all style issues
- add all needed capi files, including capi_time. 
- move capi  stm32 platform drivers to no-OS/capi/platform 
- drop all remaining refs to no_os api from capi stm32 drivers

cristina-suteu:
v6:
- style capi sources

cristina-suteu:
v7: 
- stm32_capi_uart : add set/get_line_config, transmit/receive_async, register_callback, isr, line_status, interrupt_reason - supported on all stm32 chip families 
- add stm32_capi_time to replace stm32_delay
- add stm32_capi_pwm driver

capi/platform/stm32/stm32_capi_i2c.c
```diff
@@ -0,0 +1,796 @@
...
+	struct capi_i2c_controller_handle *i2c_handle;
+	struct stm32_i2c_priv_handle *priv_handle;
+	struct stm32_i2c_extra_config *extra_config;
+	I2C_TypeDef *base;
+	int ret;
+
+	if (!handle || !config)
+		return -EINVAL;
+
+	i2c_handle = capi_calloc(1, sizeof(*i2c_handle));
```
buha: We need to check here for the value of the handle and if it's non NULL, use it (app owns it), if it's NULL, then allocate it. Please apply this change to all init functions, maybe get a good i2c_init that handles both cases and then propagate to others via claude.

See this bit: https://github.com/analogdevicesinc/no-OS/pull/3090/changes/ec69d8db8869746031d1bc71e7e8ce124f104aed#diff-c5eb0379572f34544e961896cd2d9ad1588f462f85f934d19d64f3e256f0feebR154-R158

capi/platform/stm32/stm32_capi_spi.c
```diff
@@ -0,0 +1,1648 @@
...
+#endif
+
+#define STM32_SPI_CRC_POLYNOMIAL		10
+#define STM32_SPI_DMA_DEFAULT_TIMEOUT_MS	1000
+#define STM32_SPI_GPIO_PINS_PER_PORT		16
+
+/**
+ * @brief STM32 SPI private handle with full feature support
+ */
+struct stm32_spi_priv_handle {
```
buha: move this to a new file, something like this:
```
=== NEW FILE: capi/platform/stm32/stm32_capi_spi_priv.h =========================
+ #ifndef _STM32_CAPI_SPI_PRIV_H_
+ #define _STM32_CAPI_SPI_PRIV_H_
+
+ #include "capi_spi.h"
+ #include "stm32_capi_spi.h"   /* extra-config types, ops */
+
+ /* Moved verbatim out of stm32_capi_spi.c — now the cross-TU "internal" type,
+  * so the _INIT() macro below can see its real size. */
+ struct stm32_spi_priv_handle {
+         SPI_HandleTypeDef hspi;
+         uint32_t input_clock;
+         ... /* (all existing fields, unchanged) */
+ };
+
+ /* App-owned / zero-allocation counterpart to the dynamic path.
+  * Mirror of CAPI_SPI_CONTROLLER_HANDLE_SPI_01_INIT in bm-drivers. */
+ #define CAPI_SPI_CONTROLLER_HANDLE_STM32_INIT()                       \
+         (&(struct capi_spi_controller_handle){                        \
+                 .ops = NULL,                                          \
+                 .init_allocated = false,                             \
+                 .lock = NULL,                                         \
+                 .priv = &(struct stm32_spi_priv_handle){0}})
+
+ #endif /* _STM32_CAPI_SPI_PRIV_H_ */
```

The CAPI_SPI_CONTROLLER_HANDLE_STM32_INIT is very important, it will handle static allocation for the user, it has to exist.

Do this separation and amcro for all peripheral drivers that have ops.

cristina-suteu:
v8: 
- check for init_allocated and properly allocate handle in init functions
- add _priv separation

buha:
I2C slave not supported, I'm ok to implement this later, in the future! Just writing it out in clear here, will be added later.

cristina-suteu:
> I2C slave not supported, I'm ok to implement this later, in the future! Just writing it out in clear here, will be added later. 
@buha  currently working on tests for i2c, could implement this, test it and add it to this PR

capi/platform/stm32/stm32_capi_time.c
```diff
@@ -0,0 +1,92 @@
...
+	start = DWT->CYCCNT;
+	while (DWT->CYCCNT - start < cycles)
+		;
+}
+#pragma GCC pop_options
+#else
+void capi_wait_us_impl(uint32_t us)
+{
+	(void)us;
+	HAL_Delay(1);
```
buha:
```
	if (us) HAL_Delay((us - 1) / 1000 + 1);

```

This was wrong in the original no-OS implementation, here's a better version

cristina-suteu:
v9: 
- stm32_capi_time: fix capi_wait_us_impl

buha (dismissed):
I think this is largely correct and is a good template for other implementations. I'm planning to merge it for now, and any fixes and new features (i2c slave, tested irq implementation) will be added in a new PR, possibly as part of a `capi_selftest` project PR that runs on a single MCU eval board with some flywires.

capi/platform/stm32/stm32_capi_i2c_priv.h
```diff
@@ -0,0 +1,42 @@
...
+	/** Current transfer for async operations */
+	struct capi_i2c_transfer *current_transfer;
+};
+
+#define CAPI_I2C_CONTROLLER_HANDLE_STM32_INIT()                                \
+	(&(struct capi_i2c_controller_handle){                                 \
+		.ops = NULL,                                                   \
+		.init_allocated = false,                                       \
+		.priv = &(struct stm32_i2c_priv_handle){0}})
+
```
rbolboac: these look messy, a clean-up would be nice (see all instances where it applies)

capi/platform/stm32/stm32_capi_dma.c
```diff
@@ -0,0 +1,499 @@
...
+		chan_priv->trig = chan_extra->trig;
+
+		if (chan_extra->hdma)
+			memcpy(&chan_priv->hdma, chan_extra->hdma, sizeof(DMA_HandleTypeDef));
+	}
+
+#if defined(STM32F2) || defined(STM32F4) || defined(STM32F7)
+	chan_priv->hdma.Init.Channel = chan_priv->ch_num;
+#else
+	chan_priv->hdma.Instance = (DMA_Channel_TypeDef *)(uintptr_t)chan_priv->ch_num;
```
rbolboac: not sure if this will work, worth checking it out at least. see the below comment from claude
`On L4/G4/G0/F0/L0/U5/H7 this stomps a valid stream/channel pointer with a small integer (e.g. 1); HAL_DMA_Init() then dereferences 0x1 → hard fault. DMA non-functional on every ST family outside F2/F4/F7.`

capi/platform/stm32/stm32_capi_spi.c
```diff
@@ -0,0 +1,1610 @@
...
+	while (timeout--) {
+		capi_wait_ms(1);
+		if (priv_handle->dma_done)
+			break;
+	}
+
+	if (timeout == 0) {
+		stm32_capi_spi_dma_abort(device);
+		return -ETIMEDOUT;
+	}
```
rbolboac: will never happen, replace the above while loop with 
`while(timeout) {timeout--;}`
otherwise, timeout will be 0xFFFFFFFF when a timeout will occur

capi/platform/stm32/stm32_capi_gpio.c
```diff
@@ -0,0 +1,690 @@
...
+	if (pin->number >= STM32_GPIO_PINS_PER_PORT)
+		return -EINVAL;
+
+	priv = pin->port_handle->priv;
+
+	/* Use BSRR for atomic set/reset */
+	if (value)
+		priv->port->BSRR = (1 << (pin->number));
+	else
+		priv->port->BSRR = (1 << (pin->number)) << STM32_BSRR_RESET_SHIFT;
```
rbolboac: For pin->number == 15: (1<<15)<<16 on signed int → UB and a -Werror=shift-overflow build break. Use 1U. Same (1 << pin->number) pattern at lines 519, 525, 529, 530, 538, 565, 666.

capi/platform/stm32/stm32_capi_dma.c
```diff
@@ -0,0 +1,499 @@
...
+#endif
+
+	if (xfer->src_inc == CAPI_DMA_BYTE_INCREMENT)
+		chan_priv->hdma.Init.MemInc = DMA_MINC_ENABLE;
+	else
+		chan_priv->hdma.Init.MemInc = DMA_MINC_DISABLE;
+
+	if (xfer->dst_inc == CAPI_DMA_BYTE_INCREMENT)
+		chan_priv->hdma.Init.PeriphInc = DMA_PINC_ENABLE;
```
rbolboac: `MemInc is tied to src_inc and PeriphInc to dst_inc regardless of xfer_type. For DEV_TO_MEM (peripheral=src, memory=dst) this enables peripheral increment and disables memory increment → corrupted capture. (Note: the SPI driver itself sets the fields to match its hardcoded direction at lines 1152-1165, so SPI happens to work; the bug bites any direct DMA user / future peripheral.)`

I know this is taken from the old implementation, but it is worth to solve this bug here.

capi/platform/stm32/stm32_capi_spi.c
```diff
@@ -0,0 +1,1610 @@
...
+		rx_ch_xfer[i].src = (capi_dma_glbl_addr_t) & (SPIx->DR);
+#endif
+		rx_ch_xfer[i].xfer_type = CAPI_DMA_DEV_TO_MEM;
+		rx_ch_xfer[i].src_inc = CAPI_DMA_NO_INCREMENT;
+		rx_ch_xfer[i].dst_inc = CAPI_DMA_BYTE_INCREMENT;
+		rx_ch_xfer[i].length = transfers[i].rx_size;
+	}
+
+	priv_handle->tx_ch_xfer = tx_ch_xfer;
+	priv_handle->rx_ch_xfer = rx_ch_xfer;
```
rbolboac, line 1194: in case of a following error, should we also nullifies these?

capi/platform/stm32/stm32_capi_uart.c
```diff
@@ -0,0 +1,780 @@
...
+ */
+static int stm32_capi_uart_init(struct capi_uart_handle **handle,
+				const struct capi_uart_config *config)
+{
+	struct capi_uart_handle *uart_handle;
+	struct stm32_uart_priv_handle *uart_priv_handle;
+	struct stm32_uart_extra_config *uart_extra_config;
+	int ret = 0;
+
+	if (!handle || !config)
```
rbolboac: should also check for config->extra (it is optional in capi and it is not checked), check to see in all drivers that this check is performed before accessing config->extra stuff

capi/platform/stm32/stm32_capi_i2c.c
```diff
@@ -0,0 +1,791 @@
...
+					    transfer->len,
+					    HAL_MAX_DELAY);
+	} else if (transfer->no_stop) {
+		hal_ret = HAL_I2C_Master_Seq_Transmit_IT(&priv_handle->hi2c,
+				dev_addr,
+				transfer->buf,
+				transfer->len,
+				I2C_FIRST_FRAME);
+		if (hal_ret == HAL_OK) {
+			while (HAL_I2C_GetState(&priv_handle->hi2c) != HAL_I2C_STATE_READY)
```
rbolboac: a timeout would be nice, see everywhere where applicable

capi/platform/stm32/stm32_capi_spi.c
```diff
@@ -0,0 +1,1610 @@
...
+	if (dev_extra && dev_extra->cs_delay_first) {
+		capi_wait_us(dev_extra->cs_delay_first);
+	}
+
+	/* Use HAL functions for reliable operation */
+	if (transfer->tx_buf && transfer->rx_buf) {
+		hal_ret = HAL_SPI_TransmitReceive(&priv_handle->hspi,
+						  (uint8_t *)transfer->tx_buf,
+						  transfer->rx_buf,
+						  max(transfer->tx_size, transfer->rx_size),
```
rbolboac: here we have a tricky situation, because until now we stated that buffers should have the same length, now the capi implementation takes the maximum size, but if rx size is smaller than tx size, we will corrupt memory. we have to implement a separate usecase for tx_size!rx_size, for now, let's just not allow it if it is too difficult to implement

rbolboac (commented):
some other remarks from my side, maybe it is worth to also ask Claude for a final review after these are implemented

capi/src/capi_alloc.c
```diff
@@ -0,0 +1,55 @@
+/*
+ * Copyright (c) 2025 Analog Devices, Inc.
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+#include "capi_alloc.h"
+
+/* Default weak implementations: no-op implementations that return NULL/do nothing */
+
+__attribute__((weak)) void *capi_malloc_impl(size_t size)
```
edelweiseescala, line 10: Hello just a question about these weak implementations. These always return NULL. I think stm32 also needs to implement strong implementations for these functions.

cristina-suteu: you are right, i forgot to add stm32_capi_alloc.

cristina-suteu:
v10: 
- added stm32_capi_alloc
- added config->extra safe guard in init functions where appropriate
- stm32_capi_dma: support all stm32 chips, fix MemInc issue
- stm32_capi_spi: support different TX and RX buffer sizes, nullify priv->tx/rx_ch_xfer, use proper timeout loop
- stm32_capi_i2c: use proper timeout loop
During a more complete test of the new CAPI platform drivers, several new issues were identified for SPI, DMA, GPIO, TIMER which were also addressed in this version

rbolboac (approved):

PR #3092, opened 2026-05-06
drivers: power: ltm4700: Add LTM4700 dual-channel regulator driver

cjones-adi:
## Pull Request Description

Add driver support for LTM4700/LTM4777 dual-channel digital power system manager family with complete PMBus 1.3 protocol implementation, IIO integration, and extensive unit testing.

## PR Type

## PR Checklist

### DIFF
--- a/drivers/power/ltm4700/iio_ltm4700.c
+++ b/drivers/power/ltm4700/iio_ltm4700.c
@@ -1,0 +1,754 @@
+/***************************************************************************//**
+ *   @file   iio_ltm4700.c
+ *   @brief  Source file for the LTM4700 IIO Driver
+ *   @author Carlos Jones Jr (carlosjr.jones@analog.com)
+********************************************************************************
+ * Copyright 2024(c) Analog Devices, Inc.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice,
+ *    this list of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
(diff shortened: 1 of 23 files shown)


drivers/power/ltm4700/ltm4700.c
```diff
@@ -0,0 +1,923 @@
...
+	case LTM4700_SPECIAL_ID_VALUE:
+		dev->id = ID_LTM4700;
+		break;
+	case LTM4700_LTM4777_SPECIAL_ID_VALUE:
+		dev->id = ID_LTM4777;
+		break;
+	default:
+		ret = -ENODEV;
+		goto error_i2c;
+	}
```
buha, line 262: it would probably be worth comparing the ID you readback from device with the one the driver API user requested, just in case there is a mismatch, and return some error somewhere here

drivers/power/ltm4700/iio_ltm4700.c
```diff
@@ -0,0 +1,754 @@
...
+		.priv = LTM4700_IIO_OPERATION,
+		.show = ltm4700_iio_read_avail,
+	},
+	END_ATTRIBUTES_ARRAY
+};
+
+/* LTM4700 IIO current input attributes */
+static struct iio_attribute ltm4700_iio_current_input_attrs[] = {
+	{
+		.name = "raw",
```
buha, line 493: the only gripe i have about this iio device is that these channels (voltage, current, temperature etc.) should also implement the offset and the scale attributes, so that a user can read the value and convert to corresponding measurement units like Volt, Ampere, Celsius

cjones-adi: thanks for the review @buha , I will reupload the project after doing some refactors.

PR #3101, opened 2026-05-19
Frequency PLLs mod2 fix

skucukar:
## Pull Request Description

Issue : #3025 
Fix MOD2 handling for ADF5611, ADF4368, and ADF4382 frequency drivers to avoid invalid fractional computation paths.

This PR updates three drivers so MOD2 is initialized to 1 (instead of 0) in default/non-fractional paths, and adds a guard in RF output readback to prevent division by zero when MOD2 is 0.
When MOD2 is 0, FRACT2 contribution is treated as 0 and the calculation continues safely.

## PR Type

## PR Checklist

### DIFF
--- a/drivers/frequency/adf4368/adf4368.c
+++ b/drivers/frequency/adf4368/adf4368.c
@@ -1098,8 +1098,12 @@ int adf4368_get_rfout(struct adf4368_dev *dev, uint64_t *val)
 		return ret;
 	mod2 |= tmp;
 
-	freq = frac2 * pfd;
-	freq = no_os_div_u64(freq, mod2);
+	if (mod2) {
+		freq = frac2 * pfd;
+		freq = no_os_div_u64(freq, mod2);
+	} else {
+		freq = 0;
+	}
 	freq = freq + (frac1 * pfd);
 	freq = no_os_div_u64(freq, ADF4368_MOD1WORD);
 	freq = freq + (n_value * pfd);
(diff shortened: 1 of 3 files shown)


amiclaus (commented):
the commits are lacking a body and signature

drivers/frequency/adf4368/adf4368.c
```diff
@@ -1098,8 +1098,12 @@ int adf4368_get_rfout(struct adf4368_dev *dev, uint64_t *val)
 		return ret;
 	mod2 |= tmp;
 
-	freq = frac2 * pfd;
-	freq = no_os_div_u64(freq, mod2);
+	if (mod2) {
+		freq = frac2 * pfd;
+		freq = no_os_div_u64(freq, mod2);
+	} else {
+		freq = 0;
```
amiclaus, line 1105: shouldn't we return an error code here? if not please explain in the commits why you are doing it like this

skucukar: No, because we are still continue to calculate rf out freq on the line 1109.

skucukar: And also value can be zero, even though we dont want to set it to zero.

skucukar:
Commits are updated 
Signature added.

buha (approved):
These seem correct to me, build issues are unrelated, will merge.

PR #3110, opened 2026-05-26
Staging/adxl37x support

RaulGeo289:
## Pull Request Description

Changed ADXL371 driver to support ADXL372
Some things I changed:
ADXL371 was writing 3 bits in the INT1 and INT2 register
which were reserved,like it says in the datasheet

Followed a pattern that supports multiple chips by
looking at ADXL38X and AD5686 drivers, I detailed more 
about the way I used them in the commit message for the driver

Called set_internal_sync() in the workaround function to set it 
directly, not just verify it with the getter method

## PR Type

## PR Checklist


rbolboac:
I advise to take a look at each commit and organize the files accordingly - commit history should be as clean as possible. You can change it such that each commit handles a new feature - e.g.:
1. driver + driver doc
2. iio driver + documentation update with iio features
3. basic example + basic example doc
4. iio example + documentation update with iio example

It also looks like there are some merge comments in the files so there is still a little bit of clean-up to be done.

drivers/accel/adxl371/adxl37x.h
```diff
@@ -0,0 +1,484 @@
...
+                                                                               * INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+                                                                               * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA,
+                                                                               * OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF
+                                                                               * LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING
+                                                                               * NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE,
+                                                                               * EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+                                                                               *******************************************************************************/
+
+#ifndef ADXL371_H_
```
RaduSabau1: shouldn't this be ADXL37X_H_?

drivers/accel/adxl371/adxl37x.h
```diff
@@ -0,0 +1,484 @@
...
+#define ADXL371_INT2_MAP_INACT_MSK NO_OS_BIT(4)
+#define ADXL371_INT2_MAP_INACT_MODE(x) (((x) & 0x1) << 4)
+#define ADXL371_INT2_MAP_ACT_MSK NO_OS_BIT(5)
+#define ADXL371_INT2_MAP_ACT_MODE(x) (((x) & 0x1) << 5)
+#define ADXL371_INT2_MAP_AWAKE_MSK NO_OS_BIT(6)
+#define ADXL371_INT2_MAP_AWAKE_MODE(x) (((x) & 0x1) << 6)
+#define ADXL371_INT2_MAP_LOW_MSK NO_OS_BIT(7)
+#define ADXL371_INT2_MAP_LOW_MODE(x) (((x) & 0x1) << 7)
+
+//specific for ADXL372 INT1 and INT2 registers
```
RaduSabau1: you mean ADXL37X, right?

drivers/accel/adxl371/adxl37x.h
```diff
@@ -0,0 +1,484 @@
...
+	ADXL37x_FIFO_OLD_SAVED
+};
+
+enum adxl37x_type {
+	ADXL371,
+	ADXL372
+};
+
+struct adxl37x_chip_info {
+	int8_t wur;
```
RaduSabau1: why not use the enum types?

drivers/accel/adxl371/adxl37x.h
```diff
@@ -0,0 +1,484 @@
+/***************************************************************************/ /**
+                                                                               *   @file   adxl37x.h
+                                                                               *   @brief  Header file for adxl371 Driver.
```
RaduSabau1: adxl37x?

drivers/accel/adxl37x/adxl37x.c
```diff
@@ -0,0 +1,983 @@
...
+
+	ret = adxl37x_set_op_mode(dev, init_param.op_mode);
+	if (ret < 0)
+		goto error;
+
+	*device = dev;
+	printf("adxl371 successfully initialized\n");
+	return ret;
+
+error:
```
RaduSabau1, line 951: resource leak. if gpio, spi/i2c are initialized before the failing point, those resources are never released. use multiple error labels for that.

drivers/accel/adxl37x/adxl37x.c
```diff
@@ -0,0 +1,983 @@
...
+/**
+ * @brief Wrapper used of multibyte reads
+ *
+ * @param dev - The device structure
+ * @param reg_addr - The register address
+ * @param reg_data - The register data
+ * @param count - Number of bytes to read
+ * @return 0 in case of success, negative error code otherwise
+ */
+int adxl37x_read_reg_multiple(struct adxl37x_dev *dev, uint8_t reg_addr,
```
RaduSabau1, line 107: perhaps I am wrong about this, but I can't see the declaration of this in the header file.

drivers/accel/adxl37x/adxl37x.c
```diff
@@ -0,0 +1,983 @@
...
+		if ((ADXL371_STATUS_1_FIFO_RDY(status1)) ||
+		    (ADXL371_STATUS_1_FIFO_FULL(status1))) {
+			/*
+			 * When reading data from multiple axes from the FIFO,
+			 * to ensure that data is not overwritten and stored out
+			 * of order, at least one sample set must be left in the
+			 * FIFO after every read.
+			 */
+			*fifo_entries -= 3;
+			ret = adxl37x_get_fifo_xyz_data_workaround(dev, fifo_data, *fifo_entries);
```
RaduSabau1: buffer overflow? *fifo_entries is passed up here and then read as cnt * 6 bytes into dev->fifo_raw[1024]. with cnt = 509, that would be 3054 bytes written into a 1024 byte-buffer. I think *fifo_entries / 3 (for XYZ format) could be passed or divided by the appropriate number of axes based on the fifo_config (up to you). could you please test this theory out if wasn't already?

drivers/accel/adxl37x/adxl37x.c
```diff
@@ -0,0 +1,983 @@
...
+	if (dev->fifo_config.fifo_mode != ADXL37x_FIFO_BYPASSED) {
+		if ((ADXL371_STATUS_1_FIFO_RDY(status1)) ||
+		    (ADXL371_STATUS_1_FIFO_FULL(status1))) {
+			/*
+			 * When reading data from multiple axes from the FIFO,
+			 * to ensure that data is not overwritten and stored out
+			 * of order, at least one sample set must be left in the
+			 * FIFO after every read.
+			 */
+			*fifo_entries -= 3;
```
RaduSabau1: there is also an underflow risk I think. *fifo_entries -= 3 is done unconditionally on a uint16_t. if *fifo_entries < 3, it could wrap to a large positive value, a guard would necessary I would say.

drivers/accel/adxl371/adxl37x.c
```diff
@@ -0,0 +1,983 @@
...
+
+/**
+ * @brief Set timing register to external sync setting
+ *
+ * @param dev adxl371 device data structure
+ * @return int 0 if success, else negative error code
+ */
+int adxl37x_set_external_sync(struct adxl37x_dev *dev)
+{
+	return adxl37x_write_reg(dev, ADXL371_TIMING,
```
RaduSabau1: both set_external_sync and set_internal_sync functions may clear WUR bits. use write_mask instead.

drivers/accel/adxl37x/adxl37x.c
```diff
@@ -0,0 +1,983 @@
...
+	if (!internal_sync)
+		return -EFAULT;
+
+	enum adxl37x_op_mode op_mode;
+	ret = adxl37x_get_op_mode(dev, &op_mode);
+	if (ret < 0)
+		return ret;
+
+	/* @workaround Set the timing register to external sync along with odr */
+	adxl37x_set_external_sync(dev);
```
RaduSabau1: why discarding the return values?

drivers/accel/adxl371/adxl37x.c
```diff
@@ -0,0 +1,983 @@
...
+	uint8_t buf[6];
+	uint8_t status1, status2;
+	uint16_t fifo_entries;
+	int ret;
+
+	do {
+		ret = adxl37x_get_status(dev, &status1, &status2, &fifo_entries);
+		if (ret < 0)
+			return ret;
+	} while (!(ADXL371_STATUS_1_DATA_RDY(status1)));
```
RaduSabau1: a timeout should be used here. perhaps a counter would be the simplest way.

drivers/accel/adxl37x/adxl37x.c
```diff
@@ -0,0 +1,983 @@
...
+		dev->reg_read = adxl37x_i2c_reg_read;
+		dev->reg_write = adxl37x_i2c_reg_write;
+		dev->reg_read_multiple = adxl37x_i2c_reg_read_multiple;
+
+		ret = adxl37x_read_reg(dev, ADXL371_REVID, &rev_id);
+		if (ret < 0)
+			goto error;
+		/* Starting with the 3rd revision an I2C chip bug was fixed */
+		if (rev_id < 3) {
+			printf("I2C might not work properly with other "
```
RaduSabau1: use pr_info/pr_debug instead

drivers/accel/adxl37x/iio_adxl37x.c
```diff
@@ -0,0 +1,764 @@
...
+	if (!dev)
+		return -EINVAL;
+
+	iio_adxl37x = (struct adxl37x_iio_dev *)dev;
+
+	if (!iio_adxl37x->adxl37x_dev)
+		return -EINVAL;
+
+	adxl37x = iio_adxl37x->adxl37x_dev;
+	active_channels = iio_adxl37x->no_of_active_channels;
```
RaduSabau1: where is active_channels used?

drivers/accel/adxl37x/iio_adxl37x.c
```diff
@@ -0,0 +1,764 @@
...
+	if (!desc)
+		return -ENOMEM;
+
+	desc->iio_dev = &adxl37x_iio_dev;
+
+	ret = adxl37x_init(&desc->adxl37x_dev, *(init_param->adxl37x_dev_init));
+	if (ret)
+		goto error_init;
+
+	ret = adxl37x_set_op_mode(desc->adxl37x_dev, ADXL37x_FULL_BW_MEASUREMENT);
```
RaduSabau1: seems like the op_mode from init_param is ignored here.

drivers/accel/adxl371/iio_adxl37x.c
```diff
@@ -0,0 +1,764 @@
...
+				 (uint8_t)writeval);
+}
+
+/***************************************************************************//**
+ * @brief Handles the read request for raw attribute.
+*******************************************************************************/
+static int adxl37x_iio_read_raw(void *dev, char *buf, uint32_t len,
+				 const struct iio_ch_info *channel, intptr_t priv)
+{
+      int32_t ret;
```
RaduSabau1: use tabs for identation

drivers/accel/adxl371/README.rst
```diff
@@ -0,0 +1,820 @@
...
+>>>>>>> c08359952 (accel: adxl371: Add ADXL371 driver)
++----------------------------+---------------------+---------------------+---------------------+
+| ADXL371_FULL_BW_MEASUREMENT| Full-bandwidth      | 3                   | When full-bandwidth |
+|                            | measurement mode    |                     | sampling is required|
+|                            |                     |                     |                     |
+|                            |                     |                     |                     |
+|                            |                     |                     |                     |
++----------------------------+---------------------+---------------------+---------------------+
+
+<<<<<<< HEAD
```
RaduSabau1: please remove this unresolved merge conflicts.

drivers/accel/adxl371/README.rst
```diff
@@ -0,0 +1,820 @@
...
+=======
+
+>>>>>>> c08359952 (accel: adxl371: Add ADXL371 driver)
+Device Configuration
+--------------------
+
+Initialization Functions
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+The ``adxl371_init`` function sets up ADXL371 accelerometer by
```
RaduSabau1: the whole documentation seems to use old adxl371 names instead of adxl37x.

drivers/accel/adxl371/README.rst
```diff
@@ -0,0 +1,820 @@
+ADXL371 no-OS Driver
+====================
+
+Supported Devices
+-----------------
+
+- :adi:`ADXL371`
```
RaduSabau1: ADXL372 too, no?

drivers/accel/adxl371/README.rst
```diff
@@ -0,0 +1,820 @@
...
++----------------------------+---------------------+---------------------+---------------------+
+| Mode Name                  | Description         | Configuration Bits  | Typical Use Case    |
++----------------------------+---------------------+---------------------+---------------------+
+| ADXL371_STANDBY            | Standby mode for    | 0                   | When the device     |
+|                            | lowest power        |                     | needs to conserve   |
+|                            | consumption.        |                     | energy and data     |
+|                            |                     |                     | acquisition is not  |
+|                            |                     |                     | required            |
++----------------------------+---------------------+---------------------+---------------------+
+| ADXL371_WAKE_UP            | Standby mode for    | 1                   | Ideal for simple    |
```
RaduSabau1: This one seems to have the same description as STANDBY.

drivers/accel/adxl371/README.rst
```diff
@@ -0,0 +1,820 @@
...
+                  ADXL371_POWER_CTL,
+                  ADXL371_POWER_CTL_HPF_DIS_MSK,
+                  ADXL371_POWER_CTL_HPF_DIS_MODE(0b1));
+
+       /*
+        * Enable full-BW Measurement to set the reference
+        * if using internally referenced act/inact thresholds
+        */
+       adxl371_set_op_mode(dev, ADXL371_FULL_BW_MEASUREMENT);
+       MXC_Delay(MXC_DELAY_MSEC(10));
```
RaduSabau1: This is platform specific. replace with no_os_mdelay(10) instead.

drivers/accel/adxl371/README.rst
```diff
@@ -0,0 +1,820 @@
...
+       /*
+        * Enable full-BW Measurement to set the reference
+        * if using internally referenced act/inact thresholds
+        */
+       adxl371_set_op_mode(dev, ADXL371_FULL_BW_MEASUREMENT);
+       no_os_mdelay(10);
+       adxl371_set_op_mode(dev, ADXL371_STANDBY);
+
+       /* Clear fifo completely before enabling Instant On mode */
+       err = adxl371_get_status(adxl_device, &status1, &status2, &fifo_entries);
```
RaduSabau1: adxl_device is not declared in the code snippet, multiple undeclared variables in the code snippet seem to appear in the documentation, please check those other occurrences as well.

drivers/accel/adxl371/README.rst
```diff
@@ -0,0 +1,820 @@
...
+            .bw =  ADXL371_BW_2560HZ,
+            .fifo_config = {
+                .fifo_format = ADXL371_XYZ_PEAK_FIFO,
+                .fifo_mode = ADXL371_FIFO_STREAMED,
+                .fifo_samples = ACCEL_MAX_SAMPLES
+            },
+            .op_mode = ADXL371_FULL_BW_MEASUREMENT,
+            .act_proc_mode = ADXL371_LOOPED
+        };
+       ret = adxl371_init(&adxl371_desc, adxl371_ip);
```
RaduSabau1: nit: there is a name mismatch here with the previously declared ip.

projects/eval-adxl37x/src/examples/basic_example/basic_example.c
```diff
@@ -0,0 +1,88 @@
...
+        
+        for(int i=0;i<numberSamples;i++){
+            x=(int16_t)(recorded_data[i].x<<4)>>4;//sign extension
+            y=(int16_t)(recorded_data[i].y<<4)>>4;
+            z=(int16_t)(recorded_data[i].z<<4)>>4;
+            pr_info("Datele accel sunt la sample %d: x=%d,y=%d,z=%d\r\n",i,x,y,z);
+        }
+    }
+    error_init:
+    no_os_free(dev);
```
RaduSabau1: you are double-freeing the device here.

projects/eval-adxl37x/src/examples/basic_example/basic_example.c
```diff
@@ -0,0 +1,88 @@
...
+    error:
+    pr_info("Error!\n");
+    return 0;
+
+}
```
RaduSabau1: EOF line

projects/eval-adxl37x/src/examples/basic_example/basic_example.c
```diff
@@ -0,0 +1,88 @@
...
+            x=(int16_t)(recorded_data[i].x<<4)>>4;//sign extension
+            y=(int16_t)(recorded_data[i].y<<4)>>4;
+            z=(int16_t)(recorded_data[i].z<<4)>>4;
+            pr_info("Datele accel sunt la sample %d: x=%d,y=%d,z=%d\r\n",i,x,y,z);
+        }
+    }
+    error_init:
+    no_os_free(dev);
+    error:
+    pr_info("Error!\n");
```
RaduSabau1: the identation seems wrong in this file. also, use tabs (8 spaces long) instead of whitespaces.

projects/eval-adxl37x/src/examples/basic_example/basic_example.c
```diff
@@ -0,0 +1,88 @@
...
+    return 0;
```
RaduSabau1: error path should return the ret value, 0 means success which is not the case in an error label.

projects/eval-adxl37x/src/examples/basic_example/basic_example.c
```diff
@@ -0,0 +1,88 @@
...
+
+        ret=adxl37x_get_fifo_xyz_data_workaround(dev,recorded_data,numberSamples);
+        if(ret!=0)
+            goto error;
+        
+        for(int i=0;i<numberSamples;i++){
+            x=(int16_t)(recorded_data[i].x<<4)>>4;//sign extension
+            y=(int16_t)(recorded_data[i].y<<4)>>4;
+            z=(int16_t)(recorded_data[i].z<<4)>>4;
+            pr_info("Datele accel sunt la sample %d: x=%d,y=%d,z=%d\r\n",i,x,y,z);
```
RaduSabau1: I think english would be nice here instead.

projects/eval-adxl37x/src/examples/basic_example/basic_example.c
```diff
@@ -0,0 +1,88 @@
...
+        goto error_init;
+    volatile uint8_t pwr_ctl;
+
+
+    adxl37x_read_reg(dev,0x3F,&pwr_ctl);
+    pr_info("pwr_ctl register after init of adxl37x: 0x%02X\n\r",pwr_ctl);
+
+
+    //extract data
+    struct adxl37x_xyz_accel_data recorded_data[128];
```
RaduSabau1: usually, variables are declared at the start of the function.

projects/eval-adxl37x/src/examples/basic_example/basic_example.c
```diff
@@ -0,0 +1,88 @@
...
+    int count=0;
```
RaduSabau1: space between variable name, '=' sign and variable value is preferred. check for occurrences like this throughout the whole PR please.

projects/eval-adxl37x/src/examples/basic_example/basic_example.c
```diff
@@ -0,0 +1,88 @@
...
+
+	no_os_uart_stdio(uart);
+
+    ret=adxl37x_init(&dev,dev_init);
+    if(ret!=0)
+        goto error_init;
+    volatile uint8_t pwr_ctl;
+
+
+    adxl37x_read_reg(dev,0x3F,&pwr_ctl);
```
RaduSabau1: 0x3F seems like a magic number. use a macro for it please. in this case I think ADXL371_POWER_CTL (which could also be renamed to use the ADXL37X prefix - check for other occurrences please).

projects/eval-adxl37x/src/examples/basic_example/basic_example.c
```diff
@@ -0,0 +1,88 @@
...
+	ret = no_os_uart_init(&uart, &adxl37x_uart_ip);
+	if (ret)
+		return ret;
+
+	no_os_uart_stdio(uart);
+
+    ret=adxl37x_init(&dev,dev_init);
+    if(ret!=0)
+        goto error_init;
+    volatile uint8_t pwr_ctl;
```
RaduSabau1: why volatile?

projects/eval-adxl37x/src/examples/basic_example/basic_example.c
```diff
@@ -0,0 +1,88 @@
...
+
```
RaduSabau1, line 40: extra white space.

projects/eval-adxl37x/src/examples/iio_example/iio_example.c
```diff
@@ -28,54 +29,58 @@
...
+	struct iio_data_buffer accel_buff = {
+		.buff = (void *)iio_data_buf,
+		.size = ADXL37X_MAX_BUFFER_SAMPLES * 3 * sizeof(int),
+	};
 
-	ltc2983_iio_ip.ltc2983_desc_init_param = &ltc2983_ip;
-	ret = ltc2983_iio_init(&ltc2983_iio_dev, &ltc2983_iio_ip);
+	adxl37x_iio_ip.adxl37x_dev_init = &adxl37x_init_param;
+	ret = adxl37x_iio_init(&adxl37x_iio_desc, &adxl37x_iio_ip);
+	//adxl37x_set_internal_sync(adxl37x_iio_desc->adxl37x_dev);
```
RaduSabau1: this seems like commented-out dead code, I think it should be removed.

projects/eval-adxl37x/src/platform/maxim/parameters.h
```diff
@@ -0,0 +1,53 @@
...
+#define UART_DEVICE_ID  0
+#endif
+#define INTC_DEVICE_ID  0
+#define UART_BAUDRATE   57600
+#define UART_EXTRA      &adxl37x_uart_extra
+#define UART_OPS        &max_uart_ops
+
+extern struct max_uart_init_param adxl37x_uart_extra;
+
+#define SPI_BAUDRATE    100000
```
RaduSabau1: is this enough considering the ODR settings used elsewhere?

RaduSabau1 (changes requested):
I think the commits could be cleaned up a bit:
- There's a lot of unnecessary diff especially in the first one. I think there could be a single clean commit that adds support for both parts. (currently you are having the adxl37x.c driver in the adxl371 folder, this looks wrong).
- Commit ordering, usually the first one would be driver, IIO driver, driver RST, example, example RST.

projects/eval-adxl37x/src/examples/basic_example/basic_example.c
```diff
@@ -0,0 +1,88 @@
...
+    int32_t ret;
+	struct no_os_uart_desc *uart;
+
+	ret = no_os_uart_init(&uart, &adxl37x_uart_ip);
+	if (ret)
+		return ret;
+
+	no_os_uart_stdio(uart);
+
+    ret=adxl37x_init(&dev,dev_init);
```
amiclaus: the `adxl37x_init` already handles the no_os_free internally. nit: missing spaces between operands.

RaulGeo289:
Changelog V1:
-cleaned up the commit order and the contents

drivers/accel/adxl371/adxl37x.h
RaulGeo289: changed to ADXL37X

drivers/accel/adxl371/adxl37x.h
RaulGeo289: changed to ADXL37X

drivers/accel/adxl371/adxl37x.h
RaulGeo289: changed to use enum types

drivers/accel/adxl371/adxl37x.h
RaulGeo289: changed it to adxl37x and fixed indentation

drivers/accel/adxl37x/adxl37x.c
RaulGeo289: fixed leaks using error labels and changed the device struct allocation to use calloc instead of malloc, because i need them for the checks in the label bodies

projects/eval-adxl37x/src/examples/basic_example/basic_example.c
RaulGeo289: right.The compiler does not have a reason to optimize that variable.Removed volatile

projects/eval-adxl37x/src/examples/basic_example/basic_example.c
RaulGeo289: removed it

projects/eval-adxl37x/src/examples/basic_example/basic_example.c
RaulGeo289: fixed it  and found some other occurrences as well

projects/eval-adxl37x/src/examples/basic_example/basic_example.c
RaulGeo289: fixed it

RaulGeo289:
Changelog V2:
-fixed memory leaks in adxl37x_init()
-removed volatile from example as it is not necessary here
-fixed some project and driver parts that had the stale adxl371 name

projects/eval-adxl37x/src/examples/basic_example/basic_example.c
RaulGeo289: fixed it

projects/eval-adxl37x/src/examples/basic_example/basic_example.c
RaulGeo289: you're right
Now it returns ret value.

projects/eval-adxl37x/src/examples/basic_example/basic_example.c
RaulGeo289: fixed it

RaulGeo289:
Changelog V3:
-implemented a way to avoid entering the undefined memory zone in the adxl37x_get_fifo_xyz_data_workaround() function and avoid possible overflow in adxl37x_service_fifo_ev(), but in the meantime I found a better way so i'm gonna implement that

drivers/accel/adxl37x/adxl37x.c
RaulGeo289: fixed it

drivers/accel/adxl371/README.rst
RaulGeo289: fixed them

drivers/accel/adxl371/README.rst
RaulGeo289: fixed them and looked for other instances and fixed them as well

drivers/accel/adxl371/README.rst
RaulGeo289: yes, added ADXL372 as well

drivers/accel/adxl371/README.rst
RaulGeo289: removed the impact detection and triggered examples, as they were not necessary
updated the documentation as well

projects/eval-adxl37x/src/examples/basic_example/basic_example.c
RaulGeo289: fixed it

projects/eval-adxl37x/src/examples/iio_example/iio_example.c
RaulGeo289: removed the unused line of code

projects/eval-adxl37x/src/examples/basic_example/basic_example.c
RaulGeo289: fixed it and applied astyle guidelines

drivers/accel/adxl371/README.rst
RaulGeo289: fixed the code snippets

drivers/accel/adxl371/README.rst
RaulGeo289: fixed the code snippets

drivers/accel/adxl37x/adxl37x.c
RaulGeo289: you're right.I fixed it.

projects/eval-adxl37x/src/examples/basic_example/basic_example.c
RaulGeo289: you're right.Removed error_init label in basic_example.c .

projects/eval-adxl37x/src/examples/basic_example/basic_example.c
RaulGeo289: you're right.Removed error_init label in basic_example.c .

drivers/accel/adxl371/iio_adxl37x.c
RaulGeo289: used tabs for indentation

drivers/accel/adxl371/README.rst
RaulGeo289: fixed it

drivers/accel/adxl37x/iio_adxl37x.c
RaulGeo289: active_channels is not used here, but i followed adxl355 IIO driver, which initialized this attribute, even if it did not used it.
The general purpose of that is to update the number of channels when streaming data in iio oscilloscope.

drivers/accel/adxl37x/iio_adxl37x.c
RaulGeo289: you're right.I forgot to change it after testing i think.I changed it to op_mode attribute used in init_param.

drivers/accel/adxl37x/adxl37x.c
RaulGeo289: changed to pr_info

drivers/accel/adxl371/adxl37x.c
RaulGeo289: you're right.The function should write only the bits needed for that exact feature.

RaulGeo289:
Changelog V4:
-solved possible access to undefined memory areas, caused by the fact that the driver assumes 512 sample for 3 axes used, which could be more than the FIFO's depth.It needs, in this case,no more than 170 samples.I did this trough if elif macro guards in common_data.c .Also used AXES_USED macro line to calculate how many samples the driver actually gets from the FIFO, as there is a constraint that it should leave at least one sample in the FIFO for data integrity, so it eliminates the risk of underflow in adxl37x_get_fifo_xyz_data_workaround() function.
-did some IIO driver, doc, basic_example and iio_example clean-up

drivers/accel/adxl37x/adxl37x.c
RaulGeo289: fixed it with a macro that calculates the number of extracted samples based on the axes used and eliminated.With this change, I removed this line as well

drivers/accel/adxl37x/adxl37x.c
RaulGeo289: fixed it with another macro that sets the number of extracted samples and maximum samples, so now it won't access that undefined memory area.I tested it on IIO oscilloscope.On the system testing, it looked mostly fine, even though there may have been some slight delays, or wrong values(probably these values were extracted from the undefined memory zone), but mostly everything was fine.

RaulGeo289:
Changelog v5:
-handled selection for the number of axes used
-clean-up ,testing ,solving driver bugs

RaulGeo289:
Changelog v6:
-documented the selection of the axes in the project readme

RaulGeo289:
Changelog v6:
-fixed bugs in IIO driver

RaulGeo289:
ready for review

amiclaus (dismissed):

buha:
please rebase on latest main and check into adding cmake build support

RaulGeo289:
Changelog V7
-rebased with main
-added CMake support

drivers/accel/adxl37x/adxl37x.c
```diff
@@ -0,0 +1,1117 @@
...
+
+	*device = dev;
+	pr_info("adxl37x successfully initialized\n");
+	return ret;
+
+error_GPIO:
+	if (dev->gpio_int1 != NULL) {
+		ret = no_os_gpio_remove(dev->gpio_int1);
+		if (ret < 0)
+			return ret;
```
RaduSabau1: memory leak. each descriptor that gets initialized should have its own error label if possible. also call the remove function without returning the errno code, otherwise next error labels will not get called and thus the memory leak.

drivers/accel/adxl37x/adxl37x.c
```diff
@@ -0,0 +1,1117 @@
...
+		cnt++;
+	} while (cnt != 100 && !(ADXL37X_STATUS_1_DATA_RDY(status1)));
+
+	ret = adxl37x_read_reg_multiple(dev, ADXL37X_X_MAXPEAK_H, buf,
+					NO_OS_ARRAY_SIZE(buf));
+	if (ret)
+		return ret;
+
+	max_peak->x = (buf[0] << 4) | (buf[1] >> 4);
```
RaduSabau1: isn't sign extension needed here?

drivers/accel/adxl37x/adxl37x.c
```diff
@@ -0,0 +1,1117 @@
...
+		return -EINVAL;
+
+	/* @workaround must be in internal sync mode */
+	ret = adxl37x_get_internal_sync(dev, &internal_sync);
+	if (ret < 0)
+		return ret;
+	if (!internal_sync)
+		return -EFAULT;
+
+	enum adxl37x_op_mode op_mode;
```
RaduSabau1, line 1070: I think its ok for this one to be moved at the top of the function for consistency

drivers/accel/adxl37x/adxl37x.c
```diff
@@ -0,0 +1,1117 @@
...
+
+	/* @workaround clear the fifo (bypass mode) */
+	ret = adxl37x_configure_fifo(dev, ADXL37x_FIFO_BYPASSED,
+				     dev->fifo_config.fifo_format,
+				     dev->fifo_config.fifo_samples, op_mode);
+	if (ret < 0)
+		return ret;
+
+	/* @workaround set the desired fifo mode */
+	ret = adxl37x_configure_fifo(dev, fifo_mode, dev->fifo_config.fifo_format,
```
RaduSabau1, line 1102: if (ret) return ret; after this function otherwise the error code gets lost and it could also result in unexpected behavior perhaps

drivers/accel/adxl37x/adxl37x.h
```diff
@@ -0,0 +1,474 @@
...
+
+/* Specific for ADXL372 INT1 and INT2 registers */
+#define ADXL37X_INT1_MAP_FIFO_RDY_MSK NO_OS_BIT(1)
+#define ADXL37X_INT1_MAP_FIFO_RDY_MODE(x) (((x) & 0x1) << 1)
+#define ADXL37X_INT1_MAP_FIFO_FULL_MSK NO_OS_BIT(2)
+#define ADXL37X_INT1_MAP_FIFO_FULL_MODE(x) (((x) & 0x1) << 2)
+#define ADXL37X_INT1_MAP_FIFO_OVR_MSK NO_OS_BIT(3)
+#define ADXL37X_INT1_MAP_FIFO_OVR_MODE(x) (((x) & 0x1) << 3)
+
+enum adxl37x_axis { ADXL37x_X_AXIS, ADXL37x_Y_AXIS, ADXL37x_Z_AXIS };
```
RaduSabau1: nit: higher case X for consistency with the macros, styling purposes, up to you.

drivers/accel/adxl37x/iio_adxl37x.c
```diff
@@ -0,0 +1,842 @@
...
+		} else {
+			used_freq_table = adxl372_iio_odr_table;
+		}
+
+		//iterate trough the table and get each value in buf,then concatanate
+		// with the buffer used by IIO
+		for (uint8_t i = 0; i < ODR_MODES; i++) {
+			val = used_freq_table[i];
+			ret = iio_format_value(buffer, sizeof(buffer), IIO_VAL_INT, 1, &val);
+			if (ret != 0)
```
RaduSabau1: was this tested? I think iio_format_value returns bytes written (> 0) on success. It would exit early because ret > 0. read_lpf_freq_avail function correctly uses "< 0".

drivers/accel/adxl37x/iio_adxl37x.c
```diff
@@ -0,0 +1,842 @@
...
+	struct adxl37x_dev* adxl37x;
+	if (!dev)
+		return -EINVAL;
+	iio_adxl37x = (struct adxl37x_iio_dev*) dev;
+	if (!iio_adxl37x->adxl37x_dev)
+		return -EINVAL;
+	adxl37x = iio_adxl37x->adxl37x_dev;
+	uint8_t raw_buf[3];
+	ret = adxl37x->reg_read_multiple(adxl37x, ADXL37X_OFFSET_X, raw_buf, 3);
+
```
RaduSabau1, line 323: unnecessary blank line

drivers/accel/adxl37x/iio_adxl37x.c
```diff
@@ -0,0 +1,842 @@
...
+static int adxl37x_iio_read_calibbias(void *dev, char *buf, uint32_t len,
+				      const struct iio_ch_info *channel, intptr_t priv)
+{
+	int val;
+	int ret;
+	struct adxl37x_iio_dev* iio_adxl37x;
+	struct adxl37x_dev* adxl37x;
+	if (!dev)
+		return -EINVAL;
+	iio_adxl37x = (struct adxl37x_iio_dev*) dev;
```
RaduSabau1, line 318: nit: I know its a styling choice mostly but a blank line could exist between this line and the previous one it makes things more readable. however a blank line before ret = function_call() and the next if (ret) line should not exist since the existing of it would make things less readable.

drivers/accel/adxl37x/iio_adxl37x.c
```diff
@@ -0,0 +1,842 @@
...
+						  3);			//use 0xF as mask to get only the lower nibble
+			return iio_format_value(buf, len, IIO_VAL_INT, 1, &val);
+		case ACCEL_AXIS_Z:
+			val = no_os_sign_extend32(raw_buf[2] & 0xF, 3);
+			return iio_format_value(buf, len, IIO_VAL_INT, 1, &val);
+		default:
+			pr_err("No axis found!\n\r");
+			return -EINVAL;
+		}
+
```
RaduSabau1: unnecessary blank line, check other occurences. mostly try and look up other code from the repo in order to have a better grasp on where to use the blank lines and where not to.

drivers/accel/adxl37x/iio_adxl37x.c
```diff
@@ -0,0 +1,842 @@
...
+	int ret;
+	struct adxl37x_iio_dev* iio_adxl37x;
+	struct adxl37x_dev* adxl37x;
+	if (!dev)
+		return -EINVAL;
+	iio_adxl37x = (struct adxl37x_iio_dev*) dev;
+	if (!iio_adxl37x->adxl37x_dev)
+		return -EINVAL;
+	adxl37x = iio_adxl37x->adxl37x_dev;
+	uint8_t raw_buf[3];
```
RaduSabau1: declaration at the start of the function, check other occurences.

drivers/accel/adxl37x/iio_adxl37x.c
```diff
@@ -0,0 +1,842 @@
...
+	if (ret)
+		return ret;
+	if (channel->type == IIO_ACCEL) {
+		switch (channel->address) {
+		case ACCEL_AXIS_X:
+			val = no_os_sign_extend32(raw_buf[0] & 0xF, 3);
+			return iio_format_value(buf, len, IIO_VAL_INT, 1, &val);
+		case ACCEL_AXIS_Y:
+			val = no_os_sign_extend32(raw_buf[1] & 0xF,
+						  3);			//use 0xF as mask to get only the lower nibble
```
RaduSabau1: firstly, move this line on the same with the previous one and move the comment before the call using C style comments ( /* ... */ ). also, why only the lower nibble?

drivers/accel/adxl37x/iio_adxl37x.c
```diff
@@ -0,0 +1,842 @@
...
+	adxl37x = iio_adxl37x->adxl37x_dev;
+
+	int i = 0;
+	ret = adxl37x_get_fifo_xyz_data_workaround(adxl37x, data_buffer, samples);
+	if (ret) {
+		pr_err("Error in getting the samples!\r\n");
+		return ret;
+	}
+
+	for (int s = 0; s < samples; s++) {
```
RaduSabau1: s could also be declared at the start of the function

drivers/accel/adxl37x/iio_adxl37x.c
```diff
@@ -0,0 +1,842 @@
...
+}
+
+/***************************************************************************//**
+ * @brief Reads N samples in polling mode (read_dev path).
+ *        Burst-reads X_DATA_H..Z_DATA_L for each sample, fills buff[].
+*******************************************************************************/
+static int adxl37x_iio_read_samples(void *dev, int *buff, uint32_t samples)
+{
+	static struct adxl37x_xyz_accel_data data_buffer[ADXL37X_FIFO_MAX_SAMPLES];
+	if (samples > ADXL37X_FIFO_MAX_SAMPLES)
```
RaduSabau1: couldn't this be moved from where variables are declared?

projects/eval-adxl37x/src/common/common_data.h
```diff
@@ -0,0 +1,33 @@
+
```
RaduSabau1, line 10: License header missing.

projects/eval-adxl37x/src/common/common_data.h
```diff
@@ -0,0 +1,33 @@
...
+#if AXES_USED==3
+#define MAX_SAMPLES 170 //the maximum number of samples in the fifo, according to the number of axes used
+#elif AXES_USED==2
+#define MAX_SAMPLES 256
+#elif AXES_USED==1
+#define MAX_SAMPLES 512
+#endif
+
+#define EXTRACTED_SAMPLES MAX_SAMPLES-1  //at least 1 sample left in the FIFO for sample integrity
```
RaduSabau1: MAX_SAMPLES-1 should be in parentheses.

projects/eval-adxl37x/src/examples/basic_example/basic_example.c
```diff
@@ -0,0 +1,92 @@
...
+		goto error;
+	}
+
+	while (1) {
+		ret = adxl37x_get_fifo_xyz_data_workaround(dev, recorded_data, numberSamples);
+		if (ret != 0)
+			goto error;
+
+		for (int i = 0; i < numberSamples; i++) {
+			x = (int16_t)(recorded_data[i].x << 4) >> 4; //sign extension
```
RaduSabau1: I am not 100% sure but this looks redundant and perhaps incorrect. I am saying this because I suppose the FIFO parse functions already produce correctly-signed 12-bit values in int16_t so the manual re-sign-extension is unnecessary and also left-shifting a signed value that overflows is undefined behaviour in C, was this tested?

projects/eval-adxl37x/src/examples/basic_example/basic_example.c
```diff
@@ -0,0 +1,92 @@
...
+			y = (int16_t)(recorded_data[i].y << 4) >> 4;
+			z = (int16_t)(recorded_data[i].z << 4) >> 4;
+			pr_info("Accelerometer data for sample number %d is: x=%d,y=%d,z=%d\r\n", i, x,
+				y, z);
+		}
+	}
+
+	no_os_free(dev);
```
RaduSabau1: memory leak. code can't get here unless it exits while (1) which seems unlikely :). have this in a free_dev error label or something like that up to you. also the UART desc doesn't get freed as well

projects/eval-adxl37x/src/common/common_data.c
```diff
@@ -0,0 +1,58 @@
+#include "common_data.h"
```
RaduSabau1, line 11: missing licence header, check other occurences

projects/eval-adxl37x/src/examples/basic_example/basic_example.c
```diff
@@ -0,0 +1,92 @@
...
+
+#include "no_os_print_log.h"
+#include "no_os_delay.h"
+#include "no_os_alloc.h"
+#include "adxl37x.h"
+#include "common_data.h"
+
+int example_main()
+{
+	//adxl37x init and start SPI
```
RaduSabau1: c style comments are preferred

projects/eval-adxl37x/src/examples/iio_example/iio_example.c
```diff
@@ -0,0 +1,92 @@
...
+	app_init_param.devices = iio_devices;
+	app_init_param.nb_devices = NO_OS_ARRAY_SIZE(iio_devices);
+	app_init_param.uart_init_params = adxl37x_uart_ip;
+
+	ret = iio_app_init(&app, app_init_param);
+	if (ret)
+		goto error;
+
+	return iio_app_run(app);
+
```
RaduSabau1, line 75: call iio_app_remove before returning otherwise you have a memory leak.

RaulGeo289:
Changelog V8:
-fixed memory leaks
-fixed maximum peak results by sign extending them

RaulGeo289:
Changelog V9:
-parsed the samples using uint8_t variables to avoid undefined behavior
-removed sign extension from basic example, because the driver does it now

amiclaus (dismissed):

drivers/accel/adxl37x/adxl37x.c
```diff
@@ -0,0 +1,1123 @@
...
+ * THIS SOFTWARE IS PROVIDED BY ANALOG DEVICES, INC. “AS IS” AND ANY EXPRESS OR
+ * IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF
+ * MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO
+ * EVENT SHALL ANALOG DEVICES, INC. BE LIABLE FOR ANY DIRECT, INDIRECT,
+ * INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA,
+ * OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF
+ * LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING
+ * NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE,
+ * EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
```
rbolboac: please add a SPDX BSD-3 tag instead in all new files
```
 * Copyright 2026(c) Analog Devices, Inc.
 *
 * SPDX-License-Identifier: BSD-3-Clause
```

RaulGeo289:
Changelog v10
-changed license for all files and added docuementation for the functions in IIO driver
-added a minimal calibration procedure in basic example to make the ADXL behave more adequate
-implemented a method trough which the user reads the sample susing the data_h and data_l registers
-verified 1 axis ,2 axes modes in IIO driver

RaulGeo289:
I checked with the linux driver for ADXL372 and did a simillar approach in my IIO driver, when trying to switch the number of axes used by ADXL.In more detail, for example ,I want to switch to the x-axis only samples in the FIFO, i'm choosing in the main IIO oscilloscope window to view only the x-axis.This is done in linux as well.Please let me know if this is still the correct approach.

drivers/accel/adxl37x/adxl37x_i2c.c
```diff
@@ -0,0 +1,95 @@
...
+				      uint8_t *reg_data, uint16_t count)
+{
+	uint8_t buf[512];
+	int32_t ret;
+
+	if (count > 512)
+		return -EINVAL;
+
+	buf[0] = reg_addr;
+	memset(&buf[1], 0x00, count - 1);
```
rbolboac: uint8_t buf[512] = {0}; instead

drivers/accel/adxl37x/adxl37x_i2c.c
```diff
@@ -0,0 +1,95 @@
...
+
+	ret = no_os_i2c_write(dev->i2c_desc, buf, 1, 0);
+	if (ret < 0)
```
rbolboac: if (ret) instead, see everywhere where applicable

drivers/accel/adxl37x/adxl37x_spi.c
```diff
@@ -0,0 +1,88 @@
...
+				      uint8_t *reg_data, uint16_t count)
+{
+	uint8_t buf[1025];
+	int32_t ret;
+
+	if (count > 1024)
+		return -EINVAL;
+
+	buf[0] = ADXL37X_REG_READ(reg_addr);
+	memset(&buf[1], 0x00, count);
```
rbolboac: same buf[1025] = {0}' could work

drivers/accel/adxl37x/adxl37x.c
```diff
@@ -0,0 +1,1134 @@
...
+ *
+ * @param dev - The device structure
+ * @param reg_addr - The register address
+ * @param reg_data - The register data
+ * @return 0 in case of success, negative error code otherwise
+ */
+int adxl37x_read_reg(struct adxl37x_dev *dev, uint8_t reg_addr,
+		     uint8_t *reg_data)
+{
+	return (int)dev->reg_read(dev, reg_addr, reg_data);
```
rbolboac, line 78: dev should be checked for null, se everywhere

drivers/accel/adxl37x/adxl37x.c
```diff
@@ -0,0 +1,1134 @@
...
+				   uint16_t thresh, bool referenced,
+				   bool enable)
+{
+	uint8_t th_val_h, th_val_l, i;
+	int ret = 0;
+
+	th_val_h = (thresh >> 3);
+	th_val_l = (thresh << 5) | (referenced << 1) | enable;
+
+	for (i = 0; i < 3; i++) {
```
rbolboac: don't use magic numbers, use macros instead

drivers/accel/adxl37x/adxl37x.c
```diff
@@ -0,0 +1,1134 @@
...
+ */
+int adxl37x_configure_fifo(struct adxl37x_dev *dev, enum adxl37x_fifo_mode mode,
+			   enum adxl37x_fifo_format format,
+			   uint16_t fifo_samples,
+			   enum adxl37x_op_mode op_mode)
+{
+	uint8_t fifo_config;
+	int ret;
+
+	if (fifo_samples > 512)
```
rbolboac: don't use magic values, use macros instead

drivers/accel/adxl37x/adxl37x.c
```diff
@@ -0,0 +1,1134 @@
...
+
+	if (ADXL37X_STATUS_1_FIFO_OVR(status1)) {
+		pr_err("FIFO overrun\n");
+		return -EFAULT;
+	}
+
+	if (dev->fifo_config.fifo_mode != ADXL37X_FIFO_BYPASSED) {
+		if ((ADXL37X_STATUS_1_FIFO_RDY(status1)) ||
+		    (ADXL37X_STATUS_1_FIFO_FULL(status1))) {
+			ret = adxl37x_get_fifo_xyz_data_workaround(dev, fifo_data, *fifo_entries);
```
rbolboac: can return directly

drivers/accel/adxl37x/iio_adxl37x.c
```diff
@@ -0,0 +1,1029 @@
...
+	if (!dev)
+		return -EINVAL;
+
+	iio_adxl37x = (struct adxl37x_iio_dev*) dev;
+	if (!iio_adxl37x->adxl37x_dev)
+		return -EINVAL;
+
+	adxl37x = iio_adxl37x->adxl37x_dev;
+
+	ret = adxl37x->reg_read_multiple(adxl37x, ADXL37X_OFFSET_X, raw_buf, 3);
```
rbolboac: can we not read a single register per requested channel?

drivers/accel/adxl37x/iio_adxl37x.c
```diff
@@ -0,0 +1,1029 @@
...
+
+	iio_adxl37x = (struct adxl37x_iio_dev *)dev;
+
+	if (!iio_adxl37x->adxl37x_dev)
+		return -EINVAL;
+
+	adxl37x = iio_adxl37x->adxl37x_dev;
+
+	if (channel->type == IIO_ACCEL) {
+		val = (adxl37x->adxl_type == ADXL371) ? adxl37x_iio_odr_table[adxl37x->odr] :
```
rbolboac: this could be a switch case just in case we add another chip in the future, then we can easily add something else in the switch case

rbolboac (commented):
some remarks from my side, please check if the marked issues can be found in other places and update those as well

drivers/accel/adxl37x/adxl37x.c
```diff
@@ -0,0 +1,1134 @@
...
+	uint16_t need = dev->fifo_config.extracted_samples *
+			dev->fifo_config.bytes_per_sample;
+	uint16_t j = 0;
+	uint16_t s;
+	uint8_t a;
+
+	if (need > sizeof(dev->fifo_raw))
+		return -EINVAL;
+
+	for (s = 0; s < dev->fifo_config.extracted_samples; s++) {
```
RaduSabau1: why is cnt parameter ignored? using extracted_samples, which can exceed ADXL37X_FIFO_MAX_SAMPLES (170) will overflow the static data_buffer.

projects/eval-adxl37x/src/examples/basic_example/basic_example.c
```diff
@@ -0,0 +1,109 @@
...
+	if (ret)
+		goto error_dev;
+	ret = adxl37x_write_reg(dev, ADXL37X_OFFSET_Y, 0 & 0x0F);
+	if (ret)
+		goto error_dev;
+	ret = adxl37x_write_reg(dev, ADXL37X_OFFSET_Z, 1 & 0x0F);
+	if (ret)
+		goto error_dev;
+
+	pr_info("offset trims: OFFSET_X=%d , OFFSET_Y=%d , OFFSET_Z=%d \n\r",
```
RaduSabau1: off_x, off_y and off_z appear to be printed but never intialized/assigned, wrong values might be printed here

drivers/accel/adxl37x/README.rst
```diff
@@ -0,0 +1,338 @@
...
+
+.. code-block:: C
+
+   #include "adxl37x.h"
+   #include "no_os_i2c.h"
+   #include "no_os_spi.h"
+   #include "no_os_alloc.h"
+   #include "no_os_delay.h"
+
+   #include "adxl37x_noos_data.h"
```
RaduSabau1: where is this header? remove if it doesn't exist

drivers/accel/adxl37x/README.rst
```diff
@@ -0,0 +1,338 @@
...
+   	/* Initialize the device */
+   	struct adxl37x_init_param adxl_ip = {
+   		.spi_init = spiInitParams,
+   		.gpio_int1 = int1Init,
+   		.int1_config = {},
+   		.gpio_int2 = int2Init,
+   		.int2_config = {},
+   		.comm_type =  ADXL37X_SPI,
+   		.odr =  ADXL37X_ODR_5120HZ,
+   		.bw =  ADXL37X_BW_2560HZ,
```
RaduSabau1: these params still exist?

drivers/accel/adxl37x/README.rst
```diff
@@ -0,0 +1,338 @@
...
+   	{
+   		.name        = "adxl37x",
+   		.dev         = adxl37x_iio_desc,
+   		.dev_descriptor = adxl37x_iio_desc->iio_dev,
+   		.read_buff   = &accel_buff,
+   		.write_buff  = NULL,
+   	}
+   };
+
+   return iio_app_run(iio_devices, NO_OS_ARRAY_SIZE(iio_devices));
```
RaduSabau1: the signature of the function is wrong here

RaulGeo289:
Changelog V11:
-implemented remarks from @rbolboac and @RaduSabau1 checked for other occurrences
-i detailed the process of switching the number of axes and which axes are registered by the accelerometer

RaulGeo289:
-solved astyle issues

drivers/accel/adxl37x/adxl37x_i2c.c
```diff
@@ -0,0 +1,106 @@
...
+int32_t adxl37x_i2c_reg_read_multiple(struct adxl37x_dev *dev, uint8_t reg_addr,
+				      uint8_t *reg_data, uint16_t count)
+{
+	uint8_t buf[ADXL37X_I2C_MAX_XFER_BYTES] = {0};
+	int32_t ret;
+
+	if (!dev)
+		return -EINVAL;
+
+	if (count > ADXL37X_I2C_MAX_XFER_BYTES)
```
rbolboac, line 90: should this work with fifo? fifo size is maximum 1024 so I expect some problems here

drivers/accel/adxl37x/iio_adxl37x.c
```diff
@@ -0,0 +1,1105 @@
...
+	struct adxl37x_iio_dev *iio_adxl37x;
+	struct adxl37x_dev *adxl37x;
+	uint32_t i = 0;
+	uint32_t s;
+	int32_t ret;
+
+	if (!dev)
+		return -EINVAL;
+
+	if (samples > ADXL37X_FIFO_MAX_SAMPLES)
```
rbolboac, line 1008: this is valid for 3 axis, how about if the client request only 1 or 2 axis reads?

drivers/accel/adxl37x/iio_adxl37x.c
```diff
@@ -0,0 +1,1105 @@
...
+		max_samples = ADXL37X_FIFO_MAX_ENTRIES;
+		break;
+	case 2:
+		max_samples = ADXL37X_FIFO_MAX_ENTRIES / 2;
+		break;
+	default:
+		max_samples = ADXL37X_FIFO_MAX_SAMPLES;
+		break;
+	}
+	adxl37x->fifo_config.extracted_samples = max_samples - 1;
```
rbolboac, line 969: I don't see this value being used anywhere

drivers/accel/adxl37x/adxl37x.c
```diff
@@ -0,0 +1,1205 @@
...
+			   uint16_t fifo_samples,
+			   enum adxl37x_op_mode op_mode)
+{
+	uint8_t fifo_config;
+	int ret;
+
+	if (!dev)
+		return -EINVAL;
+
+	if (fifo_samples > ADXL37X_FIFO_MAX_ENTRIES)
```
rbolboac, line 660: fifo samples can be 512 and this value is used as is for watermak which is only 9-bit

drivers/accel/adxl37x/adxl37x.h
```diff
@@ -0,0 +1,496 @@
...
+		.bw = ADXL371_BW_320HZ,
+		.filter_settle = ADXL371_FILTER_SETTLE_462p5
+	},
+	[ADXL372] = {
+		.wur = ADXL372_WUR_104ms,
+		.odr = ADXL372_ODR_800HZ,
+		.bw = ADXL372_BW_200HZ,
+		.filter_settle = ADXL372_FILTER_SETTLE_16
+	}
+};
```
rbolboac, line 318: move definition to .c

projects/eval-adxl37x/src/examples/basic_example/basic_example.c
```diff
@@ -0,0 +1,114 @@
...
+	ret = adxl37x_read_reg(dev, ADXL37X_OFFSET_Y, &off_y);
+	if (ret)
+		goto error_dev;
+	ret = adxl37x_read_reg(dev, ADXL37X_OFFSET_Z, &off_z);
+	if (ret)
+		goto error_dev;
+
+
+
+	pr_info("offset trims expressed in two's complement: OFFSET_X=%d , OFFSET_Y=%d , OFFSET_Z=%d \n\r",
```
rbolboac, line 83: perform the conversion yourself and print accordingly

drivers/accel/adxl37x/iio_adxl37x.c
```diff
@@ -0,0 +1,1105 @@
...
+		reg = ADXL37X_Y_DATA_H;
+		break;
+	case ACCEL_AXIS_Z:
+		reg = ADXL37X_Z_DATA_H;
+		break;
+	default:
+		return -EINVAL;
+	}
+
+	ret = adxl37x->reg_read_multiple(adxl37x, reg, raw_buf,
```
rbolboac, line 254: use adxl37x_read_reg_multiple instead

projects/eval-adxl37x/CMakeLists.txt
```diff
@@ -0,0 +1,53 @@
...
+
+target_link_libraries(eval-adxl37x no-os)
+
+# The basic example keeps recorded_data[MAX_SAMPLES] on the stack (3072 bytes at
+# AXES_USED 1) and stdout is unbuffered, so newlib routes printf through
+# __sbprintf, which puts a further BUFSIZ scratch buffer on the stack. Together
+# with _vfprintf_r that overruns the 4 kB default from startup_max${TARGET_NUM}.S
+# and corrupts the string being printed. __STACK_SIZE is read by the
+# preprocessor in the startup assembly, which is a source of the no-os library,
+# so the define has to go on that target rather than on this executable.
```
rbolboac, line 44: is this up to date?

drivers/accel/adxl37x/adxl37x_spi.c
```diff
@@ -0,0 +1,99 @@
...
+
+	if (!dev)
+		return -EINVAL;
+
+	if (count > ADXL37X_SPI_MAX_XFER_BYTES)
+		return -EINVAL;
+
+	buf[0] = ADXL37X_REG_READ(reg_addr);
+
+	ret = no_os_spi_write_and_read(dev->spi_desc, buf, count + 1);
```
rbolboac, line 92: consider using no_os_spi_transfer and see that  buf is no longer needed

drivers/accel/adxl37x/adxl37x_i2c.c
```diff
@@ -0,0 +1,106 @@
...
+		return -EINVAL;
+
+	buf[0] = reg_addr;
+
+	ret = no_os_i2c_write(dev->i2c_desc, buf, 1, 0);
```
rbolboac, line 95: buf is not needed here

rbolboac (commented):
some new remarks from my side

drivers/accel/adxl37x/iio_adxl37x.h
```diff
@@ -0,0 +1,36 @@
...
+
+#define ADXL37X_FIFO_MAX_SAMPLES   170
+#define ADXL37X_MAX_BUFFER_SAMPLES 400
+
+
+struct adxl37x_iio_dev {
+	struct adxl37x_dev *adxl37x_dev;
+	struct iio_device *iio_dev;
+	uint32_t active_channels;
+	uint8_t no_of_active_channels;
```
RaduSabau1, line 24: nit: doesn't look like this is read anywhere, just written in update_channels(). looks like a leftover, here read_samples() uses the bitmask directly instead. remove if not needed

PR #3116, opened 2026-06-01
platform: stm32: fix SPI DMA timeout underflow

### LINKED ISSUE #3115: Bug in stm32 SPI DMA driver: Integer underflow in timeout masks DMA fault
### Description
There is a logic error in the STM32 platform SPI driver (`stm32_spi_transfer_dma`) regarding the timeout handling. The use of a post-decrement operator in the while loop condition leads to an unsigned integer underflow. 

As a result, if a DMA transfer hangs or times out, the function silently returns 0 (Success) instead of `-ETIME`. Furthermore, it introduces a race condition where a last-millisecond success is wrongly reported as a timeout.

### Driver Information
* **File:** `drivers/platform/stm32/stm32_spi.c` (or similar depending on the exact no-OS branch)
* **Function:** `stm32_spi_transfer_dma()`

### Detailed Explanation
Here is the current implementation snippet:
```c
	timeout = msgs->bytes_number;
	while (timeout--) {
		no_os_mdelay(1);
		if (sdesc->stm32_spi_dma_done)
			break;
	};

	/* need some cleanup here? */
	if (timeout == 0)
		return -ETIME;

	return 0;
```

### There are two critical issues with this logic:

- **The Underflow (False Success)**: If the DMA transfer never completes, the loop evaluates timeout at 0, breaks the loop, but still executes the post-decrement (--). Since timeout is a uint32_t, it underflows to `0xFFFFFFFF`. The subsequent check `if (timeout == 0)` evaluates to false, and the function incorrectly returns 0 (Success). The DMA hang is completely masked.

- **The Race Condition (False Timeout):** If timeout evaluates to 1, is decremented to 0, and in that exact delay cycle the DMA finishes (stm32_spi_dma_done becomes true), the loop breaks. The variable timeout is now exactly 0. The code checks `if (timeout == 0)` and incorrectly returns -ETIME, even though the transfer was successful.

### Proposed Fix
The loop should avoid the post-decrement in the condition and the final validation should check the actual DMA completion flag, not the timeout counter.

```c
int32_t stm32_spi_transfer_dma(struct no_os_spi_desc* desc,
			       struct no_os_spi_msg* msgs,
			       uint32_t len)
{
	uint32_t timeout;
	struct stm32_spi_desc* sdesc = desc->extra;

	sdesc->stm32_spi_dma_done = false;
	stm32_config_dma_and_start(desc, msgs, len, stm32_spi_dma_callback, desc);
	
	timeout = msgs->bytes_number;
	
	/* Fix: Safe loop condition */
	while (timeout > 0) {
		no_os_mdelay(1);
		if (sdesc->stm32_spi_dma_done) {
			break;
		}
		timeout--;
	}

	/* Fix: Check the actual completion flag instead of the counter */
	if (!sdesc->stm32_spi_dma_done) {
		/* Optional: Add HAL_DMA_Abort_IT() or similar cleanup here */
		return -ETIME;
	}

	return 0;
}
```
### Steps to Reproduce
- Initiate an SPI DMA transfer using the STM32 no-OS wrapper.
- Force a DMA stall (e.g., by misconfiguring the DMA channels, or physically preventing the clock/data lines).
- Observe the return value of `stm32_spi_transfer_dma`. It will return 0 instead of the expected `-ETIME`.

morleman1:
This fixes an issue where an integer underflow in the timeout loop masks DMA failures and creates a race condition that reports false timeouts.

Resolves #3115

### Problem
The loop `while (timeout--)` in `stm32_spi_transfer_dma()` causes two issues:
1. **Integer Underflow:** If DMA stalls, `timeout` decrements past 0 to `0xFFFFFFFF`. The final `if (timeout == 0)` check fails, and the function incorrectly returns `0` (Success).
2. **Race Condition:** If DMA completes during the very last iteration, the loop breaks but `timeout` is already `0`, causing a false `-ETIME` return.

### Solution
- Changed loop condition to `while (timeout > 0)` to prevent underflow.
- Verified completion using the actual hardware flag `!sdesc->stm32_spi_dma_done` instead of the counter.

### Testing
Verified via hardware debugging on STM32 with an ADIN1110 by simulating a DMA stall. The driver now correctly catches the stall and returns `-ETIME`.

## PR Type

## PR Checklist

### DIFF
--- a/drivers/platform/stm32/stm32_spi.c
+++ b/drivers/platform/stm32/stm32_spi.c
@@ -672,16 +672,18 @@ int32_t stm32_spi_transfer_dma(struct no_os_spi_desc* desc,
 	sdesc->stm32_spi_dma_done = false;
 	stm32_config_dma_and_start(desc, msgs, len, stm32_spi_dma_callback, desc);
 	timeout = msgs->bytes_number;
-	while (timeout--) {
+	while (timeout > 0) {
 		no_os_mdelay(1);
 		if (sdesc->stm32_spi_dma_done)
 			break;
+		timeout--;
 	};
 
-	/* need some cleanup here? */
-	if (timeout == 0)
+  /* Fix: Check the actual completion flag instead of the counter */
+	if (!sdesc->stm32_spi_dma_done) {
+    /* Optional: Add HAL_DMA_Abort_IT() or similar cleanup here */
 		return -ETIME;
-
+	}
 	return 0;
 }
 
--- a/drivers/platform/stm32/stm32_spi.c
+++ b/drivers/platform/stm32/stm32_spi.c
@@ -679,9 +679,8 @@ int32_t stm32_spi_transfer_dma(struct no_os_spi_desc* desc,
 		timeout--;
 	};
 
-  /* Fix: Check the actual completion flag instead of the counter */
 	if (!sdesc->stm32_spi_dma_done) {
-    /* Optional: Add HAL_DMA_Abort_IT() or similar cleanup here */
+    /* need some cleanup here? */
 		return -ETIME;
 	}
 	return 0;
--- a/drivers/platform/stm32/stm32_xspi.c
+++ b/drivers/platform/stm32/stm32_xspi.c
@@ -693,14 +693,14 @@ static int32_t stm32_xspi_transfer_dma(struct no_os_spi_desc *desc,
 		return ret;
 
 	timeout = msgs->bytes_number;
-	while (timeout--) {
+	while (timeout > 0) {
 		no_os_mdelay(1);
 		if (sdesc->stm32_xspi_dma_done)
 			break;
+		timeout--;
 	};
 
-	/* need some cleanup here? */
-	if (timeout == 0) {
+	if (!sdesc->stm32_xspi_dma_done) {
 		no_os_dma_xfer_abort(sdesc->dma_desc, sdesc->dma_ch);
 		return -ETIME;
 	}


amiclaus (dismissed):

buha:
Thanks for catching it and submitting the fix. While you are at it, and because they seem identical, could you fix the same issue in xspi implementation ? You can put the xspi change in the same commit as they are tightly related. 
https://github.com/analogdevicesinc/no-OS/blob/main/drivers/platform/stm32/stm32_xspi.c#L696

drivers/platform/stm32/stm32_spi.c
```diff
@@ -672,16 +672,18 @@ int32_t stm32_spi_transfer_dma(struct no_os_spi_desc* desc,
...
+	while (timeout > 0) {
 		no_os_mdelay(1);
 		if (sdesc->stm32_spi_dma_done)
 			break;
+		timeout--;
 	};
 
-	/* need some cleanup here? */
-	if (timeout == 0)
+  /* Fix: Check the actual completion flag instead of the counter */
```
buha: remove either the "Fix:" or remove the whole comment for it to make sense when looking at raw code

drivers/platform/stm32/stm32_spi.c
```diff
@@ -672,16 +672,18 @@ int32_t stm32_spi_transfer_dma(struct no_os_spi_desc* desc,
...
+	if (!sdesc->stm32_spi_dma_done) {
+    /* Optional: Add HAL_DMA_Abort_IT() or similar cleanup here */
```
buha: remove this comment as well

morleman1: i reverted to the original comment `/* need some cleanup here? */`, as a cleanup still should be implemented similar to the xspi implementation. here it is more complex, because of seperate handles for rx and txdma_ch

buha: You are right, just squash the commits for now into a single commit and leave the original question there. We can deal with that separately in another PR and merge this one.

amiclaus (approved):

rbolboac:
squashed the commits here: https://github.com/analogdevicesinc/no-OS/pull/3131

rbolboac:
changes merged with https://github.com/analogdevicesinc/no-OS/pull/3131, closing this one

PR #3119, opened 2026-06-03
AD4630-20 support

machschmitt:
## Pull Request Description

Extend AD463x device driver and project to also support AD4630-20 and similar 20-bit precision ADCs.

## PR Type

## PR Checklist

### DIFF
--- a/drivers/adc/ad463x/ad463x.c
+++ b/drivers/adc/ad463x/ad463x.c
@@ -46,6 +46,24 @@
 
 #define ADAQ4224_GAIN_MAX_NANO 6670000000ULL
 
+/**
+ * Lookup table for amount of precison bits on default output mode.
+ */
+static const uint8_t ad463x_chip_precision[] = {
+	[ID_AD4630_24] = 24,
+	[ID_AD4630_20] = 20,
+	[ID_AD4630_16] = 16,
+	[ID_AD4631_24] = 24,
+	[ID_AD4631_20] = 20,
+	[ID_AD4631_16] = 16,
+	[ID_AD4632_24] = 24,
+	[ID_AD4632_20] = 20,
+	[ID_AD4632_16] = 16,
+	[ID_AD4030] = 24,
+	[ID_ADAQ4216] = 16,
+	[ID_ADAQ4224] = 24,
+};
+
 /*
  * Gains computed as fractions of 1000 so they can be expressed by integers.
  */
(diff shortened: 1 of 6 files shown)


buha (dismissed):

drivers/adc/ad463x/ad463x.c
```diff
@@ -760,9 +760,17 @@ int32_t ad463x_init(struct ad463x_dev **device,
 	switch (dev->output_mode) {
 	case AD463X_24_DIFF:
 		switch (dev->device_id) {
+		case ID_AD4630_16:
+		case ID_AD4631_16:
+		case ID_AD4632_16:
```
rbolboac: I would change this in a different commit

rbolboac (commented):
just a nit from my side

drivers/adc/ad463x/ad463x.c
machschmitt: okay, will update accordingly.

drivers/adc/ad463x/iio_ad463x.c
```diff
@@ -52,7 +52,11 @@ static int32_t _iio_ad463x_read_dev(struct iio_ad463x *desc, uint32_t *buff,
 static int32_t _iio_ad463x_prepare_transfer(struct iio_ad463x *desc,
 		uint32_t mask);
 #define BITS_PER_SAMPLE 32
+#if AD4630_20_DEV
+#define REAL_BITS 20
+#else
```
CiprianRegus: Maybe set this at runtime based on the chip type. As of currently, I don't think you can have both devices in the same build.

machschmitt:
Change log v1 -> v2:
- Simplified ADC precision initialization logic.
- Now setting IIO data scan properties at runtime according to chip/design specs.

drivers/adc/ad463x/ad463x.c
```diff
@@ -46,6 +46,25 @@
 
 #define ADAQ4224_GAIN_MAX_NANO 6670000000ULL
 
+/**
+ * @enum ad463x_chip_precision
```
amiclaus: doesn't look like an enum type

amiclaus (commented):
one nit otherwise lgtm.

drivers/adc/ad463x/ad463x.c
machschmitt: Good catch. Updated the comment to clarify that's a lookup table for amount of precision bits.
Thanks.

RaduSabau1 (approved):
I think the Xilinx CI error is unrelated, otherwise LGTM!

PR #3151, opened 2026-06-22
Staging/capi xilinx

Voc94:
## Pull Request Description

Adds the Xilinx platform implementation of the Common API (CAPI) under `drivers/platform/xilinx/`, providing board-agnostic backends for the core peripheral classes on PS-controlled ARM targets and MicroBlaze designs.

The drivers are SDT-only and use base-address `LookupConfig`. Interrupts route through the IRQ controller singleton via `capi_irq_connect()`, with a soft fallback to polled operation when no IRQ is wired in the XSA.

Each backend has its own `.c` file and exports its own operations table. The configuration structures do not use a type enum.

## Drivers Added

- **GPIO**
  - `XGpioPs` (PS): global pin IDs, with an EMIO offset of 54 on Zynq-7000 and 78 on ZynqMP. Bank boundaries are validated during initialization.
  - `XGpio` (PL): per-channel, 32-bit operation.
  - Both drivers account for the inverted direction polarity relative to the CAPI convention.

- **UART**
  - `XUartPs` (PS)
  - `XUartNs550` (AXI 16550)
  - `XUartLite` (AXI, fixed 8N1)

- **SPI**
  - `XSpiPs` (PS): master mode, with native CS0/1/2 represented as one-hot masks.
  - `XSpi` (AXI Quad SPI): master mode.

- **I2C**
  - `XIicPs` (PS): 7-bit and 10-bit addressing with repeated-start support.
  - `XIic` (AXI IIC).

- **Timer**
  - `XTtcPs`: up-counting.
  - `XScuTimer`: 32-bit down-counting, A9-private.
  - `XTmrCtr`: AXI timer with two channels.

- **IRQ**
  - `XScuGic` (GIC).
  - `XIntc` (AXI INTC), either as a root controller or cascaded to a GIC SPI line.
  - IRQ IDs are bit-packed: controller in the upper half, input in the lower half.

- **Time / Platform**
  - Strong `capi_time` hooks:
    - Uptime via the A9 global timer.
    - Delays via `usleep()` / `usleep_MB()`.
  - `CAPI_PLATFORM_IO_*` macros implemented using `Xil_In*` / `Xil_Out*`.

## What Was Tested

Validated on:

- ZedBoard (Zynq-7000), across GIC, INTC, and no-IRQ XSA variants.
- ZCU102 (ZynqMP).
- Cora Z7 (MicroBlaze).

Testing used a structured hardware test environment and covered:

- Initialization verified through a real follow-up operation.
- Deinitialization verified by reading hardware registers back, including confirmation that interrupts were disabled.
- Functional and loopback paths:
  - SPI internal loopback (`rx == tx`).
  - UART loopback.
  - Real timer counter movement in both up-counting and down-counting modes.
  - IRQ delivery through GIC, INTC cascade, INTC root, and no-IRQ `-ENOTSUP`.
- Asynchronous and interrupt-driven paths:
  - PS and PL SPI.
  - UART asynchronous TX completion.
  - No-IRQ polled fallback.
- Applied-configuration round trips:
  - Bus speed.
  - Line configuration.
  - Initialization clock frequency.
  - Channel configuration matching.
  - Runtime UART settings: baud rate, data bits, parity, and stop bits.
- Contract and error paths:
  - Per-backend contracts.
  - Null configuration and null handle.
  - Null and zero-length TX/RX buffers.
  - Reinitialization stress.
- Real-device smoke tests where hardware was wired, including the ADXL355 over I2C.

## Known Limitations

The following features are not hardware-dependent and are explicitly asserted to return `-ENOTSUP`:

- **DMA**
  - DMA is not supported yet.
  - A non-null `dma_handle` is rejected during initialization.

- **I2C**
  - Duty-cycle configuration.
  - Clock stretching.
  - 10-bit addressing on PL.

- **SPI PS**
  - LSB-first operation.
  - Per-transfer counter fields.

- **UART**
  - RTS and RTS/CTS flow control.
  - 9-bit and multidrop addressing.
  - Sticky, mark, and space parity.
  - `UartLite` is fixed at 8N1, with baud rate, word length, parity, and stop bits configured at synthesis time. Non-default line configurations are rejected.

- **Timer**
  - Non-rollover configuration.
  - Non-zero minimum values.

- **IRQ**
  - Priority and trigger configuration apply only to GIC inputs; INTC inputs reject these settings.

## PR Type

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

## PR Checklist

- [x] I have built all projects affected by the changes in this PR
- [x] I have tested in hardware affected projects, at the relevant boards:
  - ZedBoard (Zynq-7000: GIC, INTC, and no-IRQ)
  - ZCU102 (ZynqMP)
  - Cora Z7 (MicroBlaze)
- [ ] I have updated the documentation (wiki pages, ReadMe etc), if applicable (N/A, driver-only change)

### DIFF
--- a/capi/platform/xilinx/xilinx_capi_gpio_pl.c
+++ b/capi/platform/xilinx/xilinx_capi_gpio_pl.c
@@ -1,0 +1,697 @@
+/*
+ * Copyright (c) 2026 Analog Devices, Inc.
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+/**
+ * @file
+ * @brief Xilinx platform GPIO driver implementation
+ *
+ * Supports XGpio (PL GPIO).
+ */
+#include "capi_gpio.h"
+#include <xilinx_capi_gpio_priv.h>
+#include <capi_alloc.h>
+#include <string.h>
+#include <errno.h>
+#include <stdint.h>
+
+#ifdef XPAR_XGPIO_NUM_INSTANCES
+/* AXI GPIO exposes one 32-bit data/tristate register per channel. */
+#define CAPI_GPIO_PL_MAX_PINS	32U
+
+/* Function prototypes */
+static int capi_gpio_xilinx_port_init(struct capi_gpio_port_handle **handle,
+				      const struct capi_gpio_port_config *config);
+static int capi_gpio_xilinx_port_deinit(struct capi_gpio_port_handle **handle);
+static int capi_gpio_xilinx_port_set_direction(struct capi_gpio_port_handle
+		*handle,
+		uint64_t direction_bitmask);
+static int capi_gpio_xilinx_port_get_direction(struct capi_gpio_port_handle
+		*handle,
+		uint64_t *direction_bitmask);
+static int capi_gpio_xilinx_port_set_value(struct capi_gpio_port_handle *handle,
+		uint64_t value_bitmask);
+static int capi_gpio_xilinx_port_get_value(struct capi_gpio_port_handle *handle,
+		uint64_t *value_bitmask);
+static int capi_gpio_xilinx_port_set_raw_value(struct capi_gpio_port_handle
+		*handle,
+		uint64_t value_bitmask);
+static int capi_gpio_xilinx_port_get_raw_value(struct capi_gpio_port_handle
+		*handle,
+		uint64_t *value_bitmask);
+static int capi_gpio_xilinx_pin_set_direction(struct capi_gpio_pin *pin,
+		uint8_t direction);
+static int capi_gpio_xilinx_pin_get_direction(struct capi_gpio_pin *pin,
+		uint8_t *direction);
+static int capi_gpio_xilinx_pin_set_value(struct capi_gpio_pin *pin,
+		uint8_t value);
+static int capi_gpio_xilinx_pin_get_value(struct capi_gpio_pin *pin,
+		uint8_t *value);
+static int capi_gpio_xilinx_pin_set_raw_value(struct capi_gpio_pin *pin,
+		uint8_t value);
+static int capi_gpio_xilinx_pin_get_raw_value(struct capi_gpio_pin *pin,
+		uint8_t *value);
+static int capi_gpio_xilinx_port_toggle(struct capi_gpio_port_handle *handle,
+					uint64_t pins_bitmask);
+static int capi_gpio_xilinx_pin_toggle(struct capi_gpio_pin *pin);
+
+/* GPIO operations structure. */
+const struct capi_gpio_ops capi_gpio_xilinx_pl_ops = {
+	.port_init = capi_gpio_xilinx_port_init,
+	.port_deinit = capi_gpio_xilinx_port_deinit,
+	.port_set_direction = capi_gpio_xilinx_port_set_direction,
+	.port_get_direction = capi_gpio_xilinx_port_get_direction,
+	.port_set_value = capi_gpio_xilinx_port_set_value,
+	.port_get_value = capi_gpio_xilinx_port_get_value,
+	.port_set_raw_value = capi_gpio_xilinx_port_set_raw_value,
+	.port_get_raw_value = capi_gpio_xilinx_port_get_raw_value,
+	.port_toggle = capi_gpio_xilinx_port_toggle,
+	.pin_set_direction = capi_gpio_xilinx_pin_set_direction,
+	.pin_get_direction = capi_gpio_xilinx_pin_get_direction,
+	.pin_set_value = capi_gpio_xilinx_pin_set_value,
+	.pin_get_value = capi_gpio_xilinx_pin_get_value,
+	.pin_set_raw_value = capi_gpio_xilinx_pin_set_raw_value,
+	.pin_get_raw_value = capi_gpio_xilinx_pin_get_raw_value,
+	.pin_toggle = capi_gpio_xilinx_pin_toggle
+};
+
+static uint64_t logical_to_raw(uint64_t logical_value, uint64_t active_low_mask)
+{
+	return logical_value ^ active_low_mask;
+}
+
+static uint64_t raw_to_logical(uint64_t raw_value, uint64_t active_low_mask)
+{
+	return raw_value ^ active_low_mask;
+}
+
+static uint32_t capi_gpio_get_max_pins(void)
+{
+	return CAPI_GPIO_PL_MAX_PINS;
+}
+
+static void xilinx_gpio_free_allocated_handle(
+	struct capi_gpio_port_handle **handle)
+{
+	if (handle == NULL || *handle == NULL)
+		return;
+
+	capi_free((*handle)->priv);
+	capi_free(*handle);
+	*handle = NULL;
+}
+
+static void xilinx_gpio_clear_app_handle(struct capi_gpio_port_handle *handle)
+{
+	if (handle != NULL)
+		handle->ops = NULL;
+}
+
+/* Dispatcher already checks pin != NULL and port_handle != NULL before calling us */
+static int gpio_validate_pin(const struct capi_gpio_xilinx_port_handle *xhandle,
+			     const struct capi_gpio_pin *pin)
+{
+	if (pin->number >= xhandle->num_pins)
+		return -EINVAL;
+
+	return 0;
+}
+
+static uint32_t gpio_pin_flags(const struct capi_gpio_xilinx_port_handle
+			       *xhandle,
+			       const struct capi_gpio_pin *pin)
+{
+	uint32_t flags = pin->flags;
+
+	if (xhandle->pin_flags != NULL)
+		flags |= xhandle->pin_flags[pin->number];
+
+	return flags;
+}
+
+/**
+ * @brief Initialize a GPIO port.
+ * @note PL: XGpio_Initialize().
+ *
+ * @param[out] handle GPIO port handle pointer.
+ * @param[in] config GPIO port configuration.
+ * @return 0 on success, -EINVAL, -ENODEV, -ENOMEM, or -EIO.
+ */
+static int capi_gpio_xilinx_port_init(struct capi_gpio_port_handle **handle,
+				      const struct capi_gpio_port_config *config)
+{
+	if (handle == NULL || config == NULL) {
+		return -EINVAL;
+	}
+	struct capi_gpio_xilinx_pl_config *xcfg =
+		(struct capi_gpio_xilinx_pl_config *)config->extra;
+	if (xcfg == NULL)
+		return -EINVAL;
+
+	if (xcfg->channel == 0 || xcfg->channel > 2)
+		return -EINVAL;
+
+	if (config->num_pins == 0 ||
+	    config->num_pins > capi_gpio_get_max_pins())
+		return -EINVAL;
+
+	bool alloc_handle = (*handle == NULL);
+	struct capi_gpio_port_handle *h = *handle;
+	struct capi_gpio_xilinx_port_handle *xhandle;
+	XGpio *pl_inst = NULL;
+	int ret;
+
+	if (alloc_handle) {
+		h = capi_malloc(sizeof(*h));
+		if (h == NULL)
+			return -ENOMEM;
+		memset(h, 0, sizeof(*h));
+
+		xhandle = capi_malloc(sizeof(*xhandle));
+		if (xhandle == NULL) {
+			capi_free(h);
+			return -ENOMEM;
+		}
+		h->priv = xhandle;
+		*handle = h;
+	} else {
+		xhandle = h->priv;
+		if (xhandle == NULL)
+			return -EINVAL;
+		if (h->ops != NULL || xhandle->instance != NULL ||
+		    xhandle->pin_flags != NULL)
+			return -EBUSY;
+	}
+
+	memset(xhandle, 0, sizeof(*xhandle));
+	h->init_allocated = alloc_handle;
+	h->ops = config->ops;
+	h->priv = xhandle;
+	xhandle->channel = xcfg->channel;
+	xhandle->num_pins = config->num_pins;
+
+	/* Allocate and copy pin flags */
+	xhandle->pin_flags = capi_malloc(config->num_pins * sizeof(uint32_t));
+	if (xhandle->pin_flags == NULL) {
+		ret = -ENOMEM;
+		goto err_handle;
+	}
+	if (config->flags != NULL)
+		memcpy(xhandle->pin_flags, config->flags,
+		       config->num_pins * sizeof(uint32_t));
+	else
+		memset(xhandle->pin_flags, 0, config->num_pins * sizeof(uint32_t));
+
+	/* Compute active-low mask once at init (eliminates iteration per value call) */
+	xhandle->active_low_mask = 0;
+	for (int i = 0; i < config->num_pins; i++) {
+		if (xhandle->pin_flags[i] & CAPI_GPIO_ACTIVE_LOW)
+			xhandle->active_low_mask |= (1ULL << i);
+	}
+
+	pl_inst = capi_malloc(sizeof(XGpio));
+	if (pl_inst == NULL) {
+		ret = -ENOMEM;
+		goto err_flags;
+	}
+	memset(pl_inst, 0, sizeof(XGpio));
+	if (XGpio_Initialize(pl_inst, (UINTPTR)config->identifier) != XST_SUCCESS) {
+		ret = -EIO;
+		goto err_inst;
+	}
+	xhandle->instance = pl_inst;
+
+	/*
+	 * Reject channel 2 on a single-channel core. XGpio_* assert
+	 * (Channel == 2) && (IsDual == TRUE), so without this guard the first
+	 * op on a non-dual core traps inside embeddedsw.
+	 */
+	if (xhandle->channel == 2 && pl_inst->IsDual == 0) {
+		ret = -ENOTSUP;
+		goto err_inst;
+	}
+
+	/*
+	 * Seed the direction cache from the actual TRI register instead of
+	 * leaving it zeroed. A per-pin direction change rewrites the whole
+	 * channel TRI register from this cache, so a zero seed would silently
+	 * turn every other pin into an output. XGpio TRI polarity (1=input,
+	 * 0=output) matches the CAPI cache convention, so copy it directly.
+	 */
+	xhandle->direction = XGpio_GetDataDirection(pl_inst, xhandle->channel);
+
+	return 0;
+
+err_inst:
+	capi_free(pl_inst);
+	xhandle->instance = NULL;
+err_flags:
+	capi_free(xhandle->pin_flags);
+	xhandle->pin_flags = NULL;
+err_handle:
+	if (alloc_handle)
+		xilinx_gpio_free_allocated_handle(handle);
+	else
+		xilinx_gpio_clear_app_handle(h);
+	return ret;
+}
+
+/**
+ * @brief Deinitialize a GPIO port.
+ * @note PL: Frees XGpio instance.
+ *
+ * @param[in,out] handle GPIO port handle pointer, set to NULL on success.
+ * @return 0 on success, -EINVAL if handle is NULL.
+ */
+static int capi_gpio_xilinx_port_deinit(struct capi_gpio_port_handle **handle)
+{
+	if (handle == NULL || *handle == NULL) {
+		return -EINVAL;
+	}
+	struct capi_gpio_port_handle *h = *handle;
+	struct capi_gpio_xilinx_port_handle *xhandle = h->priv;
+	if (xhandle == NULL)
+		return -EINVAL;
+
+	/* Free the Xilinx driver instance */
+	if (xhandle->instance != NULL) {
+		capi_free(xhandle->instance);
+		xhandle->instance = NULL;
+	}
+	/* Free dynamically allocated pin flags */
+	if (xhandle->pin_flags != NULL) {
+		capi_free(xhandle->pin_flags);
+		xhandle->pin_flags = NULL;
+	}
+	if (h->init_allocated) {
+		capi_free(xhandle);
+		capi_free(h);
+		*handle = NULL;
+	} else {
+		h->ops = NULL;
+	}
+	return 0;
+}
+
+/**
+ * @brief Set GPIO port direction.
+ * @note PL: XGpio_SetDataDirection().
+ *
+ * XGpio_SetDataDirection uses inverted polarity (1=output, 0=input) vs CAPI (1=input, 0=output).
+ *
+ * @param[in] handle GPIO port handle.
+ * @param[in] direction_bitmask 1=input, 0=output per pin.
+ * @return 0 on success, -EINVAL.
+ */
+int capi_gpio_xilinx_port_set_direction(struct capi_gpio_port_handle *handle,
+					uint64_t direction_bitmask)
+{
+	if (handle == NULL)
+		return -EINVAL;
+
+	struct capi_gpio_xilinx_port_handle *xhandle =
+			handle->priv;
+
+	uint64_t mask = (1ULL << xhandle->num_pins) - 1;
+	xhandle->direction = (xhandle->direction & ~mask) |
+			     (direction_bitmask & mask);
+
+	XGpio *inst = (XGpio *)xhandle->instance;
+	XGpio_SetDataDirection(inst, xhandle->channel, (uint32_t)xhandle->direction);
+	return 0;
+}
+
+/**
+ * @brief Get GPIO port direction (cached).
+ * @note PL: Returns cached direction.
+ *
+ * @param[in] handle GPIO port handle.
+ * @param[out] direction_bitmask 1=input, 0=output per pin.
+ * @return 0 on success, -EINVAL if NULL.
+ */
+static int capi_gpio_xilinx_port_get_direction(struct capi_gpio_port_handle
+		*handle,
+		uint64_t *direction_bitmask)
+{
+	if (handle == NULL || direction_bitmask == NULL) {
+		return -EINVAL;
+	}
+	struct capi_gpio_xilinx_port_handle *xhandle =
+			handle->priv;
+
+	uint64_t mask = (1ULL << xhandle->num_pins) - 1;
+	*direction_bitmask = xhandle->direction & mask;
+	return 0;
+}
+/**
+ * @brief Set GPIO port raw output value (no ACTIVE_LOW inversion).
+ *
+ * @note PL: XGpio_DiscreteWrite().
+ *
+ * @param[in] handle GPIO port handle.
+ * @param[in] value_bitmask Raw value per pin.
+ * @return 0 on success, -EINVAL, or -ENOTSUP.
+ */
+static int capi_gpio_xilinx_port_set_raw_value(struct capi_gpio_port_handle
+		*handle,
+		uint64_t value_bitmask)
+{
+	if (handle == NULL) {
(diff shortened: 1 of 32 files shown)


buha:
1. git rebase -i HEAD~10
2. mark every commit of yours with edit or e
3. ci/astyle.sh
4. if it produces changes, git add -u, git commit --amend 
5. git rebase --continue, repeat from step 3

Voc94:
## Changelog v1

- IRQ
  - INTC can now act as the root interrupt controller, enabling MicroBlaze support.
  - The IRQ backend is selected at compile time; linking multiple root backends now fails intentionally.
  - Replaced the old scheme (`irq < 32` = INTC, `irq >= 32` = GIC) with bit-packed IDs: the controller in the upper half, the input in the lower half.
  - Added helper macros for building and decoding IRQ IDs, so callers no longer do the bit math by hand.

- I2C
  - Fixed the async busy-wait loop that could spin badly.
  - AXI IIC can now set bus speed through optional timing registers instead of always returning `-ENOTSUP`.
  - Xilinx status codes now map to real errnos.
  - Controllers now reset cleanly after failed transfers.

- SPI
  - Fixed the async `read_command` path so it uses the controller correctly.

- SPI and I2C
  - Added comments explaining the async flow.
  - Documented the embeddedsw quirk where registers must be cleared manually.

- Handle and ISR plumbing
  - `capi_*_init()` now populates `handle->priv` consistently instead of relying on first-member casts.
  - `capi_*_isr()` now recovers driver-private state through `handle->priv` before touching Xilinx instances.
  - App-owned and driver-allocated handles now use the same private-state access pattern.

- Housekeeping
  - Split each header into a public config header and a private `*_priv.h` for internals.
  - Time hooks are SDT-only now; the legacy `xtime_l.h` path was removed.
  - Removed UARTPSV from the PR description until it is actually tested.
  - Added CoraZ7 to the hardware test list.
  - Rebased onto the CAPI core from PR #3090.
  - Ran `astyle` on every commit, all clean.

Voc94:
## Changelog v2
- moved capi drivers from drivers/platform/xilinx to capi/platform/xilinx 
- changed commit message titles

Voc94:
Changelog v3

1. I2C PL: Fix sync 10-bit transfers using IRQ-backed XIic master APIs, fix callback cleanup race, fix target-mode polling, and honor transfer->target_addr.
2. I2C PS: Use transfer->target_addr as a per-transfer target address when provided, otherwise fall back to the device’s default device->address.

3. Timers: Add TTC/SCU output_freq_hz prescaler handling, clear stale IRQ status before enabling interrupts, fix SCU max-1 reload behavior, and document overflow-safe tick/ns math.
4. Timer PL: Clear any stale AXI Timer interrupt flags before enabling IRQs, and use the BSP-reported fabric clock when available, falling back to the caller-provided input_clock_hz otherwise.

5. UART PS/NS550: Validate line configs strictly, propagate init set_line_config failures, and fix init cleanup paths.
6. UART PL Lite: Treat zero-length transfers as no-op success and remove unnecessary SDT status-bit guards.

7. Private headers: Formatting-only macro alignment.

Voc94:
## Changelog v4

### I2C
- PS sync TX no longer times out mid-`no_stop` chain by skipping the bus-wait when HOLD is already set.
- Reads with `sub_address` but no `repeated_start` now return `-ENOTSUP` instead of silently dropping the register pointer. Applies to PS and PL.
- Reject `dma_handle` with `-ENOTSUP`; Xilinx I2C has no DMA support.
- PL rejects unsupported `clk_stretch`.

### SPI
- Native CS is deselected after async completion, error, or abort; previously it could remain asserted.
- `CAPI_SPI_CS_AUTO` is accepted as a no-op instead of returning `-EINVAL`.
- PL rejects per-device `max_speed_hz` because the backend cannot honor it.

### 64-bit
- Change `LookupConfig` identifier casts from `u32` to `UINTPTR` for I2C PS and TTC timer to avoid truncating base addresses on ZynqMP/Versal.

### Build
- Add `xilinx_capi_alloc.c` with strong `capi_malloc/free/calloc/realloc` hooks using libc passthrough, matching STM32.

### Misc
- Rebased onto latest `main`.
- Expanded doc comments for async flow, read-command temp buffers, and target mode being init-time rather than runtime-switchable.

Voc94:
Changelog v5

- Moved the Xilinx backend ops declarations from the private `*_priv.h` headers into the normal public `*.h` headers. This means users can choose a backend with `config->ops` without including private driver headers.
- Marked all Xilinx ops tables as `const`, like STM32 does. These tables are not meant to change at runtime, so they can safely live in read-only memory.
- Kept the internal driver data, stack allocation helper macros, and Xilinx BSP includes such as `xgpiops.h` and `xspips.h` in the private headers. The public headers now stay simpler and do not expose Xilinx-specific driver internals.

Voc94:
### Changelog v6

#### I2C

This PR makes I2C target/slave registration runtime-switchable for both PS and PL controllers, removing the previous init-locked role restriction.

- `register_target()` now switches an initialized controller into target mode:
  - PS: `XIicPs_SetupSlave`
  - PL: `XIic_Stop()` -> `XIic_SetAddress()` -> `XIic_Start()`
- `unregister_target()` restores initiator/master mode.
- The ISR now dispatches based on the controller's current role, so a controller opened as an initiator can be switched to target mode and back.
- PL target registration is now functional and has been removed from Known Limitations.
- Added comments in both drivers documenting the runtime role switch and role-based ISR dispatch.

capi/platform/xilinx/xilinx_capi_gpio_ps.c
```diff
@@ -0,0 +1,627 @@
...
+		memcpy(xhandle->pin_flags, config->flags, config->num_pins * sizeof(uint32_t));
+	else
+		memset(xhandle->pin_flags, 0, config->num_pins * sizeof(uint32_t));
+
+	for (int i = 0; i < config->num_pins; i++) {
+		if (xhandle->pin_flags[i] & CAPI_GPIO_ACTIVE_LOW)
+			xhandle->active_low_mask |= (1ULL << i);
+	}
+
+	XGpioPs_Config *ps_cfg = XGpioPs_LookupConfig((UINTPTR)config->identifier);
```
buha, line 215: @cristina-suteu made this helper to distinguish between base address and id (dev0, dev1, dev2...) https://github.com/analogdevicesinc/no-OS/blob/main/capi/platform/stm32/stm32_capi_gpio.c#L38C22-L38C51

From what I could gather, even though the XGpioPs_LookupConfig function is not well documented in xilinx docs (confusion about argument being only base address), it takes either ID or Base and gives you a valid config that contains both. This is what we want. Can you confirm you tested with both and works ?

drivers/Makefile
```diff
@@ -24,6 +24,7 @@ INCLUDES = -I../include/ \
 	 -I./axi_core/clk_axi_clkgen \
 	 -I./axi_core/jesd204 \
 	 -I./platform/xilinx \
+	 -I../capi/inc \
```
rbolboac: was this added by mistake?

capi/platform/xilinx/xilinx_capi_timer_ps_scu.c
```diff
@@ -0,0 +1,872 @@
...
+	XScuTimer_EnableAutoReload(xi);
+
+	/*
+	 * Configure prescaler: SCU prescaler value N divides by N+1 (0..255).
+	 * If output_freq_hz is requested, compute the closest prescaler.
+	 */
+	uint32_t src_clk = (uint32_t)config->input_clock_hz;
+	if (config->output_freq_hz > 0) {
+		if (src_clk == 0) {
+			ret = -EINVAL;
```
rbolboac: does this build? ret does not look to be defined

capi/platform/xilinx/xilinx_capi_timer_ps_scu.c
```diff
@@ -0,0 +1,872 @@
...
+			goto err_inst;
```
rbolboac: err_inst does not exist as well

capi/platform/xilinx/xilinx_capi_i2c_pl.c
```diff
@@ -0,0 +1,1289 @@
...
+				      I2C_PL_DEFAULT_THIGH_OFFSET;
+		xh->pl_tlow_offset = xcfg->tlow_offset != 0 ?
+				     xcfg->tlow_offset :
+				     I2C_PL_DEFAULT_TLOW_OFFSET;
+		xh->pl_default_duty_percent = xcfg->default_duty_percent != 0 ?
+					      xcfg->default_duty_percent :
+					      I2C_PL_DEFAULT_DUTY_PERCENT;
+	}
+
+	XIic_Config *cfg = XIic_LookupConfig((UINTPTR)config->identifier);
```
buha, line 441: We've tested this on a project and it doesn't seem to work actually. Passing 0/1 id's for i2c0/i2c1 works, but while passing base addresses for i2c0 and i2c1, the actual bus that was used was i2c0, so this needs some investigation. I've also checked capi-selftest, there is no test for this scenario, it's worth adding.

Voc94: referencing the previous comment about the "helper to distinguish between base address and id" , 
I've taken a look at the testing(remote-bus) bsp , it is legacy, it has device_id and there's a clear split in xilinx's embeddedsw on this with the SDT defined constants, this gives us two options 

1. make a helper to convert deviceID to BaseAddress on SDT's to allow old apps to still work with the New BSP(but this would break Legacy BSP code like the remote-bus that was used in the review you mentioned) (0->0x40ff)
 or
2. ignore it as it is and let the dev choose the right one for the job
SDT Vitis BSPs (2023.2 and newer) uses only BaseAddress, there is no confusion, here's a snippet from the embeddedsw header of i2c PL:
```
  typedef struct {
  #ifndef SDT
        u16 DeviceId;     /**< Unique ID  of device */
  #else
        char *Name;
  #endif
        UINTPTR BaseAddress;  /**< Device base address */
        ...
        }
```
so DeviceId has no part in the LookupConfig functions unless a helper is designed to convert it, but this would invalidate any older BSP versions that still use it, my suggestion is to make a comment to inform this in the code and the developer should be concerned about this split

Voc94: p.s. :  no-OS uses Vitis 2025.1 (changed by Popa Stefan recently), so we are SDT only from now on, but other projects may not be unfortunately

Voc94:
## Changelog v7:

Review fixes from @rbolboac:
  - **timer_ps_scu.c**: fixed the init compile error where `ret` was undeclared and `err_inst` didn't exist. Now frees the instance and returns -EINVAL like the other
  error paths.
  - **drivers/Makefile**: dropped the stray `-I../capi/inc`, it was unused, stemmed from old location of the drivers.

  While in there, I also:
  - **timer_ps_scu.c**: added the missing `!xh` NULL checks so all ops are consistent.
  - **timer_ps_ttc.c**: added a comment noting `input_clock_hz` is overridden by the BSP config (no behavior change).

Voc94:
Changelog V8
- GPIO/UART API sync: GPIO and UART core-API additions (capi_uart, capi_gpio).
- Re-entrancy across SPI/UART/I2C-PL: the vendor ISR now only records the terminal event; cleanup and the user callback run at the ISR-wrapper tail. 
(tldr for the drivers) if a callback triggers the next async transmit, we start that transmission on trigger instead of returning -EBUSY. This was fixed by moving the busy-flag updates into our drivers, since embeddedsw updates isBusy(this is a simplification) too late. UART, SPI, and I2C now have async busy flags; UART has separate flags for TX and RX.
- I2C held-bus deadlock: held-bus no_stop continuations are routed through the ISR path, since polled WaitBusFree would deadlock. So far, the “synchronous transfers use only XIic_Send and also allow repeated start” approach does not seem possible: MasterSend is the only path that does not call XIic_WaitBusFree(). We might need to revisit this after a datasheet analysis of the I2C IP core. For now, I used Ruben’s implementation with a bit more guarding.(Only PL I2C Has this Issue)

Voc94:
## Changelog v9

- added ifdef `XPAR_*_NUM_INSTANCES` guard to all peripheral .c files, the reason : cmake `CONFIG_*_XILINX`requires only 1 peripheral name and xilinx is a special case where we have 2 or more types for each peripheral , so that if the bsp doesn't have it , it will be silently dropped and not cause linker issues.

Voc94:
## Changelog v10

  - Timer TTC (xilinx_capi_timer_ps_ttc.c): compare channel now uses the match register + match-0 interrupt (XTTCPS_IXR_MATCH_0_MASK) instead of the interval register/interrupt; overflow keeps the interval interrupt. Compare and overflow are now distinct hardware events and dispatched independently in the ISR.
  - Timer SCU (xilinx_capi_timer_ps_scu.c) & AXI PL (xilinx_capi_timer_pl.c): hardware has a single shared "expired" interrupt with no separate compare register, so arming compare and overflow together now returns -EBUSY (on channel 0 for PL) instead of silently firing both callbacks off one expiry.

Voc94:
## Changelog v11

  I2C target-mode fixes, found by adding target-transmit and master-async coverage to the loopback test.

  PS I2C
  - Clear the TX FIFO when arming a slave send. TransmitFifoFill() sizes each burst from TRANS_SIZE, but in slave-transmitter mode that register means "bytes left after the master
  terminates", not current occupancy. Leftover values made it under-fill anything 16 bytes or longer.
  - Wait for the bus to go idle before that clear. COMPLETE_SEND fires when the last byte enters the FIFO, not when it reaches the wire, so re-arming on the callback could wipe a
  transfer still in progress.
  - Stop treating TIME_OUT as an error while acting as target. It is a master-side watchdog and fires on our own clock stretching. Still an error when we are the initiator.
  - Added an ISR-side FIFO refill for transfers longer than the FIFO. It never blocks, since the initiator may be polled on the same CPU.

  PL I2C
  - init() was missing XIic_Start() for target mode, so a controller opened as a target never ACKed. Only reachable when PL is the target.
  - Replaced XIic_Recv() with a bounded version. Every wait in the BSP original is an unbounded while, which hangs the board when a polled initiator meets a slow target. Same
  register sequence, now returns -ETIMEDOUT or -EIO. Worth a review since it is hand-copied BSP logic.

  Both
  - recover_bus() now cancels in-flight work instead of returning -EBUSY. It was refusing at exactly the moment recovery was needed, which left the handle stuck for good. Reports
  -ECANCELED after the reset.
  - deinit() clears async state so a reused handle does not come back busy.

all components we're tested and passed on capi-selftest, i think this capi platform impl. is mostly done

capi/platform/xilinx/xilinx_capi_uart_ps.c
```diff
@@ -0,0 +1,925 @@
...
+	if (handle == NULL || buf == NULL)
+		return -EINVAL;
+	if (len == 0)
+		return 0;
+
+	struct capi_uart_xilinx_handle *xh = handle->priv;
+	if (ps_xfer_busy(xh))
+		return -EBUSY;
+
+	uint32_t sent = XUartPs_Send(inst(xh), buf, len);
```
rbolboac: I don't have much experience with Xilinx implementation, so I asked Claude to perform a review. maybe this is worth mentioning: 
_XUartPs_Send pushes only what fits in the ~64-byte hardware TX FIFO and relies on the interrupt handler to feed the rest; in this polled path the remainder is never sent. Any blocking capi_uart_transmit() larger than the FIFO (an ordinary log line) transmits only the first ~64 bytes. The NS550 and Lite backends both loop (while (total < len)) — PS is the outlier and should do the same. (Aside: XUartPs_Send returns the requested count, so the sent == len / -EIO branch is effectively dead — the bug is silent truncation, not a reported error.)_

capi/platform/xilinx/xilinx_capi_gpio_pl.c
```diff
@@ -0,0 +1,695 @@
...
+	XGpio *inst = (XGpio *)xhandle->instance;
+
+	/*
+	 * Toggling a raw bit is equivalent to toggling the logical value
+	 * regardless of polarity, so no active-low correction is needed here:
+	 * (raw = logical ^ active_low_mask) => (logical ^ 1) ^ active_low_mask
+	 * == raw ^ 1.
+	 */
+	uint32_t current = XGpio_DiscreteRead(inst, xhandle->channel);
+	uint32_t to_toggle = (uint32_t)pins_bitmask & ((1U << xhandle->num_pins) - 1);
```
rbolboac: When num_pins == 32, 1U << 32 is undefined behavior; on x86/ARM the shift count wraps to 0, yielding mask 0, so the port toggles nothing. The sibling port_set_raw_value already guards this exact case (num_pins == 32U ? UINT32_MAX : ...). Apply the same guard here.

capi/platform/xilinx/xilinx_capi_i2c_pl.c
```diff
@@ -0,0 +1,1596 @@
...
+
+		status = XIic_MasterSend(inst(xh), tx_buf, (int)tx_len);
+		if (status != XST_SUCCESS) {
+			xiic_cancel_rejected_async_start(xh);
+			xh->callback = saved_cb;
+			ret = xiic_status_to_errno(status);
+			goto out_free;
+		}
+
+		uint32_t timeout = I2C_PL_BUS_BUSY_TIMEOUT;
```
rbolboac: what does this timeout amount to (in time)? in general I2C is quite slow, is this enough if a larger transfer is performed and there is a high CPU frequency?

capi/platform/xilinx/xilinx_capi_i2c_pl.c
```diff
@@ -0,0 +1,1596 @@
...
+			status = XIic_MasterRecv(inst(xh), transfer->buf,
+						 transfer->len);
+		}
+		if (status != XST_SUCCESS) {
+			xiic_cancel_rejected_async_start(xh);
+			xh->callback = saved_cb;
+			return xiic_status_to_errno(status);
+		}
+
+		uint32_t timeout = I2C_PL_BUS_BUSY_TIMEOUT;
```
rbolboac: same remark with transmit... is this high enough if more bytes are received?

capi/platform/xilinx/xilinx_capi_irq_gic.c
```diff
@@ -0,0 +1,560 @@
...
+ * @brief Return the active XScuGic instance.
+ * @note PS: Returns active XScuGic instance.
+ *
+ * @return XScuGic instance pointer, or NULL if not initialized.
+ */
+XScuGic *capi_irq_get_scugic_instance(void)
+{
+	return g_initialized ? &g_scugic : NULL;
+}
+#endif /* XPAR_XSCUGIC_NUM_INSTANCES */
```
rbolboac: no newline at EOF

capi/platform/xilinx/xilinx_capi_gpio_pl.c
```diff
@@ -0,0 +1,695 @@
...
+	 */
+	XGpio *inst = (XGpio *)xhandle->instance;
+	uint32_t current = XGpio_DiscreteRead(inst, xhandle->channel);
+	uint32_t raw = current ^ (1U << pin->number);
+
+	XGpio_DiscreteWrite(inst, xhandle->channel, raw);
+	return 0;
+}
+
+#endif /* XPAR_XGPIO_NUM_INSTANCES */
```
rbolboac: no newline at EOF

.github/scripts/astyle.sh
liviutomoiaga, line 1: LGTM

.github/workflows/build-common.yaml
liviutomoiaga, line 1: LGTM

liviutomoiaga (commented):
On CI side, the changes are fine.

liviutomoiaga (dismissed):

capi/platform/xilinx/xilinx_capi_i2c_pl.c
Voc94: no, it is not, you're right, the reasoning is not ok, for the slowest one would give timeout for a transfer of over 8 bytes, should I make a precise timeout function based on speed of SCL(using xh->clk_freq_hz & transfer_length, faster timeout, but calculation cost) or a compile time larger timeout constant based on CPU_FREQ defined value (less precise but easier)?

capi/platform/xilinx/xilinx_capi_i2c_pl.c
Voc94: zynq has a 650 Mhz CPU , 5cycles/iter while loop , 100k iterations constant
  time = iterations × cycles/iter / CPU_freq = 0.77ms(timeout wait)
SCL = 0.1Mhz(slowest) =>  for 8 bytes(ADDR+7 DATA) & 9 clk/byte(data&ACK) we get => 8\*9\*0.1 = 0.72 ms(transfer) anything bigger would timeout

capi/platform/xilinx/xilinx_capi_i2c_pl.c
rbolboac: @buha any input? for me it's fine either way, I just want to avoid false timeout errors

Voc94:
## Changelog V12

from review:

  @rbolboac
  - UART PS: transmit() now loops until the whole buffer is queued. XUartPs_Send only pushes what fits in the ~64-byte TX FIFO; the polled path never sent the rest, so anything over ~64 bytes truncated silently.
  - GPIO PL: port_toggle() guards the 32-pin mask. 1U << 32 is UB and toggled nothing. Matches the sibling port_set_raw_value.
  - I2C PL: replaced the fixed iteration-count timeouts with real time-based ones. The old limit was a raw loop counter (~0.77 ms on this CPU), so it fired falsely on transfers longer than ~8 bytes and got shorter as CPU speed went up. Each transfer now gets a deadline scaled to its actual length and bus speed **roughly 3× the wire time + 2 ms**, never less than 5 ms measured against a real clock. Backends with no capi_uptime clock use the same as before. (MicroBlaze) (microblaze CLK is usually slower than that of the ARM CPU so with i2c it's not that much of a problem)
  - EOF newlines: irq_gic.c, gpio_pl.c.

  @buha
  - I2C PS/PL: use_irq setup failure is now a hard error, not a silent downgrade to polled. A target can't answer without its ISR.
 
regarding the changes needed for xilinx to handle changing bus speeds for target tests in capi-selftest:

- I2C PS configure_bus_speed: made it work on a controller that's mid-HOLD / repeated-start / busy , it now resets-and-restores before SetSClk (which silently fails on a non-idle), so re-speeding a target actually takes. This is what lets a PS target be reconfigured to the initiator's bus; a 100k target can't sample 400k until this runs. ( slave clk_freq config PS-only must match master)
- I2C PS/PL docs: rewrote configure_bus_speed notes to the verified story. PS SetSClk sizes a divisor shared by master SCL and slave sampling, failure was a divider mismatch, not electrical as claude claimed :/. 
  - I2C header: added the same PS/PL sampling-window note to the xilinx_capi_i2c.h banner, pointing at capi_i2c_{ps,pl}_configure_bus_speed.

Voc94:
## Changelog v13

  I2C PS

  - Target receive now uses a bounce buffer when the caller's buffer is smaller than
    the hardware's max transfer size (252 bytes). XIicPs_SlaveRecv() discards the
    ByteCount parameter and SlaveRecvData() drains the RX FIFO unbounded, so an
    initiator writing more than transfer->len overwrites the caller's buffer. The
    bounce catch-and-copy on completion protects it.

  UART Lite

  - lite_xfer_busy() and the async guards now test RequestedBytes instead of
    RemainingBytes. RemainingBytes counts what's left to hand to the FIFO and clears
    early on transmit, leaving the tail looking idle; RequestedBytes spans the whole
    transfer. In external loopback RX cannot finish until TX completes, so the RX
    completion runs first and would pull the shared interrupt enable out from under
    the pending TX on a RemainingBytes check, losing the TX-empty edge outright on
    this edge-triggered line.
  - IRQ trigger changed from CAPI_IRQ_LEVEL_HIGH to CAPI_IRQ_EDGE_RISING. The BSP
    XPAR_XUARTLITE_0_INTERRUPTS trigger nibble is 0x10 (type 1, edge rising), not
    level high like NS550 (0x40) or I2C (0x4022).
  - Polled transmit() and receive() now zero RequestedBytes on the way out (vendor
    Send() sets it and only the interrupt path clears it).
  - ISR re-services until status register is quiet, capped at 8 passes. Both
    directions are OR-ed onto one edge-triggered pin, so a source rising while the
    other holds the line produces no edge and would be missed by a single-read
    InterruptHandler().
  - transmit_async() only disables interrupts when no async receive is pending
    (shared single-bit enable for both directions).
  - receive_async() completes synchronously when XUartLite_Recv() drains the FIFO
    on entry (RX pre-buffered by a prior blocking transmit).
  - read_byte op implemented: polls XUartLite_IsReceiveEmpty(), reads XUL_RX_FIFO_OFFSET.
    Returns 0 while lite_xfer_busy() holds (a buffered transfer owns the FIFO).

rbolboac (dismissed):

Voc94:
## Changelog v14

  I2C PL

  - Target receive now arms an internal bounce buffer (XIIC_TARGET_RX_ABSORB_LEN, 256) when the caller's buffer is shorter, and hands back only the requested length on completion. 
  - XIic counts the armed ByteCount down and asserts NO_ACK the moment it hits zero, so arming the caller's exact length made an initiator that sent more get NAKed mid-frame and fail its write. Judged on bytes actually arrived (armed - remaining), not the reported remainder: short of caller_len is truncation (-EIO), at or over it the request is satisfied and the surplus dropped. Buffer freed on the failure path too (xiic_complete_transfer,xiic_reset_async_state), never leaked.

  I2C PS

  - capi_i2c_ps_receive_async() now zeros inst(xh)->RecvByteCount before arming the target receive. It is a free-running up-counter in slave-receive mode that no vendor call resets per transfer, and xiicps_async_complete() reads it to size the bounce copy-back. Without the reset the second and later target receives copied
  back a stale tail from an earlier frame.
  - xiicps_async_complete() now judges a bounced target receive on bytes actually arrived: got >= caller_len hands back exactly caller_len, short of it returns -EIO / CAPI_I2C_NONE and copies nothing. Was memcpy(min(got, caller_len)) reported as success, so a truncated frame looked like a good short read. Now matches the PL backend, so a short target receive is identical on both.

  Private headers

  - async_rx_bounce / async_rx_caller_len doc updated: no longer "PS target only". Both backends need it for opposite hardware reasons -- XIicPs ignores ByteCount and overruns an unbounced short buffer, XIic honours it and NAKs one bouncing gives both the same CAPI-visible result.

Timer PL

 - Up-counter now gets exactly the period the caller asked for. The AXI timer always counts up and rolls over at 2^32 -- that rollover point is fixed in hardware, all you control is the value it starts from. So instead of loading TLR (Timer/Counter Load Register, the start value reloaded on every rollover) with 0 and counting the full 2^32, it now preloads 2^32 - config->max, so the counter climbs from there to the rollover in exactly max ticks. A full-span request (0 or UINT32_MAX) still loads 0. Supersedes the earlier "max cannot cap an up-counter, not applicable" reasoning.

  UART PS

  - Split ps_xfer_busy() into ps_tx_busy() / ps_rx_busy(); transmit() and receive() each guard only their own direction. TX and RX are independent, and the async API holds a receive armed while TX moves the other way, so the combined check refused exactly the arm-RX-then-send-TX sequence full-duplex callers rely on.
  - RECV_TOUT is terminal only when rx_remaining == 0 (routes to ps_rx_done()); with bytes still outstanding it reports RX_TIMEOUT as progress and leaves the receive armed. The old code masked RXOVR/RXFULL/TOUT unconditionally, disabling the very interrupt the remaining bytes needed, so RX_DONE could never follow and the transfer stalled forever.

buha (approved):

PR #3157, opened 2026-06-24
drivers: adc: ad4170: Add support for AD4190-8

rbudai98:
## Description

Extend the existing `ad4170` driver (which already supports AD4170/AD4171/AD4172/AD4190/AD4195) with support for the **AD4190-8**.

The AD4190-8 is register-compatible with the already-supported AD4190-4. Comparing the two datasheets, the register maps are identical. The differences are:

- **Inputs:** 8 differential / 16 pseudo-differential (AD4190-8) vs. 4 differential / 8 pseudo-differential (AD4190-4), in a 40-lead LFCSP package. The driver already handles up to 16 channels generically, so no change is required here.
- **PRODUCT_ID:** reads `0x0049` on the AD4190-8 vs. `0x0048` on the AD4190-4. This is the only programming-visible difference and is what the driver checks during device verification at init.

## Changes

- Add `AD4190_8_PRODUCT_ID_L_VALUE` (`0x49`) / `AD4190_8_PRODUCT_ID_H_VALUE` (`0x0`).
- Add `ID_AD4190_8` to `enum ad4170_id`.
- Handle `ID_AD4190_8` in `ad4170_init()` so the PRODUCT_ID verification accepts the new part.

This follows the same minimal pattern used when AD4195 support was added (commit d01e5b93b).

## Testing

- Verified register maps of AD4190-4 vs AD4190-8 datasheets are identical (only PRODUCT_ID differs: 0x48 → 0x49).
- Change mirrors the established variant-add pattern in this driver.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


amiclaus (dismissed):

rbudai98:
v2:
- solved merge conflicts

rbolboac (approved):

PR #3159, opened 2026-06-24
drivers: gnss-gps: nmea_ubx: fix gnss_ubx_get_val

DiaconuGabriel:
UBX-CFG-VALGET encodes the configuration layer as an enum (RAM=0, BBR=1, FLASH=2), unlike UBX-CFG-VALSET which uses a bitmask (RAM=1, BBR=2, FLASH=4). gnss_ubx_get_val reused the VALSET bitmask values, so it requested the wrong layer. Add dedicated VALGET_LAYER_* constants and use them in the VALGET path.

The module also sends an ACK after the VALGET response. That ACK in gnss_ubx_get_val needs to be consumed so it will not be mistaken as an actual message by the next command.

Fixes: 257c700 ("drivers: gnss-gps: nmea_ubx: initial commit")

## Pull Request Description

## PR Type

## PR Checklist

### DIFF
--- a/drivers/gnss-gps/nmea_ubx/nmea_ubx.c
+++ b/drivers/gnss-gps/nmea_ubx/nmea_ubx.c
@@ -717,16 +717,17 @@ int gnss_ubx_get_val(struct gnss_dev *dev, uint32_t key_id,
 	if (value_size != 1 && value_size != 2 && value_size != 4 && value_size != 8)
 		return -EINVAL;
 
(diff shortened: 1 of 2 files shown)


amiclaus (approved):

rbolboac:
while at it, could you also add a builds.json file for the projects using these files? I only found eval-ublox-gnss

buha (approved):

PR #3160, opened 2026-06-25
gcc14 fixes for stm32 projects (and used drivers)

buha:
## Pull Request Description

We bumped CubeIDE on the CI from 1.7.0 to 2.1.1 which implies a bump in gcc from gcc9 to gcc14. This change reveals some compilation errors that were previously treated as warnings.

## PR Type

## PR Checklist


buha:
force pushed a v2 which contains a fix for ad405x which was a bit special, it builds for another family than was previously tested and that revealed more compiler adaptation type of changes needed in the stm32 implementation

rbolboac (dismissed):

amiclaus (dismissed):

build_projects.yml
```diff
@@ -87,8 +87,26 @@ jobs:
     clean: true
     persistCredentials: true
   - script: |
-      echo "Using STM32CubeIDE version: $(STM32CUBEIDE_VERSION)"
-      echo "Using STM32CubeMX version: $(STM32CUBEMX_VERSION)"
+      [ -f /no-OS_agent/.stm32_environment.sh ] && . /no-OS_agent/.stm32_environment.sh
+      STM32CUBEIDE="${STM32CUBEIDE:-/opt/stm32cubeide}"
```
liviutomoiaga, line 91: I think the fallback path should a different one. The available versions are on this path: `opt/st/` with the following subfolders
stm32cubeide_1.7.0/   stm32cubeide_2.1.1/   stm32cubeide_2.1.1_2/

build_projects.yml
```diff
@@ -87,8 +87,26 @@ jobs:
...
+      STM32CUBEMX="${STM32CUBEMX:-/opt/stm32cubemx}"
```
liviutomoiaga, line 92: Same here - the fallback path do not exist.

liviutomoiaga (commented):
Small notes from my side for `build_projects.yml`.

build_projects.yml
buha: i'll leave it as-is, /opt/stm32cubeide and /opt/stm32cubemx are embedded in no-OS build system, they should've been installed there in the first place but it's okay, it's irrelevant where they are installed as long as there is a .stm32_environment.sh which specifies it

rbolboac (approved):

buha:
adt4000 will still fail because: https://github.com/analogdevicesinc/precision-converters-library/pull/23

PR #3206, opened 2026-07-13
Staging/capi selftest

cristina-suteu:
## Pull Request Description

  This PR implements a platform-agnostic system test of the CAPI platform drivers.

  Adds the CAPI selftest project, a comprehensive integration suite for the Common API platform drivers across Xilinx (Zynq-7000) and STM32 (NUCLEO-F767ZI) targets.

  The selftest validates CAPI peripherals through a type-agnostic test framework that remains inert until each driver is mapped on the target platform. Every peripheral can be independently gated via Kconfig
  (`CONFIG_CAPI_*`); tests compile only when their dependencies are met, and skip gracefully at runtime when a peripheral is absent.

  ### Examples

  - **basic_example**: Minimal UART transmit sanity check. Entry point for new users and CI smoke tests.
  - **capi_loopback**: Comprehensive suite exercising GPIO, IRQ, SPI, Timer, I2C, DMA, and UART through real hardware loopback and the test framework. Selectable per-peripheral skip modes for partial runs.

  ### Test Framework

  - Structured result tracking (pass / fail / skip with per-subtest granularity).
  - Watchdog hooks for cleanup on failure, configurable per-section.
  - Console UART transport for all output, with chunked polled writes to avoid TX FIFO stalls.

  ### Peripherals Tested

  Each test exercises loopback, data integrity, timing, and async delivery (IRQ or DMA where supported):

  - **GPIO**: Port and pin loopback over a wired pair (PS EMIO on Xilinx, PE0↔PC0 on STM32).
  - **IRQ**: GPIO-driven integration path validation (arm / ack / disarm hooks).
  - **SPI**: Clock modes, LSB-first, transfer sizes, and async abort.
  - **Timer**: Counter movement, compare/capture, overflow IRQ, rate calibration.
  - **I2C**: Two-bus setup (initiator + target on cross-wired SCL/SDA); address reconfig, repeated-start, NACK handling, and async target IRQ.
  - **DMA**: Memory-to-memory transfers (STM32 only).
  - **UART**: Async loopback (second instance, never console) on PS/PL backends on Xilinx.

  ### Platform Mappings

  **Xilinx:**
  - PS and PL variants coexist where available (gpio_ps + gpio_pl, uart_ps + uart_pl_lite + uart_pl_ns550, etc.), selected at compile time via `XPAR_*_NUM_INSTANCES`.
  - IRQ controller is auto-detected (GIC vs. AXI INTC) or manually selected.
  - Per-peripheral async support gated on IRQ availability.

[...]

PR #3207, opened 2026-07-13
STM32 CAPI platform drivers fixes

cristina-suteu:
## Pull Request Description

This PR contains fixes for stm32 capi platform drivers discovered during the testing process, using the capi_selftest here : https://github.com/analogdevicesinc/no-OS/pull/3206

## PR Type

## PR Checklist

### DIFF
--- a/capi/platform/stm32/stm32_capi_spi.c
+++ b/capi/platform/stm32/stm32_capi_spi.c
@@ -100,7 +100,7 @@ static int stm32_capi_spi_config_peripheral(struct stm32_spi_priv_handle
 	int ret = 0;
 
 	/* Verify SPI instance is configured */
-	if (!priv_handle->hspi.Instance)
+	if (!priv_handle->hspi->Instance)
 		return -EINVAL;
 
 	/* automatically select prescaler based on max_speed_hz */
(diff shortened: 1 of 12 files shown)


capi/platform/stm32/stm32_capi_uart.c
```diff
@@ -538,6 +538,518 @@ static int stm32_capi_uart_get_line_status(struct capi_uart_handle *handle,
...
+		;
+
+	frame = (uint16_t)((data & 0xFFU) | (is_address ? 0x100U : 0x000U));
+#if defined(USART_TDR_TDR)
+	priv->huart->Instance->TDR = frame;
+#else
+	priv->huart->Instance->DR = frame;
+#endif
+
+	while (!__HAL_UART_GET_FLAG(priv->huart, UART_FLAG_TC))
```
buha: can we bound this in any way ? alternatively, can we use stm32 hal to do the same thing and not bother with bounding the loop ?

capi/platform/stm32/stm32_capi_spi.c
```diff
@@ -899,25 +982,24 @@ static int stm32_capi_spi_abort_async(struct capi_spi_device *device)
...
 
-		if (priv_handle->callback)
-			priv_handle->callback(CAPI_SPI_EVENT_ERROR,
-					      priv_handle->callback_arg,
-					      -ECANCELED);
-		return -EIO;
-	}
+	if (priv_handle->callback)
+		priv_handle->callback(CAPI_SPI_EVENT_ERROR,
+				      priv_handle->callback_arg, 0);
```
buha, line 1002: is there a reason why you changed -ECANCELED -> 0 ?

cristina-suteu: when asking the HAL to abort the SPI transfer, this callback was reached when the abort itself failed, returning -ECANCELLED. now, the abort is done at the register level, which should always work, and this callback is reached when an abort was completed successfully.  the event is still a CAPI_SPI_EVENT_ERROR , because the from the transfer's point of view, this is still an error, but the 0 signifies that the transfer was cancelled successfully.

cristina-suteu:
v2: 
- stm32_capi_uart: use stm32 HAL with built-in timeout for 9bit mode
- stm32_capi_uart:  added parity safeguard for 9bit mode

buha (approved):

PR #3220, opened 2026-07-17
Migrate pico, maxim and xilinx platform to github actions

Lewluu:
## Pull Request Description

Based on decisions regarding migration of no-os platforms from azure-pipelines to github actions, the following changes will take place: Pico and maxim will run on github cloud hosted runners(ubuntu-latest) in docker containers Xilinx will run on github self-hosted runners with mounted tools needed Checkers + docs publishing will run on github cloud hosted runners as well

## PR Type

## PR Checklist


.github/workflows/build-xilinx-platform.yaml
```diff
@@ -0,0 +1,113 @@
+name: Build NO-OS projects
```
liviutomoiaga: Change the name to `Build Xilinx platform`

.github/workflows/build-light-platforms.yaml
```diff
@@ -0,0 +1,126 @@
+name: Build NO-OS projects
```
liviutomoiaga: Change eventually to `Build Pico and Maxim`

.github/workflows/build-xilinx-platform.yaml
```diff
@@ -0,0 +1,113 @@
...
+        uses: actions/upload-artifact@v7
+        with:
+          path: ${{ env.RELEASES_DIR }}
+          name: release_${{ matrix.platform }}_${{ github.job }}
+      - name: Upload logs artifact
+        uses: actions/upload-artifact@v7
+        with:
+          path: ${{ env.LOGS_DIR }}
+          name: logs_${{ matrix.platform }}_${{ github.job }}
+  unit-tests:
```
liviutomoiaga: I think it's better to separate Unit Test job from `Build Xilinx platform`

.github/workflows/build-common.yaml
```diff
@@ -0,0 +1,49 @@
+name: Main build workflow for no-os
+
```
liviutomoiaga, line 2: Leave a suggestive name about what is going to be built on this flow

.github/workflows/build-light-platforms.yaml
liviutomoiaga: or even separate the platforms in independent yml files

.github/workflows/build-common.yaml
```diff
@@ -0,0 +1,49 @@
...
+        uses: actions/checkout@v5
+        with:
+          submodules: recursive
+      - name: Build drivers
+        shell: sh
+        run: |
+          sudo apt-get install -y gcc-arm-none-eabi libnewlib-arm-none-eabi
+          make -j${NUM_JOBS} -C ${{ github.workspace }}/drivers -f Makefile
+
+
```
liviutomoiaga: extra new line

.github/workflows/build-common.yaml
liviutomoiaga: checkers and drivers

Lewluu:
Status: **maxim**, **pico** and **xilinx**(https://github.com/analogdevicesinc/no-OS/actions/runs/29842535838/job/88674770500?pr=3220) platforms are working.
Next changes are will only affect optimizations and fix checker/unit-tests etc.

liviutomoiaga:
> Status: **maxim**, **pluto** and **xilinx**(https://github.com/analogdevicesinc/no-OS/actions/runs/29842535838/job/88674770500?pr=3220) platforms are working. Next changes are will only affect optimizations and fix checker/unit-tests etc.

STM32 and ADuCM3029 will be as well migrated towards GitHub Actions but in separated PR.

build_projects.yml
```diff
@@ -1,19 +1,22 @@
-trigger:
-- main
-- next_stable
-- 20*
+trigger: none
```
liviutomoiaga: To set back the triggers to it's original state right before merging this PR into main.

.github/config/_ubuntu.sources
liviutomoiaga, line 1: Is this file still needed ? and if yes, why ?

Lewluu: not anymore. will be deleted in the next changes for optimizations. since the checkers will take long again to run for a single file deletion, it's not worth it to remove it now. thanks for observation!

SRaus:
> ## PR Type
> * [ ]  Bug fix (change that fixes an issue)
> * [ ]  New feature (change that adds new functionality)
> * [ ]  Breaking change (has dependencies in other repos or will cause CI to fail)
> 
> ## PR Checklist
> * [ ]  I have followed the [Coding style guidelines](http://analogdevicesinc.github.io/no-OS/drivers_guide.html#coding-style)
> * [ ]  I have complied with the [Submission Checklist](http://analogdevicesinc.github.io/no-OS/contributing.html#submission-checklist)
> * [ ]  I have performed a self-review of the changes
> * [ ]  I have commented my code, at least hard-to-understand parts
> * [ ]  I have build all projects affected by the changes in this PR
> * [ ]  I have tested in hardware affected projects, at the relevant boards
> * [ ]  I have signed off all commits from this PR
> * [ ]  I have updated the documentation (wiki pages, ReadMe etc), if applies

Please Check the boxes that apply :)

.github/config/_ubuntu.sources
SRaus: So - if I understood it properly - in order to delete the file, you'll just need to re-run the whole CI. 
How long is this taking? Couldn't we wait with merging the PR for another couple of hours?

.github/scripts/astyle.sh
SRaus, line 1: It is a good practice to have file renaming and functional changes on the same file in different commits (reasons such as: if a bug appears later, it's easier to identify whether it came from the rename or the changes; each commit should have a single purpose; easy to track history across files; easy to review etc)

.github/scripts/astyle.sh
```diff
@@ -26,21 +26,46 @@
...
 ASTYLE_EXT_LIST="${ASTYLE_EXT_LIST} .c .h"
-
 COMMIT_RANGE="$1"
 
+parse_commit_range() {
+	local operation_name="${1:-check}"
+
+	if [ -z "$COMMIT_RANGE" ]; then
+		COMMIT_RANGE="${COMMIT_RANGE}"
+	fi
```
SRaus, line 54: Not clear for me what this IF statement should do. 
I think it can be deleted, but if it's really necessary, add a comment to describe what's doing.

.github/scripts/astyle.sh
```diff
@@ -26,21 +26,46 @@
...
+
+	if [ -z "$COMMIT_RANGE" ]  && [ -n "$TARGET_BRANCH" ]; then
+		git fetch --depth=50 origin $TARGET_BRANCH
```
SRaus, line 57: Add a comment explaining the "magic number" 50 (why do you need to use exactly "50")

.github/scripts/astyle.sh
```diff
@@ -26,21 +26,46 @@
...
+		git branch $TARGET_BRANCH origin/$TARGET_BRANCH
+		COMMIT_RANGE="${TARGET_BRANCH}.."
+	fi
+
+	if [ -z "$COMMIT_RANGE" ]; then
```
SRaus, line 62: You have a similar checking at line 38 :
```
if [ -z "$COMMIT_RANGE" ]; then
		bla bla 
fi
```
Couldn't you just merge the two IFs ? If not, why ?

.github/scripts/astyle.sh
```diff
@@ -26,21 +26,46 @@
...
+	local operation_name="${1:-check}"
+
+	if [ -z "$COMMIT_RANGE" ]; then
+		COMMIT_RANGE="${COMMIT_RANGE}"
+	fi
+
+	if [ -z "$COMMIT_RANGE" ]  && [ -n "$TARGET_BRANCH" ]; then
+		git fetch --depth=50 origin $TARGET_BRANCH
+		git branch $TARGET_BRANCH origin/$TARGET_BRANCH
+		COMMIT_RANGE="${TARGET_BRANCH}.."
```
SRaus, line 59: It seems you are creating a local branch at every run, with the same name "$TARGET_BRANCH". Isn't enough to just clone that branch ?
Or, if this is used just to compare against the main, it can be simplified to something like `COMMIT_RANGE="origin/${TARGET_BRANCH}..HEAD"`

.github/scripts/astyle.sh
```diff
@@ -26,21 +26,46 @@
...
+	fi
+
+	if [ -z "$COMMIT_RANGE" ]; then
+		echo "Using only latest commit, since there is no Pull Request"
+		COMMIT_RANGE=HEAD~1
+	fi
+
+	echo "Running $operation_name on commit range '$COMMIT_RANGE'"
+	echo "Commits should be:"
+	if ! git rev-parse $COMMIT_RANGE ; then
```
SRaus, line 69: `git rev-parse` command returns you some output (because it actually process revisions). You would probably want to change it, to get just the exit code, maybe somethign like
 `git rev-list "$COMMIT_RANGE" >/dev/null 2>&1`

.github/scripts/astyle.sh
```diff
@@ -26,21 +26,46 @@
...
+		git branch $TARGET_BRANCH origin/$TARGET_BRANCH
+		COMMIT_RANGE="${TARGET_BRANCH}.."
+	fi
+
+	if [ -z "$COMMIT_RANGE" ]; then
+		echo "Using only latest commit, since there is no Pull Request"
+		COMMIT_RANGE=HEAD~1
+	fi
+
+	echo "Running $operation_name on commit range '$COMMIT_RANGE'"
```
SRaus, line 67: You have some variables quoted with " " , some quoted with ' ', and some unquoted.
Add them to the same standard, which " " (ask internet/AI the difference between all 3 cases and why this is the best).

.github/scripts/astyle.sh
```diff
@@ -26,21 +26,46 @@
...
 	[[ -f ".astyleignore" ]] &&
-	while read -r fpath
-	do
+	while read -r fpath; do
 		[[ -z "$fpath" ]] && continue
 		[[ "$file" == *"$fpath"* ]] && return 1
 	done < ".astyleignore"
+
 	return 0
 }
```
SRaus, line 84: You have mixed into the same commit: file renaming, functional changes and coding style changes. Mixing them makes reviews harder and git history noisier. 
It would make more sense to have them in separated commits.

.github/scripts/setup-platform-env.sh
```diff
@@ -0,0 +1,16 @@
...
+if [[ "$PLATFORM" == "pico" ]]; then
+    echo "/home/runner/py-env/bin" >> $GITHUB_PATH
+    echo "VIRTUAL_ENV=/home/runner/py-env" >> $GITHUB_PATH
+    echo "PICO_SDK_PATH=${WORKSPACE}/libraries/pico-sdk" >> $GITHUB_ENV
+elif [[ "$PLATFORM" == "maxim" ]]; then
+    echo "/home/runner/py-env/bin" >> $GITHUB_PATH
+    echo "VIRTUAL_ENV=/home/runner/py-env" >> $GITHUB_PATH
+    echo "CMAKE_GENERATOR=Ninja" >> $GITHUB_ENV
+elif [[ "$PLATFORM" == "xilinx" ]]; then
+    cloudsmith download ${TOOLS_REPO} hardware.tar.gz --tag latest --outfile ${BUILDS_DIR}_${HDL_BRANCH}/hardware.tar.gz
```
SRaus: Using CloudSmith in this case will just add extra time for upload/download the archive.
You can keep it locally, somewhere on the server:
  - just copying it will take shorter time 
  - keep the flow simpler (just a cp command instead of generating CloudSmith OIDC, upload and download commands etc).

.github/workflows/build-common.yaml
```diff
@@ -0,0 +1,78 @@
+name: Workflow for checkers, drivers and unit-tests
+
+on:
+  pull_request:
+    branches:
+      - main
+      - staging/*
+      - next_stable
```
SRaus, line 8: Check with dev team if "next_stable" is still needed here.

.github/workflows/build-common.yaml
```diff
@@ -0,0 +1,78 @@
...
+      - name: Checkout code
+        uses: actions/checkout@v5
+        with:
+          submodules: recursive
+      - name: Build drivers
+        shell: sh
+        run: |
+          sudo apt-get install -y gcc-arm-none-eabi libnewlib-arm-none-eabi
+          make -j${NUM_JOBS} -C ${{ github.workspace }}/drivers -f Makefile
+  unit-tests:
```
SRaus, line 52: Shouldn't unit tests run also on pushed commits?

.github/workflows/build-light-platforms.yaml
```diff
@@ -0,0 +1,98 @@
+name: Build Pico and Maxim platforms
+
+on:
+  pull_request:
+    branches:
+      - main
+      - next_stable
```
SRaus, line 7: Same comment on 'next_stable' as above.

.github/workflows/build-light-platforms.yaml
```diff
@@ -0,0 +1,98 @@
...
+      matrix:
+        include:
+          - runs-on: "ubuntu-latest"
+            platform: "pico"
+            container: "pico-platform:1.0.0"
+          - runs-on: "ubuntu-latest"
+            platform: "maxim"
+            container: "maxim-platform:1.0.0"
+    env:
+      CLOUDSMITH_API_KEY: ${{ secrets.CLOUDSMITH_API_KEY }}
```
SRaus, line 56: Add a comment where to explain the CloudSmith workaround you've implemented here. 
Maybe it can be kept simpler than that.

.github/workflows/build-light-platforms.yaml
```diff
@@ -0,0 +1,98 @@
...
+        run: |
+          ${{ github.workspace }}/.github/scripts/setup-platform-env.sh
+        env:
+          PLATFORM: ${{ matrix.platform }}
+          WORKSPACE: ${{ github.workspace }}
+      - name: Build project ${{ matrix.platform }}
+        shell: sh
+        run: |
+          echo "Setup cloudsmith helper ..."
+          wget -P ./tools/scripts/ \
```
SRaus: -P and -O params don't make sense in the same wget command.

.github/workflows/build-xilinx-platform.yaml
```diff
@@ -0,0 +1,87 @@
...
+        env:
+          PLATFORM: xilinx
+          WORKSPACE: ${{ github.workspace }}
+      - name: Build Xilinx projects
+        shell: sh
+        run: |
+          echo "Setup cloudsmith helper ..."
+          wget -P ./tools/scripts/ \
+            https://raw.githubusercontent.com/analogdevicesinc/wiki-scripts/refs/heads/main/utils/cloudsmith_utils/cloudsmith_helper.py -O ./tools/scripts/cloudsmith_temp.py
+          cat tools/scripts/cloudsmith_temp.py | sed 's#/usr/bin/python3#/usr/bin/env python3#' > tools/scripts/cloudsmith_helper.py
```
SRaus, line 78: Explain why you need the sed command here (I know that '/usr/bin/env python3' enables by default the venv, but it also depends on the default user's PATH)

.github/workflows/build-xilinx-platform.yaml
```diff
@@ -0,0 +1,87 @@
...
+      - name: Upload Xilinx build logs
+        if: always()
+        uses: actions/upload-artifact@v7
+        with:
+          path: ${{ env.LOGS_DIR }}
+          name: logs_xilinx_${{ github.job }}
+      - name: Remove logs, new_hardware and release output
+        shell: sh
+        run: |
+          rm -rf ${{ env.RELEASES_DIR }} ${{ env.LOGS_DIR }} ${{ env.BUILDS_DIR }}_${{ env.HDL_BRANCH }}/new_hardware
```
SRaus, line 106: If you delete new_hardware, there will be no baseline available for comparing next versions of XSAs (and other HDL build output files).

.github/workflows/deploy-docs.yaml
```diff
@@ -0,0 +1,50 @@
+name: Deploy docs for ToF
```
SRaus: ToF?

.github/workflows/deploy-docs.yaml
```diff
@@ -0,0 +1,50 @@
...
+      - name: Deploy doxygen
+        if: ${{ steps.verify_changed_files.outputs.docs == 'true' || steps.verify_changed_files.outputs.drivers == 'true' }}
+        shell: sh
+        run: |
+          ${{ github.workspace }}/.github/scripts/deploy-docs.sh
+        env:
+          TOP_DIR: ${{ github.workspace }}
+          NUM_JOBS: ${{ env.NUM_JOBS }}
+          VERSION: '1.13.2'
+          UPDATE_GH_PAGES: ${{ github.event == 'workflow_dispatch' && inputs.update-gh-pages || 'true' }}
```
SRaus: Wrong EOF character (probably from Windows).

tools/scripts/build_projects.py
```diff
@@ -164,16 +164,25 @@ def cmake_cache_source_mismatch(build_dir, noos):
 
 def process_blacklist():
 	blacklist = []
-	err = os.system('curl -L -H "Accept: application/vnd.github.v3.raw" -H "Authorization: Bearer {}" \'{}\' -o blacklist.txt >> {} 2>&1'
+	if not TOKEN:
+		log_err('TOKEN (GH_GHDL_TOKEN) environment variable is not set')
+		return blacklist
```
SRaus: Maybe you can optimize this - an empty blacklist is returned if:
 - cannot connect to ghdl
 - blacklist url is not set
 - can connect to ghdl, but blacklist failed to download etc
For example - if blacklist_url was not given, or it was intentionally set to 'None', you my still return an empty blacklist. Is this desired behavior?

tools/scripts/build_projects.py
```diff
@@ -164,16 +164,25 @@ def cmake_cache_source_mismatch(build_dir, noos):
...
-	for line in file.readlines():
-		project = line.split('#')[0].rstrip().replace('.', '_')
-		if project != '':
-			blacklist.append(project)
+	if not os.path.isfile('blacklist.txt'):
+		log_err('Blacklist file was not created after download')
+		return blacklist
+	with open('blacklist.txt', 'r') as file:
+		for line in file.readlines():
+			project = line.split('#')[0].rstrip().replace('.', '_')
```
SRaus: Why `rstrip` (removing trailing whitespace only from the right side) instead of just `strip` (removing trailing whitespace from both ends) ?

tools/scripts/build_projects.py
```diff
@@ -164,16 +164,25 @@ def cmake_cache_source_mismatch(build_dir, noos):
...
-	file = open('blacklist.txt', 'r')
-	for line in file.readlines():
-		project = line.split('#')[0].rstrip().replace('.', '_')
-		if project != '':
-			blacklist.append(project)
+	if not os.path.isfile('blacklist.txt'):
+		log_err('Blacklist file was not created after download')
+		return blacklist
+	with open('blacklist.txt', 'r') as file:
+		for line in file.readlines():
```
SRaus: You've already opened the file for read by using "with ...", this  means the for statement can be simplified to `for line in file`.
`readlines()` will read the file line by line, in cash, one more time.

SRaus (dismissed):

.github/scripts/astyle.sh
Lewluu: it's an already existing script migrated from azure-pipelines, not created by me. modifying it now is not the purpose of this migration.

.github/scripts/astyle.sh
Lewluu: it's an already existing script migrated from azure-pipelines, not created by me. modifying it now is not the purpose of this migration.

.github/scripts/astyle.sh
Lewluu: it's an already existing script migrated from azure-pipelines, not created by me. modifying it now is not the purpose of this migration.

.github/scripts/setup-platform-env.sh
Lewluu: the download and unzipping doesn't take that much. using the same shared folder is considered for the next optimizations for now.

.github/workflows/build-common.yaml
Lewluu: unit tests upload the data in the PR. if there's a wish for running this workflow, there need to be implemented a condition to detect which kind of event it is (pr or push)

.github/workflows/build-light-platforms.yaml
Lewluu: that is needed for the cloudsmith_helper.py

.github/workflows/build-xilinx-platform.yaml
Lewluu: it replaces the #!/usr/bin/python3 with #/usr/bin/env python3 in the script. xilinx is running in docker with default python path to its virtual environment

.github/workflows/build-xilinx-platform.yaml
Lewluu: hardware is needed for the build. new_hardware is generated during the build. but i'm not 100% familiar with the full buildchain behavior.

tools/scripts/build_projects.py
Lewluu: @liviutomoiaga

tools/scripts/build_projects.py
Lewluu: @liviutomoiaga

tools/scripts/build_projects.py
Lewluu: @liviutomoiaga - he is more familiar with the blacklist implementation

.github/scripts/setup-platform-env.sh
SRaus: I haven't said anything about shared folder. It's just about keeping the files locally versus remote.

.github/workflows/build-xilinx-platform.yaml
SRaus: I think it's better to fully understand the role of `new_hardware` in the overall build flow before removing it. 
That will ensure the change remains a pure refactoring and that we don't accidentally losing functionality or introducing any unintended issues.

.github/workflows/build-xilinx-platform.yaml
```diff
@@ -0,0 +1,87 @@
...
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
+
+jobs:
+  build-xilinx-platform:
+    env:
+      CLOUDSMITH_API_KEY: ${{ secrets.CLOUDSMITH_API_KEY }}
+      TOOLS_REPO: ${{ secrets.TOOLS_REPO }}
+      BLACKLIST_URL: ${{ secrets.BLACKLIST_URL }}
+      HDL_BRANCH: 'main'
```
liviutomoiaga, line 59: we need to clarify  the if `main` is always needed or needs to be calculated

.github/workflows/build-light-platforms.yaml
```diff
@@ -0,0 +1,98 @@
...
+      - next_stable
+      - release
+      - '20*'
+    paths-ignore:
+      - '**/*.rst'
+      - '**/*.md'
+  push:
+    branches:
+      - main
+    paths-ignore:
```
liviutomoiaga, line 29: you might include .github/config/** as well. I do not think `config.cppcheck` or `patch_docy.patch` should trigger a build if those are changed, right ?

.github/workflows/build-xilinx-platform.yaml
```diff
@@ -0,0 +1,87 @@
+name: Build Xilinx platform
+
+on:
+  pull_request:
+    branches:
+      - main
+      - next_stable
+      - release
+      - '20*'
+    paths-ignore:
```
liviutomoiaga, line 10: you might include .github/config/** as well. I do not think `config.cppcheck` or `patch_docy.patch` should trigger a build if those are changed, right ?

.github/workflows/build-xilinx-platform.yaml
```diff
@@ -0,0 +1,87 @@
...
+      - '**/*.rst'
+      - '**/*.md'
+  push:
+    branches:
+      - main
+    paths-ignore:
```
liviutomoiaga, line 29: you might include .github/config/** as well. I do not think `config.cppcheck` or `patch_docy.patch` should trigger a build if those are changed, right ?

tools/scripts/build_projects.py
liviutomoiaga: I'll take care of it

tools/scripts/build_projects.py
liviutomoiaga: I'll take care of it

tools/scripts/build_projects.py
liviutomoiaga: I'll take care of it

.github/workflows/build-xilinx-platform.yaml
SRaus: Good catch! 
And it's not only about the blacklist, but also about the HDL binaries to use - from main or from different release branches.

The same for this CI - let's keep in mind that, in the end, it should be able to build multiple No-OS branches. 
But I think we can postpone this aspect until after the decision about release is taken (if there will be a release branch, or just a release tag in main etc)

.github/workflows/build-xilinx-platform.yaml
SRaus: in this case, there should be also ".gitignore", ".claude/**" be ignored?

tools/scripts/build_projects.py
```diff
@@ -164,26 +164,50 @@ def cmake_cache_source_mismatch(build_dir, noos):
...
-				.format(TOKEN, blacklist_url, log_file))
-	if err != 0 or (not os.path.isfile('blacklist.txt')):
-		log_err('Can not download blacklist file')
-		return blacklist
-	file = open('blacklist.txt', 'r')
-	for line in file.readlines():
-		project = line.split('#')[0].rstrip().replace('.', '_')
-		if project != '':
-			blacklist.append(project)
+	log('Fetching blacklist from %s' % blacklist_url)
```
liviutomoiaga, line 167: Reworked process_blacklist so it doesn't just give up when the branch-specific blacklist file isn't found. Now it tries blacklist_{BRANCH}.txt first, and if that 404s (e.g. on a feature branch that doesn't have its own blacklist), it falls back to blacklist_main.txt automatically. This way we always get blacklist filtering on every branch, not just on main and next_stable. If even the main fallback fails, it logs the error and proceeds without filtering same as before. Is it useful when you want to run an independent execution of Xilinx on a feature or test branch which do not have a blacklist in place.

liviutomoiaga: example:
 -> Blacklist for branch test/pico not found, falling back to main

SRaus: The blacklist should not even checked for non-Xilinx. 
For Xilinx projects, the blacklist file is used to signal somehow what hdl binaries are missing/failing, so there is no sense to build  their corresponding No-OS projects (hdl binary being used as input).

Lewluu:
Updates:
Updated some features on unit-tests and checkers, separated platform by workflows, created some minimal paths checkers to detected which platform to build and re-worked with @liviutomoiaga on blacklist features.
Xilinx builds may take less time on next builds since the cache exists in the runners build workspaces now.
For further builds optimizations, other more granulated features should be implemented or is already work-in progress by both developers and the devops.
Please check the logs and builds artifacts for Pico, Maxim and Xilinx from the CIs.
Thank you!

buha (dismissed):
from my POV, artefacts are published, cache is used for xilinx and stm32 builds, looks good, can be merged

liviutomoiaga:
Just a small clarification: STM32 and ADuCM will be migrated to GitHub Actions in a separate PR, as the new environment is still being prepared. This PR only covers the migration of the Pico, Maxim, and Xilinx platforms to GitHub Actions.

> from my POV, artefacts are published, cache is used for xilinx and stm32 builds, looks good, can be merged

ci/astyle_config
liviutomoiaga, line 1: where is this config now ?

buha:
yeah sorry, i was talking only about xilinx, i know stm32 is still on azure

rbolboac:
it looks like for pico platform uf2 files are not being built

ci/astyle_config
Lewluu: adding it

.github/scripts/documentation.sh
```diff
@@ -244,4 +259,8 @@ build_doxygen
 
 build_sphinx
 
-update_gh_pages
+if [[ "UPDATE_GH_PAGES" == 'true' ]]; then
```
liviutomoiaga: Missing $ in variable check. Should be  `"$UPDATE_GH_PAGES"`

.github/scripts/documentation.sh
```diff
@@ -26,15 +26,43 @@
 # NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE,
 # EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
-. ./ci/lib.sh
-
 COMMIT_RANGE="$1"
 
+parse_commit_range() {
```
liviutomoiaga, line 33: remove echo_green and echo_red if are not used anymore. those were defined in `ci/lib.sh`

.github/scripts/deploy-docs.sh
```diff
@@ -0,0 +1,50 @@
...
+mkdir -p ${DEPS_DIR}/doxygen ${DEPS_DIR}/doxygen-theme
+cd ${DEPS_DIR}
+
+# Download doxygen tools
+wget ${DOXYGEN_URL} >/dev/null
+tar --strip-components=1 -xvf doxygen-${VERSION}.src.tar.gz -C doxygen
+
+# Download doxygen theme
+wget ${DOXYGEN_THEME_URL} >/dev/null
+tar --strip-components=1 -xvf adi-harmonic-doxygen-theme.tar.gz -C doxygen
```
liviutomoiaga: Something feels odd here... you unzip in line 21 doxygen-${VERSION} under `doxygen` - then in line 25 you unzip doxygen-theme to same doxygen directory, right ? Is this correct ? 

maybe you can take another look over https://github.com/analogdevicesinc/no-OS/pull/3220/changes#diff-a56d4dad614a857b0fce97f12910fe0f196c4ced26bb21e217145c2c0c3ca35fL88-L95

.github/workflows/deploy-docs.yaml
```diff
@@ -0,0 +1,53 @@
...
```
liviutomoiaga: I think here you wanted to use `github.event_name` . `github.event` returns the full webhook payload, so comparing it to workflow_dispatch will always be false.

.github/scripts/documentation.sh
```diff
@@ -180,57 +208,44 @@ build_sphinx() {
...
-
-                        echo_green "Documetation updated!"
+        # Create .nojekyll file
+        touch ${TOP_DIR}/.nojekyll
+        CURRENT_COMMIT=$(git log -1 --pretty=%B)
+        
+        if [[ ${CURRENT_COMMIT:(-7)} != ${MAIN_COMMIT:0:7} ]]; then
+                git add --all .
+                git commit --allow-empty -m "Update documentation to ${MAIN_COMMIT:0:7}"
+                if [ -n "$GITHUB_DOC_TOKEN" ] ; then
```
liviutomoiaga, line 244: is `GITHUB_DOC_TOKEN` passed somehow to this script ? do I miss something here  ?

.github/scripts/deploy-docs.sh
Lewluu: yes. that's correct. two different archives are unzipped in the same folder. there is actually the "--strip-components=1" which skips one directory from archive, if that's confusing. it was tested and it works.

.github/scripts/deploy-docs.sh
```diff
@@ -0,0 +1,50 @@
+#!/bin/bash
+
```
liviutomoiaga, line 2: `set -e` to catch possible errors .

.github/workflows/build-common.yaml
liviutomoiaga: This still seems  problematic 🗡️

.github/workflows/build-xilinx-platform.yaml
```diff
@@ -0,0 +1,97 @@
...
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v5
+        with:
+          submodules: false
+          clean: false
+      - name: Building for platform Xilinx 
+        shell: sh
+        run: |
+          echo "Setup cloudsmith helper ..."
```
liviutomoiaga, line 75: In all 3 platform workflow files, you wget cloudsmith_helper.py. Not really necessary, but is it worth it having it under a composite action ?

.github/scripts/documentation.sh
Lewluu: thanks! missed some of them

.github/workflows/build-common.yaml
```diff
@@ -0,0 +1,78 @@
...
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v5
+        with:
+          fetch-depth: 50
+          clean: true
+      - name: Astyle check
+        shell: sh
+        run: |
+          ${{ github.workspace }}/.github/scripts/astyle.sh
```
liviutomoiaga, line 32: Compared with old azure pipeline, target branch is parsed to astyle.sh -> check [here](https://github.com/analogdevicesinc/no-OS/pull/3220/changes#diff-7915b9b726a397ae7ba6af7b9703633d21c031ebf21682f3ee7e6a4ec52837a5L24-L26)

.github/scripts/documentation.sh
Lewluu: that's in case you wish to publish in your name, using your token. by default it takes cseci
image

.github/scripts/astyle.sh
```diff
@@ -26,21 +26,46 @@
...
-set -e
 
-. ./ci/lib.sh
+set -e
 
 ASTYLE_EXT_LIST="${ASTYLE_EXT_LIST} .c .h"
-
 COMMIT_RANGE="$1"
 
+parse_commit_range() {
```
liviutomoiaga, line 49: If target branch is not passed, the parse_commit_range will be HEAD~1. In case a PR has more then 1 commit, this check becomes irrelevant.

liviutomoiaga (changes requested):
Just a few small tweaks from my side :D

.github/scripts/documentation.sh
liviutomoiaga: Oh ok! Great. Missed that config 👍 . Looks good

.github/scripts/deploy-docs.sh
liviutomoiaga: ok, looked confusing at first sight. as long as we do not have a direct reference to `doxygen-theme` i think we are safe here.

.github/workflows/build-xilinx-platform.yaml
Lewluu: until we have a an official centralized adi opensource repo for shared composite actions, repetitive steps like these may appear

rbolboac (dismissed):

.github/config/astyle_config
```diff
@@ -7,3 +7,4 @@
 --pad-header       # Add spaces between 'if', 'for', 'while' 
 --unpad-paren      # Remove spaces inside parentheses
 --pad-comma        # Add space after commas
+
```
SRaus: In the end, you'll see that the single change from this file (astyle_config) is just an extra line. 
Is this really needed?

.github/workflows/build-common.yaml
SRaus: The issue is not the extra line itself; a non-Unix EOF was introduced.

Image

SRaus (changes requested):

.github/config/astyle_config
Lewluu: that was an extra indeed. thanks!

.github/scripts/documentation.sh
```diff
@@ -180,57 +210,44 @@ build_sphinx() {
...
-                        fi
-
-                        echo_green "Documetation updated!"
+        # Create .nojekyll file
+        touch ${TOP_DIR}/.nojekyll
+        CURRENT_COMMIT=$(git log -1 --pretty=%B)
+        
+        if [[ ${CURRENT_COMMIT:(-7)} != ${MAIN_COMMIT:0:7} ]]; then
+                git add --all .
+                git commit --allow-empty -m "Update documentation to ${MAIN_COMMIT:0:7}"
```
SRaus: As an optimization - it's enough to change this line with 
`git diff --cached --quiet || git commit -m "Update documentation to ${MAIN_COMMIT:0:7}"` - which will not even create the commit if there is no file touched.
In the current implementation an empty commit will be created and pushed, even if there is no file changed.

liviutomoiaga (dismissed):
LGTM! Thanks for taking care of this migration 🎆

.github/workflows/build-xilinx-platform.yaml
SRaus: Please add a comment here, in code, where to explain how 'hardware' / 'new_hardware' folders are used. The comment got closed, but there is no explanation.

SRaus (dismissed):
Please collect all remaining small improvements / suggestions - just to not forgot about them - and we'll treat them in a separate PR.

buha (dismissed):

rbolboac (dismissed):

Lewluu:
Updates: rebased and added 'deps_cache' features from remove submodules changes

buha (approved):

SRaus (commented):
You are very close to the final shape of this PR :)

Just some final actions:
- I'm seeing some commits that contain just feedback implementation (or are very small) and they should be squashed - i.e. https://github.com/analogdevicesinc/no-OS/pull/3220/changes/e1c089293327d382122f345ba6200062f61e1069 just removes two empty lines. It should be squashed into another commit that touched astyle_config and build-common files. The same with commits https://github.com/analogdevicesinc/no-OS/pull/3220/changes/0cd1cc645f17e5aa2427db7f7fce44abd23e2f4e and https://github.com/analogdevicesinc/no-OS/pull/3220/changes/b847060d708a06b400223260ae0b54b19aa964a4 that have very small changes.
- You might want to add some comments in code to explain complex implementation (also squashed into implementation commits), as you said here: 

Image

- And don't forgot to collect a list with all remaining suggestions

SRaus (commented):
You are very close to the final shape of this PR :)

Just some final actions:
- I'm seeing some commits that contain just feedback implementation (or are very small) and they should be squashed - i.e. https://github.com/analogdevicesinc/no-OS/pull/3220/changes/e1c089293327d382122f345ba6200062f61e1069 just removes two empty lines. It should be squashed into another commit that touched astyle_config and build-common files. The same with commits https://github.com/analogdevicesinc/no-OS/pull/3220/changes/0cd1cc645f17e5aa2427db7f7fce44abd23e2f4e and https://github.com/analogdevicesinc/no-OS/pull/3220/changes/b847060d708a06b400223260ae0b54b19aa964a4 that have very small changes.
- You might want to add some comments in code to explain complex implementation (also squashed into implementation commits),  as you said here: 

Image

- And don't forgot to collect a list with all remaining suggestions

rbolboac (approved):

PR #3221, opened 2026-07-18
projects: adalm-lsmspg: Add support for max32655fthr

thorenscientific:
Both tinyiiod and curvetrace_examples.

## Pull Request Description

Add support for the MAX32655FTHR , typically used in no-OS hands-on workshops.

## PR Type
- [X] New feature (change that adds new functionality)

## PR Checklist
- [X] I have complied with the [Submission Checklist](http://analogdevicesinc.github.io/no-OS/contributing.html#submission-checklist)
- [X] I have build all projects affected by the changes in this PR
- [X] I have tested in hardware affected projects, at the relevant boards
- [X] I have updated the documentation (wiki pages, ReadMe etc), if applies

### DIFF
--- a/projects/adalm-lsmspg/src/common/common_data.h
+++ b/projects/adalm-lsmspg/src/common/common_data.h
@@ -34,7 +34,6 @@
 #define __COMMON_DATA_H__
 
 extern struct no_os_spi_init_param ad5592r_spi_ip;
-extern struct no_os_gpio_init_param ad5592r_spi_ss_ip;
 extern struct no_os_uart_init_param uart_ip;
 extern struct no_os_irq_init_param timer_irq_ip;
 extern struct no_os_timer_init_param timer_ip;
--- a/projects/adalm-lsmspg/boards/curvetrace_example/max32655fthr.conf
+++ b/projects/adalm-lsmspg/boards/curvetrace_example/max32655fthr.conf
@@ -1,0 +1,1 @@
+# max32655fthr uses standard UART (no USB UART)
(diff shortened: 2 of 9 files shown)


thorenscientific:
@nacuna-adi  can you take a peek at this? I started with your original MAX32665 target and "vibe coded" in support for MAX32655. There were quite a few subtleties - CS selection, and getting the heartbeat timer to initialize properly. Surprisingly the acual UART (as opposed to the MAX32666's USB UART) worked pretty much the first time. I've tested out all four builds - IIO/standalone curve tracer on both boards and they look good, but would be great to get confirmation from you as well.

@amiclaus your input always appreciated, and I believe you reviewed Neil's original project. Can you take a look at this for style, correctness, etc., and that it's set up to generate four configurations in the release - iio and curvetrace_example for each target board? One concern is the MSDK version - I believe there is still a bug with the I2C driver in the latest release, we had to use the June 2023 release for this. Which version is used for building Maxim targets here?

If it doesn't look too terrible I'll de-draft the PR. Thanks!

nielacuna:
Looks good to me @thorenscientific

projects/adalm-lsmspg/src/common/common_data.h
```diff
@@ -34,7 +34,9 @@
 #define __COMMON_DATA_H__
 
 extern struct no_os_spi_init_param ad5592r_spi_ip;
+#if (TARGET_NUM == 32665)
```
buha: board/target specific code should not go into src/common but in src/platform/maxim in this case

thorenscientific: Should be all set now, thanks! I rebuilt and tested all four target configurations.

amiclaus (approved):

buha (approved):

PR #3231, opened 2026-07-28
Selective builds proof of concept

buha:
This is more like a discussion starter not a serious PR at this point, the idea is that if a PR edits a particular driver under driver/adc/[name], only projects that include and build this driver should be built by CI. If a new driver is added to drivers/afe/Kconfig, build all projects that include drivers in drivers/afe because of that Kconfig change. If a project is edited, only build that project. And so on.

This is hooked into existing Azure CI but I don't expect this to be merged before we finish the Github CI migration so that's already one thing that needs to be changed.

Is the json filter okay or should we somehow implement it directly in no_os_builds.py ? 
Do you have ideas for other rules or ideas for implementing this in a simpler way ? 
Perhaps the test script should be hooked to unit tests really and run by CI always on some historical commits.

### DIFF
--- a/build_projects.yml
+++ b/build_projects.yml
@@ -46,6 +46,7 @@ jobs:
     persistCredentials: true
   - script: 'echo "Using Xilinx version: $(XILINX_VERSION)"'
     displayName: 'Show Xilinx version'
+  - template: ci/steps-select-builds.yml
   - script: |
       export BRANCH=$(branch)
       wget -P ./tools/scripts/ \
@@ -55,7 +56,8 @@ jobs:
         -log_dir $(LOG_DIR) \
         -builds_dir /no-OS_builds/builds \
         -platform xilinx \
-        -hdl_branch $(HDL_BRANCH)
+        -hdl_branch $(HDL_BRANCH) \
+        -combos_file $(COMBOS_FILE)
     env:
       CLOUDSMITH_API_KEY: $(CLOUDSMITH_API_KEY)
       TOKEN: $(TOKEN)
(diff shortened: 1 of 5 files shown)


SRaus:
> Do you have ideas for other rules or ideas for implementing this in a simpler way ?

Is the `make` command on the repo root level available?

On the HDL projects, we are comparing the change (either from PR or a push) with latest built workspace:
```
make -q
MAKE_QUESTION_CODE="$?"
echo "Make Question Code: $MAKE_QUESTION_CODE (0 - already up to date / 1 - re-make is required / 2 - error)"
```
will give us the list of projects affected by a change.

buha:
There is no way we can do this with make -n. no-OS clone does not even survive CI runs, so you don't have an already built cached version of whatever the base version was on which you apply the PR diffs and then run make -n. This is a dead end. Maybe there is a better way than what I attempt to do here but I'm still waiting for something better to come up.

SRaus:
> This is a dead end.

Let's try to not put "Dead End" stamp from start :)

>no-OS clone does not even survive CI runs

This shouldn't be considered a blockage / we can "help them to survive".  There is a similar mechanism in place for hdl binaries.

buha:
I dont think no-OS clone should survive CI runs, otherwise sooner or later we run into nasty cached version issues for no-OS files. I think it should always be cloned from scratch and the only thing that should remain in cache is the sdk generated bsp files to cut ci run time.

SRaus:
> I dont think no-OS clone should survive CI runs, otherwise sooner or later we run into nasty cached version issues for no-OS files. I think it should always be cloned from scratch and the only thing that should remain in cache is the sdk generated bsp files to cut ci run time.

If we are able to detect which projects needs to be re-built due to a specific change, we can keep a workspace to compare against it. This comparison should give us the list of affected projects and we can build them from scratch in a new clone.

PR #3253, opened 2026-08-06
various fixes when using cmake on windows

buha:
- tools/scripts/no_os_build.cmd — adds a CMD wrapper so the build script can be invoked on Windows without the py prefix, matching the Linux usage.

 - OpenOCD discovery on Windows — the pico toolchain now probes the xPack scripts layout (openocd/scripts) in addition to the standard one
   (share/openocd/scripts); the STM32 toolchain now finds openocd.exe inside CubeIDE plugins, gated by CMAKE_HOST_WIN32 so WSL builds never pick up a
   non-executable PE binary.

 - FindSTM32CubeIDE: fix CUBEIDE_DIR export — PARENT_SCOPE is a no-op outside a function; replaced with CACHE FORCE so the GCC and OpenOCD discovery in the STM32
   toolchain correctly see the CubeIDE install path.

 - Docs — remove the obsolete MSYS2/Makefile build_win.rst; add a common CMake ≥ 3.28 and Ninja prerequisites section with winget install commands; expand
   build_pico_cmake.rst with Windows-specific toolchain and OpenOCD install steps and drop the stale "Linux only" note.

## PR Type

## PR Checklist


buha:
- tested by me and @kister-jimenez , fixes an issue on stm32 where openocd is not picked up correctly when using --flash
- tested by me with rpi-pico builds, fixes a similar problem about openocd when using --flash

CiprianRegus (approved):

drivers/platform/stm32/toolchain.cmake
```diff
@@ -58,7 +58,18 @@ endif()
...
+    # cannot be executed there and would cause an 'Exec format error' at flash time.
+    if(CMAKE_HOST_WIN32)
+        set(_ocd_names openocd openocd.exe)
+    else()
+        set(_ocd_names openocd)
+    endif()
+    foreach(_ocd_name IN LISTS _ocd_names)
+        file(GLOB_RECURSE _found ${CUBEIDE_DIR}/plugins/*/${_ocd_name})
+        list(APPEND _cubeide_openocd ${_found})
+    endforeach()
```
CiprianRegus, line 72: Can we use `find_package()` for this instead? It should handle the executable extension difference automatically.

PR #3254, opened 2026-08-06
drivers projects: fix CodeQL-flagged defects

rbolboac:
## Pull Request Description

Resolve static-analysis findings reported by CodeQL across drivers and example projects. All fixes are behavior-preserving except where the previous behavior was itself a bug (sd.c, adin1110 ethertype print).

- sd-card: sd.c: guard against uint32_t underflow when response_len is 0 (compare response_len > 1 instead of response_len - 1 > 0), which previously passed a huge length to memset.

- Widen narrow loop indices compared against wider bounds so the loop cannot wrap: ad5706r, ad9517, ad9523, ad9528, max96792_csi, iio_adis, ade9113, adpd410x, stm32 dma/gpdma/spi/xspi, nmea_ubx, eval-adxl367z dummy_example.

- Match printf/snprintf format specifiers to their argument types (cast uint32_t/int32_t args to (unsigned) long for %l): ad9144, ad9523, nmea_ubx, iio_adis, ad413x, ad469x_evb, ad713x_fmcz, ad796x_fmcz, ad9656_fmc, aducm3029_flash_demo, cn0561, cn0565, pulsar-adc. ad413x dropped a stray argument; adin1110 printed a two-byte ethertype array with a single %02X - now prints both bytes.

- Cast an operand to the wider type before multiplying to avoid overflow before the widening conversion: iio_dac_demo, ad9545, adf4156, adf4157, adv7511 cf_hdmi.

- cn0565 eit: check sscanf return against the expected conversion count (== 1); the uint8_t result stored EOF as 255, so parse failures were treated as success.

## PR Type

## PR Checklist

### DIFF
--- a/drivers/dac/ad5706r/ad5706r.c
+++ b/drivers/dac/ad5706r/ad5706r.c
@@ -60,7 +60,8 @@ static int ad5706r_transfer_with_crc(struct ad5706r_dev *dev,
 	int32_t addr;
 	uint8_t cnt = 0;
 	uint8_t crc_cnt = 0;
-	uint8_t i, crc_seed;
+	uint32_t i;
+	uint8_t crc_seed;
 	uint8_t crc_buf[4] = {0};
 	uint8_t byte_cnt = 0;
 	int inc;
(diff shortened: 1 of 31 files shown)


amiclaus (commented):
lgtm, although build issues need to be checked if they are valid.

rbolboac:
change log: main rebase now that xilinx in building successfully on main, it should also build here

amiclaus (approved):

rbolboac:
change log: main rebase

rbolboac:
change log: main rebase

PR #3262, opened 2026-08-11
Dev/license rework part2

rbolboac:
## Pull Request Description

Continuation of https://github.com/analogdevicesinc/no-OS/pull/3261
Handles licensing issues for drivers/rf-transceiver 
When https://github.com/analogdevicesinc/no-OS/pull/3261 will be merged, 36db716a1717dc862633732987e4e32cf1bfde5a will be dropped (these changes are already available in https://github.com/analogdevicesinc/no-OS/pull/3261, but they were added here to be able to add new content on top of them)

## PR Type
- [ ] Bug fix (change that fixes an issue)
- [ ] New feature (change that adds new functionality)
- [ ] Breaking change (has dependencies in other repos or will cause CI to fail)

## PR Checklist
- [ ] I have followed the [Coding style guidelines](http://analogdevicesinc.github.io/no-OS/drivers_guide.html#coding-style)
- [ ] I have complied with the [Submission Checklist](http://analogdevicesinc.github.io/no-OS/contributing.html#submission-checklist)
- [x] I have performed a self-review of the changes
- [ ] I have commented my code, at least hard-to-understand parts
- [ ] I have build all projects affected by the changes in this PR
- [ ] I have tested in hardware affected projects, at the relevant boards
- [x] I have signed off all commits from this PR
- [ ] I have updated the documentation (wiki pages, ReadMe etc), if applies

PR #3263, opened 2026-08-11
Dev/license rework part3

### LINKED ISSUE #3245: LICENSE file is unhelpfully vague
The top-level LICENSE file is unhelpful, as it suggests that there may be non-permissive licenses scattered through the repo but does not help the user to locate them.

It would be more helpful to the user if the content were to be explicit:
* Which files (or directory trees) are under different licenses, and whether they are permissive.
* Whether those files are part of the linkable library, build/test infrastructure, documentation, or optional examples.

It would also help if the contribution guidelines were revised to require that any further contributions be under a permissive license, to avoid making the problem worse.

rbolboac:
## Pull Request Description

Continuation of https://github.com/analogdevicesinc/no-OS/pull/3262

## PR Type

## PR Checklist


drivers/adc/ad738x/iio_ad738x.c
```diff
@@ -4,11 +4,8 @@
 ********************************************************************************
 * Copyright (c) 2024-2026 Analog Devices, Inc.
 * Copyright (c) 2024 BayLibre, SAS.
-* All rights reserved.
 *
-* This software is proprietary to Analog Devices, Inc. and its licensors.
```
stephenkilbaneadi, line 9: For these files, where does BayLibre, SAS come in? What license applies to the BayLibre content?

(I'm sure we've had this discussion before, but I don't remember the details.)

I don't think it impacts the result - the ADI license applies to ADI content only - but it would be good to understand it.

drivers/adc/ad9081/api/adi_ad9081.h
```diff
@@ -1,3 +1,4 @@
+// SPDX-License-Identifier: GPL-2.0
```
stephenkilbaneadi: Why GPL-2.0 for AD9081?

drivers/adc/ad9083/ad9083_api/adi_ad9083.h
```diff
@@ -1,3 +1,4 @@
+/* SPDX-License-Identifier: LicenseRef-ADI-API-License */
```
stephenkilbaneadi, line 1: So AD9081 is GPL-2.0 and AD9083 is proprietary?

projects/ad9371/src/firmware/Mykonos_M3.h
```diff
@@ -1,3 +1,4 @@
+/* SPDX-License-Identifier: LicenseRef-ADI-AD937X-Firmware-CTSLA */
```
stephenkilbaneadi, line 1: Is this license file in this PR? I don't see it under LICENSES/proprietary.

stephenkilbaneadi (commented):
Generally looks good to me.

drivers/adc/ad9081/api/adi_ad9081.h
rbolboac: I actually searched for the correct API package license, I did not find it. What I found out was that the file sources were taken from the linux repo so I added the same license. Any advice related to these scenarios are welcomed.

drivers/adc/ad738x/iio_ad738x.c
rbolboac: Unfortunately, I do not have any expertise related to this. Any input is appreciated.

projects/ad9371/src/firmware/Mykonos_M3.h
rbolboac: indeed it is missing, I will add it.

rbolboac:
change log: added missing license refs:
LICENSES/proprietary/LicenseRef-ADI-AD937X-Firmware-CTSLA
LICENSES/proprietary/LicenseRef-ADI-ADRV9001-SLA
LICENSES/proprietary/LicenseRef-ADI-ADRV9025-SLA

drivers/adc/ad9083/ad9083_api/adi_ad9083.h
rbolboac: we can keep an open discussion for both, because there I could not find the package which was licensing the software (and my colleague said that the files were ported from linux -> GPL-2.0), but for ad9083 I could not find any such information so then I added the license based on the file copyright description.

drivers/adc/ad738x/iio_ad738x.c
stephenkilbaneadi: Okay, so some background.

First, we shouldn't change a third party's copyright assertions or license information.

If we're keeping the file's license as it is, but expressing it differently (via SPDX Identifier rather than having the license text in the header), and it's our license text that we're replacing, then that's okay.

If we're choosing to change the license of a file (e.g. from ADI's 5-clause license to a standard BSD-3-Clause), then we can do that for _our_ code, but we can't do it for anyone else's code.

So:
* If the file was entirely written by ADI, and no external third-parties have contributed to the file, then we can change _how_ express the license, and we can change the license _itself_ (because it's all our code).
* If it's our file, but someone else has contributed code to it, we can change our license header, but their code remains under the license they contributed it under - unless we contact them to get their permission to change it.
* If it's someone else's file and we've made changes, we can't change their license or their license text, we can only change the license of our contributions (unless the file is under a copyleft license, in which case our contributions have to match that license).

So I was asking in the context of BayLibre's copyright, but you've mentioned elsewhere that BayLibre made contributions to our file. So that would mean that we can change _how_ the license is expressed, but we can't change the license of the BayLibre contributions.

drivers/adc/ad9081/api/adi_ad9081.h
stephenkilbaneadi: That seems problematic to me, because it means that any application built for this platform using no-OS will have to be released under GPL-2.0, which will mean it can't be used with proprietary code. So I would suggest that we'd need to remove this code and replace it with a clean implementation that is not derived from Linux code.

However, that's an activity for a separate PR.

drivers/adc/ad9083/ad9083_api/adi_ad9083.h
stephenkilbaneadi: I feel your pain.

rbolboac:
change log:
-  BayLibre IIO drivers - commit dropped — files revert to their original ADI/BayLibre proprietary-worded headers
- ltc2312 (ltc2312.c/.h) - changed  to BSD-3 (new commit)
- jesd204 framework - changed to BSD-3 
- ad9081 API - changed to BSD-3
- removed unrelated changes (include guard and trailing newline which was added)
- license folder rework to follow part1 and part2 changes (I assume merging order will be part1->part2->part3->part4)+

rbolboac:
change log: main rebase

PR #3266, opened 2026-08-13
drivers: io-link: adiol100: add ADIOL100 driver and project

leaveyoustun:
## Pull Request Description

Add no-OS driver, project, and examples for the ADIOL100/ADIOL101
dual-channel IO-Link Master transceiver.

- Driver - SPI register access (direct + paged), L+/CQ configuration,
  framer control, FIFO messaging, cycle timer, ADC monitoring, LED control,
  CQ external FET protection, and per-channel IRQ callbacks.
- Basic example - Demonstrates a full IO-Link communication cycle
  without an external stack, manually driving STARTUP → PREOPERATE → OPERATE:
  DPP reads, ISDU parameter reads, and cyclic process data exchange.
- i-link example - Integrates the rt-labs i-link IO-Link master stack
  on FreeRTOS. Includes a custom application layer (iolink_app), a FreeRTOS
  OSAL, and an ADIOL100 port layer. Configures port A in SDCI auto mode
  and prints process data.
- ilink-glue - Shared integration code for the i-link stack: application
  layer, OSAL, stack build config. CMake fetch support via
  `resolve_library_source`.

## PR Type

## PR Checklist


leaveyoustun:
Changelog for CI error fixes:
- Added Sphinx doc stubs for driver and project (commit 2 and 4)
- Skipped adiol100 from the drivers Makefile compilation check in commit 6 (port layer requires external i-link headers)

drivers/io-link/adiol100/adiol100.c
```diff
@@ -0,0 +1,1486 @@
...
+		return 0;
+	}
+
+	total = fifo_lvl + 2;
+	if (total > ADIOL100_FIFO_MAX_LEN)
+		total = ADIOL100_FIFO_MAX_LEN;
+
+	buf[0] = (dev->chip_addr << 6) | (data_reg << 1) | ADIOL100_READ;
+	for (int i = 0; i < total; i++)
+		buf[1 + i] = 0x00;
```
rbolboac: no need to do this, just declare the buffer with {0}

drivers/io-link/adiol100/adiol100.c
```diff
@@ -0,0 +1,1486 @@
...
+	if (ret)
+		return ret;
+
+	if (buf[2] != fifo_lvl) {
+		adiol100_reset_rx_fifo(dev, ch);
+		*len = 0;
+		return -EIO;
+	}
+
+	payload = fifo_lvl;
```
rbolboac: not sure why payload is used instead of fifo_lvl directly

drivers/io-link/adiol100/adiol100.c
```diff
@@ -0,0 +1,1486 @@
...
+#include "no_os_spi.h"
+#include <errno.h>
+#include <stdint.h>
+
+/**
+ * @brief Read RevisionID and DeviceID to verify the chip is present.
+ * @param dev - The device structure.
+ * @return 0 in case of success, -ENODEV if IDs read back as zero.
+ */
+int adiol100_verify_chip(struct adiol100_dev *dev)
```
rbolboac: i would split this in 2, for each reading, also rename to read_revid, read_deviceid, because no actual verification is performed. I suppose currently we do not know the values but rev_id and dev_id should be checked against expected values

drivers/io-link/adiol100/adiol100.c
```diff
@@ -0,0 +1,1486 @@
...
+ * @return 0 in case of success, negative error code otherwise.
+ */
+int adiol100_setup_clock(struct adiol100_dev *dev,
+			 struct adiol100_init_param *ip)
+{
+	uint16_t val;
+
+	val = ADIOL100_CLKOEN |
+	      no_os_field_prep(ADIOL100_CLKDIV_MSK, ip->clk_div) |
+	      no_os_field_prep(ADIOL100_CLOCKCFG_MSK, ip->clock_src);
```
rbolboac: ip should be checked against null before trying to access its members

drivers/io-link/adiol100/adiol100.c
```diff
@@ -0,0 +1,1486 @@
...
+
+	return adiol100_write(dev, ADIOL100_REG_CLOCKCFG, val);
+}
+
+static void adiol100_irqg_isr(void *arg)
+{
+	struct adiol100_dev *dev = arg;
+
+	if (dev->irq_cb_g)
```
rbolboac: dev should be checked for null before accessing its members, make sure to check all similar behaviour. I will not leave a comment in the rest of the instances

drivers/io-link/adiol100/adiol100.c
```diff
@@ -0,0 +1,1486 @@
...
+
+	if (dev->irq_gpio_g)
+		no_os_gpio_remove(dev->irq_gpio_g);
+
+	if (dev->irq_ctrl)
+		no_os_irq_ctrl_remove(dev->irq_ctrl);
+
+	ret = no_os_spi_remove(dev->spi_desc);
+	if (ret)
+		return -EINVAL;
```
rbolboac: can return ret

drivers/io-link/adiol100/adiol100.c
leaveyoustun: Will change!

leaveyoustun:
Changelog v2

HW testing fixes (ADIOL100EVKIT + MAX78000FTHR):
- parameters.c/h: GPIO_EXTRA non-NULL (VDDIOH), IRQA defines (P0.19), IRQ_OPS
- ilink_example.c: moved all init inside FreeRTOS app_task (PendSV before scheduler start)
- adiol100_ilink_pl.c: removed mutex lock/unlock from callbacks (deadlock with iolink_pl.c wrappers)
- adiol100_ilink_pl.c: added adiol100_config_lp() in set_mode SDCI case (L+ power enable)
- adiol100.h/c: added nvic_ip/nvic_irq_id to adiol100_init_param, NVIC init inside adiol100_init()
- adiol100_ilink_pl.c/h: stripped IRQ setup from port layer, uses adiol100_register_irq_callback instead
- iolink_app.c: COMLOST guard - skip stale NO_DEV from initial probe (while in RUNNING state); smi_cb resets port state on real disconnect
- common_data.c: IRQ params wired into adiol100_ip
- projects/Kconfig: added source "projects/adiol100/Kconfig"

Driver cleanup:
- ADIOL100_FIFO_MAX_LEN corrected to 67 (MsgID + RxBytes + TxBytes + 64-byte max M-sequence)
- read_msg: removed FIFO_MAX_LEN clamp, init buf with = {0}, inlined total as fifo_lvl + 3 with comment, removed payload variable
- adiol100_remove: return ret instead of -EINVAL
- Added !dev NULL checks, added !ip NULL checks in adiol100_init and adiol100_setup_clock
- Split adiol100_verify_chip into adiol100_read_revid / adiol100_read_devid, verification in adiol100_init against ADIOL100_DEVICE_ID (0x0100) and ADIOL101_DEVICE_ID (0x0101)

Examples & README:
- Removed \r from all prints in both examples
- Added phase prints to basic example (configuring transceiver, sending EstCom, reading DPP, starting cyclic)
- README: replaced ADRSEL table with EVKIT jumper configuration (J4, J5, JMP1, J10)
- README: updated connections table (IRQA = P0.19, IRQB/IRQG = Not used)
- README: added example output for the basic variant
- README: fixed i-link key components (they were referencing older files)

drivers/io-link/adiol100/adiol100.c
```diff
@@ -0,0 +1,1643 @@
...
+ * @return 0 in case of success, negative error code otherwise.
+ */
+int adiol100_read(struct adiol100_dev *dev, uint16_t reg, uint16_t *value)
+{
+	uint8_t buf[4];
+	uint8_t len;
+	uint8_t paged;
+	int ret;
+
+	if (!dev)
```
rbolboac, line 240: value should be check as well

drivers/io-link/adiol100/adiol100.c
```diff
@@ -0,0 +1,1643 @@
...
+	return adiol100_read(dev, ADIOL100_REG_DEVICEID, dev_id);
+}
+
+/**
+ * @brief Configure the clock source, divider, and enable clock output.
+ * @param dev - The device structure.
+ * @param ip  - Initialization parameters containing clock_src and clk_div.
+ * @return 0 in case of success, negative error code otherwise.
+ */
+int adiol100_setup_clock(struct adiol100_dev *dev,
```
rbolboac, line 61: I would not give the entire init param as parameter, only the dields needed, clk_div and clock_src

drivers/io-link/adiol100/adiol100.c
```diff
@@ -0,0 +1,1643 @@
...
+		return -EINVAL;
+
+	val = ADIOL100_CLKOEN |
+	      no_os_field_prep(ADIOL100_CLKDIV_MSK, ip->clk_div) |
+	      no_os_field_prep(ADIOL100_CLOCKCFG_MSK, ip->clock_src);
+
+	return adiol100_write(dev, ADIOL100_REG_CLOCKCFG, val);
+}
+
+static void adiol100_irqg_isr(void *arg)
```
rbolboac: add doxygen doc for all APIs which are missing one

drivers/io-link/adiol100/adiol100.c
```diff
@@ -0,0 +1,1643 @@
...
+						  ip->irq_gpio_g->number,
+						  NO_OS_IRQ_EDGE_FALLING);
+		if (ret)
+			goto err_gpio_g;
+
+		ret = no_os_irq_enable(descriptor->irq_ctrl,
+				       ip->irq_gpio_g->number);
+		if (ret)
+			goto err_gpio_g;
+	}
```
rbolboac: it looks like you can create a helper function instead of repeating the some sequence of code 3 times.

drivers/io-link/adiol100/adiol100.c
```diff
@@ -0,0 +1,1643 @@
...
+		      enum adiol100_keep_msg keep)
+{
+	uint8_t buf[ADIOL100_FIFO_MAX_LEN + 1];	// +1 for SPI command byte
+	uint8_t write_len = 3 + txbytes;		// MsgID + RxBytes + TxBytes + payload
+	uint8_t read_len = 2 + rxbytes;			// MsgID + RxBytesAct + response
+	uint16_t data_reg;
+	uint16_t fc1_reg;
+	int ret;
+
+	if (!dev)
```
rbolboac, line 351: data should be checked as well

drivers/io-link/adiol100/adiol100.c
```diff
@@ -0,0 +1,1643 @@
...
+	ret = adiol100_set_burst_len(dev, ch, write_len, read_len);
+	if (ret)
+		return ret;
+
+	buf[0] = (dev->chip_addr << 6) | (data_reg << 1) | ADIOL100_WRITE;
+	buf[1] = dev->msg_id++;
+	buf[2] = rxbytes;
+	buf[3] = txbytes;
+	for (int i = 0; i < txbytes; i++)
+		buf[4 + i] = data[i];
```
rbolboac: consider using memcpy

drivers/io-link/adiol100/adiol100.c
```diff
@@ -0,0 +1,1643 @@
...
+	if (!dev)
+		return -ENODEV;
+
+	if (ch == ADIOL100_CH_A)
+		reg = ADIOL100_REG_FRAMCTRL2_A;
+	else
+		reg = ADIOL100_REG_FRAMCTRL2_B;
+
+	return adiol100_update(dev, reg, ADIOL100_TXFIFORSTN, ADIOL100_TXFIFORSTN);
+}
```
rbolboac, line 572: this could be written in less lines, it's up to you if you want to make the optimization:
```
if (ch == ADIOL100_CH_A)
       return adiol100_update(dev, ADIOL100_REG_FRAMCTRL2_A, ADIOL100_TXFIFORSTN, ADIOL100_TXFIFORSTN);

return adiol100_update(dev, ADIOL100_REG_FRAMCTRL2_B, ADIOL100_TXFIFORSTN, ADIOL100_TXFIFORSTN);

File: drivers/io-link/adiol100/adiol100.c [hunk 14]
```diff
@@ -0,0 +1,1643 @@
...
+	if (ch == ADIOL100_CH_A)
+		reg = ADIOL100_REG_FRAMCTRL2_A;
+	else
+		reg = ADIOL100_REG_FRAMCTRL2_B;
+
+	ret = adiol100_write(dev, reg, ADIOL100_CHANRST);
+	if (ret)
+		return ret;
+
+	no_os_mdelay(5);
```
rbolboac: in general we comment such delays with info taken from datasheet

drivers/io-link/adiol100/adiol100.c
```diff
@@ -0,0 +1,1643 @@
...
+
+	if (!dev)
+		return -ENODEV;
+
+	if (ch == ADIOL100_CH_A)
+		reg = ADIOL100_REG_CQCURLIM_A;
+	else
+		reg = ADIOL100_REG_CQCURLIM_B;
+
+	val |= no_os_field_prep(ADIOL100_CQVTHRH_MSK, tx_vthr_h);
```
rbolboac:
```
val |= no_os_field_prep(ADIOL100_CQVTHRH_MSK, tx_vthr_h) |
       no_os_field_prep(ADIOL100_CQVTHRL_MSK, tx_vthr_l) |
       no_os_field_prep(ADIOL100_RXVTHRH_MSK, rx_vthr_h) |
       no_os_field_prep(ADIOL100_RXVTHRL_MSK, rx_vthr_l) |
       no_os_field_prep(ADIOL100_CQCL_MSK,    current_limit) |
       no_os_field_prep(ADIOL100_CLBL_MSK,    blanking) |
       no_os_field_prep(ADIOL100_ARTTMO_MSK,  retry_tmo) |
       no_os_field_prep(ADIOL100_ARTEN,       retry_en);

File: drivers/io-link/adiol100/adiol100.c [hunk 16]
```diff
@@ -0,0 +1,1643 @@
...
+ */
+int adiol100_read_adc_chan(struct adiol100_dev *dev,
+			   enum adiol100_adc_chan source,
+			   enum adiol100_channel ch, uint16_t *value)
+{
+	uint16_t reg = (uint16_t)source;
+	uint16_t raw;
+	int ret;
+
+	if (!dev)
```
rbolboac, line 1317: value should be checkes

rbolboac (commented):
regarding to the driver part, just a few small remarks

libraries/ilink-glue/iolink_options.h
```diff
@@ -0,0 +1,96 @@
...
+ */
+#ifndef IOLINK_HW
+#define IOLINK_HW (IOLINK_HW_MAX14819)
+#endif
+
+#if IOLINK_HW == IOLINK_HW_MAX14819
+#define IOLINK_MAX14819_IRQ_GPIO
+//#define IOLINK_MAX14819_IRQ_SPI
+
+//#define IOLINK_MAX14819_RXRDY_GPIO
```
rbolboac: any reason why this is commented? any unused code should be deleted

rbolboac (commented):
one other small remark, the rest looks really good!

drivers/io-link/adiol100/adiol100.c
leaveyoustun: Will change

drivers/io-link/adiol100/adiol100.c
leaveyoustun: I personally find the original one more readable, but I don't have a strong opinion about it. Since multiple functions have similar if...else blocks for channel selection, I'm inclined to leave it like this.

drivers/io-link/adiol100/adiol100.c
leaveyoustun: I put that delay there to be defensive, but I tested without it and I doesn't seem needed, so I will remove it.

libraries/ilink-glue/iolink_options.h
leaveyoustun: This was taken from the original sample in the i-link repo, but since those defines aren't used anywhere I can delete them.

leaveyoustun:
Changelog v3

Driver (adiol100.c/h)
- Changed read_msg signature from uint8_t *len to uint8_t len, the number of response bytes is already known beforehand so there's no need for an output pointer, and it makes it clear that data must be a pre-allocated buffer
- Added NULL pointer checks for all pointer parameters across all public API functions
- Changed setup_clock to take clock_src and clk_div directly instead of the full init_param struct
- Replaced val |= field_prep(...) chains with single val = ... | ... | ... statements
- Replaced byte-copy for loops with memcpy in load_msg and read_msg
- Removed unnecessary 5ms delay from reset_channel (verified on HW)
- Updated config_lp_protection param comments to describe step sizes (matching config_cq_protection_ext style), and removed ~ from CQ ext comments since datasheet specifies exact values
- Added doxygen comments to the three static ISR callbacks
- Created adiol100_setup_gpio_irq helper to deduplicate the repeated IRQ GPIO setup in init
- Added adiol100_reset_leds function (clears sequences + resets modes), called from global_reset - LEDs were not turned off by a reset

i-link glue (libraries/ilink-glue/)
- Fixed licenses: OSAL files now include rt-labs BSD-3-Clause copyright; iolink glue/app files changed to GPL-3.0-only with rt-labs copyright
- Removed unused macros from iolink_options.h: LOG_ENABLE, WITH_MALLOC, IOLINK_NUM_DIAG_ENTRIES, and the IOLINK_HW_MAX14819 IRQ/RXRDY block
- Throttled PD console prints to 1 per second (was printing duplicates every cycle due to event loop timing)

i-link example / port layer
- Fixed licenses to GPL-3.0-only with rt-labs copyright
- Simplified get_data to match new read_msg API

Basic example
- Updated callers for new read_msg API, replaced byte-copy loop with memcpy

README
- Added GPL-3.0-only license note for the i-link combined binary

PR Description
- Removed the part saying that the i-link example was not tested on hardware

drivers/io-link/adiol100/adiol100_ilink_pl.c
```diff
@@ -0,0 +1,360 @@
...
+	uint16_t irq;
+
+	if (!drv->is_iolink[ch])
+		return;
+
+	adiol100_get_channel_irq(drv->dev, ch, &irq);
+
+	if (irq & ADIOL100_TXERRINT) {
+		if (drv->wurq_request[ch]) {
+			adiol100_get_fifo_status(drv->dev, ch, &irq);
```
rbolboac: is this a bug? it looks like irq is overwritten and then lines below it is checked against channel irq flags (overwritten values here) instead of fifo status flags

leaveyoustun: yes, thanks for catching this. I will modify!

drivers/io-link/adiol100/adiol100.c
```diff
@@ -0,0 +1,1702 @@
...
+
+	if (fifo_lvl > len) {
+		adiol100_reset_rx_fifo(dev, ch);
+		return -EIO;
+	}
+
+	buf[0] = (dev->chip_addr << 6) | (data_reg << 1) | ADIOL100_READ;
+
+	/* +3: SPI command byte + MsgID + RxBytesAct (not counted in fifo_lvl) */
+	ret = no_os_spi_write_and_read(dev->spi_desc, buf, fifo_lvl + 3);
```
rbolboac: if fifo_lvl = ADIOL100_RXFIFOLVL_MSK, then buf will be overwritten by 2, so buf should have a size of ADIOL100_RXFIFOLVL_MSK + 3

libraries/ilink-glue/app/iolink_app.c
```diff
@@ -0,0 +1,539 @@
...
+ * @param cfg - i-link stack master configuration. cb_smi, cb_pd and cb_arg
+ *              are set internally; the caller only needs to fill port_cfgs,
+ *              port_cnt, and thread priorities/stack sizes.
+ * @return 0 on success, -1 on failure.
+ */
+int iolink_app_init(struct iolink_app_master *ctx, iolink_m_cfg_t *cfg)
+{
+	uint8_t i;
+
+	memset(ctx, 0, sizeof(*ctx));
```
rbolboac, line 305: check pointers for null

drivers/io-link/adiol100/adiol100_ilink_pl.h
```diff
@@ -0,0 +1,45 @@
...
+#ifndef ADIOL100_ILINK_PL_H
+#define ADIOL100_ILINK_PL_H
+
+#include <stdint.h>
+#include <stdbool.h>
+#include "osal.h"
+#include "iolink_pl_hw_drv.h"
+#include "adiol100.h"
+
+#define ADIOL100_NUM_CHANNELS 2
```
rbolboac: already defined in adiol100.h

libraries/ilink-glue/app/iolink_app.c
```diff
@@ -0,0 +1,539 @@
...
+
+	for (i = 0; i < IOLINK_NUM_PORTS; i++) {
+		if (ctx->port[i].event != NULL)
+			port_cnt = i + 1;
+	}
+
+	while (1) {
+		uint32_t event_value;
+
+		if (os_event_wait(ctx->app_event, 0xFFFFFFFF,
```
rbolboac: OS_WAIT_FOREVER instead of 0xFFFFFFFF

later edit: actually I see this has a different meaning, did you check what  0xFFFFFFFF is doing in this case? is this the intended usage?

projects/adiol100/src/examples/ilink/ilink_example.c
```diff
@@ -0,0 +1,117 @@
...
+ * port A for IO-Link (SDCI auto mode), and runs the IO-Link application event
+ * loop. Process data from connected devices is printed to the console.
+ *
+ * @return 0 on success, negative error code otherwise.
+ */
+int ilink_example_main(void)
+{
+	pr_info("ADIOL100 i-link example\n");
+
+	xTaskCreate(
```
rbolboac: you can handle this in a smilar way: https://github.com/analogdevicesinc/no-OS/blob/main/projects/iio_demo_freeRTOS/src/platform/maxim/main.c#L97-L126

make sure you check the return and also delete the task in case of errors

libraries/ilink-glue/app/iolink_app.c
leaveyoustun: yes. it is so that we wake up on any event on any port

leaveyoustun:
Changelog v4

Commit 1: drivers: io-link: adiol100: add driver support

- adiol100.c:
  - Added txbytes/rxbytes bounds checks in load_msg()
  - read_msg buf size left as-is, FIFO only holds one frame per datasheet (single-burst SPI readout), TX resets RX before each exchange
  - Changed set_burst_len() params from int to uint8_t
  - Fixed remove() leak - spi_remove() failure no longer skips freeing dev
  - Removed all IRQ controller/NVIC/callback/ISR dispatch code (driver now only does gpio_get + direction_input for IRQ pins)
  - Added setup_gpio_input() helper with cleanup on failure
  - Renamed irq_gpio_g/a/b → gpio_irqg/irqa/irqb in init_param and dev structs
  - Removed irq_ip, nvic_ip, nvic_irq_id from init_param
  - Removed irq_ctrl, nvic_ctrl, irq_cb[], irq_cb_ctx[] from dev struct
- adiol100.h:
  - Renamed IRQ GPIO fields
  - Removed no_os_irq.h include
  - Removed IRQ-related fields from structs
- Kconfig:
  - Added select GPIO

Commit 2: drivers: io-link: adiol100: add README

- README.rst:
  - Rewrote "IRQ Callbacks" section to "IRQ GPIO Pins", documents new model where driver only sets up GPIOs, application handles IRQ controller/callbacks
  - Added updated code example using no_os_irq API
  - Documented 8 missing API functions (adiol100_read_revid, adiol100_read_devid, adiol100_setup_clock, adiol100_set_burst_len, adiol100_get_cycle_tmr, adiol100_get_fifo_status, adiol100_get_status, adiol100_reset_leds)
  - Added missing "fast blink" LED mode
  - Added LED reset mention to global_reset
  - Cleaned up example init struct (removed irq_gpio_g/a/b NULL assignments and irq_ip)

Commit 3: cmake: stm32: patch HAL_TIM_PeriodElapsedCallback as __weak

- stm32_patch_cubemx.cmake:
  - Added __attribute__((weak)) patch for HAL_TIM_PeriodElapsedCallback, when TIM6 is HAL timebase, CubeMX generates a strong definition that conflicts with stm32_irq.c

Commit 4: projects: adiol100: add project and basic example

- common_data.c:
  - Made gpio_irq_ip and nvic_ip non-static
  - Replaced hardcoded IRQA_PORT with GPIO_IRQ_CTRL_ID, added GPIO_IRQ_EXTRA field
  - Renamed irq_gpio_a_ip → gpio_irqa_ip
- common_data.h:
  - Added extern declarations for gpio_irq_ip and nvic_ip
  - Added no_os_irq.h include
- basic_example.c:
  - Fixed ISDU OOB read, replaced 3-byte IOL_ISDU_READ_CMD compound literal with zero-initialized od[IOL_MAX_OD] buffer, copying 3 ISDU bytes via memcpy using the macro (prevents reading past the array when preop_od_bytes > 3)
  - Added ISDU poll timeout check, if 100 polls exhaust without response, closes ISDU transaction and skips to next index
  - Removed redundant od_len variable (was alias for preop_od_bytes)
  - Added comments for all hardcoded delays (20ms per-frame, 3s L+ boot, 50ms poll interval)
- basic_example.h:
  - Fixed IOL_CKT_TYPE0 comment from "no OD, no PD" to "1 OD byte, no PD"
- maxim/main.c:
  - Added uart init return value check and uart_remove cleanup
  - NVIC setup only in CONFIG_ADIOL100_ILINK_EXAMPLE ifdef
- maxim/parameters.h:
  - Added NVIC_GPIO_IRQ, GPIO_IRQ_CTRL_ID, GPIO_IRQ_EXTRA defines
- stm32/main.c (new):
  - STM32 entry point with stm32_init(), get_input_clock, NVIC in ilink ifdef
- stm32/parameters.c (new):
  - STM32 extras (huart5, chip_select_port, gpio_irq port_nb)
- stm32/parameters.h (new):
  - STM32F469 pin mapping (SPI1, CS=PA15, IRQA=PG7/EXTI7)
- sdp-ck1z-basic.ioc (new):
  - CubeMX config with SysTick timebase
- boards/basic/sdp-ck1z.conf (new):
  - Points to sdp-ck1z-basic.ioc

Commit 5: projects: adiol100: add README

- README.rst:
  - Added STM32 SDP-CK1Z platform section with connections table
  - Added basic variant build commands

Commit 6: libraries: ilink-glue: add i-link stack glue

- adiol100_ilink_pl.c:
  - Added adiol100_config_lp(..., ADIOL100_LP_DIS, ADIOL100_LP_REV_DIS) to both DI/INACTIVE and DO cases in set_mode, L+ FET was left on when leaving SDCI mode
  - Extended PL cfg struct with irq_cb and irq_cb_ctx function pointer arrays
  - Fixed irq variable reuse in pl_handler — adiol100_get_fifo_status result now stored in rxstat instead of overwriting irq, which was checked against channel IRQ flags below
- adiol100_ilink_pl.h:
  - Added irq_cb and irq_cb_ctx to iolink_adiol100_cfg_t
- iolink_app.c:
  - Added os_mutex_create() NULL check with goto err_dev (was silently ignored, breaking SPI locking)
  - Added null pointer checks on all public and internal functions
  - Replaced 0xFFFFFFFF with APP_EVT_ALL in iolink_app_run event wait
- osal_freertos.c:
  - Fixed 32-bit overflow in os_get_current_time_us, changed to 64-bit intermediate: (uint64_t)xTaskGetTickCount() * 1000000ULL / configTICK_RATE_HZ, prevents non-monotonic timestamps after ~71 min

Commit 7: projects: adiol100: add i-link example

- ilink_example.c:
  - IRQ setup with generic ISR callback slots, iolink_ch_a_isr/iolink_ch_b_isr wrappers
  - GPIO IRQ controller init with goto-based cleanup
  - Added xTaskCreate return check and vTaskDelete on failure, matching iio_demo_freeRTOS pattern
- maxim/main.c:
  - Added NVIC setup in CONFIG_ADIOL100_ILINK_EXAMPLE ifdef with goto cleanup
- Kconfig:
  - Added select IRQ to ilink config
  - Added depends on FREERTOS
- maxim/FreeRTOSConfig.h (new):
  - MAX78000 FreeRTOS config
- stm32/FreeRTOSConfig.h (new):
  - STM32F469 FreeRTOS config (TIM6 timebase, configPRIO_BITS=4)
- sdp-ck1z-ilink.ioc (new):
  - CubeMX config with TIM6 timebase for FreeRTOS. FreeRTOS requires SysTick for its scheduler tick at the lowest interrupt priority, but HAL needs its timebase at a higher priority so that HAL_Delay() works inside peripheral interrupts without deadlocking. If both share SysTick, it can only have one priority, low (for FreeRTOS) freezes HAL_Delay() in higher-priority ISRs, high (for HAL) breaks RTOS context switching.
- boards/ilink/sdp-ck1z.conf (new):
  - Points to sdp-ck1z-ilink.ioc
- boards/ilink/max78000fthr.conf:
  - Added CONFIG_FREERTOS_CONF_PATH
- ilink.conf:
  - Removed CONFIG_FREERTOS_CONF_PATH (moved to per-board configs)

Commit 8: projects: adiol100: add i-link example README

- README.rst:
  - Added ilink variant build and --flash commands for both Maxim and STM32
  - Added CubeMX .ioc distinction (basic uses SysTick, ilink uses TIM6)
  - Added full repo-relative paths for ilink component files
  - Corrected "pulled in automatically via Kconfig" to "enabled automatically by the ilink variant config"

projects/adiol100/src/examples/basic/basic_example.c
```diff
@@ -0,0 +1,353 @@
...
+	pr_info("ISDU reads...\n");
+	{
+		static const int isdu_indices[] = {
+			IOL_ISDU_SERIALNUMBER, IOL_ISDU_VENDORNAME, IOL_ISDU_VENDORTEXT,
+			IOL_ISDU_PRODUCTNAME,  IOL_ISDU_PRODUCTID,  IOL_ISDU_PRODUCTTEXT,
+		};
+
+		for (int p = 0; p < (int)NO_OS_ARRAY_SIZE(isdu_indices); p++) {
+			uint8_t od[IOL_MAX_OD] = {0};
+			uint8_t result[IOL_MAX_ISDU_DATA];
```
rbolboac: result has a size of 32

leaveyoustun:
Changelog v5

Commit 1: drivers: io-link: adiol100: add driver support (additional)

- msg_id made per-channel: Changed uint8_t msg_id to uint8_t msg_id[ADIOL100_NUM_CHANNELS] in adiol100_dev. adiol100_load_msg now indexes by channel (dev->msg_id[ch]++).
- adiol100_set_cycle_tmr uses no_os_field_prep: Was passing raw value; now uses no_os_field_prep(ADIOL100_CYCLTMR_MSK, value) for consistency.
- Header @brief fixed: "Implementation of ADIOL100 Driver" → "ADIOL100 IO-Link transceiver driver definitions".
- adiol100_load_msg data parameter made const: uint8_t *data → const uint8_t *data.
- Removed no_os_delay.h from header: Only used in the .c file.

Commit 3: projects: adiol100: add project and basic example

- RX buffer overflow fix: Renamed IOL_MAX_FRAME to IOL_MAX_RX_FRAME and sized it to 2 * IOL_MAX_OD + 1 (65 bytes). TYPE_2 cyclic frames can carry up to 32 OD + 32 PD + 1 CKS bytes; the old IOL_MAX_FRAME (34 bytes) overflowed for devices with large PD.
- TX buffer sized explicitly: TX buffer in iol_send_frame now uses IOL_MAX_OD + 2 (MC + CKT + OD), instead of the shared IOL_MAX_FRAME.
- Unused label fix: Moved uart_remove: label inside the CONFIG_ADIOL100_ILINK_EXAMPLE #ifdef block (both STM32 and Maxim main.c). Previously unreferenced in basic-only builds, causing -Wunused-label warnings.
- EVKIT jumper table: Changed "MAX78000FTHR" to "an MCU board" since the table applies to any Pmod-connected board.
- Kconfig ordering fixed: Moved depends on FREERTOS before bool in ADIOL100_ILINK_EXAMPLE.

Commit 5: libraries: ilink-glue: add i-link stack glue

- VLA size capped: Changed arg_block_len from uint16_t to uint8_t in iolink_app_read_param and iolink_app_write_param, matching rt-labs' sample app. Prevents a caller-supplied length from blowing the FreeRTOS task stack via a multi-KB VLA.
- PDIn bounds check reverted: iolink_app_pdin uses the same unchecked pattern as the rt-labs sample app — pdin.len is always set by the stack within IOLINK_PD_MAX_SIZE.
- pl_handler dl_event NULL check: Added if (!drv->dl_event[ch]) return; guard in adiol100_pl_handler, matching the existing check in adiol100_pl_isr. Prevents NULL dereference if the handler runs before configure_event has stored the event handle.

Commit 6 (was 7): projects: adiol100: add i-link example

- SysTick shared timebase: Replaced TIM6-based HAL timebase with a custom SysTick_Handler that calls HAL_IncTick() unconditionally and xPortSysTickHandler() only after the FreeRTOS scheduler has started. This shares SysTick between HAL and FreeRTOS without a separate timer.
- Removed xPortSysTickHandler macro from FreeRTOSConfig.h — no longer renaming FreeRTOS's handler to SysTick_Handler.
- CubeMX .ioc switched to SysTick: Removed TIM6 timebase, TIM6 NVIC entry, and RegisterCallBack=TIM from sdp-ck1z-ilink.ioc. SysTick/SVC/PendSV handler generation remains disabled (FreeRTOS and the application provide them).
- Removed __weak HAL_TIM_PeriodElapsedCallback cmake patch: No longer needed — no TIM6 means no stm32f4xx_hal_timebase_tim.c generated, so no conflict with stm32_irq.c.
- Dropped cmake: stm32: patch HAL_TIM_PeriodElapsedCallback as __weak commit: Entire commit removed (was between driver README and project commits).
- NELEMENTS replaced with NO_OS_ARRAY_SIZE: Switched from rt-labs OSAL macro to no-OS's NO_OS_ARRAY_SIZE(). Added #include "no_os_util.h".
- SysTick handler comment in FreeRTOSConfig.h: Added comment noting SysTick_Handler is in main.c, making the split visible to maintainers.

Commit 7 (was 8): projects: adiol100: add i-link example README

- Updated .ioc description: Removed TIM6 timebase mention. Now describes the ilink .ioc as disabling SysTick/SVC/PendSV handler generation (since FreeRTOS and the application provide them).

leaveyoustun:
added missing #include "no_os_delay.h" to basic_example.c after removing it from the driver header

drivers/io-link/adiol100/adiol100.c
```diff
@@ -0,0 +1,1560 @@
...
+
+	return adiol100_send_msg(dev, ch);
+}
+
+/**
+ * @brief Read device response from RxFIFO, stripping the MsgID/RxBytesAct header.
+ * @param dev  - The device structure.
+ * @param ch   - Channel selection (A or B).
+ * @param data - Buffer to store the device response bytes (must be >= len).
+ * @param len  - Number of response bytes to read (same as rxbytes in load_msg).
```
rbolboac, line 449: what is the maximum length for len? should be well defined to fit buf[ADIOL100_RX_MAX_LEN + 1], len is never checked in the below implementation. also, the description says that is the number of response bytes to read, but the returned data contains fifo_lvl bytes, not len bytes

projects/adiol100/src/examples/ilink/ilink_example.c
```diff
@@ -0,0 +1,191 @@
...
+		pr_info("iolink_adiol100_init failed\n");
+		return;
+	}
+
+	drv = (iolink_adiol100_drv_t *)hw;
+
+	ret = no_os_irq_ctrl_init(&irq_ctrl, &gpio_irq_ip);
+	if (ret) {
+		pr_info("irq_ctrl_init failed: %d\n", ret);
+		return;
```
rbolboac, line 87: memory leak, should free what iolink_adiol100_init allocates

projects/adiol100/src/examples/basic/basic_example.c
```diff
@@ -0,0 +1,354 @@
...
+
+	pr_info("-----ADIOL100 Basic Example-----\n");
+
+	/* ════════════════════════════════════════════════════════════════════════
+	 *  STARTUP
+	 * ══════════════════════════════════════════════════════════════════════ */
+
+	/* Configure transceiver */
+	pr_info("Configuring transceiver...\n");
+	adiol100_config_lp(dev, ADIOL100_CH_A, ADIOL100_LP_EN, ADIOL100_LP_REV_EN);
```
rbolboac, line 113: you are discarding the return of the APIs, please use accordingly and check all API calls for errors

libraries/ilink-glue/app/iolink_app.c
```diff
@@ -0,0 +1,585 @@
...
+ * @param len - Number of bytes in buf.
+ * @return 0 on success, -1 on failure.
+ */
+int iolink_app_write_param(struct iolink_app_port *port, uint16_t index,
+			   uint8_t subindex, const uint8_t *buf, uint16_t len)
+{
+	if (!port || !buf)
+		return -EINVAL;
+
+	uint8_t arg_block_len = sizeof(arg_block_od_t) + len;
```
rbolboac, line 538: len is u16 and this block is u8. we should avoid this kind of truncation, if needed, return error if len is too big

liviutomoiaga:
Hey @leaveyoustun, I force-stopped the STM32 pipeline because the build was taking way too long and seemed stuck during compilation of --project adiol100 --variant basic --board sdp-ck1z. Could you test this config locally and check if STM32 IDE pops up a window? I'm thinking that might be what's hanging things up on some project configs. A full STM32 run should be done in under 10 mins.

leaveyoustun:
Changelog v6

Commit 1: drivers: io-link: adiol100: add driver support

- read_msg reworked: parameter renamed from len to data_len. Returns -EINVAL if data_len < fifo_lvl (buffer too small). No longer resets FIFO on mismatch, caller decides how to handle it.
- read_msg docstring updated: data_len is the buffer size. If larger than FIFO level, only available bytes are read. If smaller, returns -EINVAL.
- config_framer expanded: now configures both FramCtrl1 (INSCHKS, FRAMEREN) and FramCtrl2 (CKTErrInh, MsgRepeat, BDelay, DDelay, TSizeErrInh). New parameters: ckt_err_inh, msg_repeat, bdelay, ddelay, tsize_err_inh. DDelay formula documented in doxygen.
- New FramCtrl2 bit masks added: TSIZEERRINH (bit 14), CKTERRINH (bit 11), MSGREPEAT (bit 10), BDELAY_MSK (bits 9:8), DDELAY_MSK (bits 7:0).
- New enums: adiol100_ckt_err_inh, adiol100_msg_repeat, adiol100_bdelay, adiol100_tsize_err_inh.
- CQ protection defaults: adiol100_init now calls adiol100_config_cq_protection for both channels (300mA current limit, 1ms blanking, 5ms autoretry timeout, autoretry disabled, 2V thresholds).

Commit 2: drivers: io-link: adiol100: add README

- Framer Configuration section rewritten: describes FramCtrl1 + FramCtrl2 writes. Code example updated with full config_framer signature (CKTErrInh, MsgRepeat, BDelay, DDelay, TSizeErrInh parameters).
- Example code: removed unused `#include <stdio.h>`.
- Example code: SPI init updated to match common_data.c, uses SPI_BAUDRATE instead of hardcoded 10000000, added `.bit_order = NO_OS_SPI_BIT_ORDER_MSB_FIRST`.
- Example code: chip_addr changed from 0 to 0x01 to match EVKIT ADRSEL configuration.
- Example code: config_framer BDelay changed from BDELAY_0TBIT to BDELAY_3TBIT, ddelay from 0x03 to 0x08, matching basic_example.c.
- Example code: L+ delay changed from 3000ms to 300ms with updated comment referencing IO-Link spec device wake-up readiness.

Commit 3: projects: adiol100: add project and basic example

- Both .ioc files rebuilt from no-OS reference template (CubeMX 6.5.0, MxDb 6.0.50, TargetToolchain=STM32CubeIDE, AskForMigrate=false, UnderRoot=true). Fixes CI build failures caused by CubeMX version mismatch (was 6.18.1).
- Added no_os_delay.h include to basic_example.c (was relying on transitive include from driver header, which was removed in v5).
- basic_example.c: added return value checks for all function calls: config_lp, config_cq, config_framer, get_comrt, all DPP reads, MasterIdent, Preoperate command, ISDU send/poll/close, reset_rx_fifo, set cycle time, operate command, enable_cycle_timer, and operate frame. All fatal errors goto err which calls adiol100_remove. ISDU errors print and continue to next index.
- basic_example.c: config_framer call updated to new signature with FramCtrl2 parameters (CKTERR_SEND, MSGREP_DIS, BDELAY_3TBIT, ddelay=0x08, TSIZEERR_SEND).
- basic_example.c: L+ delay changed from 3000ms to 300ms with updated comment referencing IO-Link spec device wake-up readiness.

Commit 5: libraries: ilink-glue: add i-link stack glue

- read_param len parameter changed from uint16_t* to uint8_t*, matching rt-labs. Internal req_len also changed to uint8_t.
- write_param len parameter changed from uint16_t to uint8_t, matching rt-labs.
- arg_block_len changed from uint8_t to uint16_t in both read_param and write_param, matching rt-labs read and the SMI function signatures.
- PDIn bounds check reverted: iolink_app_pdin uses the same unchecked pattern as the rt-labs sample app.
- Added os_event_destroy to OSAL (osal.h declaration + osal_freertos.c implementation wrapping vEventGroupDelete).
- iolink_app_init: destroy app_event if iolink_m_init fails (was leaked).
- iolink_app_init: added NULL checks on os_event_create/os_mutex_create in port init loop, return -ENOMEM on failure.
- set_mode(SDCI): added 300ms delay after L+ enable (os_usleep(300000)) for device wake-up readiness (spec max 300ms).
- set_mode(SDCI): config_framer call updated to new signature with FramCtrl2 parameters (CKTERR_SEND, MSGREP_DIS, BDELAY_3TBIT, ddelay=0x08, TSIZEERR_SEND).

Commit 6: projects: adiol100: add i-link example

- SysTick priority set from code: added HAL_NVIC_SetPriority(SysTick_IRQn, 15, 0) in main.c after stm32_init(), guarded by CONFIG_ADIOL100_ILINK_EXAMPLE.
- ilink .ioc simplified: same as basic .ioc except handler generation disabled for SysTick/SVC/PendSV and SysTick priority 15. No more TIM6, PCC section, or CubeMX 6.18 artifacts.
- ilink_example.c: fixed resource leak, error path after iolink_adiol100_init now cleans up (os_mutex_destroy, adiol100_remove, no_os_free) via goto err_hw. Added no_os_alloc.h include.

leaveyoustun:
Changelog v7 (based on review comments from @rbolboac on #3377 

Commit 1: drivers: io-link: adiol100: add driver support

- Added `if (ip->chip_addr > ADIOL100_CHIP_ADDR_MAX)` check in `adiol100_init` with `goto err`. `ADIOL100_CHIP_ADDR_MAX` defined as 3 (2-bit SPI address field).
- Replaced magic `100` iteration count with `ADIOL100_ESTCOM_TIMEOUT_MS` (160ms). Comment documents worst case: 3 attempts × 50ms (max time for an EstCom sequence) = 150ms. Datasheet specifies a delay of 30ms per attempt, but IO-Link spec says the maximum time can be 50ms so 160ms to have a safety margin.
- Changed `@param bdelay` from "2–5 TBIT" to "0–3 TBIT" to match `enum adiol100_bdelay` values.
- Replaced all `(dev->chip_addr << 6) | (reg << 1) | READ/WRITE` with `ADIOL100_SPI_CMD(dev->chip_addr, reg, ADIOL100_READ/WRITE)`. Added `ADIOL100_ADDR_SHIFT` (6), `ADIOL100_REG_SHIFT` (1).
- Replaced hardcoded `3` and `4` with `ADIOL100_SPI_DIRECT_LEN` and `ADIOL100_SPI_PAGED_LEN`.
- Replaced `reg & 0xFF00` with `ADIOL100_REG_PAGE_MSK`, `reg & 0xFF` with `ADIOL100_REG_PAGE_OFFSET(reg)`.
- Replaced manual `value >> 8`, `value & 0xFF`, `(buf[x] << 8) | buf[y]` with `no_os_put_unaligned_be16` and `no_os_get_unaligned_be16`.
- Replaced `(read_len << 8) | write_len` with `no_os_field_prep(ADIOL100_BURST_RDLEN_MSK, read_len) | no_os_field_prep(ADIOL100_BURST_WRLEN_MSK, write_len)`.
- Replaced `+ 0x10` with `+ ADIOL100_ADC_CHB_OFFSET`.
- Replaced raw shifts (`<< 12`, `<< 8`, `<< 4`) with `no_os_field_prep(ADIOL100_LED_HI/LO_GREEN/RED_MSK, ...)`.
- Replaced `rseq_reg + 1` with `rseq_reg + ADIOL100_LED_GSEQ_REG_OFFSET`.

Commit 2: drivers: io-link: adiol100: add README

- Added rst `list-table` documenting all `adiol100_init_param` fields (spi_ip, chip_addr, clock_src, clk_div, gpio_irqg/a/b) with descriptions.
- Updated from 100ms to 160ms.

Commit 3: projects: adiol100: add project and basic example

- Replaced `(int16_t)((rx[op_od_bytes] << 8) | rx[op_od_bytes + 1])` with `(int16_t)no_os_get_unaligned_be16(&rx[op_od_bytes])`.
- Added `/* +2 for MC + CKT header */` on tx buffer declaration and on frame length argument, `/* +1 for CKS */` on OD buffer declaration.

PR #3267, opened 2026-08-17
drivers: health: Add MAXM86161 Optical PPG AFE Driver Support

jcroleda:
## Pull Request Description

The MAXM86161 is an ultra-low-power, integrated optical data acquisition device. The MAXM86161 supports 3 programmable high-current LED drivers and an analog front-end with a 19-bit ADC, Ambient-Light Cancellation, and a picket-fence detect and replace algorithm. This device is used for optical sensing applications such as heart rate detection and pulse oximetry.

This PR adds driver support for the MAXM86161 and adds example projects that show example application uses for this device.

Datasheet: [MAXM86161](https://www.analog.com/media/en/technical-documentation/data-sheets/maxm86161.pdf)

## PR Type
- [ ] Bug fix (change that fixes an issue)
- [x] New feature (change that adds new functionality)
- [ ] Breaking change (has dependencies in other repos or will cause CI to fail)

## PR Checklist
- [x] I have followed the [Coding style guidelines](http://analogdevicesinc.github.io/no-OS/drivers_guide.html#coding-style)
- [x] I have complied with the [Submission Checklist](http://analogdevicesinc.github.io/no-OS/contributing.html#submission-checklist)
- [x] I have performed a self-review of the changes
- [x] I have commented my code, at least hard-to-understand parts
- [x] I have build all projects affected by the changes in this PR
- [x] I have tested in hardware affected projects, at the relevant boards
- [x] I have signed off all commits from this PR
- [x] I have updated the documentation (wiki pages, ReadMe etc), if applies

PR #3270, opened 2026-08-18
drivers: health: Add MAXM86161 Optical PPG AFE Driver Support

jcroleda:
## Pull Request Description

The MAXM86161 is an ultra-low-power, integrated optical data acquisition device. The MAXM86161 supports 3 programmable high-current LED drivers and an analog front-end with a 19-bit ADC, Ambient-Light Cancellation, and a picket-fence detect and replace algorithm. This device is used for optical sensing applications such as heart rate detection and pulse oximetry.

This PR adds driver support for the MAXM86161 and adds example projects that show example application uses for this device.

Datasheet: [MAXM86161](https://www.analog.com/media/en/technical-documentation/data-sheets/maxm86161.pdf)

## PR Type

## PR Checklist


projects/maxm86161/CMakeLists.txt
```diff
@@ -0,0 +1,56 @@
...
+
+# Basic example
+if(CONFIG_MAXM86161_BASIC_EXAMPLE)
+	target_sources(maxm86161 PRIVATE
+		       $ {CMAKE_CURRENT_SOURCE_DIR}/src/examples/basic/basic_example.c
+		      )
+	target_include_directories(maxm86161 PUBLIC
+				   $ {CMAKE_CURRENT_SOURCE_DIR}/src/examples/basic
+				  )
+	endif()
```
edelweiseescala: I think there are unnecessary tabs here as well as the other if below

drivers/health/maxm86161/iio_maxm86161.c
```diff
@@ -0,0 +1,818 @@
...
+ * THIS SOFTWARE IS PROVIDED BY ANALOG DEVICES, INC. "AS IS" AND ANY EXPRESS OR
+ * IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF
+ * MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO
+ * EVENT SHALL ANALOG DEVICES, INC. BE LIABLE FOR ANY DIRECT, INDIRECT,
+ * INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA,
+ * OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF
+ * LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING
+ * NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE,
+ * EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
```
edelweiseescala: Following licensing changes, license should be this now

```suggestion
********************************************************************************
 * Copyright 2026(c) Analog Devices, Inc.
 *
 * SPDX-License-Identifier: BSD-3-Clause
*******************************************************************************/
````

File: drivers/health/maxm86161/iio_maxm86161.h [hunk 3]
```diff
@@ -0,0 +1,149 @@
...
+ *
+ * @param iio_desc   - Pointer that receives the address of the allocated IIO
+ *                     descriptor on success.
+ * @param init_param - Initialization parameters, including the driver init
+ *                     parameters and the FIFO watermark level.
+ * @return 0 on success, negative error code otherwise:
+ *         -EINVAL if arguments are invalid or IRQ is not enabled,
+ *         -ENOMEM if descriptor allocation fails, or a propagated error
+ *         from the driver/FIFO/IRQ setup.
+ */
```
edelweiseescala: doxygen comments are usually in the source file

edelweiseescala (commented):
Initial review in this PR https://github.com/analogdevicesinc/no-OS/pull/3267 already addressed Thank You!
Added some more comment, also the Maxim Build on the CI fails. Please also check that.

jcroleda:
Changelog V2:
- Fixed formatting issues of CMake files (also causing build errors)
- Moved doxygen comments to main code files
- Updated licensing to BSD-3-Clause 
- Removed bus_api abstraction from maxm86161.h (as only I2C is supported)

drivers/health/maxm86161/maxm86161.c
```diff
@@ -0,0 +1,2173 @@
...
+ * @param dev    - Device descriptor.
+ * @param ppg_sr - Sample rate selection code (valid range 0x00 to 0x13).
+ * @return 0 on success, -EINVAL for an out-of-range value, or another negative
+ *         error code on failure.
+ */
+int maxm86161_set_sample_rate(struct maxm86161_dev *dev, uint8_t ppg_sr)
+{
+	int ret;
+
+	if (!dev || ppg_sr > 0x13)
```
rbolboac: use macro instead of magic numbers

drivers/health/maxm86161/maxm86161.c
```diff
@@ -0,0 +1,2173 @@
...
+	uint8_t reg_val;
+
+	if (!dev || !overflow_count)
+		return -EINVAL;
+
+	ret = maxm86161_reg_read(dev, MAXM86161_REG_OVF_COUNTER, &reg_val);
+	if (ret)
+		return ret;
+
+	*overflow_count = reg_val & NO_OS_GENMASK(6, 0);
```
rbolboac: use a mask macro instead of magic number

drivers/health/maxm86161/maxm86161.c
```diff
@@ -0,0 +1,2173 @@
...
+		irq_cb.ctx = dev;
+		irq_cb.event = NO_OS_EVT_GPIO;
+		irq_cb.peripheral = NO_OS_GPIO_IRQ;
+		no_os_irq_unregister_callback(dev->irq_ctrl, dev->gpio_intb->number,
+					      &irq_cb);
+		no_os_irq_ctrl_remove(dev->irq_ctrl);
+	}
+
+	if (dev->gpio_intb) {
+		r = no_os_gpio_remove(dev->gpio_intb);
```
rbolboac: why is this r used instead of ret?

rbolboac (commented):
took a look at the driver for now, nothing too serious, mainly check the file again for all instances in which you use magic numbers and use macros instead

drivers/health/maxm86161/maxm86161.c
```diff
@@ -0,0 +1,2173 @@
...
+	return 0;
+
+err_irq:
+	if (dev->irq_ctrl)
+		no_os_irq_ctrl_remove(dev->irq_ctrl);
+err_gpio:
+	if (dev->gpio_intb)
+		no_os_gpio_remove(dev->gpio_intb);
+err_bus:
+	ret = no_os_i2c_remove(dev->i2c_desc);
```
RaduSabau1: assigning no_os_i2c_remove return value to ret would overwrite the error code stored through ret at this point and thus report a failure with (most probably, unless i2c_remove fails) a success code.

drivers/health/maxm86161/maxm86161.c
```diff
@@ -0,0 +1,2173 @@
...
+			.event = NO_OS_EVT_GPIO,
+			.peripheral = NO_OS_GPIO_IRQ,
+			.handle = init_param->irq_handle,
+		};
+
+		ret = no_os_irq_ctrl_init(&dev->irq_ctrl, init_param->irq_init);
+		if (ret)
+			goto err_gpio;
+
+		ret = no_os_irq_register_callback(dev->irq_ctrl, dev->gpio_intb->number,
```
RaduSabau1: this code path (irq_init) should also check for gpio_intb, otherwise dev->gpio_intb->number is dereferenced on a NULL gpio_intb reuslting in a NULL-pointer crash.

drivers/health/maxm86161/maxm86161.c
```diff
@@ -0,0 +1,2173 @@
...
+ */
+int maxm86161_set_led_pulse_amplitude(struct maxm86161_dev *dev,
+				      enum maxm86161_led_num led_num, uint8_t amplitude)
+{
+	uint8_t reg_addr;
+
+	if (!dev || led_num == 0 || led_num > MAXM86161_LED_NUM_PILOT_GREEN)
+		return -EINVAL;
+
+	reg_addr = MAXM86161_REG_LED1_PA + (led_num - 1);
```
RaduSabau1: if led_num is 4 (PILOT_GREEN) the computed register address is 0x26 which seems like a reserved address instead.

drivers/health/maxm86161/maxm86161.c
```diff
@@ -0,0 +1,2173 @@
...
+/**
+ * @brief Burst-read multiple bytes starting at a register address.
+ * @param dev      - Device descriptor.
+ * @param reg_addr - Starting register address for the burst read.
+ * @param reg_data - Buffer where the read bytes are stored.
+ * @param count    - Number of bytes to read.
+ * @return 0 on success, negative error code otherwise.
+ */
+int maxm86161_burst_reg_read(struct maxm86161_dev *dev, uint8_t reg_addr,
+			     uint8_t *reg_data, uint16_t count)
```
RaduSabau1, line 99: have you tested this with a near-full FIFO read? the reads are chunked which is correct. however I see the destination pointer between chunks never advances so a near-full FIFO read could corrupt the buffer. please check on this.

drivers/health/maxm86161/maxm86161.c
```diff
@@ -0,0 +1,2173 @@
...
+		return -EINVAL;
+
+	buf = no_os_calloc(1, count + 1);
+	if (!buf)
+		return -ENOMEM;
+
+	buf[0] = reg_addr;
+	memcpy(&buf[1], reg_data, count);
+
+	ret = no_os_i2c_write(dev->i2c_desc, buf, count + 1, 1);
```
RaduSabau1, line 145: the truncation of count is present here too, although no current call site exposes this.

projects/maxm86161/src/platform/maxim/main.c
```diff
@@ -0,0 +1,52 @@
...
+		goto remove_irq;
+
+	no_os_uart_stdio(uart_desc);
+
+	no_os_mdelay(100);
+	ret = example_main();
+	if (ret)
+		goto remove_uart;
+
+	no_os_uart_remove(uart_desc);
```
RaduSabau1: every successful run seems to free this uart_desc twice. no_os_uart_remove indeed checks for dev so an actual double-free wouldn't occur though the code itself looks wrong this way, this line could perhaps be removed and the if (ret) goto remove_uart; nevertheless since they become redundant in this case.

projects/maxm86161/src/examples/basic/basic_example.c
```diff
@@ -0,0 +1,261 @@
...
+	}
+
+	ret = maxm86161_set_led_cfg(dev, MAXM86161_LED_NUM_IR,
+				    MAXM86161_LED_RGE_124MA, 6, 0xFF);
+	if (ret) {
+		pr_err("Failed to configure IR LED: %d\n", ret);
+		goto cleanup;
+	}
+
+	ret = maxm86161_set_led_cfg(dev, MAXM86161_LED_SRC_PILOT_GREEN,
```
RaduSabau1: this one seems guaranteed to return -EINVAL. LED_SRC_PILOT_GREEN(8) is later used as led_num when calling set_led_range which checks for led_num > LED_NUM_RED(3) and returns -EINVAL if the condition (which is) is met.

projects/maxm86161/src/examples/iio_example/iio_example.c
```diff
@@ -0,0 +1,149 @@
...
+/*
+ * Set to 1 to run an in-code drain self-test before the IIO app starts. It
+ * exercises the same ISR -> circular buffer -> decode path that submit() uses,
+ * printing decoded tag/data over the UART stdio. Useful under GDB when no
+ * libiio host is available. Set back to 0 for normal operation.
+ */
+#define MAXM86161_SELFTEST 1
+
+#if MAXM86161_SELFTEST
+static void maxm86161_selftest(struct maxm86161_iio_desc *desc)
```
RaduSabau1: where is this called in example_main()?

drivers/health/maxm86161/iio_maxm86161.c
```diff
@@ -0,0 +1,848 @@
...
+		return maxm86161_set_picket_fence(dev, &pf);
+	}
+
+	case MAXM86161_IIO_SHUTDOWN:
+		return maxm86161_set_shutdown(dev, (bool)val);
+
+	case MAXM86161_IIO_LP_MODE:
+		return maxm86161_set_low_power_mode(dev, (bool)val);
+
+	case MAXM86161_IIO_BURST_EN:
```
RaduSabau1, line 486: seems like writing burst_en forcefully rewrites the rate to 8Hz and modifying the rate also forcefully enables the burst. a way of avoiding this (and making the two attributes independent) would be to have separate helper function for each e.g. set_burst_mode getting split in set_burst_enable and set_burst_rate or something simillar.

drivers/health/maxm86161/iio_maxm86161.c
```diff
@@ -0,0 +1,848 @@
...
+
+	desc->iio_dev = &maxm86161_iio_device;
+	*iio_desc = desc;
+
+	ret = maxm86161_set_shutdown(desc->drv_dev, true);
+	if (ret) {
+		goto err_remove_mutex;
+	}
+
+	return 0;
```
RaduSabau1: not only that this is line is duplicated, but *iio_desc =desc; which has a check itself is happening after the iio_desc assignment

RaduSabau1 (commented):
some comments from my side

projects/maxm86161/src/examples/iio_example/iio_example.c
jcroleda: This was intended mainly for debugging purposes with the GPIO IRQ during development. I will remove this from the example.

jcroleda:
Changelog V3:
- Updated magic numbers with defined values
- Updated mask macros with no-OS utils where applicable
- removed seperate return value storage for gpio_remove
- added a seperate return for i2c_remove error code
- removed redundant removes and returns (i2c, uart)
- Updated burst read/write to match buffer address after each chunk
- updated set_led_cfg to avoid setting led_range when PILOT_GREEN and helpers point to the proper PILOT_GREEN address (0x29)
- removed MAXM86161_SELFTEST from the iio_example.c
- added burst_en and burst_rate helpers to seperate the two operations in iio_attrs
- moved *iio_desc = desc; to end of init, after shutdown = true

drivers/health/maxm86161/maxm86161.c
```diff
@@ -0,0 +1,2225 @@
...
+
+		ret = no_os_i2c_write(dev->i2c_desc, buf, chunk_len + 1, 1);
+		if (ret)
+			return ret;
+
+		offset += chunk_len;
+		buf += chunk_len;
+	}
+
+	no_os_free(buf);
```
rbolboac: buf += chunk_len actually changes where buf is pointing, so freeing it after is corrupting the memory.
what is the reason behind using dynamic memory allocation? what is the maximum size of buf? I would either use stack allocation or create a buf in the dev structure to be used. dynamic allocation has a high time cost so there's a this disadvantage of using it

drivers/health/maxm86161/iio_maxm86161.c
```diff
@@ -0,0 +1,846 @@
...
+	if (!iio_dev || !iio_dev->drv_dev)
+		return;
+
+	if (!iio_dev->data_capture)
+		return;
+
+	ret = maxm86161_reg_read(iio_dev->drv_dev, MAXM86161_REG_INT_STATUS_1,
+				 &status1_data);
+	if (ret)
+		return;
```
rbolboac: this is clear on read, and it has already been read in this context in maxm86161_irq_handler(), is this working as expected?

drivers/health/maxm86161/maxm86161.c
```diff
@@ -0,0 +1,2225 @@
...
+
+	if (!dev || led_num == 0 || led_num > MAXM86161_LED_NUM_PILOT_GREEN)
+		return -EINVAL;
+
+
+	reg_addr = (led_num == MAXM86161_LED_NUM_PILOT_GREEN) ?
+		   MAXM86161_REG_LED_PILOT_PA :
+		   MAXM86161_REG_LED1_PA + (led_num - 1);
+
+	dev->led_config.pulse_amplitude[led_num - 1] = amplitude;
```
rbolboac: maximum value of led_num is 4 in your example, so you will write to pulse_amplitude[3], but pulse_amplitude has 3 elements only

drivers/health/maxm86161/maxm86161.c
```diff
@@ -0,0 +1,2225 @@
...
+	uint8_t reg_val;
+
+	if (!dev || !complete || !oor)
+		return -EINVAL;
+
+	ret = maxm86161_reg_read(dev, MAXM86161_REG_DAC_CAL_EN, &reg_val);
+	if (ret)
+		return ret;
+
+	*complete = !!no_os_field_prep(MAXM86161_CAL_DAC_COMPLETE, reg_val);
```
rbolboac: no_os_field_get instead? also below?

drivers/health/maxm86161/maxm86161.c
```diff
@@ -0,0 +1,2225 @@
...
+
+/**
+ * @brief Enable or disable burst sampling mode and set its rate.
+ *
+ * Updates the BURST_EN bit and, on success, the BURST_RATE field of PPG_CFG_3.
+ * @param dev    - Device descriptor.
+ * @param enable - true to enable burst mode, false to disable it.
+ * @return 0 on success, -EINVAL for a negative error code on failure.
+ */
+int maxm86161_set_burst_enable(struct maxm86161_dev *dev, bool enable)
```
rbolboac, line 1759: never declared in .h

drivers/health/maxm86161/maxm86161.c
```diff
@@ -0,0 +1,2225 @@
...
+/**
+ * @brief Enable or disable burst sampling mode and set its rate.
+ *
+ * Updates the BURST_EN bit and, on success, the BURST_RATE field of PPG_CFG_3.
+ * @param dev    - Device descriptor.
+ * @param rate   - Burst sampling rate selection.
+ * @return 0 on success, -EINVAL for an out-of-range rate, or another negative
+ *         error code on failure.
+ */
+int maxm86161_set_burst_rate(struct maxm86161_dev *dev,
```
rbolboac, line 1774: never declared in .h

drivers/health/maxm86161/maxm86161.c
```diff
@@ -0,0 +1,2225 @@
...
+
+	buf[0] = reg_addr;
+	memcpy(&buf[1], reg_data, count);
+
+	for (int offset = 0; offset < count;) {
+		uint32_t chunk_len = no_os_min(count - offset, MAXM86161_FIFO_BURST_RD_CAP);
+
+		ret = no_os_i2c_write(dev->i2c_desc, buf, chunk_len + 1, 1);
+		if (ret)
+			return ret;
```
rbolboac: memory leak

projects/maxm86161/src/platform/maxim/main.c
```diff
@@ -0,0 +1,48 @@
...
+	if (ret)
+		goto remove_irq;
+
+	ret = no_os_uart_init(&uart_desc, &maxm86161_uart_ip);
+	if (ret)
+		goto remove_irq;
+
+	no_os_uart_stdio(uart_desc);
+
+	no_os_mdelay(100);
```
rbolboac: any reason for this delay?

rbolboac (commented):
some other remarks, I am not sure about attribute_test_example, I will come with some feedback regarding this soon

drivers/health/maxm86161/maxm86161.c
jcroleda: Will change this to follow a constant size according to the max size used by the only burst write user (SHA challenge write = 20 bytes). As a result the burst size will no longer be an issue for the write op.

drivers/health/maxm86161/iio_maxm86161.c
jcroleda: From my tests it had worked but due to the interrupt set behavior (A_FULL) being re-enabled while the condition was fulfilled. I will remove this check from the iio_irq_handler as the base handler already checks for the correct status bit beforehand.

drivers/health/maxm86161/maxm86161.c
jcroleda: Will update the dev struct to only hold device functional attributes (IRQ, I2C, GPIO). As such these config structs will be removed in favor of runtime retrieval of attribute values instead.

drivers/health/maxm86161/maxm86161.c
jcroleda: Will update burst write as per previous comment.

projects/maxm86161/src/platform/maxim/main.c
jcroleda: This behavior was copied from an example project I was referencing without much afterthought. Upon reviewing and testing again this is unnecessary, and I will remove it in the next version. Sorry for the confusion.

jcroleda:
Changelog V4:
- Removed unused config structs from main device struct
- Removed chunked burst writes and maintained a constant buffer size
- Removed no_os_mdelay from main.c before project example call
- Removed status bitfield check from fifo_irq_handler
- Fixed no_os_bitfield_get for OOR calibration status getter
- Declared burst_enable and burst_rate in header file
- Replaced and split picket fence set/get into individual functions for each respective option

drivers/health/maxm86161/maxm86161.c
```diff
@@ -0,0 +1,2246 @@
...
+{
+	int ret = 0;
+	uint8_t buf[MAXM86161_SHA_CHALLENGE_LEN +
+						1]; /* max size of applicable burst write op */
+
+	if (!dev || !reg_data)
+		return -EINVAL;
+
+	buf[0] = reg_addr;
+	memcpy(&buf[1], reg_data, count);
```
rbolboac, line 157: count should be checked against MAXM86161_SHA_CHALLENGE_LEN to avoid memory corruption for buf

drivers/health/maxm86161/maxm86161.c
```diff
@@ -0,0 +1,2246 @@
...
+err_irq:
+	if (dev->irq_ctrl)
+		no_os_irq_ctrl_remove(dev->irq_ctrl);
+err_gpio:
+	if (dev->gpio_intb)
+		no_os_gpio_remove(dev->gpio_intb);
+err_bus:
+	r = no_os_i2c_remove(dev->i2c_desc);
+	if (r)
+		return r;
```
rbolboac: should be removed, we want to also free dev even in case of an error

drivers/health/maxm86161/maxm86161.c
```diff
@@ -0,0 +1,2246 @@
...
+					 MAXM86161_MEM_BANK_SEL, enable);
+}
+
+int maxm86161_memory_bank_get(struct maxm86161_dev *dev, bool *enable)
+{
+
+	int ret;
+	uint8_t reg_val;
+
+	if (!dev)
```
rbolboac: enable should also be checked

projects/maxm86161/src/platform/maxim/main.c
```diff
@@ -0,0 +1,47 @@
...
+
+	ret = no_os_uart_init(&uart_desc, &maxm86161_uart_ip);
+	if (ret)
+		goto remove_irq;
+
+	no_os_uart_stdio(uart_desc);
+
+	ret = example_main();
+
+remove_uart:
```
rbolboac: unused label

drivers/health/maxm86161/iio_maxm86161.c
```diff
@@ -0,0 +1,893 @@
...
+
+	if (fifo_count == 0)
+		return;
+
+	if (fifo_count > MAXM86161_FIFO_DEPTH)
+		fifo_count = MAXM86161_FIFO_DEPTH;
+
+	ret = maxm86161_fifo_overflow_check(iio_dev->drv_dev, &overflow_count);
+	if (ret)
+		return;
```
rbolboac, line 633: seems like this value is unused

rbolboac (commented):
some other remarks from my side

jcroleda:
Changelog V5:
- added bounds check for burst write to limit to MAXM86161_SHA_CHALLENGE_LEN
- removed return on i2c remove error
- added enable pointer check in memory_bank_get
- removed `remove_uart` label
- added `fifo_overflow_count` debug attribute to allow user to read overflow count cached upon FIFO read via IRQ

drivers/health/maxm86161/maxm86161.h
```diff
@@ -0,0 +1,607 @@
...
+static const uint16_t maxm86161_irq_status_bits[MAXM86161_INT_COUNT] = {
+	[MAXM86161_INT_PWR_RDY]      = MAXM86161_INT1_PWR_RDY,
+	[MAXM86161_INT_DIE_TEMP_RDY] = MAXM86161_INT1_DIE_TEMP_RDY,
+	[MAXM86161_INT_LED_COMPB]    = MAXM86161_INT1_LED_COMPB,
+	[MAXM86161_INT_PROX_INT]     = MAXM86161_INT1_PROX_INT,
+	[MAXM86161_INT_ALC_OVF]      = MAXM86161_INT1_ALC_OVF,
+	[MAXM86161_INT_DATA_RDY]     = MAXM86161_INT1_DATA_RDY,
+	[MAXM86161_INT_A_FULL]       = MAXM86161_INT1_A_FULL,
+	[MAXM86161_INT_SHA_DONE]     = (uint16_t)MAXM86161_INT2_SHA_DONE << 8,
+};
```
rbolboac: move to .c

drivers/health/maxm86161/maxm86161.h
```diff
@@ -0,0 +1,607 @@
...
+
+enum maxm86161_smp_ave {
+	MAXM86161_SMP_AVE_1 = 0,
+	MAXM86161_SMP_AVE_2,
+	MAXM86161_SMP_AVE_4,
+	MAXM86161_SMP_AVE_8,
+	MAXM86161_SMP_AVE_164,
+	MAXM86161_SMP_AVE_325,
+	MAXM86161_SMP_AVE_646,
```
rbolboac: are these typos? I expect they should be 16,32,64

jcroleda:
Changelog V6:
- moved `maxm86161_irq_status_bits` to maxm86161.c
- fixed typo of smp_ave enums

drivers/health/maxm86161/maxm86161.c
```diff
@@ -0,0 +1,2258 @@
...
+
+/**
+ * @brief Set the picket-fence IIR time constant.
+ *
+ * Updates the IIR_TC field of the PICKET_FENCE register.
+ * @param dev    - Device descriptor.
+ * @param iir_tc - IIR time constant selection.
+ * @return 0 on success, negative error code otherwise.
+ */
+int maxm86161_set_pf_iir_tc(struct maxm86161_dev *dev, uint8_t iir_tc)
```
RaduSabau1: this function writes the raw uint8_t into a 2-bit field with no range check — e.g. passing 7 silently truncates to 1 instead of returning -EINVAL

RaduSabau1 (dismissed):
only one small comment from my side, otherwise lgtm

jcroleda:
Changelog V7:
- Added picket fence bitfield enums for IIR_TC, IIR_INIT, threshold_sigma
- Updated picket fence setters to include value guards for invalid inputs

jcroleda:
Rebased to resolve merge conflict.

rbolboac:
```
health/maxm86161/iio_maxm86161.c: In function 'maxm86161_iio_fifo_irq_handler':
health/maxm86161/iio_maxm86161.c:610:17: warning: unused variable 'status1_data' [-Wunused-variable]
  610 |         uint8_t status1_data;
      |                 ^~~~~~~~~~~~
health/maxm86161/iio_maxm86161.c: At top level:
health/maxm86161/iio_maxm86161.c:770:27: warning: initialization of 'int (*)(void *, uint32_t,  uint32_t *)' {aka 'int (*)(void *, long unsigned int,  long unsigned int *)'} from incompatible pointer type 'int32_t (*)(void *, uint32_t,  uint32_t *)' {aka 'long int (*)(void *, long unsigned int,  long unsigned int *)'} [-Wincompatible-pointer-types]
  770 |         .debug_reg_read = maxm86161_iio_reg_read,
      |                           ^~~~~~~~~~~~~~~~~~~~~~
health/maxm86161/iio_maxm86161.c:770:27: note: (near initialization for 'maxm86161_iio_device.debug_reg_read')
health/maxm86161/iio_maxm86161.c:771:28: warning: initialization of 'int (*)(void *, uint32_t,  uint32_t)' {aka 'int (*)(void *, long unsigned int,  long unsigned int)'} from incompatible pointer type 'int32_t (*)(void *, uint32_t,  uint32_t)' {aka 'long int (*)(void *, long unsigned int,  long unsigned int)'} [-Wincompatible-pointer-types]
  771 |         .debug_reg_write = maxm86161_iio_reg_write,
      |                            ^~~~~~~~~~~~~~~~~~~~~~~
health/maxm86161/iio_maxm86161.c:771:28: note: (near initialization for 'maxm86161_iio_device.debug_reg_write')
health/maxm86161/iio_maxm86161.c:772:19: warning: initialization of 'int (*)(struct iio_device_data *)' from incompatible pointer type 'int32_t (*)(struct iio_device_data *)' {aka 'long int (*)(struct iio_device_data *)'} [-Wincompatible-pointer-types]
  772 |         .submit = maxm86161_iio_submit,
      |                   ^~~~~~~~~~~~~~~~~~~~
health/maxm86161/iio_maxm86161.c:772:19: note: (near initialization for 'maxm86161_iio_device.submit')
```
it looks like there are some build warnings

jcroleda:
Changelog V8:
- replaced `int32_t` with `int` for IIO function pointers causing build warnings

rbolboac (approved):
lgtm

PR #3291, opened 2026-08-19
adrv902x: xsa profile select

mpetrea:
## Pull Request Description

Select the ADRV902x JESD profile and Rx/ORx converter configuration automatically from the board XSA (adding a JESD204B ORx "nls" firmware profile), so the same project builds correctly across the supported hardware variants without manual edits. Also fixes the CMake Xilinx flow to compile the no-OS app with -DSDT on Vitis 2025+ System Device Tree BSPs, matching the BSP's ABI so peripheral lookups resolve to real base addresses (previously ad9528_setup() failed and the IIOD UART was unresponsive on zcu102).

## PR Type

## PR Checklist


mpetrea:
the new CI commits should fix a stale-cache bug on the self-hosted Xilinx runner, where clean: false and a reused build_dir let an old header/SDT decision survive and break the build with "undefined type" errors. They cure it two ways: a git reset --hard after checkout wipes stale tracked sources, and a .noos_sdt sentinel forces a fresh CMake configure whenever the BSP's SDT flag flips.

mpetrea:
temp commit introduced to exercise fixes introduced by CI commits on an isolated test runner. this will be reverted before pr merge

mpetrea:
reverted temporary CI commits

mpetrea:
d60b57f40 cmake: build no-OS with -DSDT on SDT BSP — amended to fold in an isolation fix: the -DSDT detection in config_xilinx_sdk is now gated on if(NOOS_XILINX_SDT), and only projects/adrv902x/CMakeLists.txt sets set(NOOS_XILINX_SDT ON) before config_platform_sdk(). Every other xilinx project (zed/ps7, etc.) reverts to exact main behavior.

mpetrea:
dropped the redundant _nls profile suffix and orphan bucket, and fixed the profile header output name

liviutomoiaga:
> reverted temporary CI commits

Thanks for doing so! 👍 Since there are no file changes that require my expertise, my review can be skipped.

danmois (dismissed):
lgtm

amiclaus (dismissed):

projects/adrv902x/scripts/profile_to_header.py
```diff
@@ -0,0 +1,49 @@
+#!/usr/bin/env python3
```
rbolboac, line 1: please add the according license for this file, if BSD-3 should be something like: 
```
# Copyright 2026(c) Analog Devices, Inc.
#
# SPDX-License-Identifier: BSD-3-Clause
# 
```

projects/adrv902x/scripts/xsa_profile.sh
```diff
@@ -0,0 +1,43 @@
+#!/bin/sh
```
rbolboac, line 1: same here

mpetrea: done

projects/adrv902x/scripts/profile_to_header.py
mpetrea: done

rbolboac (approved):

PR #3304, opened 2026-08-24
drivers: impedance-analyzer: ad5933: rework driver, add IIO support and examples

lerryc:
## Pull Request Description

This PR overhauls the AD5933 impedance analyzer driver to align with current no-OS conventions, adds an IIO driver layer, provides a complete EVAL-AD5933ARDZ example project for the STM32/SDP-CK1Z platform, and Sphinx documentation.

- Driver rework: fixed swapped real/imaginary registers (per BU); public functions now return int status; added I2C probe helper, ad5933_setup(), single-byte reg read/write helpers, sweep/measurement helpers, and enums for clock source, output range, PGA gain and settling multiplier; renamed device properties for clarity; infers settling cycle multiplier from settling cycles.
- IIO driver (iio_ad5933.c): adds IIO support, wired into CMake/Kconfig.
- Example project (projects/eval-ad5933ardz/): basic + IIO examples, board config, build files; sourced from projects/Kconfig.
- Docs: driver/project README.rst, new "IMPEDANCE ANALYZERS" section in the Sphinx toctree.

## PR Type

## PR Checklist


drivers/impedance-analyzer/ad5933/ad5933.c
```diff
@@ -52,30 +34,81 @@ const int32_t pow_2_27 = 134217728ul;      // 2 to the power of 27
...
+	ret = ad5933_set_system_clk(dev, dev->clock_source,
+				    dev->clock_freq);
+	if (ret)
+		return ret;
+
+	ret = ad5933_set_range_and_gain(dev, dev->output_range, dev->pga_gain);
+	if (ret)
+		return ret;
+
+	ret = ad5933_set_settling_time(dev, dev->settling_cycles);
```
edelweiseescala: You can already return here to reduce LOC

drivers/impedance-analyzer/ad5933/ad5933.c
```diff
@@ -199,215 +307,250 @@ void ad5933_set_system_clk(struct ad5933_dev *dev,
...
-	/* Ensure that incNum is a valid data. */
-	if (inc_num > AD5933_MAX_INC_NUM) {
-		inc_num_reg = AD5933_MAX_INC_NUM;
-	} else {
-		inc_num_reg = inc_num;
-	}
+	if (!dev)
+		return -EINVAL;
+
+	end_freq = start_freq + inc_freq * num_increments;
```
edelweiseescala: would be better to do this after checking the increments and start freq bounds

drivers/impedance-analyzer/ad5933/ad5933.c
```diff
@@ -199,215 +307,250 @@ void ad5933_set_system_clk(struct ad5933_dev *dev,
...
+
+	if (!gain_factor || calibration_impedance == 0)
+		return -EINVAL;
+
+	ret = ad5933_get_magnitude(dev, &magnitude);
+	if (ret)
+		return ret;
 
-	ad5933_get_data(dev, freq_function, &imag_data, &real_data);
+	if (magnitude == 0.0)
```
edelweiseescala: This floating point comparison is problematic maybe you can do something like
```suggestion
#define AD5933_MIN_MAGNITUDE  0.000001f
if (fabs(magnitude) <= AD5933_MIN_MAGNITUDE)
    return -EINVAL;
```

or do zero check inside the get_magnitude function before you squareroot

drivers/impedance-analyzer/ad5933/ad5933.c
```diff
@@ -418,47 +561,505 @@ double ad5933_calculate_gain_factor(struct ad5933_dev *dev,
...
+	if (!dev || !impedance)
+		return -EINVAL;
 
-	magnitude = sqrt((real_data * real_data) + (imag_data * imag_data));
+	ret = ad5933_get_magnitude(dev, &magnitude);
+	if (ret)
+		return ret;
 
-	impedance =  1 / (magnitude * gain_factor);
+	if (magnitude == 0.0 || gain_factor == 0.0)
```
edelweiseescala: ditto

drivers/impedance-analyzer/ad5933/ad5933.c
```diff
@@ -199,266 +307,755 @@ void ad5933_set_system_clk(struct ad5933_dev *dev,
...
+{
+	int ret;
+	uint32_t reg_val;
+
+	if (!dev)
+		return -EINVAL;
+
+	number_cycles = no_os_clamp(number_cycles, 0, AD5933_MAX_SETTLING_CYCLES);
+
+	if (number_cycles > AD5933_MAX_SETTLING_CYCLES)
```
edelweiseescala: You already used no_os_clamp, why clamp again?

drivers/impedance-analyzer/ad5933/ad5933.c
```diff
@@ -199,266 +307,755 @@ void ad5933_set_system_clk(struct ad5933_dev *dev,
...
+	if (ret)
+		return ret;
+
+	/* 2. Reset (clears status flags; sweep parameters are preserved). */
+	ret = ad5933_reset(dev);
+	if (ret)
+		return ret;
+
+	/* 3. Initialize with start frequency (excite, no measurement yet). */
+	ret = ad5933_init_start_freq(dev);
```
edelweiseescala: you can already return here, check other instances

drivers/impedance-analyzer/ad5933/README.rst
```diff
@@ -0,0 +1,251 @@
...
+against a known reference impedance is performed with
+**ad5933_calculate_gain_factor**; once the gain factor is known,
+**ad5933_calculate_impedance** returns the impedance magnitude |Z| in ohms. The
+on-chip temperature sensor is read with **ad5933_get_raw_temperature** (raw
+14-bit two's complement code) or **ad5933_get_temperature** (degrees Celsius).
+
+AD5933 Driver Initialization Example
+------------------------------------
+
+.. code-block:: bash
```
edelweiseescala: should be
`.. code-block:: c`

drivers/impedance-analyzer/ad5933/README.rst
```diff
@@ -0,0 +1,251 @@
...
+point. Requested samples beyond the number of points collected are padded with
+``INT16_MAX``. On buffer disable the device is placed in power-down mode.
+
+Since a sweep measurement can take a considerable amount of time, the client is
+responsible in setting an appropriate timeout to await a valid read.
+
+AD5933 IIO Driver Initialization Example
+----------------------------------------
+
+.. code-block:: bash
```
edelweiseescala: ditto

projects/eval-ad5933ardz/src/common/common_data.c
```diff
@@ -0,0 +1,40 @@
...
+};
+
+struct ad5933_init_param ad5933_user_init = {
+	.i2c_init = ad5933_i2c_ip,
+	.clock_freq = 16000000,
+	.clock_source = AD5933_CONTROL_EXT_SYSCLK,
+	.pga_gain = AD5933_GAIN_X1,
+	.output_range = AD5933_RANGE_2000mVpp,
+	.settling_cycles = AD5933_SETTLING_CYCLES,
+};
```
edelweiseescala: Missing new line at end of file

projects/eval-ad5933ardz/src/examples/basic/basic_example.c
```diff
@@ -0,0 +1,138 @@
...
+ *******************************************************************************/
+
+#include "common_data.h"
+#include "ad5933.h"
+#include "no_os_delay.h"
+#include "no_os_print_log.h"
+#include <stdio.h>
+
+/* Sweep definition: start frequency, increment and number of increments. */
+#define BASIC_START_FREQ_HZ 2000  /* 30 kHz start frequency. */
```
edelweiseescala: comment does not match the frequency

projects/eval-ad5933ardz/src/examples/iio_example/iio_example.c
```diff
@@ -0,0 +1,68 @@
...
+	ret = iio_app_init(&app, app_init_param);
+
+	if (ret) {
+		ad5933_iio_remove(ad5933_iio_desc);
+		return ret;
+	}
+
+
+	return iio_app_run(app);
+}
```
edelweiseescala: missing new line, check other files for missing new line

projects/eval-ad5933ardz/README.rst
```diff
@@ -0,0 +1,184 @@
...
+Used Hardware
+^^^^^^^^^^^^^
+
+* `EVAL-AD5933ARDZ <https://www.analog.com/EVAL-AD5933ARDZ>`_
+* SDP-CK1Z controller board (STM32)
+
+Connections
+^^^^^^^^^^^
+
+The EVAL-AD5933ARDZ connects to the SDP-CK1Z through the SDP connector. The
```
edelweiseescala: there should be a table to show how to connect SDP-K1 to EVAL-AD5933ARDZ

projects/eval-ad5933ardz/boards/basic/sdp-ck1z.conf
```diff
@@ -0,0 +1 @@
+CONFIG_STM32_IOC_PATH="sdp-ck1z.ioc"
```
edelweiseescala, line 1: I think this file should be under boards/basic/ and boards/iio/

drivers/impedance-analyzer/ad5933/ad5933.c
```diff
@@ -199,266 +307,755 @@ void ad5933_set_system_clk(struct ad5933_dev *dev,
...
+ * @return 0 in case of success, negative error code otherwise.
+*******************************************************************************/
+int ad5933_get_magnitude(struct ad5933_dev *dev,
+			 double *magnitude)
+{
+	int16_t real = 0;
+	int16_t imag = 0;
+	int ret;
+
+	if (!magnitude)
```
adi-ramonimbao: Missing `dev` NULL check

drivers/impedance-analyzer/ad5933/ad5933.c
```diff
@@ -199,266 +307,755 @@ void ad5933_set_system_clk(struct ad5933_dev *dev,
...
+ *         negative error code on I2C failure.
+*******************************************************************************/
+int ad5933_wait_status(struct ad5933_dev *dev, uint8_t mask,
+		       uint8_t *status_out)
+{
+	uint32_t timeout = AD5933_MAX_RETRIES;
+	uint8_t status;
+	int ret;
+
+	if (!dev)
```
adi-ramonimbao, line 649: Missing `status_out` NULL check

drivers/impedance-analyzer/ad5933/ad5933.c
```diff
@@ -199,266 +307,755 @@ void ad5933_set_system_clk(struct ad5933_dev *dev,
...
+	if (ret)
+		return ret;
+
+	*real = (int16_t)real_raw;
+	*imag = (int16_t)imag_raw;
+
+	return 0;
+}
+
+int ad5933_sweep_done(struct ad5933_dev *dev, int32_t *done)
```
adi-ramonimbao, line 1003: Missing doxygen comments

drivers/impedance-analyzer/ad5933/ad5933.c
```diff
@@ -199,266 +307,755 @@ void ad5933_set_system_clk(struct ad5933_dev *dev,
...
+ * @param dev - The device structure.
+ *
+ * @return 0 once the device acknowledges, -ETIMEDOUT otherwise.
+*******************************************************************************/
+static int ad5933_wait_powerup(struct ad5933_dev *dev)
+{
+	uint32_t retries = AD5933_POWERUP_RETRIES;
+	uint8_t status;
+
+	do {
```
adi-ramonimbao, line 1043: Missing `dev` NULL check

drivers/impedance-analyzer/ad5933/iio_ad5933.c
```diff
@@ -0,0 +1,694 @@
...
+	if (ret)
+		return ret;
+
+	memset(iio_ad5933->channel_data, 0, sizeof(iio_ad5933->channel_data));
+
+	no_os_mdelay(100);
+
+	ret = ad5933_start_sweep(iio_ad5933->ad5933_dev);
+	if (ret)
+		return 0;
```
edelweiseescala: I think this should have returned ret

drivers/impedance-analyzer/ad5933/ad5933.c
```diff
@@ -174,22 +272,32 @@ void ad5933_reset(struct ad5933_dev *dev)
...
-				  AD5933_REG_CONTROL_LB,
-				  dev->current_clock_source,
-				  1);
-}
+	int ret;
+
+	if (!dev)
+		return -EINVAL;
+
+	if (ext_clk_freq <= 0)
```
edelweiseescala: This is unsigned no need for negative check

drivers/impedance-analyzer/ad5933/iio_ad5933.c
```diff
@@ -0,0 +1,694 @@
...
+	.debug_attributes = ad5933_iio_debug_attrs,
+	.buffer_attributes = NULL,
+	.pre_enable = ad5933_iio_pre_enable,
+	.post_disable = ad5933_iio_post_disable,
+	.submit = ad5933_iio_submit,
+	.debug_reg_read = ad5933_iio_reg_read,
+	.debug_reg_write = ad5933_iio_reg_write,
+};
+
+static int ad5933_iio_reg_read(void *device, uint32_t reg_addr,
```
adi-ramonimbao, line 251: Missing doxygen comments

drivers/impedance-analyzer/ad5933/iio_ad5933.c
```diff
@@ -0,0 +1,694 @@
...
+	if (ret)
+		return ret;
+
+	*reg_data = (uint32_t)reg_val;
+
+	return 0;
+}
+
+static int ad5933_iio_reg_write(void *device, uint32_t reg_addr,
+				uint32_t reg_data)
```
adi-ramonimbao, line 284: Missing doxygen comments

drivers/impedance-analyzer/ad5933/iio_ad5933.c
```diff
@@ -0,0 +1,694 @@
...
+	uint8_t reg_val;
+
+	if (!device)
+		return -EINVAL;
+
+	return ad5933_reg_write(ad5933, (uint8_t)reg_addr, (uint8_t)reg_data);
+}
+
+static int ad5933_iio_read_raw(void *dev, char *buf, uint32_t len,
+			       const struct iio_ch_info *channel, intptr_t priv)
```
adi-ramonimbao, line 308: Missing doxygen comments

drivers/impedance-analyzer/ad5933/ad5933.c
```diff
@@ -199,266 +307,755 @@ void ad5933_set_system_clk(struct ad5933_dev *dev,
...
-	}
+	if (!dev)
+		return -EINVAL;
+
+	end_freq = start_freq + inc_freq * num_increments;
+
+	if (end_freq > AD5933_MAX_OUTPUT_FREQ)
+		return -EINVAL;
+
+	if (num_increments < 0 || num_increments > AD5933_MAX_INC_NUM)
```
edelweiseescala: This is also unsigned no need for negative check

drivers/impedance-analyzer/ad5933/iio_ad5933.c
```diff
@@ -0,0 +1,694 @@
...
+	}
+
+	return iio_format_value(buf, len, IIO_VAL_INT, 1, &raw);
+}
+
+/**
+ * @brief Read the "scale" attribute (temperature channel only).
+ */
+static int ad5933_iio_read_scale(void *dev, char *buf, uint32_t len,
+				 const struct iio_ch_info *channel, intptr_t priv)
```
adi-ramonimbao: Missing `@param`s and `@return`

drivers/impedance-analyzer/ad5933/iio_ad5933.c
```diff
@@ -0,0 +1,694 @@
...
+{
+	int32_t vals[2];
+
+	if (!dev)
```
adi-ramonimbao: Missing `channel` NULL check

drivers/impedance-analyzer/ad5933/iio_ad5933.c
```diff
@@ -0,0 +1,694 @@
...
+	}
+}
+
+
+/**
+ * @brief Read a device-global attribute.
+ */
+static int ad5933_iio_read_dev_attr(void *dev, char *buf, uint32_t len,
+				    const struct iio_ch_info *channel,
+				    intptr_t priv)
```
adi-ramonimbao: Missing `@param`s and `@return`

edelweiseescala (commented):
Hello some review from me, also noticed error on docs build please check that. Thank You!

drivers/impedance-analyzer/ad5933/iio_ad5933.c
```diff
@@ -0,0 +1,694 @@
...
+		return -EINVAL;
+	}
+}
+
+/**
+ * @brief Write a device-global attribute.
+ */
+static int ad5933_iio_write_dev_attr(void *dev, char *buf, uint32_t len,
+				     const struct iio_ch_info *channel,
+				     intptr_t priv)
```
adi-ramonimbao: Missing `@param`s and `@return`

drivers/impedance-analyzer/ad5933/iio_ad5933.c
```diff
@@ -0,0 +1,694 @@
...
+	ret = ad5933_start_sweep(iio_ad5933->ad5933_dev);
+	if (ret)
+		return 0;
+
+	iio_ad5933->state = AD5933_FUNCTION_START_SWEEP;
+
+	return 0;
+}
+
+static int ad5933_iio_post_disable(void *dev)
```
adi-ramonimbao, line 629: Missing doxygen comments

adi-ramonimbao (commented):
A few comments from me regarding doxygen comments and some missing NULL checks.

projects/eval-ad5933ardz/src/common/common_data.c
```diff
@@ -0,0 +1,40 @@
...
+	.size = NO_OS_UART_CS_8,
+	.parity = NO_OS_UART_PAR_NO,
+	.stop = NO_OS_UART_STOP_1_BIT,
+	.platform_ops = UART_OPS,
+	.extra = &uart_extra_ip,
+};
+
+const struct no_os_i2c_init_param ad5933_i2c_ip = {
+	.device_id = I2C_DEVICE_ID,
+	.max_speed_hz = 400000,
```
ARaterta-adi: Define in paramters.h same with your UART_BAUDRATE

projects/eval-ad5933ardz/src/common/common_data.c
```diff
@@ -0,0 +1,40 @@
...
+	.slave_address = AD5933_ADDRESS,
+	.platform_ops = I2C_OPS,
+	.extra = &i2c_extra_ip,
+};
+
+struct ad5933_init_param ad5933_user_init = {
+	.i2c_init = ad5933_i2c_ip,
+	.clock_freq = 16000000,
```
ARaterta-adi: Define in paramters.h same with your UART_BAUDRATE

projects/eval-ad5933ardz/src/platform/stm32/parameters.h
```diff
@@ -0,0 +1,36 @@
...
+#define __PARAMETERS_H__
+
+#include "stm32_uart.h"
+#include "stm32_uart_stdio.h"
+#include "stm32_i2c.h"
+#include "stm32_irq.h"
+#include "stm32_gpio.h"
+
+#define I2C_DEVICE_ID			1
+#define I2C_TIMING			0x00000E14 /* (Unused) */
```
ARaterta-adi, line 21: Remove unless required in future.

ARaterta-adi (commented):
Hello, please check the minor changes suggested.

drivers/impedance-analyzer/ad5933/ad5933.c
lerryc: `status_out` can be null here, it is optional

drivers/impedance-analyzer/ad5933/iio_ad5933.c
lerryc: I'm not sure if `channel` here can ever be NULL, since it is provided by the IIO library itself. Existing iio drivers do not seem have `channel` NULL checks either.

projects/eval-ad5933ardz/src/platform/stm32/parameters.h
lerryc: adf5611 and admt4000 includes this and is passed into the i2c_extra_ip for the stm32 platform. Both devices also use the sdp-ck1z board.

lerryc:
V2:

- Use epsilon comparison instead of floating point comparison for zero checks
- Early returns where applicable
- AD5933_MIN_OUTPUT_FREQ changed from 0.1 to 1000 as per datasheet, AD5933_MIN_FREQ_INC is the one that should be 0.1 Hz
- Changed start_freq, freq_increment, and current_output_freq data type from uint32_t to double, since AD5933 supports sub-Hertz increment down to 0.1 Hz as per datasheet. Updated the corresponding IIO attributes to support this change.
- Remove unnecessary negative checks for unsigned data type variables
- Add missing doxygen comments and variables
- Add missing NULL checks
- Use macros for configuration values
- Formatting

drivers/impedance-analyzer/ad5933/ad5933.c
```diff
@@ -199,266 +302,747 @@ void ad5933_set_system_clk(struct ad5933_dev *dev,
...
+/***************************************************************************//**
+ * @brief Repeat the current frequency point measurement.
+ *
+ * @param dev - The device structure.
+ *
+ * @return 0 in case of success, negative error code otherwise.
+*******************************************************************************/
+int ad5933_repeat_freq(struct ad5933_dev *dev)
+{
+	int ret;
```
amiclaus: unused. there are multiple locations, fix them all

drivers/impedance-analyzer/ad5933/ad5933.c
```diff
@@ -199,266 +302,747 @@ void ad5933_set_system_clk(struct ad5933_dev *dev,
...
+/***************************************************************************//**
+ * @brief  Default state on power-up of the AD5933 is power-down mode.
+ *
+ * @param dev - The device structure.
+ *
+ * @return 0 in case of success, negative error code otherwise.
+*******************************************************************************/
+int ad5933_power_down(struct ad5933_dev *dev)
+{
+	int ret;
```
amiclaus: ditto

drivers/impedance-analyzer/ad5933/ad5933.c
```diff
@@ -199,266 +302,747 @@ void ad5933_set_system_clk(struct ad5933_dev *dev,
...
+/***************************************************************************//**
+ * @brief Powers up the part for general operation.
+ *
+ * @param dev - The device structure.
+ *
+ * @return 0 in case of success, negative error code otherwise.
+*******************************************************************************/
+int ad5933_standby(struct ad5933_dev *dev)
+{
+	int ret;
```
amiclaus: ditto

drivers/impedance-analyzer/ad5933/ad5933.c
```diff
@@ -199,266 +302,747 @@ void ad5933_set_system_clk(struct ad5933_dev *dev,
...
+/***************************************************************************//**
+ * @brief  Enables the DDS to output the programmed start frequency for an indefinite time.
+ *
+ * @param dev - The device structure.
+ *
+ * @return 0 in case of success, negative error code otherwise.
+*******************************************************************************/
+int ad5933_init_start_freq(struct ad5933_dev *dev)
+{
+	int ret;
```
amiclaus: ditto.

drivers/impedance-analyzer/ad5933/ad5933.c
```diff
@@ -174,22 +270,29 @@ void ad5933_reset(struct ad5933_dev *dev)
...
-		dev->current_sys_clk = AD5933_INTERNAL_SYS_CLK;    // 16 MHz
-	}
-	ad5933_set_register_value(dev,
-				  AD5933_REG_CONTROL_LB,
-				  dev->current_clock_source,
-				  1);
-}
+	int ret;
+
+	if (!dev || !ext_clk_freq)
```
amiclaus: we no longer support internal clock source?

The header comment says something else: `@param ext_clk_freq - Frequency value of the external clock, if used.`

drivers/impedance-analyzer/ad5933/ad5933.c
```diff
@@ -199,266 +302,747 @@ void ad5933_set_system_clk(struct ad5933_dev *dev,
...
+ *
+ * @param dev - The device structure.
+ * @param frequency - The frequency value in Hz.
+ *
+ * @return Binary code to write into the corresponding register.
+*******************************************************************************/
+uint32_t ad5933_convert_freq_to_reg(struct ad5933_dev *dev, double frequency)
+{
+	if (!dev || dev->clock_freq == 0)
+		return 0;
```
amiclaus, line 950: what if `frequency` value is 0? how do we make a distinction?

drivers/impedance-analyzer/ad5933/ad5933.c
```diff
@@ -199,266 +302,747 @@ void ad5933_set_system_clk(struct ad5933_dev *dev,
...
+	/* Convert increment frequency to binary code. */
+	inc_freq_reg = ad5933_convert_freq_to_reg(dev, inc_freq);
+
+	if (!inc_freq_reg)
+		return -EINVAL;
 
-	/* Convert users start frequency to binary code. */
-	start_freq_reg = (uint32_t)((double)start_freq * 4 / dev->current_sys_clk *
-				    pow_2_27);
+	if (!dev)
```
amiclaus: this goes at the start of the function below the variable declarations

projects/eval-ad5933ardz/src/examples/basic/basic_example.c
```diff
@@ -0,0 +1,140 @@
...
+
+	ret = ad5933_start_sweep(dev);
+	if (ret)
+		goto error_remove;
+
+	ret = ad5933_wait_status(dev, AD5933_STAT_DATA_VALID, NULL);
+	if (ret)
+		goto error_remove;
+
+	if (ret)
```
amiclaus: duplicated.

drivers/impedance-analyzer/ad5933/ad5933.c
lerryc: I will guard against passing ext_clk_freq = 0 in ad5933_set_system_clk so that dev->clock_freq does not need to be zero-checked here. I will return -EINVAL if !dev instead.

drivers/impedance-analyzer/ad5933/ad5933.c
lerryc: I will change this to check ext_clk_freq only if clk_source is external.

projects/eval-ad5933ardz/src/examples/basic/basic_example.c
```diff
@@ -0,0 +1,140 @@
...
+
+	for (int i = 0; i < BASIC_INC_NUM; i++) {
+		freq = i == 0 ? BASIC_START_FREQ_HZ : BASIC_START_FREQ_HZ +
+		       (BASIC_FREQ_INC_HZ * i);
+
+		if (i == 0) {
+			ret = ad5933_repeat_freq(dev);
+		} else {
+			ret = ad5933_increment_freq(dev);
+		}
```
RaduSabau1, line 116: if (ret) return ret; otherwise ret value is overwritten by next operation.

projects/eval-ad5933ardz/src/examples/iio_example/iio_example.c
```diff
@@ -0,0 +1,68 @@
...
+#define FREQ_POINTS 150U
+#define SETTLING_CYCLES 15U
+#define CALIBRATION_RESISTOR_OHM 500U
+
+static struct ad5933_iio_dev_init_param ad5933_iio_ip = {
+	.ad5933_dev_ip = &ad5933_user_init,
+	.start_freq = START_FREQ,
+	.freq_increment = FREQ_INC,
+	.freq_points = FREQ_POINTS,
+	.calibration_impedance = CALIBRATION_RESISTOR_OHM,
```
RaduSabau1: where is this used?

RaduSabau1 (commented):
some minor comments from my side

projects/eval-ad5933ardz/src/examples/iio_example/iio_example.c
lerryc: start_freq, freq_increment, and freq_points are used for configuring sweep parameters. I will remove calibration_impedance as this is not used in the iio app.

lerryc:
V3:

- Set dev->clock_freq as init_param->clock_freq only if dev->clock_source is AD5933_CLOCK_SOURCE_EXTERNAL in ad5933_init
- Guard against ext_clk_freq = 0 or ext_clk_freq > AD5933_INTERNAL_SYS_CLK in ad5933_set_system_clk if clk_source = AD5933_CLOCK_SOURCE_EXTERNAL
- Change AD5933_MIN_OUTPUT_FREQ macro from constant 1kHz to dynamic frequency calculation from dev->clock_freq. Datasheet states minimum output frequency can be lowered by scaling down MCLK.
- Change ad5933_output_range_available from millivolt ranges based on linux iio to datasheet Range numbers (1-4), since output excitation range is not always exact due to it being able to scale with VDD.
- Use macros for delay magic number values
- Remove unused ret variables across various functions
- Add AD5933_STATE_SWEEP_DONE = 0x0C, only used to track sweep status during ad5933_iio_submit to prevent remeasuring in case num_samples requested by the iio client is greater than number of sweep points.
- Remove sweep_done register check in ad5933_iio_submit, in the edge case of single point measurement (num_increments = 0), sweep_done status reg can be set before ad5933_iio_submit gets called which prevents measurement acquisition.
- Early returns in ad5933_iio_write_dev_attr and -EINVAL in case of unavailable choice selected in pga_gain and output_range.
- Remove unused calibration_impedance in ad5933_iio_dev_init_param.
- Correct links in README.rst of project.
- Correct sample code blocks in README.rst of driver.
- Change data type of calibration_impedance parameter in calculate_gain_factor function to support 4-wire calibration of very small impedances.
- Add measurement_mode options in basic_example for 2-wire and 4-wire configuration of the board.
- Formatting

lerryc:
V4:

- Add additional iio debug attributes for min/max output frequency and system clock frequency.

drivers/impedance-analyzer/ad5933/iio_ad5933.c
```diff
@@ -0,0 +1,825 @@
...
+		.name = "maximum_output_frequency",
+		.show = ad5933_iio_read_dev_attr,
+		.priv = AD5933_ATTR_MAX_OUTPUT_FREQ,
+	},
+	END_ATTRIBUTES_ARRAY
+};
+
+static struct iio_attribute ad5933_iio_dev_attrs[] = {
+	{
+		.name = "in_voltage0_scale",
```
RaduSabau1: Had a deeper look in this driver and I've seen that this array is used for device attributes. This is a channel attribute name though it stands for device attributes. I can see why this is global because it handles the PGA gain which is applicable to both ch0 and ch1. However I think this should be written as .scale channel attribute which should be shared by both channels.

drivers/impedance-analyzer/ad5933/iio_ad5933.c
```diff
@@ -0,0 +1,825 @@
...
+		.store = ad5933_iio_write_dev_attr,
+		.priv = AD5933_ATTR_PGA_GAIN,
+	},
+	{
+		.name = "in_voltage0_scale_available",
+		.show = ad5933_iio_read_dev_attr,
+		.priv = AD5933_ATTR_PGA_GAIN_AVAILABLE,
+	},
+	{
+		.name = "out_altvoltage0_raw",
```
RaduSabau1: about this, not only the README.rst says it outputs range in mVpp and the actual output is an index but you seem to have a pseudo-channel through device attributes. If these attributes are actually device attributes and change to these attributes affect the whole device, nevertheless all channels/functionality, then their naming should be changed since out_altvoltage0 prefix stands for "output channel 0 of type alternating voltage". and if there actually is a channel that does that, why not have an actual altvoltage channel, and have the other altvoltage0 channel attributes in a separate struct array specially designed for this channel?

drivers/impedance-analyzer/ad5933/iio_ad5933.c
lerryc: I wanted to have channel and attribute names for this IIO driver be the same with the current Linux IIO driver implementation. Regarding the scale channel attribute concern, please see the note at https://github.com/analogdevicesinc/linux/blob/main/drivers/staging/iio/impedance-analyzer/ad5933.c#L475. If really needed, I can change them accordingly. Please let me know.

drivers/impedance-analyzer/ad5933/iio_ad5933.c
lerryc: I will change the README.rst, it should be the range index now (Range 1,2,3 or 4 from the datasheet). Previously the available values that can be set for this attribute was dynamically calculated based on the Vref chosen for the device. Currently, it is no longer possible to choose a Vref, instead the output range scales with the supplied Vdd. I will also update the README.rst to include the new debug attributes.

drivers/impedance-analyzer/ad5933/iio_ad5933.c
RaduSabau1: Checked the Linux driver — the flat IIO_DEVICE_ATTRs for out_altvoltage0/in_voltage0 aren't a deliberate naming choice, they're a workaround for a Linux-specific constraint: those two "channels" aren't in the real iio_chan_spec[] table (only temp/real/imag are), and Linux's sysfs attribute generation only works for channels that exist in that table. The comment at ad5933.c:475 confirms this ("we don't create dedicated sysfs channel attributes for out0 and in0").

no-OS doesn't have that constraint — our iio_channel struct lets any channel carry its own arbitrary attributes[] with explicit show/store, independent of any generic raw-dispatch table. So we can declare voltage0/altvoltage0 as real channels and attach raw/scale/scale_available/frequency_start etc. directly to them, and get the identical in_voltage0_scale/out_altvoltage0_raw sysfs paths for Linux compatibility — without needing the flat device-attribute workaround. I'll restructure it that way.

lerryc:
V5:

- Reworked the status poll timing to reduce poll delay but increase max retries. This keeps the same timeout duration but reduces the interval between polls, which is useful for very short frequency point measurements.
- Removed the global device attributes and added them to dedicated channel attributes for in_voltage0 and out_altvoltage0.
- Add helper macros for read-only and read/write channel attributes: AD5933_CH_RW_ATTR, AD5933_CH_RO_ATTR
- Unified attribute callbacks: removed ad5933_iio_read_dev_attr, ad5933_iio_write_dev_attr, ad5933_iio_read_scale and replaced them with ad5933_iio_read_attr, ad5933_iio_write_attr for all attribute callbacks. The temp scale attribute now uses the new callback via the added AD5933_ATTR_TEMP_SCALE priv.
- Update debug attribute sweep_initialized to accept any non-zero value instead of only the value 1.
- Change default sweep parameters in the IIO example project: 
        START_FREQ 10000→30000, FREQ_INC 100→1000, FREQ_POINTS 150→15.
- Updated README.rst

drivers/impedance-analyzer/ad5933/ad5933.h
```diff
@@ -5,58 +5,37 @@
...
+#define AD5933_REG_CONTROL_HB       0x80    // HB of the Control register, R/W 1 byte
+#define AD5933_REG_CONTROL_LB       0x81    // LB of the Control register, R/W 1 byte
+#define AD5933_REG_FREQ_START       0x82    // Start frequency, R/W 3 bytes
+#define AD5933_REG_FREQ_INC         0x85    // Frequency increment, R/W 3 bytes
+#define AD5933_REG_INC_NUM          0x88    // Number of increments, R/W 2 bytes
+#define AD5933_REG_SETTLING_CYCLES  0x8A    // Number of settling time cycles, R/W 2 bytes
+#define AD5933_REG_STATUS           0x8F    // Status, R 1 byte
+#define AD5933_REG_TEMP_DATA        0x92    // Temperature data, R 2 bytes
+#define AD5933_REG_REAL_DATA        0x96    // Real data, R 2 bytes
+#define AD5933_REG_IMAG_DATA        0x94    // Imaginary data, R 2 bytes
```
RaduSabau1, line 28: why were this two inverted? datasheet says otherwise (as before).

drivers/impedance-analyzer/ad5933/ad5933.c
```diff
@@ -199,266 +311,740 @@ void ad5933_set_system_clk(struct ad5933_dev *dev,
...
-				  2);
+	if (inc_freq > AD5933_MAX_OUTPUT_FREQ(dev->clock_freq) || inc_freq < 0)
+		return -EINVAL;
+
+	/* Convert increment frequency to binary code. */
+	inc_freq_reg = ad5933_convert_freq_to_reg(dev, inc_freq);
+
+	end_freq = start_freq + inc_freq * num_increments;
+
+	if (end_freq > AD5933_MAX_OUTPUT_FREQ(dev->clock_freq))
```
RaduSabau1, line 441: I think truncation here can allow for a silent 0Hz truncation. perhaps use >= instead

drivers/impedance-analyzer/ad5933/iio_ad5933.c
```diff
@@ -0,0 +1,787 @@
...
+			ret = ad5933_initialize_sweep(ad5933);
+			if (ret)
+				return ret;
+			iio_ad5933->state = AD5933_FUNCTION_INIT_START_FREQ;
+			iio_ad5933->current_output_freq = ad5933->start_freq;
+		}
+		return len;
+	case AD5933_ATTR_SWEEP_STARTED:
+		if (val) {
+			ret = ad5933_start_sweep(ad5933);
```
RaduSabau1, line 544: shouldn't this check if the sweep is initialized?

drivers/impedance-analyzer/ad5933/iio_ad5933.c
```diff
@@ -0,0 +1,787 @@
...
+	switch (channel->address) {
+	case AD5933_REG_TEMP_DATA:
+		ret = ad5933_get_raw_temperature(ad5933, &temp_raw);
+		if (ret)
+			return ret;
+		raw = temp_raw;
+		break;
+	case AD5933_REG_REAL_DATA:
+	case AD5933_REG_IMAG_DATA:
+		ret = ad5933_get_current_data(ad5933, &real, &imag);
```
RaduSabau1, line 333: these reads don't wait for valid data although documentation says it should.

drivers/impedance-analyzer/ad5933/iio_ad5933.c
```diff
@@ -0,0 +1,787 @@
...
+		return iio_format_value(buf, len, IIO_VAL_INT, 1, &val);
+	case AD5933_ATTR_SWEEP_STARTED:
+		val = iio_ad5933->state == AD5933_FUNCTION_START_SWEEP;
+		return iio_format_value(buf, len, IIO_VAL_INT, 1, &val);
+	case AD5933_ATTR_CURRENT_OUTPUT_FREQ:
+		val = ad5933->current_output_freq;
+		return iio_format_value(buf, len, IIO_VAL_INT, 1, &val);
+	case AD5933_ATTR_REPEAT_MEASUREMENT:
+	case AD5933_ATTR_INCREMENTED_MEASUREMENT:
+		val = ad5933->current_output_freq;
```
RaduSabau1, line 419: wrong return, most probably a copy-paste bug

drivers/impedance-analyzer/ad5933/ad5933.c
```diff
@@ -199,266 +311,740 @@ void ad5933_set_system_clk(struct ad5933_dev *dev,
...
+		return -EINVAL;
+
+	ret = ad5933_reg_write(dev, AD5933_REG_CONTROL_HB,
+			       AD5933_CONTROL_FUNCTION(AD5933_FUNCTION_START_SWEEP) |
+			       AD5933_CONTROL_RANGE(dev->output_range) |
+			       AD5933_CONTROL_PGA_GAIN(dev->pga_gain));
+	if (ret)
+		return ret;
+
+	dev->sweep_point = 0;
```
RaduSabau1, line 844: can then current_output_freq and sweep_point desync on restart then? I see sweep_point resets to 0 but current_output_freq its not reset to start_freq so I think re-arming after a partial sweep could leave the reported frequency stale.

drivers/impedance-analyzer/ad5933/ad5933.c
```diff
@@ -199,266 +311,740 @@ void ad5933_set_system_clk(struct ad5933_dev *dev,
...
-				  3);
-	ad5933_set_register_value(dev,
-				  AD5933_REG_INC_NUM,
-				  inc_num_reg,
-				  2);
+	if (inc_freq > AD5933_MAX_OUTPUT_FREQ(dev->clock_freq) || inc_freq < 0)
+		return -EINVAL;
+
+	/* Convert increment frequency to binary code. */
+	inc_freq_reg = ad5933_convert_freq_to_reg(dev, inc_freq);
```
RaduSabau1, line 437: would something like AD5933_MIN_FREQ_INC help here to avoid a small enough increment silently becoming a hardware no-op?

projects/eval-ad5933ardz/src/examples/basic/basic_example.c
```diff
@@ -0,0 +1,162 @@
...
+	ret = ad5933_init(&dev, &ad5933_user_init);
+
+	if (ret) {
+		printf("ERROR: AD5933 init failed: %d\r\n", ret);
+		goto error;
+	}
+
+	printf("AD5933 init successful\r\n\r\n");
+
+	ret = ad5933_setup(dev);
```
RaduSabau1, line 56: this code is redundant. ad5933_setup() is already called in the init() function

RaduSabau1 (commented):
some comments from my side and also an open question referring the basic example.
How does the basic example work physically? I see the loop calculates the gain factor and then gets current data and then calculate the impedance (with other operations between, of course). However wouldn't one use a known reference impedance while gain_factor is calculated (calibration purposes I assume) and then need to swap with the actually measured one at run-time (while in the loop) in some way? Otherwise we are actually double-checking the reference impedance I think, right? Perhaps there is something I am missing.

drivers/impedance-analyzer/ad5933/ad5933.h
lerryc: The datasheet needs to be updated according to BU. They will be doing the update but requests the driver to be corrected ahead of time.

drivers/impedance-analyzer/ad5933/ad5933.c
lerryc: You're right, I also updated the start_freq check above it.

drivers/impedance-analyzer/ad5933/iio_ad5933.c
lerryc: These reads only access the underlying registers. Valid data polling is already handled upon requesting repeat and incremented measurements. I will also add the polling to the sweep_started though, so the initial frequency excitation will have the data ready.

drivers/impedance-analyzer/ad5933/iio_ad5933.c
lerryc: This is intended, since on repeat measurements the current output frequency should be returned. On incremented measurements, the current_output_frequency will have been updated by the attribute write so this will also return the next output frequency.

drivers/impedance-analyzer/ad5933/ad5933.c
lerryc: I will move this sweep point reset to ad5933_initialize_sweep() instead, and also add resetting the current_output_freq to the start_freq.

drivers/impedance-analyzer/ad5933/ad5933.c
lerryc: I will guard against num_increments > 0 but inc_freq_reg = 0, so as to prevent hardware no-op.

lerryc:
V6

- Guard against truncation for end_freq and start_freq
- Guard against inc_freq_reg = 0 (hardware no-op)
- Add data_valid wait on sweep_started attr write
- Reset sweep_point and current_output_freq upon initialization
- Rebased unto main
- Reworked basic_example.c:
-store all calibration gain_factors across the configured frequency range
-use stored gain_factors for measuring the unknown impedance
-allow repeating the sweep on uart send for measuring other impedance

RaduSabau1 (dismissed):

rbolboac:
please fix the conflicts

lerryc:
> please fix the conflicts

I have rebased and fixed the conflicts.

rbolboac (approved):
lgtm, let's wait for at least another approve

edelweiseescala (approved):

PR #3307, opened 2026-08-24
scripts: xilinx: wait for Vitis Quick Build to complete

stefpopa:
## Pull Request Description

create_platform_component() triggers an asynchronous "Quick Build" that compiles the BSP. Checking for libxil.a immediately after the call may find it missing and call platform.build(), which crashes with "Application error processing RPC" because the build is already running.

Replace the immediate check with polling that waits up to 90 seconds for both libxil.a and hw0.xpfm to appear.

Fixes: 8e26bd94f4 ("scripts: add Vitis 2025 BSP gen using Unified IDE Python API")

## PR Type

## PR Checklist

### DIFF
--- a/tools/scripts/platform/xilinx/util.py
+++ b/tools/scripts/platform/xilinx/util.py
@@ -499,11 +499,25 @@ def create_project(ws, hw_path, hw_file, target):
     client.create_platform_component(
         name="hw0", hw_design=xsa, cpu=vcpu, os="standalone")
 
-    # Build the platform only if Quick Build didn't already produce the BSP.
-    # create_platform_component() sometimes triggers an internal "Quick Build"
-    # that produces libxil.a — calling platform.build() again can crash the
-    # gRPC server with "Application error processing RPC".
+    # Wait for any async Quick Build to complete before proceeding.
+    # create_platform_component() sometimes triggers an async build; calling
+    # platform.build() while it's running crashes with "Application error
+    # processing RPC".
+    from vitis import _build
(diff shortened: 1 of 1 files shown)


tools/scripts/platform/xilinx/util.py
```diff
@@ -499,19 +499,32 @@ def create_project(ws, hw_path, hw_file, target):
...
+    from vitis import _build
+    build_svc = _build.Build(client._serverObj)
+
+    print("INFO: Waiting for platform build to complete...")
+    for _ in range(90):
+        running = build_svc.listRunningBuilds()
+        if not running:
+            break
+        time.sleep(1)
+    else:
```
SRaus, line 515: Couldn't you just wait for "create_platform_component()" to complete instead of waiting 90s?
Are 90 seconds always sufficient (i.e. the build machine has less resources) ?

stefpopa: 1. Quick build is initiated during create_platform_component(), but it runs asynchronously. The function returns while the build continues in the background. 
2. I admit that 90s is arbitrary and could be insufficient. Maybe I can increase the timeout to 300s?

stefpopa: However, the build is successful. I don't know if we were just lucky this time.

rbolboac (approved):
lgtm, @SRaus has the final word

amiclaus (approved):

stefpopa:
I will close this PR for now, as it seems that the CI builds on main have been green for the past week or so.

PR #3334, opened 2026-09-08
miscellaneous driver/platform bug fixes

RaduSabau1:
## Pull Request Description

A batch of small, independent bug fixes found while auditing drivers and platform code:

- meter: ade7816: ade7816_reg_update() wrote back the raw val instead of the merged reg_val, silently discarding the update.
- platform: linux_gpio: param->port was never copied to the descriptor, so the wrong port could be reported/used at runtime.
- dac: max22017: IIO remove path freed a file-scope static/global channels array shared by all instances — freeing it on one instance's removal corrupted every other live instance.
- adc: ad7490: the final error label didn't check ret, so failures were masked and the function returned success unconditionally.
- led: max25603: max25603_set_enable() computed duty cycle inverted relative to max25603_set_beam() and the driver's own example, so duty=100 (fully on) produced a fully-off PWM.
- projects: max31827-evkit: removed two dead .mk files left over from the pre-CMake build (not tracked/referenced anywhere).
- drivers/platform/ftd2xx: removed the non-functional ftd2xx_uart driver — LibMPSSE has no UART primitives, so it never touched real hardware (init/remove just wrapped a dummy descriptor, write() printf'd to host stdout, no read op existed). Removed all build/Kconfig wiring (ftd2xx.cmake, libraries/Kconfig) and corrected inaccurate win/mac CMake comments.
- projects: max14916, max31827-evkit: since these boards have no spare UART behind the FTDI dongle, UART is now a compile-time no-op (#ifndef CONFIG_FTD2XX) instead of routing through the removed fake backend.

## PR Type

## PR Checklist


RaduSabau1:
Closing this PR and opening a new one with multiple bugs on correct new branch

PR #3335, opened 2026-09-08
ci: build_projects: add per-project build timeout

liviutomoiaga:
## Summary
 - Adds a 600 s (10 min) timeout for Xilinx and 120 s for the rest of platforms, to each per-project `no_os_build.py` invocation in `build_projects.py`
 - When a project hangs (e.g. CubeMX version-mismatch popup on STM32), the build is killed and reported as `TIMEOUT` so remaining projects still build and the shared runner is released promptly

## PR Type

## PR Checklist

### DIFF
--- a/tools/scripts/build_projects.py
+++ b/tools/scripts/build_projects.py
@@ -68,6 +68,8 @@ def parse_input():
 		args.platform, args.build_name, args.builds_dir, args.hardware, args.hdl_branch)
 
 ERR = 0
+BUILD_TIMEOUT = 120
+XILINX_BUILD_TIMEOUT = 600
 LOG_START = " -> "
 TOKEN = os.environ.get('TOKEN')
 BRANCH = os.environ.get('BRANCH')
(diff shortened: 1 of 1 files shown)


tools/scripts/build_projects.py
```diff
@@ -68,6 +68,7 @@ def parse_input():
 		args.platform, args.build_name, args.builds_dir, args.hardware, args.hdl_branch)
 
 ERR = 0
+BUILD_TIMEOUT = 600  # Per-project configure+build timeout in seconds
```
rbolboac: in my opinion, 2 minutes are more than enough

liviutomoiaga: I was thinking the same as well, but this change would affect all platforms, including Xilinx. And in case building a Xilinx projects takes longer than usual, 10 mins gives enough headroom to avoid false timeouts while still catching genuine hangs

SRaus: Since the build times are different, I would go with different timeout values for every platform, if possible.

liviutomoiaga: Hmm fair point. But since we do not really track individual build times per platform I would suggest to have a simpler approach, one timeout Xilinx specific with 10 mins value that should cover the regeneration of BSP when needed, and 2 mins for everything else. One if-statement, nothing extra to maintain when a new platform is added.

rbolboac:
I would squash the 2 commits, the rest looks good

tools/scripts/build_projects.py
```diff
@@ -68,6 +68,8 @@ def parse_input():
 		args.platform, args.build_name, args.builds_dir, args.hardware, args.hdl_branch)
 
 ERR = 0
+BUILD_TIMEOUT = 120
+XILINX_BUILD_TIMEOUT = 600
```
SRaus, line 72: Is 10 min enough for complex Xilinx projects (i.e. Versal ones) to build BSP libs and also no-os binary?

SRaus (changes requested):

tools/scripts/build_projects.py
liviutomoiaga: Yes

SRaus (approved):

rbolboac (approved):

PR #3340, opened 2026-09-09
Bugfix/admt4000 misc issues

kister-jimenez:
## Pull Request Description

Project was not working correctly due to:
1. Missing define for ACTIVE_PLATFORM causing EEPROM reading to read from incorrect I2C address.
2. Misplaced IIO_IGNORE_BUFF_OVERRUN_ERR in CMakelist.
3. OOB access of array in ECC.
This PR addresses these issues.

There are other issues that are also addressed by this PR:
1. conv_sync_mode availalable shows null values.
2. boolean attributes in iio_evb showing random non-zero numbers instead of "1" for true.
3. admt4000 turns channel samples truncated by 1 bit.

## PR Type
- [x] Bug fix (change that fixes an issue)
- [ ] New feature (change that adds new functionality)
- [ ] Breaking change (has dependencies in other repos or will cause CI to fail)

## PR Checklist
- [x] I have followed the [Coding style guidelines](http://analogdevicesinc.github.io/no-OS/drivers_guide.html#coding-style)
- [x] I have complied with the [Submission Checklist](http://analogdevicesinc.github.io/no-OS/contributing.html#submission-checklist)
- [x] I have performed a self-review of the changes
- [x] I have commented my code, at least hard-to-understand parts
- [x] I have build all projects affected by the changes in this PR
- [x] I have tested in hardware affected projects, at the relevant boards
- [x] I have signed off all commits from this PR
- [x] I have updated the documentation (wiki pages, ReadMe etc), if applies

PR #3341, opened 2026-09-10
Bugfix/seacse 452 admt4000 bug fix

kister-jimenez:
## Pull Request Description
Updated from the PR #3340
Project was not working correctly due to:

Missing define for ACTIVE_PLATFORM causing EEPROM reading to read from incorrect I2C address.
Misplaced IIO_IGNORE_BUFF_OVERRUN_ERR in CMakelist.
OOB access of array in ECC.
This PR addresses these issues.
There are other issues that are also addressed by this PR:

conv_sync_mode availalable shows null values.
boolean attributes in iio_evb showing random non-zero numbers instead of "1" for true.
admt4000 turns channel samples truncated by 1 bit.

## PR Type

## PR Checklist


drivers/position/admt4000/admt4000.c
```diff
@@ -44,6 +44,13 @@
 #include "no_os_alloc.h"
 #include "no_os_delay.h"
 
+/* Bytes required to hold the input data with parity. */
+#define ADMT4000_ECC_CODE_SIZE		(16)
+#define ADMT4000_ECC_PARITY_BITS	(8)
```
rbolboac, line 49: something claude observed, shouldn't this be 7?

drivers/position/admt4000/admt4000.c
```diff
@@ -405,23 +413,15 @@ static int admt4000_ecc_encode(uint8_t *code, uint8_t *input,
 	int value;
 	uint8_t position;
 	uint8_t xtract;
-	uint8_t parity_num;
-	uint8_t code_length;
+	uint8_t parity_num = ADMT4000_ECC_PARITY_BITS;
+	uint8_t code_length = ADMT4000_ECC_CODE_BIT_WIDTH;
```
rbolboac, line 417: can we derive these from input_size instead of hardcoding?

drivers/position/admt4000/admt4000.c
kister-jimenez: The total parity bits is actually 8. The 8th bit is the parity of all 7 parity bits.

drivers/position/admt4000/admt4000.c
kister-jimenez: you mean a more generalized way? The code_width and parity bits is a configuration for the ecc computation so if the input is smaller than (code_width (in bits) - parity_bits), padding should be added automatically. I'll see what I can do, this was already like this by the original developer so I only tried to touch this as little as possible.

drivers/position/admt4000/admt4000.c
kister-jimenez: Reverted the computation of redundant bits so it should be 7 parity bits + 1 overall bits at the end.

drivers/position/admt4000/admt4000.c
kister-jimenez: I modified the admt4000_ecc_encode and removed the "code" argument as this is not needed. Now the code width is generated from input size as well.

kister-jimenez:
V2:

- Reverted the calculation of count of parity bits.
- Removed unused function argument "code" in admt4000_ecc_encode and have the width computed inside the function based on input_size.
- Removed the encode array in admt4000_update_ecc since this is taken care inside admt4000_ecc_encode.
- I renamed the ADMT4000_ECC_MAX_INPUT_SIZE with ADMT4000_ECC_DATA_SIZE to make it more intuitive. ADMT4000 has a maximum of 120 bits (15 bytes) for the ECC data.

drivers/position/admt4000/admt4000.c
```diff
@@ -389,39 +393,45 @@ static int admt4000_hamming_calc(uint8_t position, uint8_t code_length,
...
-	while ((input_size * 8) > (1 << i) - (i + 1)) {
+	while (data_bits > (1 << i) - (i + 1)) {
 		parity_num += 1;
 		i++;
 	}
 
 	/* In bits */
-	code_length = parity_num + (input_size * 8);
+	code_length = parity_num + data_bits;
+	code = (uint8_t *)no_os_calloc(1, input_size + NO_OS_DIV_ROUND_UP(parity_num,
```
rbolboac, line 430: memory leak, should be freed before returning

drivers/position/admt4000/iio_admt4000.h
```diff
@@ -37,6 +37,8 @@
 #include "iio.h"
 #include "no_os_gpio.h"
 
+#define ADMT4000_NUM_CHANNELS	6
```
rbolboac, line 40: is not ADMT4000_NUM_CHAN the same?

drivers/motor/tmc5240/iio_tmc5240.c
```diff
@@ -155,8 +155,6 @@ static struct iio_attribute tmc5240_iio_acceleration_attrs[] = {
 		.priv = TMC5240_CH_ATTR_CALIBSCALE,
 		.store = tmc5240_iio_attr_store,
 		.show = tmc5240_iio_attr_show,
-		.shared = IIO_SHARED_BY_ALL,
```
rbolboac, line 158: let's leave these as they were, because now there is a single variable containing the information for all channels, what you could do is remove the store functionality which gives the impression that this value could be changed by the user, but actually can't

drivers/position/admt4000/iio_admt4000.c
rbolboac, line 652: `admt4000_conv_sync_mode_avail[mode]` should be checked for null, similarly with the fixes you already added

drivers/position/admt4000/admt4000.c
kister-jimenez: Good catch. I added the no_os_free but I might have missed this during git add -p.

kister-jimenez:
V3:

- Fix memory leak in the admt4000_ecc_encode
- Fix motor calibscale to have separate handling. Users can now independently handle calibscale in different channels depending on mechanical properties.
- Include show function of conv_sync_mode attribute in the fix invalid NULL values.
- remove unused ADMT4000_NUM_CHAN since other variants of ADMT4000 have multiple instances of the channel type, the enum should not be used for the channel count.

drivers/motor/tmc5240/iio_tmc5240.c
kister-jimenez: Updated the code to handle each channel's calibscale independently but with a default initial value of 1.0 in the driver.

drivers/position/admt4000/iio_admt4000.h
kister-jimenez: Removed the ADMT4000_NUM_CHAN since other variants of ADMT4000 have multiple instances of certain channel types.

drivers/position/admt4000/iio_admt4000.c
kister-jimenez: Done.

drivers/motor/tmc5240/tmc5240.h
rbolboac: how about deleting this as well?

rbolboac (commented):
overall looks good, just one more nit from my side

kister-jimenez:
V4:

- Removed remaining step_angle_millidegrees in the tmc5240.h

rbolboac (approved):

PR #3351, opened 2026-09-11
tools: xilinx: enable boot.bin generation

rbudai98:
During SD card booting, boot.bin is possible to be created from the cilinc projec using the added python scripts

During Xilinx platform builds, allow user to generate boot.bin by adding the fllag '--boot-bin' to the python build command

## PR Type

## PR Checklist

### DIFF
--- a/tools/scripts/no_os_build.py
+++ b/tools/scripts/no_os_build.py
@@ -514,7 +514,47 @@ def append_log(log_path, section, result):
         f.write("\n")
 
 
-def run_build(repo_root, combo, build_dir_base, jobs, clean, dry_run, probe=None, flash=False, fresh=False, hardware=None):
+def load_boot_bin_module(repo_root):
+    """Import tools/scripts/platform/xilinx/boot_bin.py by path.
+
+    Loaded on demand rather than at module import: it is only needed for
+    --boot-bin, and importing by path keeps 'platform' (which shadows a stdlib
+    module) off sys.path.
+    """
+    import importlib.util
+
+    path = repo_root / "tools" / "scripts" / "platform" / "xilinx" / "boot_bin.py"
+    spec = importlib.util.spec_from_file_location("no_os_boot_bin", path)
+    module = importlib.util.module_from_spec(spec)
+    spec.loader.exec_module(module)
+    return module
+
+
+def boot_bin_cmd(repo_root, combo, build_dir):
+    """Equivalent standalone command, for --dry-run output."""
+    script = repo_root / "tools" / "scripts" / "platform" / "xilinx" / "boot_bin.py"
+    return [sys.executable, str(script), "--build-dir", str(build_dir)]
+
+
+def run_boot_bin(repo_root, combo, build_dir):
+    """Package one combination's build into a BOOT.BIN. Returns (ok, message).
+
+    Everything the packager needs is in <build_dir>/bootinfo.json, written by
+    the Xilinx CMake flow at configure time.
+    """
+    if combo["platform"] != "xilinx":
+        return True, (f"--boot-bin: skipped, {combo['platform']} is not a "
+                      "Xilinx platform")
+
+    boot_bin = load_boot_bin_module(repo_root)
+    try:
+        boot_bin.make_boot_bin(build_dir)
+    except boot_bin.BootBinError as e:
+        return False, f"BOOT.BIN failed: {e}"
+    return True, ""
+
+
+def run_build(repo_root, combo, build_dir_base, jobs, clean, dry_run, probe=None, flash=False, fresh=False, hardware=None, boot_bin=False):
     """Run cmake configure + build (and optionally flash) for a single combination.
 
     Returns (combo, success, detail). On failure, detail is the error message.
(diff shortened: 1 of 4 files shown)


tools/scripts/platform/xilinx/boot_bin.py
```diff
@@ -0,0 +1,236 @@
...
+    for key, mapping in ARCH_MAP.items():
+        if key in arch:
+            return mapping
+    raise BootBinError(f"unsupported architecture: {arch}")
+
+
+def generate_fsbl(info, quiet=False):
+    """Build the FSBL if it is not already there. Returns its path.
+
+    create_fsbl builds a throwaway Vitis platform component just for the boot
```
SRaus, line 110: From what I know, the fsbl.elf is the same for both Linux and No-OS BOOT.BINs.
Usually, it's generated at the end of HDL build, so probably it can be exported exactly as the XSA.

tools/scripts/platform/xilinx/boot_bin.py
```diff
@@ -0,0 +1,236 @@
...
+    loader, which takes minutes. It writes into the same workspace the 'flash'
+    target uses, so the two share one FSBL.
+    """
+    fsbl = Path(info["fsbl"])
+    if fsbl.is_file():
+        return fsbl
+
+    if not quiet:
+        print("==> Generating FSBL (this takes a few minutes)", flush=True)
+    ws = Path(info["fsbl_ws"])
```
SRaus, line 120: If `generate_fsbl()` runs and really takes a couple of minutes, we'll need to revisit the 10 min timeout (PR https://github.com/analogdevicesinc/no-OS/pull/3335)

tools/scripts/platform/xilinx/boot_bin.py
```diff
@@ -0,0 +1,236 @@
...
+        subprocess.run(cmd, check=True)
+    except subprocess.CalledProcessError as e:
+        raise BootBinError(f"create_fsbl failed (exit code {e.returncode})") from e
+
+    if not fsbl.is_file():
+        raise BootBinError(f"create_fsbl reported success but {fsbl} is missing")
+    return fsbl
+
+
+def write_bif(path, fsbl, bitstream, elf, dest_cpu, pmufw=None):
```
SRaus, line 138: This method should be general enough for zynq, zynqmp and versal platforms.
What do you say about adding a clear delimitation between those.

tools/scripts/platform/xilinx/boot_bin.py
```diff
@@ -0,0 +1,236 @@
...
+    if not elf.is_file():
+        raise BootBinError(f"application ELF not found: {elf} (build it first)")
+
+    bitstream = Path(bitstream) if bitstream else (
+        Path(info["bitstream"]) if info.get("bitstream") else None)
+    if bitstream and not bitstream.is_file():
+        raise BootBinError(f"bitstream not found: {bitstream}")
+    if not bitstream and not quiet:
+        # A PL-less design is legal; a missing .bit in a design that needs one
+        # produces an image that boots and then hangs on the first PL access.
```
SRaus, line 175: It's legal...  but do we really need to generate an image that's hanging on booting?

SRaus (changes requested):

tools/scripts/platform/xilinx/boot_bin.py
rbudai98: This flag is intended as an optional boot.bin generation for users that use sd cards. In any case it should not be different from compiling the actual firmware, it might not be necessary to run again in CI

tools/scripts/platform/xilinx/boot_bin.py
rbudai98: Than it needs change of approach for xilinx builds/files, this PR is focusing only on potential possibility of boot.bin generation, in case the user needs it.

tools/scripts/platform/xilinx/boot_bin.py
rbudai98: Yes, it might need adjusting from this part, have to check it in more details.

tools/scripts/no_os_build.py
```diff
@@ -936,6 +997,13 @@ def main():
         action="store_true",
         help="Flash the firmware after a successful build (requires --probe)",
     )
+    build_parser.add_argument(
+        "--boot-bin",
+        action="store_true",
+        help="Xilinx only: package FSBL + bitstream + ELF into an SD-card "
+             "BOOT.BIN after a successful build (tools/scripts/platform/"
+             "xilinx/make_boot_bin.sh)",
```
liviutomoiaga, line 1005: I think the reference here should be `../xilinx/boot_bin.py`

tools/scripts/platform/xilinx/boot_bin.py
liviutomoiaga, line 1: Since it's a new file added by you, please take in mind to introduce the SPDX license header on it. 

Example:

https://github.com/analogdevicesinc/no-OS/blob/b33411c983a8b8a0c177703baf6779440dcb7fe9/tools/scripts/platform/xilinx/util.py#L2-L3

tools/scripts/platform/xilinx/boot_bin.py
liviutomoiaga: Since `--boot-bin` is an optional argument and is not directly used in the CI build, I would say that PR https://github.com/analogdevicesinc/no-OS/pull/3335 is save to merge as-is.

tools/scripts/platform/xilinx/boot_bin.py
SRaus: OK. Based on the description in the `xilinx_boot.cmake` header comment (no need to rediscover files, no re-parse of CMakeCache.txt, no second unzip of XSA, etc.), I see this as a significant optimization. 

So, my initial understanding was that this new flow will fully replace the current implmentation, reducing build times (and in this case "--boot-bin" would be on true by default).

PR #3354, opened 2026-09-17
scripts: xilinx: retry flaky BSP generation

liviutomoiaga:
## Pull Request Description

create_platform_component() intermittently fails generating the FSBL on Vitis 2025.1; retry on a clean workspace (a fresh run usually succeeds) and verify hw0.xpfm + libxil.a before continuing.

## PR Type

## PR Checklist


tools/scripts/platform/xilinx/util.py
```diff
@@ -460,6 +460,92 @@ def get_arch(hw_path, hw_file, target):
...
+
+    Returns the live vitis client on success (the caller reuses it for the app
+    component and disposes it at the end). Exits non-zero if all attempts fail.
+    """
+    import shutil
+    import vitis
+
+    xpfm, libxil = _hw0_artifacts(out_dir, vcpu)
+
+    for attempt in range(1, attempts + 1):
```
SRaus: you can simplify this by just using `for attempt in range(0, attempts)`

tools/scripts/platform/xilinx/util.py
```diff
@@ -460,6 +460,92 @@ def get_arch(hw_path, hw_file, target):
...
+    is the standalone BSP archive the CMake build links against. Both present
+    means the platform is usable.
+    """
+    export = os.path.join(out_dir, "hw0", "export", "hw0")
+    xpfm = os.path.join(export, "hw0.xpfm")
+    libxil = os.path.join(export, "sw", f"standalone_{vcpu}", "lib", "libxil.a")
+    return xpfm, libxil
+
+
+def _create_hw0_platform(out_dir, xsa, vcpu, attempts=2):
```
SRaus, line 480: It's a bit weird that you set `attempts=2`, but there will actually be 3 attempts (`attempt in range(1, attempts + 1)`)

tools/scripts/platform/xilinx/util.py
SRaus, line 1: Since this issue is reproducible only in CI environment and not when build locally - instead re-trying to build multiple times, I would investigate and fix the root cause.

liviutomoiaga: Agree, but sometimes there are just some random/transient issues that go away with a retry. I can lower the attempts from 3 to 2, but I think having a small retry for these kinds of cases still makes sense.

tools/scripts/platform/xilinx/util.py
liviutomoiaga: I've changed it. Now runs exactly `attempts` times

SRaus (dismissed):

rbolboac (approved):
lgtm

PR #3359, opened 2026-09-21
Ci/xilinx missing xsa warn

liviutomoiaga:
## Pull Request Description

Xilinx platform builds abort the whole run with a `FileNotFoundError` when a
target's `system_top.xsa` was not published by the HDL build this run:

```
File ".../build_projects.py", line 276, in get_hardware
    if filecmp.cmp(filename, tmp_filename):
FileNotFoundError: [Errno 2] No such file or directory:
    '.../builds_main/new_hardware/adrv903x_zcu102/system_top.xsa'
```

A missing `.xsa` is an HDL-availability issue, not a no-OS build error, so it
should not fail the run. This PR loosens that path and surfaces the condition
instead.

## PR Type

## PR Checklist


.github/workflows/build-xilinx-platform.yaml
```diff
@@ -94,6 +94,83 @@ jobs:
         env:
           CLOUDSMITH_API_KEY: ${{ env.CLOUDSMITH_API_KEY }}
           TOKEN: ${{ secrets.GH_GHDL_TOKEN }}
+      - name: Report missing XSA on PR
+        # A missing system_top.xsa is a warning, not a failure (build_projects.py
```
liviutomoiaga: I would like to move this whole logic somewhere else, but for now I'll keep it here since it's in testing phase.

SRaus (commented):
The  HDL 'main' branch is not expected to be stable... it never been. When an XSA is missing, you should retrieve in from previous HDL timestamps. In this way you'll ensure that the No-OS code builds fine against latest 'known-good' HDL artifacts. 

On the long term, every HDL PR (either a specific project that was failing got fixed - so you'll have newer XSAs, either other change) should trigger No-OS CI, quickly detecting any HDL<->No-OS dependency issues and provide feedback back to HDL devs.

tools/scripts/build_projects.py
```diff
@@ -32,6 +32,7 @@
 TBLUE =  '\033[34m' # Green Text	
 TRED =  '\033[31m' # Red Text	
 TWHITE = '\033[39m' #Withe text
```
bia1708: There's a typo in the comment, could you please fix it?

bia1708 (changes requested):

tools/scripts/build_projects.py
liviutomoiaga: Fixed! :D

liviutomoiaga:
> The HDL 'main' branch is not expected to be stable... it never been. When an XSA is missing, you should retrieve in from previous HDL timestamps. In this way you'll ensure that the No-OS code builds fine against latest 'known-good' HDL artifacts.
> 
> On the long term, every HDL PR (either a specific project that was failing got fixed - so you'll have newer XSAs, either other change) should trigger No-OS CI, quickly detecting any HDL<->No-OS dependency issues and provide feedback back to HDL devs.

I've updated the xsa search mechanism by picking the newest HDL timestamp that actually contains it, instead of scanning a fixed FOLDERS_NR window of the latest timestamps. This can be reviewed in #3376 

About long term solution, we can track it as different action.

SRaus (dismissed):

SRaus (approved):

rbolboac (approved):

tools/scripts/build_projects.py
```diff
@@ -379,6 +528,28 @@ def build_cmake_project(noos, project, _platform, _build_name, export_dir,
 				os.environ.clear(); os.environ.update(env)
 				continue
 			hardware_arg = " --hardware %s" % hardware_file
+			# Found-but-stale: the .xsa exists so the build proceeds, but if its
+			# HDL timestamp is old the design may lag recent HDL changes. Warn
```
bia1708: typo lag -> lack ?

liviutomoiaga: It's more like `lag behind`

bia1708 (approved):

PR #3371, opened 2026-09-22
Bugfix/sdgsw 1310 fix incom ptr warn

rbolboac:
## Pull Request Description

Remove `-Wincompatible-pointer-types` warnings from the drivers CI build. Each was a return width, parameter type, or struct-field type mismatch (on the ARM EABI `int32_t` is `long int`, distinct from `int`). Fixes are grouped by category:

- **drivers: iio** — make the `iio_device` op handlers (`read_dev`, `write_dev`, `pre_enable`, `debug_reg_read/write`) natively match the field types across 17 drivers instead of casting: iio-local handlers take the exact `void *`-based signature and recover the typed pointer inside, and shared core functions get small typed wrappers. The wrappers also bridge the register width (`uint8_t`/`uint16_t` vs `uint32_t`) that the old raw function-pointer casts silently truncated.
- **gmsl: common** — retype the `diag_check_*` fields of `gmsl_dev_diag_ops` from `int32_t (*)` to `int (*)` to match the  max96792/max96793 implementations.
- **drivers: clk** — return `int` (not `int32_t`) from the `clk_recalc_rate` / `clk_set_rate` / `clk_round_rate` callbacks in ad9528, ad9545, hmc7044 and adrv9009 to match `no_os_clk_platform_ops`.
- **power: ltm4700** — cast `&val` to `int32_t *` for `iio_format_value`  drop `const` from the store handler's `buf`, and return `int` from the  `debug_reg` handlers to match the field types.
- **adc: ad7799** — add the missing `intptr_t priv` parameter to the  `.show` handler and wrap `ad7799_read`/`ad7799_write` in static handlers matching the `debug_reg` field types.

Four genuine bugs surfaced by the warning are fixed in separate commits:

- **potentiometer: ad5293** — the error path passed a GPIO descriptor to `no_os_spi_remove()`; pass `spi_desc` so the SPI descriptor is freed.
- **adc: ad713x** — `gpio_cs_sync` was typed `no_os_spi_desc *` but used only with the GPIO API; retype it to `no_os_gpio_desc *`.
- **adc: ad9083** — fix `ad9083_udelay` to match `adi_delay_us_t` (`int32_t(void *, uint32_t)`).
- **afe: ad4110** — give the IRQ handler a `void *` signature matching `no_os_callback_desc.callback` and recover the typed context inside.

All changes are behavior-preserving, except the iio width-bridging wrappers, which additionally fix latent partial register reads/writes that the old raw function-pointer casts masked. Build exits 0 with no new warnings; changed lines are astyle-clean.

## PR Type

## PR Checklist


rbolboac:
change log: main rebase and fixed conflicts

